PR #20, opened 2019-08-18
Two-pass triangulation

elalish:


elalish:
Only marginally simpler and not as robust, so far.

PR #84, opened 2022-03-08
Add CGAL corefinement perf test

ochafik:
First off: amazing work, I'm super excited about Manifold and hope we can use it in OpenSCAD and beyond!

I've ported perf_test to use CGAL's corefinement after reading [your post](https://elalish.blogspot.com/2022/03/manifold-performance.html),
which I believe tests OpenSCAD more than it does CGAL itself.

I haven't had a chance to test with CUDA hardware or with OpenMP (which I expect are gonna leave CGAL in the dust, since I don't think corefinement is one of their parallelizable operations), but thought it worth doing a mono-threaded comparison as fairly as possible.

To that effect I've used the `CGAL::Epick` [kernel](https://doc.cgal.org/latest/Kernel_d/group__PkgKernelDKernels.html) (rather than the exact, [GMP](https://gmplib.org/) rationals-based `CGAL::Epeck` we use in OpenSCAD). `Epick` is only marginally faster than `Epeck` here anyway, but note that both provide exact predicates, `Epick` is only "inexact" in that it's built from double values (much like Manifold, AFAICT).

Haven't tested all the operations but hopefully you can vary the test cases using this PR as a starting point.

Some take-aways from only testing union and difference (again, only CPU-monothreaded)
- Manifold's union is up to 2.3x faster than corefinement for meshes >32k faces, but...
- Corefinement union 2-3x faster for small meshes <=2k faces
- Corefinement difference is faster at all mesh sizes I've tested

(see [raw results here](https://docs.google.com/spreadsheets/d/13ze7G1JTXOtgKhyWsBr890G-kdcGM-aG-fp6QcWManw/edit#gid=0), took best of 3 runs)

Tested with:

```bash
git clone https://github.com/ochafik/manifold.git --single-branch -b perf_test_cgal
cd manifold
docker run -v $PWD:/src:z -it nvidia/cuda:11.1-devel-ubuntu20.04 

# Then:
apt-get -y update
DEBIAN_FRONTEND=noninteractive apt install -y cmake libglm-dev libassimp-dev libboost-graph-dev libcgal-dev

cd /src
mkdir -p buildCPP && cd buildCPP
cmake -DCMAKE_BUILD_TYPE=Release -DMANIFOLD_USE_CPP=ON .. && make
tools/perfTest
tools/perfTestCGAL
```

@sloriot FYI more microbenchmarking action :-)

### DIFF
--- a/tools/perf_test_cgal.cpp
+++ b/tools/perf_test_cgal.cpp
@@ -1,0 +1,80 @@
+// Copyright 2020 Emmett Lalish & contributors.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#include <CGAL/Exact_predicates_exact_constructions_kernel.h>
+#include <CGAL/Exact_predicates_inexact_constructions_kernel.h>
+#include <CGAL/Polygon_mesh_processing/corefinement.h>
+#include <CGAL/Surface_mesh.h>
+
+#include <chrono>
+#include <fstream>
+#include <iostream>
+
+#include "manifold.h"
+
+using namespace manifold;
+
+// Epick = Exact predicates Inexact constructions. Seems fair to use to compare
+// to Manifold, which seems to store double coordinates.
+typedef CGAL::Epick Kernel;
+
+// Epeck = Exact predicates Exact constructions. What OpenSCAD uses to guarantee
+// geometry ends up where it should even after many operations. typedef
+// CGAL::Epeck Kernel;
+
+typedef CGAL::Point_3<Kernel> Point;
+typedef CGAL::Surface_mesh<Point> TriangleMesh;
+
+void manifoldToCGALSurfaceMesh(Manifold &manifold, TriangleMesh &cgalMesh) {
+  typedef boost::graph_traits<TriangleMesh> GT;
+  typedef typename GT::vertex_descriptor vertex_descriptor;
+
+  auto maniMesh = manifold.GetMesh();
+
+  std::unordered_map<size_t, vertex_descriptor> vertices;
+  for (size_t i = 0, n = maniMesh.vertPos.size(); i < n; i++) {
+    auto &vert = maniMesh.vertPos[i];
+    vertices[i] = cgalMesh.add_vertex(Point(vert.x, vert.y, vert.z));
+  }
+
+  for (auto &triVert : maniMesh.triVerts) {
+    std::vector<vertex_descriptor> polygon{
+        vertices[triVert[0]], vertices[triVert[1]], vertices[triVert[2]]};
+    cgalMesh.add_face(polygon);
+  }
+}
+
+int main(int argc, char **argv) {
+  for (int i = 0; i < 8; ++i) {
+    Manifold sphere = Manifold::Sphere(1, (8 << i) * 4);
+    Manifold sphere2 = sphere;
+    sphere2.Translate(glm::vec3(0.5));
+
+    TriangleMesh cgalSphere, cgalSphere2, cgalOut;
+    manifoldToCGALSurfaceMesh(sphere, cgalSphere);
+    manifoldToCGALSurfaceMesh(sphere2, cgalSphere2);
+
+    auto start = std::chrono::high_resolution_clock::now();
+    auto result =
+        CGAL::Polygon_mesh_processing::corefine_and_compute_difference(
+            // CGAL::Polygon_mesh_processing::corefine_and_compute_union(
+            // CGAL::Polygon_mesh_processing::corefine_and_compute_intersection(
+            cgalSphere, cgalSphere2, cgalSphere);
+
+    auto end = std::chrono::high_resolution_clock::now();
+    std::chrono::duration<double> elapsed = end - start;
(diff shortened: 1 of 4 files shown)


tools/perf_test_cgal.cpp
```diff
@@ -0,0 +1,80 @@
...
+
+using namespace manifold;
+
+// Epick = Exact predicates Inexact constructions. Seems fair to use to compare
+// to Manifold, which seems to store double coordinates.
+typedef CGAL::Epick Kernel;
+
+// Epeck = Exact predicates Exact constructions. What OpenSCAD uses to guarantee
+// geometry ends up where it should even after many operations. typedef
+// CGAL::Epeck Kernel;
```
elalish, line 34: I should probably read up on CGAL's constructions definitions. I would argue every number in a computer is inexact by definition since our Turning machines can't compute Real numbers. Still, is this just about precision, or does this choice also affect their manifoldness guarantees? And how much performance difference does it make?

tools/perf_test_cgal.cpp
```diff
@@ -0,0 +1,80 @@
...
+    Manifold sphere = Manifold::Sphere(1, (8 << i) * 4);
+    Manifold sphere2 = sphere;
+    sphere2.Translate(glm::vec3(0.5));
+
+    TriangleMesh cgalSphere, cgalSphere2, cgalOut;
+    manifoldToCGALSurfaceMesh(sphere, cgalSphere);
+    manifoldToCGALSurfaceMesh(sphere2, cgalSphere2);
+
+    auto start = std::chrono::high_resolution_clock::now();
+    auto result =
```
elalish, line 69: Is the result also a triangle mesh, or is it a polygon mesh?

elalish (approved):
Thank you! Great comparison, and saves me having to learn CGAL. Are you going to do the Sponge next? It might be nice to put some of these perf tests together so it's easier to get side-by-side comparisons (probably save some code too).

elalish:
Thanks for the raw results; super interesting! I'm curious why OpenSCAD is an order of magnitude slower than CGAL here... 

I'm stoked that mine is performing pretty nearly as well; I haven't spent much time on performance optimization and I have the feeling there is some low-hanging fruit.

tools/perf_test_cgal.cpp
ochafik: Yes [corefinement](https://doc.cgal.org/latest/Polygon_mesh_processing/group__PMP__corefinement__grp.html) consumes and produces triangles meshes, even though Surface_mesh is more generic.

tools/perf_test_cgal.cpp
ochafik: Reals are out of reach, but rationals aren’t (if you avoid sqrt): store arbitrary precision integers  for numerator and denominator, simplify fractions by finding their gcd after each operation, and you’re only limited by memory. That’s what Gmpq is doing (wrapper around [GNU MP](https://en.m.wikipedia.org/wiki/GNU_Multiple_Precision_Arithmetic_Library)), powering the Epeck kernel when ‘-DCGAL_USE_GMPXX=1’ is set.

ochafik: Re/ manifoldness guarantees Epick does magic by offering exact predicates even though numbers are ultimately represented as doubles. If the corefinement inputs are manifold (and closed), so should be the outputs.

ochafik:
> I'm curious why OpenSCAD is an order of magnitude slower than CGAL here...

I'm curious too haha! In UI mode (as opposed CLI) I haven't implemented new rendering logic so the results are converted to a mesh (PolySet) the existing renderer can process, but other things could be going wrong, I'll investigate :-)

ochafik:
> I'm stoked that mine is performing pretty nearly as well as this new CGAL; I haven't spent much time on performance optimization and I have the feeling there is some low-hanging fruit.

Super impressive indeed 👏, and exciting times ahead!

ochafik:
> saves me having to learn CGAL. Are you going to do the Sponge next?

@elalish when I have a bit more time I'll try and repurpose some CLI I wrote that reads .off files and does union/diff/inter with different kernels and now different libraries, so you can do all the tests you want ;-)

pca006132:
It seems that quite a bit of time is spent on initializing the vectors, e.g. halfedge vectors, that we will overwrite them later anyway. [Here](https://github.com/pca006132/manifold/commit/f181c0d0e53fde387c3833c6843de53d93bf3c65) is a patch that avoids such initialization (a bit ugly and unsafe, not sure if I should open a PR now) that improves the performance of single threaded manifold operations on small meshes.

Before this patch (single threaded perfTest):
```
nTri = 512, time = 0.0107632 sec
nTri = 2048, time = 0.0160586 sec
nTri = 8192, time = 0.028829 sec
nTri = 32768, time = 0.0673084 sec
nTri = 131072, time = 0.242107 sec
nTri = 524288, time = 0.941614 sec
nTri = 2097152, time = 4.21132 sec
nTri = 8388608, time = 17.4662 sec
```

After this patch:
```
nTri = 512, time = 0.00140315 sec
nTri = 2048, time = 0.00383832 sec
nTri = 8192, time = 0.0140288 sec
nTri = 32768, time = 0.0530416 sec
nTri = 131072, time = 0.236449 sec
nTri = 524288, time = 0.917415 sec
nTri = 2097152, time = 4.02338 sec
nTri = 8388608, time = 17.8529 sec
```

CGAL for comparison:
```
nTri = 512, time = 0.00349265 sec
nTri = 2048, time = 0.00595918 sec
nTri = 8192, time = 0.0176475 sec
nTri = 32768, time = 0.0586701 sec
nTri = 131072, time = 0.277333 sec
nTri = 524288, time = 1.20353 sec
nTri = 2097152, time = 5.51911 sec
nTri = 8388608, time = 25.991 sec
```

Sadly this patch cannot improve the performance for OpenMP and CUDA backend. I guess there are still quite a lot of things that can be optimized better.

elalish:
Feel free to open a PR whenever you'd like feedback on it. I definitely like the idea of performance optimizations. Part of the reason I'd like to move to C++17 parallel algorithms is that it would allow me to remove the entire `VecDH` class, as that would then be handled by the compiler internally.

PR #88, opened 2022-03-16
windows compatibility support

yetigit:
had a ton of divide by zero errors so had to replace it with some other stuff


yetigit:
the whole thing was built using vcpkg to commit b7a1088ae9bfed114dd36e0d60ddb3db3b7a4b7f (for the dependencies of cgal). 
I then modified extras missing form the cmake build , from a couple of vs projects , 
like the glm include folder path and assimp include path as well as its static lib path

elalish:
Thanks! I don't have a windows box, so haven't gotten to try with their compilers myself. Is there an easy way around the path differences in Windows vs Linux? It might be simpler to use WSL2. Also, are the two issues you opened related to your PR here? Might be easier to discuss them here. Definitely seems like something is off.

collider/src/collider.cu
```diff
@@ -15,12 +15,54 @@
...
+    }
+}
+
+uint32_t __inline clz( uint32_t value )
+{
+    DWORD leading_zero = 0;
+
+    if ( _BitScanReverse( &leading_zero, value ) )
+    {
+       return 31 - leading_zero;
```
elalish, line 56: :100:

manifold/src/boolean3.cu
```diff
@@ -114,7 +115,7 @@ __host__ __device__ thrust::pair<int, glm::vec2> Shadow01(
                           Shadows(q1ex, p0x, expandP * normalP[q1e].x)
                     : Shadows(p0x, q1ex, expandP * normalP[p0].x) -
                           Shadows(p0x, q1sx, expandP * normalP[p0].x);
-  glm::vec2 yz01(0.0f / 0.0f);
+  glm::vec2 yz01(NAN);
```
elalish, line 118: That's convenient.

test/mesh_test.cpp
```diff
@@ -145,20 +146,20 @@ TEST(Manifold, GetMesh) {
...
-  int num_overlaps = manifold.NumOverlaps(manifold1);
-  ASSERT_EQ(num_overlaps, 228677);
-
-  Mesh mesh_out = manifold.GetMesh();
-  Manifold manifold2(mesh_out);
-  Mesh mesh_out2 = manifold2.GetMesh();
-  // Identical(mesh_out, mesh_out2);
-}
+//TEST(Manifold, Regression) {
+//  Manifold manifold(ImportMesh("c:/repos/manifold/test/data/gyroidpuzzle.ply"));
```
elalish, line 150: Someone must have found way to deal with this stupid Linux vs Windows file system stuff by now, right? I think we'll want to stick with relative paths too. I don't want to merge this until we find a way to uncomment these tests.

elalish (changes requested):

yetigit:
> Also, are the two issues you opened related to your PR here? Might be easier to discuss them here. Definitely seems like something is off.

yes it's related; if you can run the full loop of perftest on your system with these changes then i guess it's okayish
but then it would mean just being on windows is the culprit , because of some different instructions ?

test/mesh_test.cpp
pentacular: Relative paths like "./foo" should work just fine in windows.

But you'll need to consider the environment you're invoking the test from.

Try printing std::filesystem::current_path() to see where the paths should be relative to.

elalish:
@yetigit Have you tried installing the CUDA toolkit and compiling with nvcc? Even if you don't have an NVidia GPU, it can compile for CPU only and might have better support for Thrust. Or is that what you're already doing?

yetigit:
@elalish  re , yes it was compiled w nvcc , windows flavored CUDA right from the toolkit

elalish:
@yetigit Hmm, I wonder what the compiler difference is between MS and gcc... I'll need to find a way to run them side-by-side to properly debug. Think you can try to get the relative path thing working in Windows so we can merge this PR?

yetigit:
> @yetigit Hmm, I wonder what the compiler difference is between MS and gcc... I'll need to find a way to run them side-by-side to properly debug. Think you can try to get the relative path thing working in Windows so we can merge this PR?

done [b65b445](https://github.com/elalish/manifold/pull/88/commits/b65b445f43415a906767c75ed458ce9c83d6e8ea)

tools/perf_test.cpp
```diff
@@ -20,7 +20,7 @@
 using namespace manifold;
 
 int main(int argc, char **argv) {
-  for (int i = 0; i < 8; ++i) {
+  for (int i = 0; i < 3; ++i) {
```
elalish, line 23: This doesn't need to change, but it doesn't really hurt anything.

elalish (approved):
Thanks!

PR #92, opened 2022-04-18
Replace boost::graph with graphlite

elalish:
Since we're only using boost::graph for a simple connected components algorithm anyway, let's simplify the dependency by taking a single-file library.


fire:
```
CMake Error at manifold/CMakeLists.txt:5 (add_library):
  CUDA_STANDARD is set to invalid value '17'
```

This is odd.

third_party/graphlite/src/connected_components.cpp
```diff
@@ -0,0 +1,42 @@
+#include <deque>
```
fire, line 1: Lacking a license header?

fire:
`manifold\src\impl.cu` is still using boost https://github.com/elalish/manifold/blob/graphlite/manifold/src/impl.cu#L19-L21

fire:
I'm having some trouble trying to get int Manifold::Impl::InitializeNewReference
 to work.

elalish:
Sorry, this is a WIP. Just letting you know I'm working on it. I need to deal with the C++17 build stuff first.

PR #94, opened 2022-04-20
updated cu internals to c++17

elalish:
This was blocking #92, so figured better to sort it out here on its own.


fire:
Looks green 👍

PR #99, opened 2022-05-01
implemented sweep

pca006132:
Implemented a generic sweep operation on polygon cross section to create manifolds. Users can specify the sweeping path with a function that returns the 3D transformation matrix on the cross section (with z = 0). Users can specify the sweeping path to be closed by setting `isClosed = true`, so the first plane (step = 0) and the last plane (step = n - 1) will be connected together.

Example:
```c++
#include "manifold.h"
#include "glm/vec3.hpp"

using namespace manifold;

int main() {
    std::vector<manifold::PolyVert> vertices;
    for (int i = 0; i < 60; i++)
        vertices.push_back({glm::vec2(sind(-i*6) + 10, cosd(-i*6) + 10), i});

    Polygons polygons({vertices});
    std::function<glm::mat4x3(int)> f = [](int s) {
        glm::mat4x3 mat(1.0f);
        mat[0][0] = cosd(6*s);
        mat[2][0] = -sind(6*s);
        mat[0][2] = -mat[2][0];
        mat[2][2] = mat[0][0];
        mat[3][1] = 5*s/60.0;
        return mat;
    };
    // creates a spring
    Manifold m = Manifold::Sweep(polygons, 1000, f);
    return 0;
}
```

### DIFF
--- a/manifold/include/manifold.h
+++ b/manifold/include/manifold.h
@@ -56,6 +56,8 @@ class Manifold {
                           glm::vec2 scaleTop = glm::vec2(1.0f));
   static Manifold Revolve(const Polygons& crossSection,
                           int circularSegments = 0);
+  static Manifold Sweep(Polygons crossSection, int steps,
+                        std::function<glm::mat4x3(int)> &transform, bool closed = false);
   ///@}
 
   /** @name Topological
--- a/manifold/src/constructors.cu
+++ b/manifold/src/constructors.cu
@@ -13,6 +13,7 @@
 // limitations under the License.
 
 #include <thrust/sequence.h>
+#include <thrust/execution_policy.h>
 
 #include "graph.h"
 #include "impl.cuh"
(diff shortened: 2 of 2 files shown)


manifold/src/constructors.cu
```diff
@@ -268,6 +269,76 @@ Manifold Manifold::Extrude(Polygons crossSection, float height, int nDivisions,
   return extrusion;
 }
 
+Manifold Manifold::Sweep(Polygons crossSection, int steps,
+        std::function<glm::mat4x3(int)> &transform, bool isClosed) {
```
elalish, line 273: So the idea is that transform takes the step int as input and returns a transform matrix? That's not necessarily going to be the easiest function for users to write. Might it be better to just have a 3D polyline as input and deduce the transforms from it? Of course the roll axis gets a little tricky. 

Also it might be possible to simplify this a bit by using extrude/revolve and the warp function - see the Knot example.

manifold/include/manifold.h
```diff
@@ -56,6 +56,8 @@ class Manifold {
                           glm::vec2 scaleTop = glm::vec2(1.0f));
   static Manifold Revolve(const Polygons& crossSection,
                           int circularSegments = 0);
+  static Manifold Sweep(Polygons crossSection, int steps,
+                        std::function<glm::mat4x3(int)> &transform, bool closed = false);
```
elalish, line 60: Would you mind adding a sample and sample test before we merge this? No rush; let's talk through the API first.

elalish (commented):
Thanks, this is a good idea! Let's chat about the API - I'd like to make this easier to use.

manifold/src/constructors.cu
pca006132: Yes, my thought is to make this a general implementation and create some helper functions/specialization for the transform function. Generating the transform function from a 3D polyline should be possible?

elalish: That's fair, but how much different is this version than an `Extrude` followed by a `Warp`? I'd like to see the helper functions and specializations. If this code is more convenient we can always build it inside as an `Impl` method to support them. I'd definitely like to see your 3D polyline version; do you have other ideas as well?

pca006132: The example I give above is a version that is a bit more difficult to create with extrude followed by a warp (perhaps you can specify a twist and calculate the vertices, but this is more complicated)

image

For polyline, I am not sure what representation should I use for the polyline, so not much idea for now.

pca006132:
Closing to use another branch and leave the master branch for CI...

elalish:
Oh, actually you really shouldn't use the master branch for anything; best practice on git is to always make a new branch for each PR - it makes merging them much smoother.

manifold/src/constructors.cu
elalish: Food for thought, this is how I would go about making your spring:
```
const int n = 24;
const int m = 1000;
const float r0 = 10;
const float r1 = 50;
const float h = 300;
const int k = 10;

Polygons circle(1);
float dPhi = 360.0f / n;
for (int i = 0; i < n; ++i) {
  circle[0].push_back(
      {r0 * glm::vec2(cosd(dPhi * i) + 2.0f, sind(dPhi * i)), 0});
}

Manifold spring = Manifold::Extrude(circle, 1, m);

spring.Warp([r1, h, k](glm::vec3& v) {
  const float phi = v.z * k * 2 * glm::pi<float>();
  v.x += v.z * h;
  const float r = v.y + r1;
  v.y = r * glm::cos(phi);
  v.z = r * glm::sin(phi);
});
```

elalish: A polyline could just be a `std::vector<glm::vec3>`, but the tricky part comes with defining the twist along the axis. Maybe a quaternion at each end?

pca006132: I guess maybe we can have a look at https://github.com/openscad/openscad/pull/2796 ? The way they specify the extrusion seems pretty easy to use.

PR #103, opened 2022-05-04
build script refactoring

### LINKED ISSUE #98: Question : Emscripten integration
Hi !

I'm trying to integrate the library with emscripten and i'm a bit struggling.
I tried a lot of stuff but nothing worked.

Does anyone know a proper way to do it ?

Thanks a lot

pca006132:
Refactored to allow building with emscripten and compilers without the need for nvcc. The wasm build can now use assimp with static assets. Most of the tests succeed in wasm, except the sponge4 test which requires too much memory, and the determinism test which failed mysteriously and requires further debugging.

Instead of specifying `MANIFOLD_USE_CPP` and `MANIFOLD_USE_OMP`, we use `THRUST_BACKEND=CUDA|OMP|CPP` now. Emscripten build will automatically choose the CPP backend. nvcc is only needed for the CUDA backend. The thrust submodule is needed for OMP/CPP backend because nvcc is not being used. The assimp and glm submodules are only needed for wasm build.

Haven't tried the GitHub Action yet because I am unable to get it running on branches other than the master branch... Waiting for permission to run it on this PR.

Closes #98


elalish:
Awesome, thank you! I'll need to take a longer look at this and try it out on my side. Don't worry about the determinism test; that was already basically a TODO for me, as it's showing there's some minor non-determinism left even on the CUDA side.

pca006132:
I also tried enabling multithread support by porting this: https://github.com/emscripten-core/emscripten/issues/13892#issuecomment-835229727

However, it seems that my port is not well optimized and runs slower than the single threaded version. (the CPU utilization is not very high, perhaps most of the time is spent on copying data or something, I am not sure about this)

elalish:
I actually haven't done much with the OMP backend; I noticed a similar thing last time I tried it. I think I'm either using it wrong or Thrust has a bug, because I could not get any simple parallel algorithms to use more than one core. Anyway, probably best to table multi-threaded emscripten until we get that worked out in the basic build path.

pca006132:
I do get OMP to use all of my 16 cores, so should not be problems in this repo.

elalish:
I just tweaked the actions settings, which I'm hoping will mean they run for you now automatically.

.github/workflows/manifold.yml
```diff
@@ -21,25 +21,54 @@ jobs:
...
+    - uses: jwlawson/actions-setup-cmake@v1.12
+    - name: Build WASM
       run: |
-        mkdir buildCUDA
-        cd buildCUDA
-        cmake -DCMAKE_BUILD_TYPE=Release .. && make
+        patch third_party/assimp/code/CMakeLists.txt < assimp.diff
+        mkdir buildWASM
+        cd buildWASM
+        git clone https://github.com/emscripten-core/emsdk.git
```
elalish, line 37: Should this just be another git submodule for consistency?

.github/workflows/manifold.yml
```diff
@@ -21,25 +21,54 @@ jobs:
...
+        cmake -DCMAKE_BUILD_TYPE=Release -DTHRUST_BACKEND=CPP .. && make -j
+    - name: Build OMP
+      run: |
+        mkdir buildOMP
+        cd buildOMP
+        cmake -DCMAKE_BUILD_TYPE=Release -DTHRUST_BACKEND=OMP .. && make -j
+    - name: Test WASM
+      run: |
+        cd buildWASM/test
+        node manifold_test.js
```
elalish, line 57: :100:

.github/workflows/manifold.yml
```diff
@@ -21,25 +21,54 @@ jobs:
...
     # Checks-out your repository under $GITHUB_WORKSPACE, so your job can access it
-    - uses: actions/checkout@v2
-    - uses: jwlawson/actions-setup-cmake@v1.12
-    
     - name: Install dependencies
       run: |
         apt-get -y update
-        DEBIAN_FRONTEND=noninteractive apt install -y libglm-dev libassimp-dev libcgal-dev
-    - name: Build CUDA
+        DEBIAN_FRONTEND=noninteractive apt install -y libglm-dev libassimp-dev libcgal-dev libomp-dev nodejs git python3
```
elalish, line 27: If we're going to submodule these, should we skip installing them?

CMakeLists.txt
```diff
@@ -12,31 +9,43 @@ set(CMAKE_VERBOSE_MAKEFILE ON)
...
 
-OPTION( MANIFOLD_USE_OMP
-  "Use OpenMP as the Thrust backend instead of CUDA."
-  OFF
-)
-OPTION( MANIFOLD_USE_CPP
-  "Use C++ (single-threaded) as the Thrust backend instead of CUDA."
-  OFF
+set(THRUST_BACKEND "CUDA" CACHE STRING
+    "Thrust backend, either \"CUDA\", \"OMP\" or \"CPP\"(single-threaded)"
```
elalish, line 13: :+1:

CMakeLists.txt
```diff
@@ -12,31 +9,43 @@ set(CMAKE_VERBOSE_MAKEFILE ON)
...
-
-IF(MANIFOLD_USE_CPP)
-    message("------------------------- Using C++ instead of CUDA.")
-    set(MANIFOLD_NVCC_FLAGS ${MANIFOLD_NVCC_FLAGS} -DTHRUST_DEVICE_SYSTEM=THRUST_DEVICE_SYSTEM_CPP)
-ENDIF(MANIFOLD_USE_CPP)
+if(THRUST_BACKEND STREQUAL "CUDA")
+    enable_language(CUDA)
+else()
+    set(THRUST_INC_DIR ${PROJECT_SOURCE_DIR}/third_party/thrust)
+    set(GLM_INC_DIR ${PROJECT_SOURCE_DIR}/third_party/glm)
```
elalish, line 38: Again, might be nice to do this for all builds for consistency.

assimp.diff
```diff
@@ -0,0 +1,16 @@
...
++++ b/third_party/assimp/code/CMakeLists.txt
+@@ -1176,13 +1176,6 @@ ADD_LIBRARY(assimp::assimp ALIAS assimp)
+ 
+ TARGET_USE_COMMON_OUTPUT_DIRECTORY(assimp)
+ 
+-# enable warnings as errors ########################################
+-IF (MSVC)
+-  TARGET_COMPILE_OPTIONS(assimp PRIVATE /WX)
+-ELSE()
+-  TARGET_COMPILE_OPTIONS(assimp PRIVATE -Werror)
```
elalish, line 12: What was the problem here? Should we file a bug on Assimp?

test/CMakeLists.txt
```diff
@@ -11,6 +11,11 @@ target_compile_features(${PROJECT_NAME} PUBLIC cxx_std_14)
 add_test(test_all ${PROJECT_NAME})
 
 file(COPY data DESTINATION . FILES_MATCHING PATTERN "*.ply")
+if(EMSCRIPTEN)
+    set_target_properties(${PROJECT_NAME} PROPERTIES LINK_FLAGS 
+        "-s ASSERTIONS=1 -s DEMANGLE_SUPPORT=1 --preload-file data --bind")
```
elalish, line 16: Just for my own information, since I've never used emscripten, what do these options do? Are they all necessary to allow this to build?

collider/CMakeLists.txt
```diff
@@ -1,8 +1,21 @@
 project (collider)
 
-add_library(${PROJECT_NAME} src/collider.cu)
+add_library(${PROJECT_NAME} src/collider.cpp)
 
-set_property(TARGET ${PROJECT_NAME} PROPERTY CUDA_ARCHITECTURES 61)
+if(THRUST_BACKEND STREQUAL "CUDA")
+    set_source_files_properties(src/collider.cpp PROPERTIES LANGUAGE CUDA)
```
elalish, line 6: I guess this is the magic line that lets us avoid `.cu`? I wonder if we should rename the `.cuh` files to `.h` as well?

elalish (commented):
This is looking great! I haven't had time to pull and test it locally yet, but the fact it passes the CI gives me confidence.

.github/workflows/manifold.yml
pca006132: This is just for setting up the build environment, not build dependency. This is the official instruction for installing emscripten on Ubuntu, but other distros have their own packages and don't have to do this.

.github/workflows/manifold.yml
pca006132: We can do this for glm. For assimp, the build script will only compile the submodule if building with emscripten to minimize build time, but we can change that if you like.

assimp.diff
pca006132: https://github.com/pca006132/manifold/runs/6302903855?check_suite_focus=true#step:6:1258

Not sure why this happens in CI, I don't get this error when compiling it locally (but can reproduce when running it in the docker instance using the exact same image). I guess some code are probably conditionally compiled and are omitted in the CI environment and cause unused variables.

collider/CMakeLists.txt
pca006132: Yes, we probably should rename the `.cuh` files to `.h`.

test/CMakeLists.txt
pca006132: The first two options are for better diagnostics, or node will just say undefined error or something when it crashes.
The last argument `--preload-file data --bind` is to load the assets into [emscripten filesystem](https://emscripten.org/docs/porting/files/file_systems_overview.html#emscripten-file-system-runtime-environment) without accessing the host file system.

pca006132:
I am not sure why the build is failing, perhaps the linker flags are changed by the submodules?

elalish:
Yeah, I wonder. Though the error at least is pretty far along and sounds actionable? `/usr/bin/ld: ../third_party/graphlite/libgraphlite.a(connected_components.cpp.o): relocation R_X86_64_PC32 against symbol `_ZSt4cerr@@GLIBCXX_3.4' can not be used when making a shared object; recompile with -fPIC`

pca006132:
Yes, but I thought modern compilers should be compiling for PIC already. Anyway I can try to add that.
btw, I think we probably should only compile assimp when the users don't have assimp installed on their own machine? compiling assimp increases the build time significantly.

elalish:
Agreed, I just tried compiling your branch locally and it's a notable difference. I also found [this](https://stackoverflow.com/questions/13812185/how-to-recompile-with-fpic), regarding the build error (which I also get on my machine). Hoping that helps?

pca006132:
Thanks, will have a look at that and perhaps setup a docker and test.

elalish:
Oh, interesting; on my second build attempt, it worked. Tests pass fine. I wonder if it's an order-of-operations thing in the build process?

pca006132:
My guess is probably this line: https://github.com/elalish/manifold/pull/103/files#diff-e6cb47aeb84b5b51a4b903ba3a9c276d9d75c9702586e41f185fc147e2ff3f00R6
But will need to get a container working to see if this is the cause of the problem, as I cannot get the build failure after a clean build.

.github/workflows/manifold.yml
```diff
@@ -21,25 +21,56 @@ jobs:
...
     # Checks-out your repository under $GITHUB_WORKSPACE, so your job can access it
-    - uses: actions/checkout@v2
-    - uses: jwlawson/actions-setup-cmake@v1.12
-    
     - name: Install dependencies
       run: |
         apt-get -y update
-        DEBIAN_FRONTEND=noninteractive apt install -y libglm-dev libassimp-dev libcgal-dev
-    - name: Build CUDA
+        DEBIAN_FRONTEND=noninteractive apt install -y libcgal-dev libomp-dev nodejs git python3
```
elalish, line 27: As I'm looking at this more, I'm thinking maybe we'd be better off with some of these builds as separate actions, rather than separate jobs. The WASM build is the most obvious example to me; no need for CUDA, but requires nodejs, etc. I might even split out the tools build that uses CGAL, since that's unrelated to this library. 

This way, when someone is looking to build, they have concise instructions for only their situation. Also, we'll get feedback on their failures separately in Actions. Anyway, this doesn't necessarily have to happen right now, just something to think about.

assimp.diff
elalish: I got this error locally until I applied the patch, FWIW.

test/mesh_test.cpp
```diff
@@ -144,6 +144,8 @@ TEST(Manifold, GetMesh) {
   Identical(mesh_out, mesh_out2);
 }
 
+// temporarily disabled for wasm due to test failure
+#ifndef __EMSCRIPTEN__
```
elalish, line 148: Let's just disable this one generally: rename it `DISABLED_Determinism`. It's not even consistent between CUDA and CPP, so it's unfair to blame WASM.

pca006132:
The build error should be fixed now.

pca006132:
They will build in parallel now, but I use the same set of dependencies for all of them to make the action script simpler. I think we should probably put the build instructions in README rather than letting users check the CI file, so they won't miss things like applying the assimp patch.

I tried building for windows but it is a painful experience with weird errors like https://github.com/pca006132/manifold/runs/6341312236?check_suite_focus=true#step:8:2399  so I just disabled it for now. Perhaps try to experiment with it later...

manifold/src/boolean3.cpp
```diff
@@ -325,6 +325,7 @@ struct Kernel02 {
       if (forward) {
         if (!Shadows(vertPos.z, z02, expandP * vertNormalP[p0].z)) s02 = 0;
       } else {
+        ALWAYS_ASSERT(closestVert != -1, topologyErr, "No closest vert");
```
elalish, line 328: So, the CI can't test CUDA since it doesn't have a GPU, so I tested it and it failed. I took a guess and commented out this line and it works. Personally, I think that's a failure of the CUDA compiler, since this line shouldn't do anything, but I think the problem stems from the fact that you can't throw an exception from inside a GPU kernel. Why it lets it compile, then totally breaks when run, I don't understand.

elalish: You'll notice I have occasional `printf` lines in these where I ought to have asserts; that's why.

manifold/src/boolean3.cpp
```diff
@@ -277,8 +277,8 @@ struct Kernel02 {
     glm::vec3 yzzRL[2];
     // Either the left or right must shadow, but not both. This ensures the
     // intersection is between the left and right.
-    bool shadows;
-    int closestVert;
+    bool shadows = false;
+    int closestVert = -1;
```
elalish, line 281: Thanks for the code review!

.github/workflows/manifold.yml
```diff
@@ -10,36 +10,109 @@ on:
...
       run: |
         apt-get -y update
-        DEBIAN_FRONTEND=noninteractive apt install -y libglm-dev libassimp-dev libcgal-dev
-    - name: Build CUDA
-      run: |
-        mkdir buildCUDA
-        cd buildCUDA
-        cmake -DCMAKE_BUILD_TYPE=Release .. && make
-    - name: Build CPP
+        DEBIAN_FRONTEND=noninteractive apt install -y libomp-dev git
```
elalish, line 25: Shall we go ahead and install assimp for these builds instead of using the submodule to keep our build times down, like you suggested?

elalish (changes requested):
This looks great; once my CUDA tests pass, I'm happy to merge this. Should be just that one line.

manifold/src/boolean3.cpp
pca006132: Shall we use this instead? https://docs.nvidia.com/cuda/cuda-c-programming-guide/index.html#assertion

elalish: Yes, probably so. But let's save that for another PR since it's in a bunch of places. Also, I should probably make that change, since I can test it. I can't remember if there was a reason I'm not already using that.

pca006132: OK I will just comment this out for now then.

.github/workflows/manifold.yml
pca006132: Just tried using `libassimp-dev`. It seems that the version is a bit outdated and does not contain `assimp/GltfMaterial.h`. I guess we can just stick to the submodule one for now. As the CI jobs are parallelized, the CI build takes about 20 minutes instead of about an hour previously, it should be a bit more acceptable.

elalish (approved):
Thanks, awesome work!

manifold/src/boolean3.cpp
elalish: I gave `assert` a try and now I remember where I got stuck. I can't seem to make Thrust catch the cuda error and throw an exception like I would expect. I can't find any documentation on it, so I wonder if it was never implemented?

Issue #105, opened 2022-05-07
Performance tracking


pca006132:
As this library is intended to be fast, I guess we should probably open an issue to track its performance?

I did some microbenchmark using the `perfTest` binary and a `unionPerfTest` code that I wrote to test lazy union (union 100 spheres with diameter=2.5 with varying separation distance, i.e. may or may not overlap). I've tested single thread using CPP backend, OpenMP backend and CUDA backend all on the same laptop with i5-8300H and GTX1050 Mobile. Here is my [spreadsheet](https://docs.google.com/spreadsheets/d/1KevjOz21uMXkYQQnsoYhkI_xvxR_GQ2yyrGaxV2B628/edit?usp=sharing), and here is my [branch](https://github.com/pca006132/manifold/tree/opt) for some optimizations that looks quite effective for CUDA and small meshes :). I will open a separate PR for the branch, after the build script PR is merged.

From the results, we can see that:

1. For small meshes, OMP > CPP > CUDA. Wondering if we can run the thrust APIs on OMP for operations involving only < 1k vertices or something.
2. For more complicated models, using the backend GPU may exhaust GPU memory. Is it possible to fallback to CPP mode after that?
3. We have to be careful about copying vector, eliminating unnecessary copying reduces the execution time for CUDA backend significantly (-30% in the microbenchmark). One way would be to disable copy constructor for `VecDH` and provide an explicit copy method.
4. It seems that the multithread cannot help much in our case, even though my computer has 4 cores 8 threads. Perhaps we should look into this because quite a lot of users may not have a GPU with CUDA (my main laptop is using AMD, so no CUDA here).

elalish:
I'm open to these optimizations, but I would recommend avoiding too much code complexity, especially as related to Thrust. I'd prefer to switch from Thrust to C++17 parallel algorithms, which will allow us to remove the `VecDH` class entirely and let the compiler handle these copies/moves internally. Hopefully the compilers might even take care of switching backends based on data size? I'm also hoping we'll get compilers that can target AMD GPUs from the parallel STL, though I'm not sure how far out that is.

pca006132:
The main performance improvement comes from these two patches: 168fd07106e5a58f739e535308e120895fcf390f and a3dbe154f0c11b2f85057a26af028734ef6f4aaf , which are not thrust specific, so this should be fine. Will submit a PR soon.

[...]

PR #106, opened 2022-05-09
Optimizations

pca006132:
optimizations as mentioned in #105

### DIFF
--- a/collider/src/collider.cpp
+++ b/collider/src/collider.cpp
@@ -266,7 +266,7 @@ Collider::Collider(const VecDH<Box>& leafBB,
  */
 template <typename T>
 SparseIndices Collider::Collisions(const VecDH<T>& querriesIn) const {
-  int maxOverlaps = 1 << 20;
+  int maxOverlaps = querriesIn.size() * 4;
   SparseIndices querryTri(maxOverlaps);
   int nOverlaps = 0;
   while (1) {
--- a/collider/src/collider.cpp
+++ b/collider/src/collider.cpp
@@ -282,7 +282,7 @@ SparseIndices Collider::Collisions(const VecDH<T>& querriesIn) const {
       break;
     else {  // if not enough memory was allocated, guess how much will be needed
       int lastQuery = querryTri.Get(0).H().back();
-      maxOverlaps *= 2 * static_cast<float>(querriesIn.size()) / lastQuery;
+      maxOverlaps *= 2;
       querryTri.Resize(maxOverlaps);
     }
   }
--- a/collider/src/collider.cpp
+++ b/collider/src/collider.cpp
@@ -282,7 +282,11 @@ SparseIndices Collider::Collisions(const VecDH<T>& querriesIn) const {
       break;
     else {  // if not enough memory was allocated, guess how much will be needed
       int lastQuery = querryTri.Get(0).H().back();
-      maxOverlaps *= 2;
+      float ratio = static_cast<float>(querriesIn.size()) / lastQuery;
+      if (ratio > 1000) // do not trust the ratio if it is too large
+        maxOverlaps *= 2;
+      else
+        maxOverlaps *= 2 * ratio;
       querryTri.Resize(maxOverlaps);
     }
   }
--- a/manifold/src/face_op.cpp
+++ b/manifold/src/face_op.cpp
@@ -53,6 +53,13 @@ void Manifold::Impl::Face2Tri(const VecDH<int>& faceEdge,
     ALWAYS_ASSERT(numEdge >= 3, topologyErr, "face has less than three edges.");
     const glm::vec3 normal = faceNormal[face];
 
+    auto linearSearch = [](const int* mapping, int value) {
+      int i = 0;
+      while (mapping[i] != value)
+        ++i;
+      return i;
+    };
+
     if (numEdge == 3) {  // Single triangle
       int mapping[3] = {halfedge[firstEdge].startVert,
                         halfedge[firstEdge + 1].startVert,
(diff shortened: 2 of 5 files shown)


manifold/src/face_op.cpp
```diff
@@ -50,12 +53,10 @@ void Manifold::Impl::Face2Tri(const VecDH<int>& faceEdge,
     ALWAYS_ASSERT(numEdge >= 3, topologyErr, "face has less than three edges.");
     const glm::vec3 normal = faceNormal[face];
 
-    std::map<int, int> vertBary;
-    for (int j = firstEdge; j < lastEdge; ++j)
-      vertBary[halfedge[j].startVert] = halfedgeBary.H()[j];
```
elalish, line 55: Good call, I bet this could be a big part of why the triangulation step has been scaling worse than it should.

manifold/src/face_op.cpp
```diff
@@ -115,9 +127,27 @@ void Manifold::Impl::Face2Tri(const VecDH<int>& faceEdge,
...
+        while (mapping[index] != tri0[k])
+          index += 1;
+        meshRelation_.triBary.H().back().vertBary[k] = halfedgeBary.H()[firstEdge + index];
+      }
+      meshRelation_.triBary.H().push_back(faceRef.H()[face]);
+      for (int k : {0, 1, 2}) {
+        int index = 0;
+        while (mapping[index] != tri1[k])
+          index += 1;
+        meshRelation_.triBary.H().back().vertBary[k] = halfedgeBary.H()[firstEdge + index];
```
elalish, line 142: Perhaps we could make a function out of this block to avoid some code repetition?

manifold/src/impl.cpp
```diff
@@ -507,17 +507,19 @@ void Manifold::Impl::CreateHalfedges(const VecDH<glm::ivec3>& triVerts) {
...
+  thrust::for_each_n(zip(countAt(0), triVerts.beginD()), numTri,
+                     Tri2Halfedges({halfedge_.ptrD(), edge.ptrD()}));
   // Stable sort is required here so that halfedges from the same face are
   // paired together (the triangles were created in face order). In some
   // degenerate situations the triangulator can add the same internal edge in
   // two different faces, causing this edge to not be 2-manifold. We detect this
   // and fix it by swapping one of the identical edges, so it is important that
   // we have the edges paired according to their face.
-  std::stable_sort(edge.begin(), edge.end());
+  thrust::stable_sort(edge.beginD(), edge.endD());
```
elalish, line 522: Is this new? I vaguely recall using `std::` because thrust didn't have a stable_sort.

utilities/include/sparse.h
```diff
@@ -161,6 +178,10 @@ class SparseIndices {
   }
 
  private:
-  VecDH<int> p, q;
+  VecDH<uninitialized_int> p, q;
+  VecDH<int>& GetP() { return reinterpret_cast<VecDH<int>&>(p); }
+  VecDH<int>& GetQ() { return reinterpret_cast<VecDH<int>&>(q); }
+  const VecDH<int>& GetP() const { return reinterpret_cast<const VecDH<int>&>(p); }
+  const VecDH<int>& GetQ() const { return reinterpret_cast<const VecDH<int>&>(q); }
```
elalish, line 185: I guess these casts what the compiler is angry about?

elalish (commented):
Looking great, thanks! Would you mind including a link to your performance spreadsheet in this PR (ideally updated once it's finished)? 30% improvement on CUDA, especially for the larger test cases is quite impressive indeed!

manifold/src/impl.cpp
pca006132: From git blame of https://github.com/NVIDIA/thrust/blame/fa54f2c6f1217237953f27ddf67f901b6b34fbdd/testing/stable_sort.cu, it seems that `stable_sort` is in thrust for over 10 years. Perhaps it is related to cudatoolkit version?

utilities/include/sparse.h
pca006132: Yes it seems so, interestingly clang++ 13 did not complain about it (I use clang because my openmp is the one from LLVM).
I was looking at other ways to do this, like union/overriding vector allocator, but it seems that I need to cast the vector type anyway. Not sure if there is any idiomatic way of doing this in C++.

pca006132: It seems that the slowness due to initialization is caused by a large initial size for sparse indices. I've reduced the initial size and fixed the growth rate of it, so this change can be dropped.

collider/src/collider.cpp
```diff
@@ -266,7 +266,7 @@ Collider::Collider(const VecDH<Box>& leafBB,
  */
 template <typename T>
 SparseIndices Collider::Collisions(const VecDH<T>& querriesIn) const {
-  int maxOverlaps = 1 << 20;
+  int maxOverlaps = querriesIn.size() * 4;
```
elalish, line 269: :+1:

elalish (approved):
Really love how much simpler this is! And `perfTest` for CUDA is indeed running ~30% faster for me: 
```
nTri = 512, time = 0.00647049 sec
nTri = 2048, time = 0.00899614 sec
nTri = 8192, time = 0.0170416 sec
nTri = 32768, time = 0.0450551 sec
nTri = 131072, time = 0.144635 sec
nTri = 524288, time = 0.524747 sec
nTri = 2097152, time = 1.89367 sec
nTri = 8388608, time = 7.68228 sec
```

collider/src/collider.cpp
```diff
@@ -282,7 +282,7 @@ SparseIndices Collider::Collisions(const VecDH<T>& querriesIn) const {
       break;
     else {  // if not enough memory was allocated, guess how much will be needed
       int lastQuery = querryTri.Get(0).H().back();
-      maxOverlaps *= 2 * static_cast<float>(querriesIn.size()) / lastQuery;
```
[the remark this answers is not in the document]
pca006132: I guess I should probably change this to long. The size overflowed to negative for large model (sponge4 in the test).

collider/src/collider.cpp
```diff
@@ -282,7 +282,11 @@ SparseIndices Collider::Collisions(const VecDH<T>& querriesIn) const {
       break;
     else {  // if not enough memory was allocated, guess how much will be needed
       int lastQuery = querryTri.Get(0).H().back();
-      maxOverlaps *= 2;
+      float ratio = static_cast<float>(querriesIn.size()) / lastQuery;
+      if (ratio > 1000) // do not trust the ratio if it is too large
```
elalish, line 286: :+1:

elalish (approved):
Thanks for the cleanup!

PR #111, opened 2022-05-11
windows compatibility

pca006132:
Tries fixing windows build error. It seems that msvc doesn't really like [alternative tokens](https://stackoverflow.com/questions/24414124/why-does-vs-not-define-the-alternative-tokens-for-logical-operators) and [getopt](https://stackoverflow.com/questions/10404448/getopt-h-compiling-linux-c-code-in-windows).

### DIFF
--- a/.github/workflows/manifold.yml
+++ b/.github/workflows/manifold.yml
@@ -71,48 +71,47 @@ jobs:
         cd build/test
         node ./manifold_test.js
 
-  # unable to get it working for now
-  # build_windows:
-  #   strategy:
-  #     matrix:
-  #       backend: [CPP, CUDA]
-  #   runs-on: windows-2019
-  #   steps:
-  #   - uses: actions/checkout@v3
-  #     with:
-  #       submodules: true
-  #   - uses: jwlawson/actions-setup-cmake@v1.12
-  #   - name: Patch assimp
-  #     shell: bash
-  #     run: patch third_party/assimp/code/CMakeLists.txt < assimp.diff
-  #   - uses: ilammy/msvc-dev-cmd@v1
-  #   - name: Install CUDA
-  #     if: matrix.backend == 'CUDA'
-  #     env: 
-  #       cuda: "11.3.0"
-  #       visual_studio: "Visual Studio 16 2019"
-  #     shell: powershell
-  #     run: |
-  #       Invoke-RestMethod https://raw.githubusercontent.com/ptheywood/cuda-cmake-github-actions/21811d232fc775fa8d44dddf7f8c5ce2cf8d2b81/scripts/actions/install_cuda_windows.ps1 -OutFile install_cuda_windows.ps1
-  #       .\install_cuda_windows.ps1
-  #   - name: nvcc check
-  #     if: matrix.backend == 'CUDA'
-  #     shell: powershell
-  #     run: |
-  #       nvcc -V
-  #       ls $env:CUDA_PATH
-  #       ls $env:CUDA_PATH\bin
-  #       ls $env:CUDA_PATH\include
-  #   - name: Build ${{matrix.backend}}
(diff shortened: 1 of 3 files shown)


test/test_main.cpp
```diff
@@ -12,39 +12,45 @@
...
 #include "test.h"
 
 Options options;
 
+void print_usage() {
+  printf("-------------------------------\n");
+  printf("manifold_test specific options:\n");
+  printf("  -h: Print this message\n");
+  printf("  -e: Export sample models\n");
+  printf("  -v: Enable verbose output\n");
```
elalish, line 24: Thanks!

third_party/graphlite/include/graph_lite.h
```diff
@@ -69,7 +69,7 @@ struct is_unordered_map<
 template <typename T>
 constexpr bool is_unordered_map_v = is_unordered_map<T>::value;
 template <typename T>
-constexpr bool is_either_map_v = is_map_v<T> or is_unordered_map_v<T>;
+constexpr bool is_either_map_v = is_map_v<T> || is_unordered_map_v<T>;
```
elalish, line 72: I was considering making this a submodule, but maybe we should leave it as is so we can make these fixes...

.github/workflows/manifold.yml
```diff
@@ -71,48 +71,47 @@ jobs:
...
-  #     if: matrix.backend != 'CUDA'
-  #     shell: bash
-  #     run: |
-  #       cd build\test
-  #       .\manifold_test.exe
-  #
+  build_windows:
+    strategy:
+      matrix:
+        backend: [CPP, CUDA]
```
elalish, line 77: Maybe do CUDA first so we can see if it builds before any CPP tests fail?

elalish (commented):
Yeah, that error is quite inscrutable. What compiler is it using? It sort of seems like that should be specified in the yml file somewhere, though I suppose the machine has a default?

elalish:
It almost looks like it's printing the environment info, but not the actual `std::out`. I wonder if there's some flag necessary to get the actual console output? Google_test doesn't generally error out without printing at least *something*.

pca006132:
It is using msvc, installed with `ilammy/msvc-dev-cmd@v1`. The error is very interesting because I tried building it on my windows laptop and it successfully builds and run the tests previously, perhaps some issue with c++ runtime or something? Quite hard to debug this as the CI is pretty much a blackbox to me...

.github/workflows/manifold.yml
pca006132: It should now do CUDA first

pca006132:
Wondering if we should disable `/Wall` (`-Wall` for msvc) considering msvc is emitting so many warnings...

pca006132:
Test passed for the CPP backend on Windows. For some reason the CUDA backend compilation will just exit without any error messages making it nearly impossible to debug, so I just disabled the CUDA build on it for now.

Added support for building with Nix, which allows us to pin the version of each dependencies (nodejs, emscripten, openmp, g++, etc.), hopefully this can make debugging CI failures on Linux a bit less painful as we can reproduce this locally.

.github/workflows/manifold.yml
```diff
@@ -71,48 +71,59 @@ jobs:
...
+
+  build_nix:
+    strategy:
+      matrix:
+        variant: [cpp, omp, cuda, js]
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v2.4.0
+    - uses: cachix/install-nix-action@v15
+    - run: nix build -L '.?submodules=1#manifold-${{matrix.variant}}'
```
elalish, line 128: Cool, I hadn't heard of nix before, but reading about it I already like it.

elalish (approved):
Looking great, thanks!

PR #113, opened 2022-05-12
python binding

pca006132:
Implemented a simple python binding and ported 2 examples from https://gist.github.com/ochafik/2db96400e3c1f73558fcede990b8a355

### DIFF
--- a/.github/workflows/manifold.yml
+++ b/.github/workflows/manifold.yml
@@ -37,8 +37,10 @@ jobs:
     - name: Test ${{matrix.backend}}
       if: matrix.backend != 'CUDA'
       run: |
+        export PYTHONHOME=$PYTHONHOME:$(pwd)/build/tools
         cd build/test
         ./manifold_test
+        python3 python/run_all.py
 
   build_wasm:
     runs-on: ubuntu-20.04
--- a/.github/workflows/manifold.yml
+++ b/.github/workflows/manifold.yml
@@ -22,7 +22,7 @@ jobs:
     - name: Install dependencies
       run: |
         apt-get -y update
-        DEBIAN_FRONTEND=noninteractive apt install -y libomp-dev git
+        DEBIAN_FRONTEND=noninteractive apt install -y libomp-dev git python3 python3-distutils
     - uses: actions/checkout@v3
       with:
         submodules: true
--- a/.github/workflows/manifold.yml
+++ b/.github/workflows/manifold.yml
@@ -23,7 +23,7 @@ jobs:
     - name: Install dependencies
       run: |
         apt-get -y update
-        DEBIAN_FRONTEND=noninteractive apt install -y libomp-dev git libtbb-dev pkg-config python3 python3-distutils
+        DEBIAN_FRONTEND=noninteractive apt install -y libomp-dev git libtbb-dev pkg-config libpython3-dev python3 python3-distutils
     - uses: actions/checkout@v3
       with:
         submodules: true
@@ -34,15 +34,16 @@ jobs:
         patch third_party/assimp/code/CMakeLists.txt < assimp.diff
         mkdir build
         cd build
-        cmake -DCMAKE_BUILD_TYPE=Release -DMANIFOLD_PAR=${{matrix.parallel_backend}} -DMANIFOLD_USE_CUDA=${{matrix.cuda_support}} .. && make
+        cmake -DCMAKE_BUILD_TYPE=Release -DBUILD_SHARED_LIBS=ON -DMANIFOLD_PAR=${{matrix.parallel_backend}} -DMANIFOLD_USE_CUDA=${{matrix.cuda_support}} .. && make
     - name: Test ${{matrix.parallel_backend}} with CUDA ${{matrix.cuda_support}}
       # note that the test for CUDA backend does not really test CUDA, as we
       # don't have CUDA GPU on GitHub Action
       run: |
-        export PYTHONHOME=$PYTHONHOME:$(pwd)/build/tools
+        export PYTHONPATH=$PYTHONPATH:$(pwd)/build/tools
         cd build/test
         ./manifold_test
-        python3 python/run_all.py
+        cd ../../
+        python3 test/python/run_all.py
 
   build_wasm:
     runs-on: ubuntu-20.04
--- a/test/python/run_all.py
+++ b/test/python/run_all.py
@@ -1,0 +1,22 @@
+import pathlib
+import sys
+import importlib
+from time import time
+
+if __name__ == "__main__":
+    current_file = pathlib.Path(__file__)
+    current_dir = current_file.parent
+    files = [f.parts[-1][:-3]
+             for f in current_dir.glob('*.py') if f != current_file]
+
+    export_models = len(sys.argv) == 2 and sys.argv[-1] == '-e'
+
+    for f in files:
+        module = importlib.import_module(f)
+        t0 = time()
+        model = module.run()
+        if export_models:
+            model.export(f'{f}.glb')
+            print(f'Exported model to {f}.glb')
+        t1 = time()
+        print(f'Took {(t1-t0)*1000:.1f}ms for {f}')
--- a/tools/pymanifold.cpp
+++ b/tools/pymanifold.cpp
@@ -55,10 +55,10 @@ PYBIND11_MODULE(pymanifold, m) {
            })
       .def(
           "translate",
-          [](Manifold self, float x, float y, float z = 0.0f) {
+          [](Manifold self, float x = 0.0f, float y = 0.0f, float z = 0.0f) {
             return self.Translate(glm::vec3(x, y, z));
           },
-          py::arg("x"), py::arg("y"), py::arg("z") = 0.0f)
+          py::arg("x") = 0.0f, py::arg("y") = 0.0f, py::arg("z") = 0.0f)
       .def("scale", static_cast<Manifold (*)(Manifold, float)>(
                         [](Manifold self, float scale) {
                           return self.Scale(glm::vec3(scale));
@@ -73,9 +73,9 @@ PYBIND11_MODULE(pymanifold, m) {
                           return self.Scale(v);
                         }), py::arg("v"))
       .def("rotate",
-           [](Manifold self, float xDegrees, float yDegrees, float zDegrees) {
+           [](Manifold self, float xDegrees = 0.0f, float yDegrees = 0.0f, float zDegrees = 0.0f) {
              return self.Rotate(xDegrees, yDegrees, zDegrees);
-           }, py::arg("xDegrees"), py::arg("yDegrees"), py::arg("zDegrees"))
+           }, py::arg("xDegrees") = 0.0f, py::arg("yDegrees") = 0.0f, py::arg("zDegrees") = 0.0f)
       .def("export",
            [](Manifold &self, std::string name) { exportManifold(self, name); },
         py::arg("filename"),
--- a/tools/pymanifold.cpp
+++ b/tools/pymanifold.cpp
@@ -39,7 +39,7 @@ PYBIND11_MODULE(pymanifold, m) {
       }), "Construct manifold as the union of a set of manifolds.")
       .def(py::self + py::self)
       .def(py::self - py::self)
-      .def("__and__", [](Manifold &a, Manifold &b) { return a ^ b; })
+      .def(py::self ^ py::self)
       .def("transform",
            [](Manifold &self, py::array_t<float> &mat) {
              auto mat_view = mat.unchecked<2>();
--- a/tools/pymanifold.cpp
+++ b/tools/pymanifold.cpp
@@ -41,7 +41,7 @@ PYBIND11_MODULE(pymanifold, m) {
       .def(py::self - py::self)
       .def(py::self ^ py::self)
       .def("transform",
-           [](Manifold &self, py::array_t<float> &mat) {
+           [](Manifold self, py::array_t<float> &mat) {
              auto mat_view = mat.unchecked<2>();
              if (mat_view.shape(0) != 3 || mat_view.shape(1) != 4)
                throw std::runtime_error("Invalid matrix shape");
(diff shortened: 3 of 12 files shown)


pca006132:
This binding is a bit less efficient than the C++ API. Apart from python interpreter overhead, functions that were operating inplace, such as the various transformations, will now clone the manifold to make the API more functional, similar to OpenSCAD.

tools/pymanifold.cpp
```diff
@@ -0,0 +1,169 @@
...
+                          if (scale_view.shape(0) != 3)
+                            throw std::runtime_error("Invalid vector shape");
+                          glm::vec3 v(scale_view(0), scale_view(1),
+                                      scale_view(2));
+                          return self.Scale(v);
+                        }), py::arg("v"))
+      .def("rotate",
+           [](Manifold self, float xDegrees, float yDegrees, float zDegrees) {
+             return self.Rotate(xDegrees, yDegrees, zDegrees);
+           }, py::arg("xDegrees"), py::arg("yDegrees"), py::arg("zDegrees"))
```
elalish, line 78: Should these have `= 0` so that they don't all need to be specified?

tools/pymanifold.cpp
```diff
@@ -0,0 +1,169 @@
...
+          },
+          py::arg("height"), py::arg("radiusLow"),
+          py::arg("radiusHigh") = -1.0f, py::arg("circularSegments") = 0)
+      .def_static(
+          "Sphere",
+          [](float radius,
+             int circularSegments =
+                 0) { return Manifold::Sphere(radius, circularSegments); },
+          py::arg("radius"), py::arg("circularSegments") = 0);
+  ;
```
elalish, line 134: intentional?

tools/pymanifold.cpp
```diff
@@ -0,0 +1,169 @@
...
+                {std::get<0>(polygons[i][j]), std::get<1>(polygons[i][j])}, j};
+          }
+          simplePolygons[i] = {vertices};
+        }
+        return PolygonsWrapper{std::make_unique<Polygons>(simplePolygons)};
+      }), py::arg("polygons"),
+      "Construct the Polygons object from a list of simple polygon, "
+      "where each simple polygon is a list of points (pair of floats).")
+      .def(
+          "extrude",
```
elalish, line 152: :+1:

tools/pymanifold.cpp
```diff
@@ -0,0 +1,169 @@
...
+typedef std::tuple<float, float> Float2;
+typedef std::tuple<float, float, float> Float3;
+
+struct PolygonsWrapper {
+  std::unique_ptr<Polygons> polygons;
+};
+
+PYBIND11_MODULE(pymanifold, m) {
+  m.doc() = "Python binding for the manifold library. Please check the C++ documentation for APIs.\n"
+      "This binding will perform copying to make the API more familiar to OpenSCAD users.";
```
elalish, line 31: Can we talk about this in a little more depth? As an OpenSCAD user, I'm not sure this applies, since the OpenSCAD language is so different (yes it's functional, but it's also different in lots of other ways). Do you have a sense of how performance-impacting this decision is? Are there other considerations impacting this choice? 

It would be great to have a cross-compiler that took OpenSCAD language and converted it to an equivalent python or JS program using Manifold, but that seems like the right layer at which to account for the functional nature of OpenSCAD's language, rather than our bindings themselves. What do you think?

tools/pymanifold.cpp
```diff
@@ -0,0 +1,169 @@
...
+      .def(py::init<>())
+      .def(py::init([](std::vector<Manifold> &manifolds) {
+        Manifold result;
+        for (Manifold &manifold : manifolds)
+          result += manifold;
+        return result;
+      }), "Construct manifold as the union of a set of manifolds.")
+      .def(py::self + py::self)
+      .def(py::self - py::self)
+      .def("__and__", [](Manifold &a, Manifold &b) { return a ^ b; })
```
elalish, line 42: I take it python doesn't let us overload the `^` operator?

test/python/run_all.py
```diff
@@ -0,0 +1,22 @@
...
+
+    for f in files:
+        module = importlib.import_module(f)
+        t0 = time()
+        model = module.run()
+        if export_models:
+            model.export(f'{f}.glb')
+            print(f'Exported model to {f}.glb')
+        t1 = time()
+        print(f'Took {(t1-t0)*1000:.1f}ms for {f}')
```
elalish, line 22: Thanks for bringing in these tests!

.github/workflows/manifold.yml
```diff
@@ -33,12 +33,15 @@ jobs:
         patch third_party/assimp/code/CMakeLists.txt < assimp.diff
         mkdir build
         cd build
-        cmake -DCMAKE_BUILD_TYPE=Release -DTHRUST_BACKEND=${{matrix.backend}} .. && make
+        echo Python3 Path $(which python3)
+        cmake -DCMAKE_BUILD_TYPE=Release -DTHRUST_BACKEND=${{matrix.backend}} -DPYTHON_EXECUTABLE=$(which python3) .. && make
```
elalish, line 37: Looks like the cmake step is having trouble. Maybe it needs some kind of `FindPython` script in the CMakeLists.txt?

tools/pymanifold.cpp
```diff
@@ -0,0 +1,169 @@
...
+
+using namespace manifold;
+
+void exportManifold(Manifold &m, std::string name) {
+  Mesh out = m.GetMesh();
+  manifold::ExportOptions options;
+  options.faceted = true;
+  options.mat.roughness = 0.2;
+  options.mat.metalness = 0.0;
+  manifold::ExportMesh(name, out, options);
```
elalish, line 19: Of course this is only a subset of our functionality (no MeshRelation for textures and such), but it does seem like a good, simple place to start.

elalish (commented):
Thanks, this is looking promising!

tools/pymanifold.cpp
pca006132: yes, will change

tools/pymanifold.cpp
pca006132: It allows us to overload that, but I thought intersection is more similar to and, and was using this in my previous code and forgot to change. Will change that back to `^`.

tools/pymanifold.cpp
pca006132: Regarding performance, I did some simple tests with a few examples and they are all faster (~5x faster on my machihne) than OpenSCAD with fast-csg enabled, at least according to the figures in https://gist.github.com/ochafik/2db96400e3c1f73558fcede990b8a355. You can try running those examples and check the performance.

For more complicated models, I think we should do something like lazy union, which tries to compose most of the manifolds that do not intersect each other. From my benchmark with the [bricks](https://gist.github.com/deckar01/ef11def51de7e71d9f288c6e5819fdb7) example, lazy union is a lot faster for larger models (took about 8s for width=height=length=100). The lazy union feature will also share the underlying list of manifolds, so objects that only differ in transformation will be lazily cloned. When that feature is added, this copying will basically be free.

If we don't do copying, users will have to manually copy the manifolds via constructor/copy function call, which may not be very ergonomic to use. But I agree, maybe I should split it into a low-level binding and a user facing module that provides this kind of features.

elalish: @rsaccon Since you seem to another interested Python user, would you care to weigh in here, regarding ergonomics and performance?

.github/workflows/manifold.yml
pca006132: I guess CMake is detecting a wrong version of Python here, which is a common problem: https://pybind11.readthedocs.io/en/stable/faq.html#cmake-doesn-t-detect-the-right-python-version

Will need some trial and error to fix this in the CI.

tools/pymanifold.cpp
pca006132: Another consideration for this: If copying causes performance problem, it should not be too hard for us to modify the c++ code to add more laziness to the module using some sort of wrapper, even if lazy union is not yet implemented. But if users depend on the API that will cause mutation for the manifolds, it would be hard for us to apply optimizations later and changing such behavior will be a breaking change.

rsaccon: I have only little experience with Python (I wrote a Freecad addon a while ago), but I have a very specific use case why I need  Python bindings: I wanna give it a try to write a Blender addon which will provide a Manifold based way to create precise shapes. So my wish list is: The bindings should be minimal (full manifold core API, but option to skip assimp). I played with the first lazy union based approach of the bindings (from @pca006132 opt branch), worked great, but the name `ManifoldSet` is or was a bit confusing. For me, performance is very important, and lazy union seems a way to get us there, but to provide some valid feedback I need first to play around a bit more with whatever we have (and best option right now is this WIP PR)

elalish: Agreed about separating out `meshIO` (the only part with the assimp dependency). I'm not super familiar with Python bindings, but it does seem like having two: one for core Manifold and a separate one for MeshIO would be a good idea. Likewise for WASM. I'll likely use WASM with three.js and I'd prefer to use their import/export capabilities and keep the bundle size down.

tools/pymanifold.cpp
```diff
@@ -0,0 +1,169 @@
...
+        Manifold result;
+        for (Manifold &manifold : manifolds)
+          result += manifold;
+        return result;
+      }), "Construct manifold as the union of a set of manifolds.")
+      .def(py::self + py::self)
+      .def(py::self - py::self)
+      .def(py::self ^ py::self)
+      .def("transform",
+           [](Manifold &self, py::array_t<float> &mat) {
```
elalish, line 44: I think this mean transform is still in place, rather than doing an implicit copy, am I right?

tools/pymanifold.cpp
```diff
@@ -0,0 +1,169 @@
...
+             for (int i = 0; i < 3; i++) {
+               for (int j = 0; j < 4; j++) {
+                 mat_glm[j][i] = mat_view(i, j);
+               }
+             }
+             return self.Transform(mat_glm);
+           })
+      .def(
+          "translate",
+          [](Manifold self, float x = 0.0f, float y = 0.0f, float z = 0.0f) {
```
elalish, line 58: Whereas `translate`, `scale`, `rotate` are doing a copy now? I built Manifold to do these operations lazily already, so you could string many together and it would apply them all at once only when needed. I feel like doing copies between each operation, especially when there could be a long string of them before the next Boolean, would be a pretty major drag on performance.

tools/pymanifold.cpp
```diff
@@ -0,0 +1,169 @@
...
+             return self.Warp([&f](glm::vec3 &v) {
+               Float3 fv = f(std::make_tuple(v.x, v.y, v.z));
+               v.x = std::get<0>(fv);
+               v.y = std::get<1>(fv);
+               v.z = std::get<2>(fv);
+             });
+           }, py::arg("f"))
+      .def("refine", [](Manifold self, int n) { return self.Refine(n); }, py::arg("n"))
+      .def_static(
+          "FromMesh",
```
elalish, line 94: Seems like we should add a "ToMesh" method for symmetry. This will allow the bindings to work with the likes of Blender. Still, they'll also need the `MeshRelation` and the rest of the API in order to manage their texture coordinates and material properties.

tools/pymanifold.cpp
elalish: Sorry, I walk talking about the one-line `;` on 134.

tools/pymanifold.cpp
pca006132: Yes, I will have a look at how to split it into something like `pymanifold.core` and `pymanifold.assimp` or something like that.

tools/pymanifold.cpp
pca006132: Thanks for catching this.

tools/pymanifold.cpp
pca006132: Indeed, but as I've said we can use something like a `shared_ptr` to share the underlying manifold and do CoW. Will implement that if needed when the above discussion about the API settles.

tools/pymanifold.cpp
rsaccon: and the "Smooth" method, which I consider as one of the main selling points of this lib !

tools/pymanifold.cpp
pca006132: oh, that was not intentional. deleted

tools/pymanifold.cpp
elalish: Ah, I'm starting to understand this better. That's an interesting idea. It still seems like it would be preferable to keep the API consistent among the different language bindings. Would you propose we change the C++ API to function this way as well? This project is young, so better to make breaking API changes now than when there are more users.

pca006132: Yes, it would be ideal to chang the C++ API as well. In fact the prototype implementation `ManifoldSet` in my fork contains some code duplication for handling transforms, so it would be nice to integrate it to the Manifold class directly.

elalish: Okay, we have two sort of parallel discussions going here about the API and copying. I'd like to unblock this PR. How about we stick with the API you have for now, and look at modifying the C++ API to match in a following PR? Do you want to go ahead and get the tests passing and get this merged?

tools/pymanifold.cpp
```diff
@@ -0,0 +1,205 @@
...
+          py::arg("twist_degrees") = 0.0f,
+          py::arg("scale_top") = std::make_tuple(1.0f, 1.0f))
+      .def(
+          "revolve",
+          [](PolygonsWrapper &self, int circularSegments = 0) {
+            return Manifold::Revolve(*self.polygons, circularSegments);
+          },
+          py::arg("circular_segments") = 0);
+
+  py::class_<Mesh>(m, "Mesh")
```
elalish, line 157: :100:

elalish:
I'd like to get this merged before it gets too out-of-date; there's good stuff in here. Is there anything you feel is blocking this now?

pca006132:
CI failure, I will have a look at this today.

elalish (approved):
Thanks, I think these bindings will be super useful! Do you have ideas of how to split this into two Python packages so the Assimp part can be an add-on? That would be a really helpful follow-on, and I'm hoping to follow that pattern for trimming down the WASM build too.

tools/pymanifold.cpp
elalish: With this merged, it should be fairly easy to expand the Python API; would you like to try adding `Smooth`?

pca006132: `Smooth` is already added in line 94.

pca006132:
I am not sure how we can split the methods into two Python packages, but this should be possible considering python is very dynamic. Will have to check later.

tools/pymanifold.cpp
elalish: Thanks! Totally missed that.

PR #123, opened 2022-05-16
stop building assimp

elalish:
@pca006132 I see why you went this route, but I'm impatient when it comes to building, plus the patch makes building a bit complicated. I don't love having to wait to rebuild assimp every time I make a random change to CMakeLists. The only trick is I'm not sure the best way to install it for Windows. I'm wondering if we should switch to `vcpkg` for our dependencies so we have a consistent system between Linux and Windows. Assimp [recommends](https://github.com/assimp/assimp/blob/master/Build.md#install-on-all-platforms-using-vcpkg) it. What do you think?


pca006132:
Never tried `vspkg` before, but sounds good to me if its version is more up to date, comparing to the one installed with the OS. I think we still need to build assimp for emscripten to run our tests though.

If what we need is faster build time, I think there are several things that we can do to speed up compilation without removing assimp entirely:
1. Reduce the number of import and export formats. Currently we build every format supported by assimp, but we don't need most of them just for our tests.
2. Try ccache? Haven't tried that before but I've heard that it can speed up compilation a lot.

elalish:
Hmm, in the same way we've been talking about separating `meshIO` out of Python/WASM, maybe we should separate it from our tests as well. It's not really used by very many. I'd be pretty comfortable not even bothering with `meshIO`/`assimp` at all for the WASM build. What do you think?

If someone does want `meshIO`, it's nice to have it automatically working for all the formats, but we could fall back to reducing that if we really need to keep building it ourselves.

pca006132:
I'm fine with separating it from our tests.

elalish:
Okay @pca006132, I'm going to leave this for now and focus on fixing the bugs you've reported. If you happen to get interested in this PR, you're welcome to take it over. Not a huge priority, but I think separating `meshIO` as much as we can and using pre-built `assimp` everywhere else will simplify things.

pca006132:
Maybe you can try this patch and see if it improves your build time?

```diff
diff --git a/CMakeLists.txt b/CMakeLists.txt
index 8f32750..70f2138 100644
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -33,6 +33,17 @@ if(NOT THRUST_BACKEND STREQUAL "CUDA" AND NOT THRUST_BACKEND STREQUAL "OMP" AND
     message(FATAL_ERROR "Invalid value for THRUST_BACKEND: ${THRUST_BACKEND}. Should be one of \"CUDA\", \"OMP\" or \"CPP\"")
 endif()
 
+option(ASSIMP_FAST_BUILD "build ASSIMP just for tests" ON)
+if(ASSIMP_FAST_BUILD)
+    option(ASSIMP_INSTALL FALSE)
+    option(ASSIMP_BUILD_ALL_IMPORTERS_BY_DEFAULT FALSE)
+    option(ASSIMP_BUILD_ALL_EXPORTERS_BY_DEFAULT FALSE)
+    foreach(FMT OBJ;PLY;STL;GLTF)
+        set(ASSIMP_BUILD_${FMT}_IMPORTER TRUE)
+        set(ASSIMP_BUILD_${FMT}_EXPORTER TRUE)
+    endforeach()
+endif()
```

Just disabled most of the importers and exporters. The build is significantly faster on my computer.

fire:
I am not that hot on vcpkg :/

elalish:
This has been superseded by trimming down the number of Assimp importer/exporters to build. That's much simpler for now.

PR #125, opened 2022-05-19
Thread safety

pca006132:
- Added lock for the static variable `meshID2Original_` to make manifold copy constructors thread safe, as users may copy manifolds concurrently (I guess gtest do run tests concurrently?). A `allMeshID` cache is added to cache all mesh IDs referenced in the manifold `pImpl_`, so we just need to iterate over this set while holding the lock to `meshID2Original_`, instead of having to iterate over the entire `triBary` vector while holding the lock.
- Added TBB backend and changed `atomicAdd` to use c++ atomics instead of OpenMP pragma.

Todo: Is there a way to wrap the vector with a mutex? So we can return `meshID2Original_` protected by a mutex in the `MeshID2Original` function.


manifold/src/impl.cpp
```diff
@@ -362,13 +362,37 @@ Manifold::Impl::Impl(Shape shape) {
...
+  char* seenD = seen.ptrD();
+  thrust::for_each(thrust::device, meshRelation_.triBary.begin(), meshRelation_.triBary.end(),
+    [=](BaryRef& ref) {
+      int index = thrust::lower_bound(begin, end, ref.meshID) - begin;
+      seenD[index] = 1;
+      ref.meshID = first + index;
+    });
+  for (int i = 0; i < oldIDs.size(); i++) {
+    if (!seen[i])
+      meshRelation_.allMeshID.erase(first + i);
```
elalish, line 395: This function has gotten quite a bit more complicated; maybe a comment about what we're doing here? I'm having some trouble following.

manifold/src/impl.cpp
```diff
@@ -382,9 +406,14 @@ int Manifold::Impl::InitializeNewReference(
     const std::vector<float>& properties,
     const std::vector<float>& propertyTolerance) {
   meshRelation_.triBary.resize(NumTri());
+
+  meshID2OriginalMutex_.lock();
   const int nextMeshID = meshID2Original_.size();
   meshID2Original_.push_back(nextMeshID);
+  meshID2OriginalMutex_.unlock();
```
elalish, line 413: Could we make this locking simpler by making `meshID2Original` a hash table or something? The IDs don't have to be sequential; I'd be fine with a GUID if that would make the code simpler.

manifold/src/boolean_result.cpp
```diff
@@ -674,6 +674,10 @@ Manifold::Impl Boolean3::Result(Manifold::OpType op) const {
   Timer simplify;
   simplify.Start();
 
+  outR.meshRelation_.allMeshID = inP_.meshRelation_.allMeshID;
+  outR.meshRelation_.allMeshID.insert(
+    inQ_.meshRelation_.allMeshID.begin(),
+    inQ_.meshRelation_.allMeshID.end());
```
elalish, line 680: Some of these meshIDs might not be referenced anymore, right? I take it that doesn't matter since this is an internal detail?

manifold/src/impl.cpp
```diff
@@ -362,13 +362,37 @@ Manifold::Impl::Impl(Shape shape) {
...
+
+  auto begin = oldIDs.begin();
+  auto end = oldIDs.end();
+  VecDH<char> seen(oldIDs.size(), 0);
+  char* seenD = seen.ptrD();
+  thrust::for_each(thrust::device, meshRelation_.triBary.begin(), meshRelation_.triBary.end(),
+    [=](BaryRef& ref) {
+      int index = thrust::lower_bound(begin, end, ref.meshID) - begin;
+      seenD[index] = 1;
+      ref.meshID = first + index;
```
elalish, line 391: I think CUDA doesn't like your lambda functions. This is why I use struct functors for all my device functions.

elalish (commented):
I'm hoping there's a simpler way to get thread safety for `MeshID2Original`. Let's not be too picky about breaking changes; I doubt anyone has used that feature yet anyway.

manifold/src/boolean_result.cpp
pca006132: Yes, the cache is meant to be an over approximation of the set of mesh ids. IDs that are not being referenced will be removed when we duplicate the impl.

manifold/src/impl.cpp
pca006132: Ah, forgot to add `__host__ __device__` to the lambda. Will fix.

manifold/src/impl.cpp
pca006132: I think hash table in the stl do not support concurrent access.

manifold/src/impl.cpp
```diff
@@ -28,12 +29,10 @@ __host__ __device__ void AtomicAddVec3(glm::vec3& target,
...
     atomicAdd(&target[i], add[i]);
-#elif defined(_OPENMP)
-#pragma omp atomic
-    target[i] += add[i];
 #else
-    // should be executed with single thread on the host
-    target[i] += add[i];
+    std::atomic<float> &tar = reinterpret_cast<std::atomic<float>&>(target[i]);
+    float old_val = tar.load();
+    while (!tar.compare_exchange_weak(old_val, old_val + add[i]));
```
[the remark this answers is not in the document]
pca006132: Because `atomic<float>` does not have `fetch_add`, iirc it is added in c++17 or 20. For older versions, only integers can do fetch add.

pca006132:
Actually regarding `MeshID2Original`, I am thinking if it is possible to make it reference the old copy using weak pointers or something, and make the ID local to the objects (and their parents). This way we don't need a static variable for storing the IDs, but may be a bit more complicated. The benefit is that there will be no static buffer leaking memory (we never shrink the static buffer), and we don't need a lock for the buffer.

elalish:
I suppose another option, which I think is in the same vein, is to just add an `OriginalMeshID` alongside `meshID` in `BaryRef` and just skip all of this entirely. Not sure why that didn't occur to me before; I think that would keep everything simpler. What do you think?

elalish:
Then I think the only thing you'd need for thread safety is an atomic to increment the next `meshID` when you need one.

pca006132:
Great, so when we copy A -> B -> C, C's `OriginalMeshID` should point to A right? Unless we do `SetAsOriginal`.

pca006132:
Sorry I was busy with finals recently. I have a bit trouble understanding the purpose of `meshID`.

https://github.com/elalish/manifold/blob/a7791d09b64afdd6aa029fe7e63680e237afad0c/test/mesh_test.cpp#L67-L72

It seems that the original mesh ID used here is the immediate predecessor of the copy? i.e. when we copy A -> B -> C, `meshID2Original[meshID]` of C would point to B rather than A. Is this intended?

If we follow this behavior, we should make C's `OriginalMeshID` point to B instead of A. However, if we don't have a static mapping and if B is just an intermediate object that is not accessible by the user, the user will not be able to find out that C comes from A, because the mapping A -> B is only stored in B and is lost. And if we set C's `OriginalMeshID` to A's mesh ID, we will lose the mapping `B -> C`.

elalish:
No problem at all! I got halfway through fixing the bug you found and then got COVID. Hoping to get back to it soon. 

You actually had it right before, original points all three way back to the first mesh, unless set as original is called. You get a new mesh id any time the mesh is copied, which includes booleans. It's like an instance id. Perhaps I should have named it that.

pca006132:
Sorry to hear that, hope that you can recover soon.

Regarding the mesh ID problem, I'm still not sure what is the purpose for unique ID for each copy. It seems to me that the major usage of this ID is to find the original mesh and apply the properties of the original mesh to the copy, which does not need a unique ID for each copy.

elalish:
Honestly, that's a good point. I had some reasons, but they may have been superceded by the set as original function. I was thinking it would help with finding the intersection between copies, but that can probably be deduced from which verts are original. It's been hard to explain on the documentation, which is probably reason enough to remove it.

pca006132:
Great, I will just rename `meshID` to `originalMeshID` and remove the `DuplicateMeshIDs` function. This way the code will be simpler, and lazy evaluation now makes more sense as different copies may share the same underlying mesh, it is 'lazier' to avoid having to create a new set of mesh IDs.

pca006132:
Is the `meshID` is also used to avoid collapsing edges from different meshes? I tried removing the `DuplicateMeshIDs`, setting flipping the 30th bit to indicate whether the triangle comes from P or Q and then reset it when we finish the boolean operation, which works for the boolean tests but failed the sample tests.

It seems that if we have to keep the `meshID` from P and Q distinct even after a boolean operation, because subsequent operations may try to simplify the mesh and will corrupt it if the distinct `meshID` property is lost. Not sure if this behavior is intended or not. Here is my patch that you can try:

<details>

```diff
diff --git a/manifold/include/manifold.h b/manifold/include/manifold.h
index 16e0e0b..b101c34 100644
--- a/manifold/include/manifold.h
+++ b/manifold/include/manifold.h
@@ -105,7 +105,6 @@ class Manifold {
   MeshRelation GetMeshRelation() const;
   std::vector<int> GetMeshIDs() const;
   int SetAsOriginal();
-  static std::vector<int> MeshID2Original();
   ///@}
 
   /** @name Modification
@@ -164,4 +163,4 @@ class Manifold {
   static float circularEdgeLength_;
 };
 /** @} */
-}  // namespace manifold
\ No newline at end of file
+}  // namespace manifold
diff --git a/manifold/src/boolean_result.cpp b/manifold/src/boolean_result.cpp
index 807727f..6d302df 100644
--- a/manifold/src/boolean_result.cpp
+++ b/manifold/src/boolean_result.cpp
@@ -442,6 +442,9 @@ struct CreateBarycentric {
     const BaryRef oldRef = halfedgeRef.PQ == 0 ? triBaryP[tri] : triBaryQ[tri];
 
     faceRef[halfedgeR.face] = oldRef;
+    if (halfedgeRef.PQ != 0) {
+      faceRef[halfedgeR.face].meshID |= 1 << 30;
+    }
 
     if (halfedgeR.startVert < firstNewVert) {  // retained vert
       int i = halfedgeRef.vert;
@@ -674,10 +677,18 @@ Manifold::Impl Boolean3::Result(Manifold::OpType op) const {
   Timer simplify;
   simplify.Start();
 
-  outR.DuplicateMeshIDs();
+  outR.meshRelation_.triBary.Dump();
 
   outR.SimplifyTopology();
 
+  thrust::for_each(
+      thrust::device,
+      outR.meshRelation_.triBary.begin(),
+      outR.meshRelation_.triBary.end(),
+      []__host__ __device__(BaryRef& b) {
+        b.meshID &= ~(1 << 30);
+      });
+
   simplify.Stop();
   Timer sort;
   sort.Start();
diff --git a/manifold/src/constructors.cpp b/manifold/src/constructors.cpp
index e0be155..c71c339 100644
--- a/manifold/src/constructors.cpp
+++ b/manifold/src/constructors.cpp
@@ -426,7 +426,6 @@ Manifold Manifold::Compose(const std::vector<Manifold>& manifolds) {
     nextBary += impl.meshRelation_.barycentric.size();
   }
 
-  combined.DuplicateMeshIDs();
   combined.Finish();
   return out;
 }
diff --git a/manifold/src/impl.cpp b/manifold/src/impl.cpp
index d6ca915..a188e52 100644
--- a/manifold/src/impl.cpp
+++ b/manifold/src/impl.cpp
@@ -281,7 +281,7 @@ struct EdgeBox {
 
 namespace manifold {
 
-std::vector<int> Manifold::Impl::meshID2Original_;
+std::atomic<int> Manifold::Impl::meshIDCounter_(1);
 
 /**
  * Create a manifold from an input triangle Mesh. Will throw if the Mesh is not
@@ -355,22 +355,6 @@ Manifold::Impl::Impl(Shape shape) {
   InitializeNewReference();
 }
 
-/**
- * When a manifold is copied, it is given a new unique set of mesh relation IDs,
- * identifying a particular instance of a copied input mesh. The original mesh
- * ID can be found using the meshID2Original mapping.
- */
-void Manifold::Impl::DuplicateMeshIDs() {
-  std::map<int, int> old2new;
-  for (BaryRef& ref : meshRelation_.triBary) {
-    if (old2new.find(ref.meshID) == old2new.end()) {
-      old2new[ref.meshID] = meshID2Original_.size();
-      meshID2Original_.push_back(meshID2Original_[ref.meshID]);
-    }
-    ref.meshID = old2new[ref.meshID];
-  }
-}
-
 void Manifold::Impl::ReinitializeReference(int meshID) {
   thrust::for_each_n(thrust::device, zip(meshRelation_.triBary.begin(), countAt(0)), NumTri(),
                      InitializeBaryRef({meshID, halfedge_.cptrD()}));
@@ -381,8 +365,7 @@ int Manifold::Impl::InitializeNewReference(
     const std::vector<float>& properties,
     const std::vector<float>& propertyTolerance) {
   meshRelation_.triBary.resize(NumTri());
-  const int nextMeshID = meshID2Original_.size();
-  meshID2Original_.push_back(nextMeshID);
+  const int nextMeshID = meshIDCounter_.fetch_add(1);
   ReinitializeReference(nextMeshID);
 
   const int numProps = propertyTolerance.size();
diff --git a/manifold/src/impl.h b/manifold/src/impl.h
index 27c21b1..def5e66 100644
--- a/manifold/src/impl.h
+++ b/manifold/src/impl.h
@@ -40,7 +40,7 @@ struct Manifold::Impl {
   Collider collider_;
   glm::mat4x3 transform_ = glm::mat4x3(1.0f);
 
-  static std::vector<int> meshID2Original_;
+  static std::atomic<int> meshIDCounter_;
 
   Impl() {}
   enum class Shape { TETRAHEDRON, CUBE, OCTAHEDRON };
@@ -56,7 +56,6 @@ struct Manifold::Impl {
       const std::vector<float>& properties = std::vector<float>(),
       const std::vector<float>& propertyTolerance = std::vector<float>());
 
-  void DuplicateMeshIDs();
   void ReinitializeReference(int meshID = -1);
   void CreateHalfedges(const VecDH<glm::ivec3>& triVerts);
   void CreateAndFixHalfedges(const VecDH<glm::ivec3>& triVerts);
diff --git a/manifold/src/manifold.cpp b/manifold/src/manifold.cpp
index 642d2ac..dc04011 100644
--- a/manifold/src/manifold.cpp
+++ b/manifold/src/manifold.cpp
@@ -62,14 +62,11 @@ Manifold::~Manifold() = default;
 Manifold::Manifold(Manifold&&) noexcept = default;
 Manifold& Manifold::operator=(Manifold&&) noexcept = default;
 
-Manifold::Manifold(const Manifold& other) : pImpl_(new Impl(*other.pImpl_)) {
-  pImpl_->DuplicateMeshIDs();
-}
+Manifold::Manifold(const Manifold& other) : pImpl_(new Impl(*other.pImpl_)) {}
 
 Manifold& Manifold::operator=(const Manifold& other) {
   if (this != &other) {
     pImpl_.reset(new Impl(*other.pImpl_));
-    pImpl_->DuplicateMeshIDs();
   }
   return *this;
 }
@@ -320,15 +317,6 @@ int Manifold::SetAsOriginal() {
   return meshID;
 }
 
-/**
- * Returns a vector that maps a given unique MeshID to the MeshID of the
- * original Mesh it came from, to easily identify separate copies of the same
- * thing.
- */
-std::vector<int> Manifold::MeshID2Original() {
-  return Manifold::Impl::meshID2Original_;
-}
-
 /**
  * Should always be true. Also checks saneness of the internal data structures.
  */
diff --git a/manifold/src/sort.cpp b/manifold/src/sort.cpp
index 7a0a67a..69541c6 100644
--- a/manifold/src/sort.cpp
+++ b/manifold/src/sort.cpp
@@ -315,7 +315,6 @@ void Manifold::Impl::GatherFaces(const Impl& old,
                  old.meshRelation_.triBary.begin(),
                  meshRelation_.triBary.begin());
   meshRelation_.barycentric = old.meshRelation_.barycentric;
-  DuplicateMeshIDs();
 
   if (old.faceNormal_.size() == old.NumTri()) {
     faceNormal_.resize(numTri);
diff --git a/test/mesh_test.cpp b/test/mesh_test.cpp
index 7fc0b7e..8213488 100644
--- a/test/mesh_test.cpp
+++ b/test/mesh_test.cpp
@@ -63,12 +63,9 @@ void Related(const Manifold& out, const std::vector<Mesh>& input,
              const std::map<int, int>& meshID2idx) {
   Mesh output = out.GetMesh();
   MeshRelation relation = out.GetMeshRelation();
-  std::vector<int> meshID2Original = Manifold::MeshID2Original();
   for (int tri = 0; tri < out.NumTri(); ++tri) {
     int meshID = relation.triBary[tri].meshID;
-    int meshIdx = meshID2idx.find(meshID) != meshID2idx.end()
-                      ? meshID2idx.at(meshID)
-                      : meshID2idx.at(meshID2Original[meshID]);
+    int meshIdx = meshID2idx.at(meshID);
     ASSERT_LT(meshIdx, input.size());
     const Mesh& inMesh = input[meshIdx];
     int inTri = relation.triBary[tri].tri;
@@ -869,4 +866,4 @@ TEST(Boolean, DISABLED_Cylinders) {
   EXPECT_TRUE(m1.IsManifold());
   EXPECT_TRUE(m1.MatchesTriNormals());
   EXPECT_LE(m1.NumDegenerateTris(), 12);
-}
\ No newline at end of file
+}
```

</details>

pca006132:
Final changes:

The meshID in the internal API is purely internal, usually starting from
0, which is used for differentiating between triangles coming from
different meshes. originalID can map this internal meshID to the
external meshID, and such mapping is automatically done when calling
`GetMeshRelation`, so users don't have to care about this mapping.

As a bonus, we don't need to deal with atomics when manipulating meshIDs
internally: we can use whatever IDs we want as long as the mapping is
preserved and distinct for triangles coming from different meshes.
I also made the meshID modification code run in the device, which should
be faster for larger meshes.

elalish:
Yes, you are correct, sorry I forgot to mention that issue regarding edge collapse. And yes, I like your idea of making that an internal detail to simplify the API. I'm still feeling not great, so I can't promise when I'll do a proper code review. Hopefully soon. Thanks!

CMakeLists.txt
```diff
@@ -28,8 +28,15 @@ set(GLM_INC_DIR ${PROJECT_SOURCE_DIR}/third_party/glm)
...
 set_property(CACHE THRUST_BACKEND PROPERTY STRINGS CUDA OMP CPP)
 
-if(NOT THRUST_BACKEND STREQUAL "CUDA" AND NOT THRUST_BACKEND STREQUAL "OMP" AND NOT THRUST_BACKEND STREQUAL "CPP")
-    message(FATAL_ERROR "Invalid value for THRUST_BACKEND: ${THRUST_BACKEND}. Should be one of \"CUDA\", \"OMP\" or \"CPP\"")
+option(ASSIMP_FAST_BUILD "build ASSIMP just for tests" ON)
+if(ASSIMP_FAST_BUILD)
+    option(ASSIMP_INSTALL FALSE)
+    option(ASSIMP_BUILD_ALL_IMPORTERS_BY_DEFAULT FALSE)
+    option(ASSIMP_BUILD_ALL_EXPORTERS_BY_DEFAULT FALSE)
+    foreach(FMT OBJ;PLY;STL;GLTF)
```
elalish, line 36: :+1:

CMakeLists.txt
```diff
@@ -41,21 +48,30 @@ set(MANIFOLD_NVCC_FLAGS -Xcudafe --diag_suppress=esa_on_defaulted_function_ignor
...
-    endif()
+    # clear THRUST_INC_DIR, we use the one from nvcc
+    set(THRUST_INC_DIR "")
+elseif (THRUST_BACKEND STREQUAL "OMP")
+    find_package(OpenMP REQUIRED)
+    set(MANIFOLD_INCLUDE OpenMP::OpenMP_CXX)
+    set(MANIFOLD_FLAGS ${MANIFOLD_FLAGS} -DTHRUST_DEVICE_SYSTEM=THRUST_DEVICE_SYSTEM_OMP -fopenmp)
+elseif (THRUST_BACKEND STREQUAL "CPP")
+    set(MANIFOLD_FLAGS ${MANIFOLD_FLAGS} -DTHRUST_DEVICE_SYSTEM=THRUST_DEVICE_SYSTEM_CPP)
+elseif (THRUST_BACKEND STREQUAL "TBB")
```
elalish, line 66: :100:

elalish: Also, is it possible to run, or at least compile TBB in our CI?

manifold/src/boolean_result.cpp
```diff
@@ -442,6 +442,11 @@ struct CreateBarycentric {
     const BaryRef oldRef = halfedgeRef.PQ == 0 ? triBaryP[tri] : triBaryQ[tri];
 
     faceRef[halfedgeR.face] = oldRef;
+    // Mark the meshID as coming from P by flipping the 30th bit.
+    // The 30th bit is used to avoid flipping the sign.
+    if (halfedgeRef.PQ == 0) {
+      faceRef[halfedgeR.face].meshID |= 1 << 30;
```
elalish, line 448: Let's save `1 << 30` as a `constexpr`, maybe `kFromP` or something?

manifold/src/boolean_result.cpp
```diff
@@ -674,10 +679,49 @@ Manifold::Impl Boolean3::Result(Manifold::OpType op) const {
...
+  // in general, meshIDs should be small
+  thrust::sort_by_key(thrust::host, meshIDs.begin(), meshIDs.end(), original.begin());
+
+  const int numMesh = meshIDs.size();
+  const int* meshIDsPtr = meshIDs.cptrD();
+  int* originalPtr = original.ptrD();
+
+  thrust::for_each(thrust::device, outR.meshRelation_.triBary.begin(),
+                   outR.meshRelation_.triBary.end(),
+                   [=] __host__ __device__(BaryRef &b) {
```
elalish, line 710: I need to get used to using these device lambdas, at least for simple functions like this. Would probably clean up the code a lot. When I started, I couldn't get these to work.

manifold/src/constructors.cpp
```diff
@@ -420,13 +420,47 @@ Manifold Manifold::Compose(const std::vector<Manifold>& manifolds) {
...
+                                   meshIDsPtr;
+                       b.meshID = index + meshIDStart;
+                       originalPtr[index] |= 1 << 30;
+                     });
+
+    for (int i = 0; i < numMesh; ++i) {
+      if (original[i] & (1 << 30)) {
+        combined.meshRelation_.originalID[i+meshIDStart] = original[i] & ~(1 << 30);
+      }
+    }
```
elalish, line 456: This 30-line block looks a lot like the block added to boolean-result. Can we make a function we can reuse instead?

manifold/src/impl.cpp
```diff
@@ -174,6 +173,7 @@ struct InitializeBaryRef {
     int tri = thrust::get<1>(inOut);
 
     // Leave existing meshID if input is negative
+    // TODO: Why do we need this?
```
elalish, line 176: Yeah, not super clear. It makes it so when I make a `sphere` and call `ReinitializeReference` without an argument, it doesn't bother renumbering. I can't remember a reason that's important. If you'd like to simplify it and nothing breaks, be my guest.

manifold/src/manifold.cpp
```diff
@@ -264,16 +261,18 @@ Curvature Manifold::GetCurvature() const { return pImpl_->GetCurvature(); }
  * >= 0, indicating it is a new vertex. If the index is < 0, this indicates it
  * is an original vertex, the index + 3 vert of the referenced triangle.
  *
- * Every time a manifold is copied or combined to form a new manifold it gets a
- * new meshID to indicate that particular instance of the mesh. In order to look
- * up which input mesh a given instance came from, simply use the
- * MeshID2Original() static vector.
+ * In order to look up which input mesh a given instance come from, use the
+ * `originalID` field, i.e. `originalID[meshID]`. Users should not rely on the
+ * raw value of `meshID`.
```
elalish, line 266: Is this comment out of date? It looks like you've now taken care of this so it's no longer exposed in the API, right?

manifold/src/manifold.cpp
```diff
@@ -286,17 +285,11 @@ MeshRelation Manifold::GetMeshRelation() const {
...
-      thrust::device, zip(meshIDs.begin(), pImpl_->meshRelation_.triBary.begin()), NumTri(),
-      GetMeshID());
-
-  thrust::sort(thrust::device, meshIDs.begin(), meshIDs.end());
-  int n = thrust::unique(thrust::device, meshIDs.begin(), meshIDs.end()) - meshIDs.begin();
-  meshIDs.resize(n);
-
   std::vector<int> out;
-  out.insert(out.end(), meshIDs.begin(), meshIDs.end());
+  out.reserve(pImpl_->meshRelation_.originalID.size());
```
elalish, line 289: Is this still a tight set, or could it include `meshIDs` that aren't in the `meshRelation` anymore?

manifold/include/manifold.h
```diff
@@ -105,7 +105,6 @@ class Manifold {
   MeshRelation GetMeshRelation() const;
   std::vector<int> GetMeshIDs() const;
   int SetAsOriginal();
-  static std::vector<int> MeshID2Original();
```
elalish, line 108: :+1:

test/mesh_test.cpp
```diff
@@ -63,12 +63,9 @@ void Related(const Manifold& out, const std::vector<Mesh>& input,
              const std::map<int, int>& meshID2idx) {
   Mesh output = out.GetMesh();
   MeshRelation relation = out.GetMeshRelation();
-  std::vector<int> meshID2Original = Manifold::MeshID2Original();
   for (int tri = 0; tri < out.NumTri(); ++tri) {
     int meshID = relation.triBary[tri].meshID;
-    int meshIdx = meshID2idx.find(meshID) != meshID2idx.end()
-                      ? meshID2idx.at(meshID)
-                      : meshID2idx.at(meshID2Original[meshID]);
```
elalish, line 71: Simplifying the test is a good indication we're moving in the right direction on API shape.

utilities/include/structs.h
```diff
@@ -229,7 +230,13 @@ struct Curvature {
...
  */
 struct BaryRef {
-  /// Reference to original Mesh, matching Manifold.GetMeshIDs().
+  /// Identifier of the triangle, can be mapped to the original mesh ID.
+  /// - In `MeshRelation`: The original mesh triangle index.
+  /// - In `MeshRelationD`: The original mesh triangle index =
+  /// `originalID[meshID]`
+  ///
+  /// @note Triangles coming from different manifolds should have different mesh
+  /// ID, otherwise `SimplifyTopology` will not work properly.
```
elalish, line 239: This is public API documentation (parsed by Doxygen) so it shouldn't refer to anything internal like `MeshRelationD` or `SimplifyTopology`. Let's move these comments to the `Impl`.

elalish (commented):
Looking great, thanks!

manifold/src/manifold.cpp
pca006132: Yes, forgot to update this one.

manifold/src/manifold.cpp
pca006132: This should be tight. I use the 30th bit to indicate whether the meshID was referenced in the result of the boolean operation/compose.

pca006132: Ah, forgot to make it tight for decompose. Will fix.

manifold/src/constructors.cpp
pca006132: Yes, should be possible.

elalish (approved):
This looks great, thanks for helping to simplify the API here!

pca006132:
Revisiting this after (mostly) finished the lazy boolean implementation. Keeping the meshIDs from different meshes distinct can fix the collapseEdge problem mentioned above, but it also prohibits reducing the set of vertices after boolean operations (e.g. when you union n cubes together as a larger cube, the number of vertices will grow to n*8).

Perhaps we should try to use the originalID instead and see how to fix the issue regarding edge collapse. Do you have any documentation about the simplify topology code that I can have a look at?

PR #132, opened 2022-06-04
clang-format

### LINKED ISSUE #130: Add clang-format pre-commit hook
I'm pretty sure there's a way to tell Github to run clang-format on any submitted PR, which would be nice both with viewing diffs and with keeping the code base consistent. I have VSCode set up to use the standard `google` style with clang-format, so that's what I'd like automatically applied for other contributors.

pca006132:
Closes #130

Not sure if we want to merge this now or after the lazy boolean PR. This would probably create some conflicts with that.

### DIFF
--- a/.clang-format
+++ b/.clang-format
@@ -1,0 +1,1 @@
+BasedOnStyle: Google
--- a/.github/workflows/check_format.yml
+++ b/.github/workflows/check_format.yml
@@ -1,0 +1,17 @@
+name: Check code formatting
+
+on:
+  push:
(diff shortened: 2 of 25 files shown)


pca006132:
Ah clicked on the wrong button

elalish (approved):
I'd like to merge it now, as it'll make my own PRs easier, but it depends how long the lazy boolean will be open. How hard do you think it'll be to update that PR?

pca006132:
It should be fine, you can merge this first

PR #140, opened 2022-06-18
build script update

pca006132:
- Simplified build script by exposing compiler options as public in utilities, so we don't have to repeat them in collider/manifold.
- Use file glob instead of adding source files one by one, so we don't have to change the cmake script to add more source files later.
- Added code coverage report.

Note that you can use `build/compile_commands.json` instead of fiddling with vscode settings :).

### DIFF
--- a/collider/CMakeLists.txt
+++ b/collider/CMakeLists.txt
@@ -1,29 +1,15 @@
 project (collider)
 
-add_library(${PROJECT_NAME} src/collider.cpp)
+file(GLOB_RECURSE SOURCE_FILES CONFIGURE_DEPENDS *.cpp)
+add_library(${PROJECT_NAME} ${SOURCE_FILES})
 
 if(MANIFOLD_USE_CUDA)
-    set_source_files_properties(src/collider.cpp PROPERTIES LANGUAGE CUDA)
+    set_source_files_properties(${SOURCE_FILES} PROPERTIES LANGUAGE CUDA)
     set_property(TARGET ${PROJECT_NAME} PROPERTY CUDA_ARCHITECTURES 61)
-    target_compile_options(${PROJECT_NAME} 
-        PRIVATE ${MANIFOLD_NVCC_FLAGS}
-    )
-    target_compile_options(${PROJECT_NAME} 
-        PRIVATE "$<$<CONFIG:RELEASE>:${MANIFOLD_NVCC_RELEASE_FLAGS}>" "$<$<CONFIG:DEBUG>:${MANIFOLD_NVCC_DEBUG_FLAGS}>"
-    )
-else()
-    target_compile_options(${PROJECT_NAME} 
-        PRIVATE ${MANIFOLD_FLAGS}
-    )
 endif()
 
-target_include_directories( ${PROJECT_NAME}
-    PUBLIC ${PROJECT_SOURCE_DIR}/include
-)
-target_link_libraries( ${PROJECT_NAME}
-    PUBLIC utilities
-    PRIVATE ${MANIFOLD_INCLUDE}
-)
+target_include_directories(${PROJECT_NAME} PUBLIC ${PROJECT_SOURCE_DIR}/include)
+target_link_libraries(${PROJECT_NAME} PUBLIC utilities)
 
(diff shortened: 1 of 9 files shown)


manifold/CMakeLists.txt
```diff
@@ -1,34 +1,15 @@
...
     set_property(TARGET ${PROJECT_NAME} PROPERTY CUDA_ARCHITECTURES 61)
-    target_compile_options(${PROJECT_NAME} 
-        PRIVATE ${MANIFOLD_NVCC_FLAGS}
-    )
-    target_compile_options(${PROJECT_NAME} 
-        PRIVATE "$<$<CONFIG:RELEASE>:${MANIFOLD_NVCC_RELEASE_FLAGS}>" "$<$<CONFIG:DEBUG>:${MANIFOLD_NVCC_DEBUG_FLAGS}>"
-    )
-else()
-    target_compile_options(${PROJECT_NAME} 
-        PRIVATE ${MANIFOLD_FLAGS}
```
elalish, line 26: Great simplification!

collider/CMakeLists.txt
```diff
@@ -1,29 +1,15 @@
 project (collider)
 
-add_library(${PROJECT_NAME} src/collider.cpp)
+file(GLOB_RECURSE SOURCE_FILES CONFIGURE_DEPENDS *.cpp)
+add_library(${PROJECT_NAME} ${SOURCE_FILES})
```
elalish, line 4: Agreed this looks cleaner, but I avoided it after reading posts like this: https://stackoverflow.com/a/18538444

utilities/CMakeLists.txt
```diff
@@ -1,20 +1,43 @@
...
-    set_source_files_properties(src/detect_cuda.cpp PROPERTIES LANGUAGE CUDA)
+    set_source_files_properties(${SOURCE_FILES} PROPERTIES LANGUAGE CUDA)
     set_property(TARGET ${PROJECT_NAME} PROPERTY CUDA_ARCHITECTURES 61)
     target_compile_options(${PROJECT_NAME} 
-        PRIVATE ${MANIFOLD_NVCC_FLAGS}
-    )
-    target_compile_options(${PROJECT_NAME} 
-        PRIVATE "$<$<CONFIG:RELEASE>:${MANIFOLD_NVCC_RELEASE_FLAGS}>" "$<$<CONFIG:DEBUG>:${MANIFOLD_NVCC_DEBUG_FLAGS}>"
+        PUBLIC ${MANIFOLD_FLAGS} -DMANIFOLD_USE_CUDA
+        "$<$<COMPILE_LANGUAGE:CUDA>:${MANIFOLD_NVCC_FLAGS}>"
```
elalish, line 35: This is slick, setting the compile options once via utilities. :+1:

elalish (changes requested):
Thanks for the CMake cleanup! The coverage report is nice, though I can't see the details, I think because you force-pushed since it ran? Do you know how to re-trigger it?

elalish:
I'm excited about this browser extension that visualizes the code coverage: https://docs.codecov.com/docs/browser-extension

I've got it installed, but I think we'll have to merge this before it'll work.

collider/CMakeLists.txt
pca006132: But the second answer mentioned that

> In CMake 3.12, the [`file(GLOB ...)` and `file(GLOB_RECURSE ...)`](https://cmake.org/cmake/help/v3.12/command/file.html#glob) commands gained a `CONFIGURE_DEPENDS` option which reruns cmake if the glob's value changes.

which should not have the problem in the first answer. But I'm fine with reverting back to the original one if you prefer that.

pca006132:
I think I will try to merge into my fork first and see how the codecov thing works.

collider/CMakeLists.txt
elalish: Oh, that's great to know. Okay, let's go with this then.

elalish (approved):
Looks good to me. Should we go ahead and merge it?

pca006132:
OK. And you have to add the status badge by yourself: https://docs.codecov.com/docs/status-badges

PR #142, opened 2022-06-20
disable branch coverage

elalish:
I really like the coverage report, especially in combination with the codecov extension to the sourcegraph browser extension that let's me visualize the coverage line-by-line right in Github! But it was coming up with a lot of things having ~50% coverage, and it was claiming lots of lines were "partial coverage". It seems this is largely because of internal exception handling code and probably has something to do with why branch coverage is disabled by default in `lcov`: https://stackoverflow.com/a/43726240

I'm switching it here to see if this gives us a less messy coverage report that's more actionable. What do you think, @pca006132?


pca006132:
Oh I'm fine with this, iirc I just copied some lcov commands from the web randomly. Btw maybe you can add the coverage badge to the README file, that badge is also a link to the coverage report so others can look at it easily.

Issue #73, opened 2022-02-28
Feature request: JS interface + WASM module


bhouston:
This is a very interesting endeavor.  Congratulations @elalish !

This is a wishlist item:

* JS interface so one can write JS/TS and integrate it with Three.js for display and manipulation.
* WASM module so you can use it directly in a browser.

elalish:
Here here! Contributions welcome!

FishOrBear:
Compile directly into wasm interface and publish npm package

elalish:
Agreed; would you like to give it a shot?

elalish:
@pca006132 Now that you got the emscripten build working for the tests, do you know how to make a build that's callable as a library? I'd love to make an example of calling this WASM blob from a simple three.js project, a la #75. Probably best to skip `MeshIO` and `assimp` in that build, to keep size down, but also because plenty of JS libraries like Three already support import/export.

pca006132:
I'm having a look at this, but it seems that the API exposed to JS will not be very convenient to use as you will need to call delete everywhere: https://emscripten.org/docs/porting/connecting_cpp_and_javascript/embind.html#memory-management , or you will leak memory pretty quickly.

Writing the entire thing in C++ and compile to WASM is already possible for now.

pca006132:
OK it seems that three.js also require calling dispose explicitly, so this should not be a problem. Wondering if it will be better to just create a C FFI, so binding to other languages will be easier (if we want to do that in the future).

elalish:
three.js requires `dispose` to free up any GPU memory, as WebGL is a C-style API that requires explicit memory management. I think when it comes to C FFI, you know more than I do.

makc:
perhaps [this comment](https://github.com/elalish/manifold/issues/128#issuecomment-1157091891) should have been made here. TL;DR right now emmake produces a bunch of .a files that can be put through [embind](https://emscripten.org/docs/porting/connecting_cpp_and_javascript/embind.html) if someone writes that one js file - then the result will be directly useable in either three.js or any other js library of user's choice.

pca006132:
Indeed the structure looks like the one in pybind11. I guess we should do something similar to https://github.com/elalish/manifold/blob/master/tools/pymanifold.cpp

elalish:
Agreed, but let's leave out `ExportMesh` this time, since we plan to factor that out of the python module eventually too.

elalish:
[...]

Issue #141, opened 2022-06-19
Removing exceptions


fire:
- [ ] remove STL containers exceptions
- [ ] remove Thrust exceptions
- [ ] remove Manifold exceptions

Create strategy to change exception code to return codes.

Some discussion is in the Godot issue.

elalish:
So, I'm not yet convinced that return codes are superior. Please convince me here! The gist I have is that exceptions weren't supported well on old compilers, but mostly are now: https://github.com/elalish/manifold/issues/91#issuecomment-1104147535, and that the STL is somehow okay even if exceptions are not, which I also don't understand: https://github.com/elalish/manifold/issues/91#issuecomment-1105199004

It sounds to me like the underlying issue is you'd like for Manifold to successfully return a reasonable result for all manifold input, which I would like as well. There are probably a few bugs left to fix, so the best thing there is to submit tests that reproduce them. The bigger issue is I don't currently make any guarantees about handling self-overlapping input meshes, and I don't have an efficient way to test for them either. I would like to make an algorithm to fix these, but that'll take considerable time. Until then, general input meshes aren't going to be safe.

fire:
Returning an empty mesh is an non-reasonable result for all manifold input, but it technically satisfy the criteria of not crashing.

fire:
STL is ok in the sense that it still crashes Godot Engine on exception throw but there's a feature to disable it. https://stackoverflow.com/questions/553103/can-i-disable-exceptions-in-stl

pca006132:
Not sure how game devs view this, but from Rust we categorize errors into two types: expected (input check, file exists etc.) and not expected (internal asserts that indicates bugs when failed)

For the first kind, we will use error code or `Result<T>` here, and users are expected to check for errors (or unwrap to abort if error). For the second kind, users are not expected to handle them and we can just abort.

I guess we can do something similar here. For user errors, we return error code. For assertion failure, we throw if exception is enabled and abort otherwise. I think using error code for all assert checks will make the API very complicated and these checks should be internal checks and not exposed to users.

fire:
From content creation testing.

1. Expected (input check, file exists etc.)

If the mesh is not manifold, return an empty mesh.

3. Not expected

Boolean merge operation on two manifold meshes returns a non manifold mesh causing the next operator to fail too.

pca006132:
For #125, do you mean it fixed a crash or it causes a new crash?

[...]

PR #146, opened 2022-06-30
Wasm

makc:
This is not a real pull request but rather a simple test for #73


pca006132:
It's great that there is a three.js demo (?). I think you can add the generated files to `.gitignore` so they won't be added to the commit.

makc:
ok, noted, but how would I integrate it into the build? e g how to run this long ass emcc command during/after make?

wasm/bindings.cpp
```diff
@@ -0,0 +1,113 @@
...
+        Mesh empty;
+        return empty;
+    }
+}
+
+void subtract(Manifold& a, Manifold& b) {
+    a -= b;
+}
+
+EMSCRIPTEN_BINDINGS(whatever) {
```
elalish, line 66: Love it; does this name even show up anywhere in the actual WASM blob?

wasm/bindings.cpp
```diff
@@ -0,0 +1,113 @@
...
+
+Manifold makeManifoldFromMesh(Mesh& mesh) {
+    //logMesh(mesh);
+
+    try {
+        Manifold manifold(mesh);
+        return manifold;
+    }
+    catch(const std::runtime_error& e) {
+        puts(e.what());
```
elalish, line 40: This is interesting, and gets at some of the discussion around removing or changing exception handling: #141. Is there no way for C++ exceptions to automatically throw JS exceptions with WASM?

wasm/index.html
```diff
@@ -0,0 +1,106 @@
...
+        v = mesh.triVerts.get(i);
+        x.push(v[0], v[1], v[2]);
+      }
+      geometry.setAttribute('position', new THREE.BufferAttribute(new Float32Array(p), 3));
+      geometry.setAttribute('normal', new THREE.BufferAttribute(new Float32Array(n), 3));
+      geometry.setIndex(new THREE.BufferAttribute(new Uint8Array(x), 1));
+      return geometry;
+    }
+
+    // most of three.js geometries arent manifolds, so...
```
elalish, line 95: :+1:

wasm/index.html
```diff
@@ -0,0 +1,106 @@
+<!doctype html>
+<html>
+  <script src="https://unpkg.com/three@0.142.0/build/three.js"></script>
+  <script src="https://unpkg.com/three@0.142.0/examples/js/utils/BufferGeometryUtils.js"></script>
+  <script>
+    var Module = {
+      onRuntimeInitialized: function() {
+        // we have manifold module, let's do some three.js
```
elalish, line 8: :100:

elalish (commented):
This is a great demo, thank you! Let's see if we can get this building in the CI. I wonder what the easiest way to test it end-to-end would be?

elalish:
The CI isn't running because this is a draft PR. Feel free to mark it as ready for review so we can play with the automated builds. Don't worry, I'm not going to merge it in it's current state :)

wasm/bindings.cpp
makc: @elalish it does throw all the way to js but the message you get is just some random number - so, with puts(e.what()) you can at least see what was actually wrong on c++ side

wasm/bindings.cpp
makc: it is not in manifold.js at least; emscripten docs [say](https://emscripten.org/docs/api_reference/bind.h.html#_CPPv419EMSCRIPTEN_BINDINGS4name) this:
> a label to mark a group of related bindings (for example EMSCRIPTEN_BINDINGS(physics), EMSCRIPTEN_BINDINGS(components), etc.)

PR #149, opened 2022-07-06
Copyright updates

elalish:
I'd like some feedback on this one before I merge it, as I want to make sure none of this will concern any of the contributors and users of this project. Basically, I'm trying to get more time to work on this by making it my official 20% project at Google. A day a week will give me a lot more time to apply to it than my current "after my kids go to bed and before I pass out". As part of that, I need to add Google to the copyrights (I'm doing that through an AUTHORS file, which also allows contributors to add themselves, since this has clearly grown beyond being solely my own project - thank you!). 

Of course this remains just as open source, under the same Apache 2.0 license. The one annoying part is I may need to require contributors to sign a [CLA](https://cla.developers.google.com/about/google-individual), which is pretty innocuous, but legalese always presents a barrier. I'm going to ask for an exception, but I'm curious if anyone feels strongly that this would stop them from wanting to contribute?


pca006132:
Yes I am fine with this, just submitted the CLA. The entry in the AUTHORS file is good, thanks.

pca006132:
OK I guess the compilation error is due to thrust using some deprecated std functions (`__has_trivial_copy`) which emits warnings and turned into errors. We probably should update the thrust version and see if it works.

PR #150, opened 2022-07-09
use radix sort instead of merge sort

pca006132:
Sort `uint64_t` instead of `TmpEdge`, so thrust can use radix sort via template specialization. This gives some tiny performance improvement over merge sort. (perfTest last case went from 4.3s -> 4s on my machine, which is about 7%)


pca006132:
The coverage report is interesting, removing two lines that were covered before resulting in a minor reduction in total coverage... I understand what they are doing here but perhaps they should use other metrics that will not report a negative in such cases

manifold/src/impl.cpp
```diff
@@ -449,18 +449,20 @@ void Manifold::Impl::CreateHalfedges(const VecDH<glm::ivec3>& triVerts) {
   // drop the old value first to avoid copy
   halfedge_.resize(0);
   halfedge_.resize(3 * numTri);
-  VecDH<TmpEdge> edge(3 * numTri);
+  VecDH<uint64_t> edge(3 * numTri);
+  VecDH<int> ids(3*numTri);
```
elalish, line 453: It looks like our code formatting check is missing the spacing here. Any idea why?

elalish (commented):
Nice, 7% is a pretty decent improvement!

elalish:
And yes, the coverage report putting up a red X for -0.01% is pretty silly. We don't have it marked to block a merge, so that's good. I guess we just have to remember to take it with a grain of salt.

manifold/src/impl.cpp
pca006132: https://google.github.io/styleguide/cppguide.html#Horizontal_Whitespace

Operator:

```
// Other binary operators usually have spaces around them, but it's
// OK to remove spaces around factors.  Parentheses should have no
// internal padding.
```

maybe we should set it explicitly that these operators should have spaces around them.

pca006132: I found that there is no options in clang-format to force that.

pca006132:
Removed the remove cuda from devShell change, as that is now contained in the wasm PR.

manifold/src/impl.cpp
elalish: That's strange, since clang-format automatically changes that kind of spacing. I sort of thought the check just ran auto-format on the files and checked that it was a no-op.

pca006132: The action is just a check, because I don't think it is a good idea to make the bot commit automatically. That will mess up git blame.

So I guess clang-format will probably change that spacing, but it will not treat that as an error.

PR #151, opened 2022-07-10
wasm

makc:
Basically repeat of #146 with manifold.yml modified to include a step to build it, let's see how it goes


makc:
@pca006132 so I did like you said, but [there seem to be no output after leaving build folder](https://github.com/elalish/manifold/runs/7270173610?check_suite_focus=true#step:6:1070) - as if it still uses original yml file and not the one in the fork

pca006132:
Try this:

```diff
diff --git a/CMakeLists.txt b/CMakeLists.txt
index b3843c4..9571283 100644
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -83,3 +83,7 @@ add_subdirectory(meshIO)
 add_subdirectory(samples)
 add_subdirectory(test)
 add_subdirectory(tools)
+
+if(EMSCRIPTEN)
+add_subdirectory(wasm)
+endif()
diff --git a/flake.nix b/flake.nix
index 8d917a1..4d9bbbe 100644
--- a/flake.nix
+++ b/flake.nix
@@ -63,6 +63,8 @@
             nativeBuildInputs = (with pkgs; [ cmake python38 ]);
             buildInputs = [ pkgs.nodejs ];
             configurePhase = ''
+              mkdir -p .emscriptencache
+              export EM_CACHE=$(pwd)/.emscriptencache
               mkdir build
               cd build
               emcmake cmake -DCMAKE_BUILD_TYPE=Release ..
@@ -77,9 +79,8 @@
             '';
             installPhase = ''
               mkdir -p $out
-              cd tools
-              cp *.js $out/
-              cp *.wasm $out/
+              cp {tools,wasm}/*.js $out/
+              cp {tools,wasm}/*.wasm $out/
             '';
           };
           devShell = devShell { };
diff --git a/wasm/CMakeLists.txt b/wasm/CMakeLists.txt
new file mode 100644
index 0000000..34ad11f
--- /dev/null
+++ b/wasm/CMakeLists.txt
@@ -0,0 +1,7 @@
+project (wasm)
+
+add_executable(manifoldjs bindings.cpp)
+target_link_libraries(manifoldjs manifold)
+target_compile_options(manifoldjs PRIVATE ${MANIFOLD_FLAGS} -fexceptions)
+target_link_options(manifoldjs PUBLIC --bind)
+target_compile_features(manifoldjs PUBLIC cxx_std_14)
```

You need to add it as a target to CMake.

pca006132:
btw your branch is outdated from the master, maybe you can do a merge/rebase first and cleanup the log code in the binding file.

makc:
@pca006132 it does not seem to be happy with master:
> error: patch failed: flake.nix:63
> error: flake.nix: patch does not apply

pca006132:
The failures are quite weird, I will have a look again.

makc:
Ahaha, I tried to exclude assimp from emscripten build, and it worked - however that also broke "Test WASM" thingy

pca006132:
I don't think it is due to the assimp dependency. Try the following commit:
```diff
diff --git a/.github/workflows/manifold.yml b/.github/workflows/manifold.yml
index 64bcad4..34ec32b 100644
--- a/.github/workflows/manifold.yml
+++ b/.github/workflows/manifold.yml
@@ -94,6 +94,7 @@ jobs:
         mkdir build
         cd build
         emcmake cmake -DCMAKE_BUILD_TYPE=Release .. && emmake make
+        cd ../
     - name: Test WASM
       run: |
         cd build/test
diff --git a/flake.lock b/flake.lock
index 46bb235..8068c7c 100644
--- a/flake.lock
+++ b/flake.lock
@@ -17,16 +17,16 @@
     },
     "nixpkgs": {
       "locked": {
-        "lastModified": 1651310835,
-        "narHash": "sha256-MLk/zsLlbPhwFucxL64Fr+oIrvQC2/76Ap2F7ekbPNI=",
+        "lastModified": 1657296039,
+        "narHash": "sha256-Ghh39+aS+pw5sTP/ZO8VIKE6sBhMadDaQZtf+3yu4Vc=",
         "owner": "NixOS",
         "repo": "nixpkgs",
-        "rev": "fd3e33d696b81e76b30160dfad2efb7ac1f19879",
+        "rev": "71d7a4c037dc4f3e98d5c4a81b941933cf5bf675",
         "type": "github"
       },
       "original": {
         "id": "nixpkgs",
-        "ref": "nixos-21.11",
+        "ref": "nixos-22.05",
         "type": "indirect"
       }
     },
diff --git a/flake.nix b/flake.nix
index 1d7dd79..d7f8b74 100644
--- a/flake.nix
+++ b/flake.nix
@@ -1,6 +1,6 @@
 {
   inputs.flake-utils.url = "github:numtide/flake-utils";
-  inputs.nixpkgs.url = "nixpkgs/nixos-21.11";
+  inputs.nixpkgs.url = "nixpkgs/nixos-22.05";
 
   outputs = { self, nixpkgs, flake-utils }:
     flake-utils.lib.eachDefaultSystem
@@ -118,7 +118,7 @@
               '';
             };
           };
-          devShell = devShell { additional = [ pkgs.cudatoolkit_11_4 ]; };
+          devShell = devShell { };
           devShells.cuda = devShell {
             additional = [ pkgs.cudatoolkit_11_5 ];
           };
```

It seems that the "Test WASM" one break due to some path error, I am not entirely sure how github action handles current directory in different tasks, so I try to cd back now and see if it works. The nix changes are to upgrade to the new 22.05 so the emscripten is newer and probably can handle `EM_CACHE` properly (21.11 was deprecated).

makc:
Nah, see - [it passes now](https://github.com/elalish/manifold/runs/7271406633?check_suite_focus=true) that I brought assimp back

makc:
ah build_nix  is still no go

pca006132:
OK this is really weird. I will have a look at it tmr to see what is happening to GH Action...

makc:
Sure, I will just try to change `error: cudatoolkit_11_5 has been renamed to cudaPackages_11_5.cudatoolkit` this bit, and give up if it still fails

makc:
yeah that did not do anything 🤷

makc:
btw I think you can directly commit to this branch since PR references it

pca006132:
No, I don't have permission to commit to this:
```
ERROR: Permission to makc/manifold.git denied to pca006132.
fatal: Could not read from remote repository.

Please make sure you have the correct access rights
and the repository exists.
```

I'm now testing the changes on my repo: https://github.com/pca006132/manifold/pull/4

pca006132:
oh ok, forgot that assimp is required to build the tests, we cannot remove it for now.

pca006132:
You only need to add this commit: https://github.com/pca006132/manifold/pull/4/commits/0a4caaff32b5bd5d1935bf9edc42076e01658613

The problem with your previous commit is that you used `pkgs.cudatoolkit_11_5.cudatoolkit` instead of `pkgs.cudaPackages_11_5.cudatoolkit` which looks very similar but are actually different.

makc:
> No, I don't have permission to commit to this:

github lied to me
lol

pca006132:
I am a contributor but not a maintainer. I guess maintainer means having write permission to the repository.

pca006132:
sorry you need this as well: https://github.com/pca006132/manifold/pull/4/commits/e846a79ba040f6af82f8dd2df959a4ee6c35afed

I just tested the cuda compilation so it should work.

due to some weird dependencies error. somehow they packaged cuda 11.5 with gcc 10 while the compiler uses gcc 11 for other files, causing linker errors.

makc:
All green now!

super-green

makc:
A [thing](https://github.com/elalish/manifold/pull/151/files#diff-93b95633822d23c59559b98755ea2163389875310888c662886ae10250562893R68) to discuss here unrelated to the build process:
```cpp
    value_object<glm::ivec3>("ivec3")
        .field("0", &glm::ivec3::x)
        .field("1", &glm::ivec3::y)
        .field("2", &glm::ivec3::z)
```
it seemed like 0,1,2 make more sense than x,y,z for vertex index in a face, at the same time glm::vec3 and vec4 go with x,y,z,w notation so this does not seem consistent. I wish we could have both but (I think ?) if we do, emscripten will require js objects to have both, too.

makc:
an incomplete list of things to do (feel free to expand):
- decide 012 vs xyz notation above
- decide if we keep makeManifoldFromMesh/makeMeshFromManifold wrappers or let the users do new Module.Manifold(mesh)/manifold.GetMesh() and deal with exceptions themselves
- do we need `cd ../` [in yml](https://github.com/elalish/manifold/pull/151/files#diff-723c1253d0c94075d3d7c80504b08943c02209f016433937ccf803a3ca6eb5feR97)
- expose remaining boolean operators

pca006132:
For me, I think 0, 1, 2 for `ivec3` and x, y, z for `vec3` is perfectly fine as they are representing different concepts, the first one is merely an index while the second one has a geometric interpretation.

We don't need the `cd ../` in yaml, I added it because I thought the error was due to some weird GitHub CI thing, but it turns out that it is due to the unit tests not being built.

wasm/bindings.cpp
```diff
@@ -0,0 +1,110 @@
...
+    catch(const std::runtime_error& e) {
+        puts(e.what());
+
+        Mesh empty;
+        return empty;
+    }
+}
+
+void subtract(Manifold& a, Manifold& b) {
+    a -= b;
```
elalish, line 63: Probably better to stick to a more functional-style API and just return `a - b` with const inputs. I wish JS had operator overloading...

wasm/bindings.cpp
```diff
@@ -0,0 +1,110 @@
...
+    register_vector<glm::vec4>("Vector_vec4");
+
+    value_object<Mesh>("Mesh")
+        .field("vertPos", &Mesh::vertPos)
+        .field("triVerts", &Mesh::triVerts)
+        .field("vertNormal", &Mesh::vertNormal)
+        .field("halfedgeTangent", &Mesh::halfedgeTangent)
+        ;
+
+    function("makeEmptyMesh", &makeEmptyMesh);
```
elalish, line 100: Is there a reason this can't be the constructor of `Mesh`?

elalish (commented):
Nice teamwork! Do you want to build bindings for our whole API in this PR, or pick an MVP and add to it later?

wasm/bindings.cpp
makc: the thing is, this gets exposed as a's method. but well, we can do it just a function.

makc:
ok imma close this and come back with iteration 3 that will hopefully be merge-able.

pca006132:
@makc note that you can drop the last 4 commits related to nix because they are now included in #150

Issue #148, opened 2022-07-05
Triangulation error near precision limit


elalish:
@pca006132 I'm breaking you're repro out here, since I think it's separate from the issue it's in currently, but definitely worth looking into. I can also repro it locally, which is helpful.

Copied from https://github.com/elalish/manifold/issues/139#issuecomment-1171886763, which was inspired from https://github.com/elalish/manifold/issues/141#issuecomment-1171871261

```
TEST(Boolean, Close) {
  Manifold a = Manifold::Sphere(10, 256);
  Manifold result = a;
  for (int i = 0; i < 10; i++) {
    std::cout << i << std::endl;
    result ^= a.Translate({a.Precision()/10 * i, 0.0, 0.0});
    EXPECT_TRUE(result.IsManifold());
    EXPECT_TRUE(result.MatchesTriNormals());
  }
}
```

What I need to check: is this polygon actually epsilon-overlapped, or does the triangulator just have a bug? If it is overlapped, this may imply the precision is not well-bounded (due to nearly-parallel planes?), in which case it may be better to look into making the triangulator robust to overlapping input rather than intentionally failing.

elalish:
The first problem polygon is not overlapped, but it is entirely colinear within precision. It's an error in my handling of uncertain East/West positioning. Once I fix that we'll see if this test exposes any other errors.

pca006132:
Some thoughts about tracking precision:

If we consider the precision as 2 terms: inherent error and floating point error, the updated precision after transformation `M` should probably be `oldPrecision * det(M), kTolerence*bbox.Scale()` because the inherent error is scaled according to the transformation and is independent of the floating point error, while the floating point error is independent of the percentage error in the original model.

The current way of scaling the precision by `glm::max(1.0f, newScale / oldScale)` looks a bit weird to me: consider a line `(0, 0, 0)` to `(0, 0, 1)` with precision 0.01, if we transform it to `(0, 0, -1)` to `(0, 0, 1)`, the precision should be 0.02, but the current calculation will only give 0.01 because the maximum absolute coordinate is still 1. Also, when we scale the model down, we should get better precision.

pca006132:
Also, when we perform boolean operations, should we take the precision into account? It seems that the boolean operation code do not use precision when checking for overlaps, including both the shadow code and the bounding box overlap check.

elalish:
[...]

PR #152, opened 2022-07-14
Precision calculation

pca006132:
Modified precision calculation. Instead of scaling the precision using change in max coordinate, we scale the precision according to the transformation matrix, taking the max along three axis.

### DIFF
--- a/manifold/src/impl.cpp
+++ b/manifold/src/impl.cpp
@@ -506,12 +506,12 @@ Manifold::Impl Manifold::Impl::Transform(const glm::mat4x3& transform_) const {
   // axis-aligned.
   if (!result.collider_.Transform(transform_)) result.Update();
 
-  const float oldScale = result.bBox_.Scale();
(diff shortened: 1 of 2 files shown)


fire:
Does this fix the tests that were broken previously? What is the effect of this?

pca006132:
No, this doesn't fix those broken tests. This just makes sure that the precision calculation is correct, e.g. for the new test that I added. The broken test in #148 requires work on the triangulator.

elalish (approved):
Thank you, that makes much more sense than what I had. Honestly not sure what I was thinking there...

pca006132:
Weird, the CI failed after merging. Is the compiler version in GitHub Action changed or something else?

elalish:
Yeah, I just noticed that. No idea. Want to open a trivial PR and see if it repros?

pca006132:
Maybe you can update your copyright update PR to check that?

Issue #138, opened 2022-06-16
Remove thrust


fire:
To make manifold easier to build there was a proposal to remove thrust for C++ primitives.

This is tracking that thrust removal proposal.

pca006132:
In fact what we are doing with https://github.com/elalish/manifold/blob/master/utilities/include/par.h is building a proxy for some of the thrust APIs, although there are still some stuff like thrust transform iterators elsewhere in the code. I think we can make it a compile time option to use stl implementation instead of thrust and use `std::execution::par` if the compiler support it, so users who don't want to build the CUDA variant don't have to pull in the thrust dependencies.

I think it will take a long time for stl to support GPU offloading for parallel algorithms, and I doubt if the performance will be portable. In our code, we often switch between CPU and GPU processing, and we use CUDA unified memory to try to minimize data transfer. For a general implementation, e.g. based on OpenCL, iirc there is no unified memory available.

fire:
I don't think the game industry uses opencl. There seems to be move towards technologies like onnx and vulkan compute.

elalish:
I believe someone here already got Manifold compiling on Apple Macs with the Thrust C++ backend. The nice thing about what @pca006132 did switching everything over to .cpp files and pulling in thrust as a proper third_party dep is that it works on a bunch of different compilers now, no longer needing nvcc or CUDA, since Thrust is just a C++ header library. The only thing you need an Nvidia GPU for is if you want the CUDA backend.

fire:
The other issues worth mentioning:

- [ ] I made a script with degit to copy the dependencies without git submodules https://github.com/V-Sekai/godot/blob/csg-manifold-07/modules/csg/checkout.sh
- [ ] Need to remove exceptions from thrust since they cause warnings and code with warnings are not mergable.

pca006132:
For removing exceptions, can you open a separate issue to track that? I think I can work on that as I have experience using rust, which does not use exceptions. I think the good way to do this might be to abort on some invariant check (assertion fault is considered a bug on the library but not user input error), and data validation failure will return some error code.

Removing exceptions from thrust would be a bit tricky though, I guess the only way to do it is to write a proxy module that provides those iterator helpers that we use, and use a compile-time option to decide whether or not to include thrust.

elalish:
[...]

PR #153, opened 2022-07-16
misc changes for thrust

pca006132:
1. Fix wasm build error.
2. Use std implementation where possible, so we can make thrust an optional dependency later (#138)

Problems:
1. We need some iterator library that is compatible with thrust in order to make thrust optional. The problem with https://github.com/dpellegr/ZipIterator is that it does not work with some thrust functions, but I'm not sure why, perhaps due to not implementing random access iterator? I haven't yet dig into it. And we need implementation for permutation iterator, transform iterator, etc.
2. Currently, the type of zip iterator has to be `thrust::zip_iterator<thrust::tuple<...>>` as they require something like `tail_type` in their zip iterator implementation which is not provided by `std::tuple`. This forces all functors to accept `thrust::tuple` and use thrust functions like `thrust::get` everywhere. We should be able to fix this by finding some other implementation of zip iterator.

### DIFF
--- a/thrust.diff
+++ b/thrust.diff
@@ -1,0 +1,49 @@
+diff --git a/third_party/thrust/thrust/detail/type_traits.h b/third_party/thrust/thrust/detail/type_traits.h
+index d147f832..60b57654 100644
+--- a/third_party/thrust/thrust/detail/type_traits.h
++++ b/third_party/thrust/thrust/detail/type_traits.h
+@@ -144,11 +144,11 @@ template<typename T> struct has_trivial_constructor
+       is_pod<T>::value
+ #if THRUST_HOST_COMPILER == THRUST_HOST_COMPILER_MSVC || \
+     THRUST_HOST_COMPILER == THRUST_HOST_COMPILER_CLANG
+-      || __has_trivial_constructor(T)
++      || __is_trivially_constructible(T)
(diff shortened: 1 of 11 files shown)


thrust.diff
```diff
@@ -0,0 +1,32 @@
+diff --git a/third_party/thrust/thrust/detail/type_traits.h b/third_party/thrust/thrust/detail/type_traits.h
+index d147f832..60b57654 100644
+--- a/third_party/thrust/thrust/detail/type_traits.h
++++ b/third_party/thrust/thrust/detail/type_traits.h
+@@ -144,11 +144,11 @@ template<typename T> struct has_trivial_constructor
+       is_pod<T>::value
+ #if THRUST_HOST_COMPILER == THRUST_HOST_COMPILER_MSVC || \
+     THRUST_HOST_COMPILER == THRUST_HOST_COMPILER_CLANG
+-      || __has_trivial_constructor(T)
++      || __is_trivially_constructible(T)
```
elalish, line 10: Do we think thrust has a bug? Any idea what changed on the CI; can you repro locally?

pca006132:
oh the patch file is wrong, will fix tmr

thrust.diff
pca006132: they use a deprecated API: https://github.com/NVIDIA/thrust/issues/1743

elalish (approved):
Thanks!

PR #155, opened 2022-07-18
wasm

makc:
Yet another iteration on #151
Cleaned up bindings.cpp, added all three operations to 3js demo.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -9,6 +9,7 @@ set(MANIFOLD_PAR "TBB" CACHE STRING "Parallel backend, either \"TBB\" or \"OpenM
 
 if(EMSCRIPTEN)
     message("Building for Emscripten")
+    set(MANIFOLD_FLAGS -fexceptions)
     set(CMAKE_EXE_LINKER_FLAGS ${CMAKE_EXE_LINKER_FLAGS} -sALLOW_MEMORY_GROWTH=1)
     set(MANIFOLD_USE_CUDA OFF)
     set(MANIFOLD_PAR "NONE")
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -78,3 +78,7 @@ add_subdirectory(meshIO)
 add_subdirectory(samples)
 add_subdirectory(test)
 add_subdirectory(tools)
+
+if(EMSCRIPTEN)
+add_subdirectory(wasm)
+endif()
--- a/flake.nix
+++ b/flake.nix
@@ -97,6 +97,8 @@
               nativeBuildInputs = (with pkgs; [ cmake python38 ]);
               buildInputs = [ pkgs.nodejs ];
               configurePhase = ''
+                mkdir -p .emscriptencache
+                export EM_CACHE=$(pwd)/.emscriptencache
                 mkdir build
                 cd build
                 mkdir cache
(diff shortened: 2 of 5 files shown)


makc:
@pca006132 build_nix is not green again, halp

pca006132:
```
error: undefined symbol: __cxa_is_pointer_type (referenced by $CatchInfo__deps: ['$ExceptionInfo','__cxa_is_pointer_type'], referenced by __cxa_find_matching_catch_2__deps: ['$exceptionLast','$ExceptionInfo','$CatchInfo','__resumeException','__cxa_can_catch'], referenced by top-level compiled C/C++ code)
warning: ___cxa_is_pointer_type may need to be added to EXPORTED_FUNCTIONS if it arrives from a system library
```

ah interesting

pca006132:
Related: https://github.com/emscripten-core/emscripten/issues/14313

I guess we somehow disabled linking exceptions in the build script?

makc:
build_wasm is fine though. so you say this is because build_nix uses newer emscripten?

pca006132:
Fix:

```diff
diff --git a/CMakeLists.txt b/CMakeLists.txt
index 27128dc..d7373fe 100644
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -9,6 +9,7 @@ set(MANIFOLD_PAR "TBB" CACHE STRING "Parallel backend, either \"TBB\" or \"OpenM
 
 if(EMSCRIPTEN)
     message("Building for Emscripten")
+    set(MANIFOLD_FLAGS -fexceptions)
     set(CMAKE_EXE_LINKER_FLAGS ${CMAKE_EXE_LINKER_FLAGS} -sALLOW_MEMORY_GROWTH=1)
     set(MANIFOLD_USE_CUDA OFF)
     set(MANIFOLD_PAR "NONE")
```

makc:
pure green

elalish (approved):
Excellent! Thanks for putting in the effort here, and the API looks great. As a follow-on, we should make sure to serve your index.html file on our Github pages so we can link it in our docs. That will make a compelling example to see live!

makc:
@elalish btw, is there an example of using [this](https://github.com/elalish/manifold/wiki/Manifold-Library#vertex-properties) for applying uvs in c++ ?

elalish:
Not really; the closest I have at the moment is [this](https://github.com/elalish/manifold/blob/master/test/mesh_test.cpp#L62-L85). The reason is that this is so dependent on your engine/format. Say you union two models with different textures. Do you atlas the images together and offset the UVs? Which way do you want to do that? What if the two models don't have the same texture channels? I give you all the math you need, but that's just the beginning. I think now that you've got a three.js example hooked up, this would be a good time to make a simple demo of atlasing textures. Or I suppose we could just attach copies of a single model so that they can all share the same texture - that's by far the simplest.

makc:
to avoid changing uv layout, in the context of 3js there is an option to assign different materials to areas that used to belong to different operands. in another lib, the user could split the result into multiple meshes. I suppose I would need to step-by-step those tests to understand GetMeshIDs() and GetMeshRelation()

elalish:
Yes, that's a good point, I forgot about their multiple material support. I believe it involves duplicating verts (pretty unavoidable in graphics) which makes it hard to come back to a manifold. However, my APIs contain all the info so we can apply the material stuff after all the operations have completed, so it should be fine. Would you like to start a PR? I'd be happy to help with the `MeshRelation` part.

PR #159, opened 2022-07-21
WIP: demo website

pca006132:
#158 

This contains a commit to upload the wasm related files, so the deploy task can download these files. @elalish How should I modify the deploy task to use the wasm files?

### DIFF
--- a/.github/workflows/manifold.yml
+++ b/.github/workflows/manifold.yml
@@ -100,6 +100,14 @@ jobs:
       run: |
         cd build/test
         node ./manifold_test.js
+    - name: 'Upload WASM files'
+      uses: actions/upload-artifact@v3
+      with:
+        name: demo
+        path: |
+          build/bindings/wasm/manifold.js
+          build/bindings/wasm/manifold.wasm
+        retention-days: 1
 
   build_windows:
     strategy:


.github/workflows/manifold.yml
```diff
@@ -100,6 +100,15 @@ jobs:
       run: |
         cd build/test
         node ./manifold_test.js
+    - name: 'Upload WASM files'
+      uses: actions/upload-artifact@v3
+      with:
+        name: demo
+        path: |
+          build/bindings/wasm/index.html
```
elalish, line 108: I don't think we need to upload this, since it's already served by `deploy.sh`: see here (the page is blank, since it can't find `manifold.js`): https://elalish.github.io/manifold/bindings/wasm/examples/index.html

.github/workflows/manifold.yml
```diff
@@ -100,6 +100,15 @@ jobs:
...
+          build/bindings/wasm/manifold.js
+          build/bindings/wasm/manifold.wasm
```
elalish, line 110: Okay, so running build causes everything to build, so I guess the idea is that we get both a manifold-test.wasm (with assimp, etc) and a manifold.wasm without and just JS bindings. That sounds good.

.github/workflows/manifold.yml
```diff
@@ -100,6 +100,15 @@ jobs:
       run: |
         cd build/test
         node ./manifold_test.js
+    - name: 'Upload WASM files'
+      uses: actions/upload-artifact@v3
```
elalish, line 104: I guess we just need to add `actions/download-artifact` to `deploy.yml` and then add that output path to this list: https://github.com/elalish/manifold/blob/master/deploy.sh#L23, which copies is to `docs/html`, which in turn becomes the root of our served github pages. I guess the shell script will need to take a couple of input paths.

pca006132: Yes, I think maybe you can modify the `deploy.yml` as I'm not familiar with this.

elalish: Sure; I can take over this PR if you like. Thanks for pointing me to that action.

PR #160, opened 2022-07-21
Make triangulator exceptions optional

### LINKED ISSUE #148: Triangulation error near precision limit

elalish:
Fixes #148 

The Boolean.Close test is no longer disabled. I added an execution param: processOverlaps, which tells the triangulator to be robust to overlapping input instead of throwing exceptions. The danger of course is that the triangulator can't actually do anything geometrically "good" with overlapped input, so it'll just make a basically arbitrary manifold triangulation. This has the potential to make small overlaps much worse, but it also solves this issue of small pieces of geometry getting past their precision limit. 

I've looked at the output sphere; it seems generally fine (of course lots of small triangles along the diameter where the intersections are taking place). It's a bit hard to tell how overlapped they are, but hopefully its decent for most uses. Also, it's a pretty extreme input in the first place.

This now sets the stage for having a compile-time option to remove exceptions entirely. 

One strange thing: I cut the test down from 10 to 6, because on 6 it was failing with `munmap_chunk(): invalid pointer`. I tried running valgrind, but it was so slow I gave up. We should look into this further; if we can manage to output the problem polygon, I should be able to debug it.

### DIFF
--- a/src/polygon/src/polygon.cpp
+++ b/src/polygon/src/polygon.cpp
@@ -28,6 +28,36 @@ using namespace manifold;
 
 ExecutionParams params;
 
+bool OverlapAssert(bool condition, const char *file, int line,
+                   const std::string &cond, const std::string &msg) {
+  if (!params.processOverlaps && !condition) {
+    std::ostringstream output;
+    output << "Error in file: " << file << " (" << line << "): \'" << cond
+           << "\' is false: " << msg;
+    throw geometryErr(output.str());
+  }
+  return condition;
+}
+
+/**
+ * Only use directly inside of the SweepForward() and SweepBack() functions! If
+ * the asserted condition is false, it implies the monotone subdivision has
+ * failed. This is most likely due to the input polygons being overlapped by
+ * more than the input precision, but if not, then it indicates a bug. Either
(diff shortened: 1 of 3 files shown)


pca006132:
I will try to look into the invalid pointer issue.

PR #166, opened 2022-07-26
remove docker image

elalish:
I think this docker image is pretty out of date, and also not really needed anymore.


pca006132:
@elalish do you have any idea about why the CI is failing after this PR is merged? It seems that the cmake step failed.

elalish:
No, it didn't really fail; the only problem on the master build was that the code coverage server failed to respond. In my next PR I've set that to not fail the CI, so it should be fine now.

PR #170, opened 2022-07-28
Make exceptions optional

### LINKED ISSUE #141: Removing exceptions

elalish:
Fixes #141 

I've introduced a compiler flag: `MANIFOLD_DEBUG` that enables exceptions, as well as error and info printing. It is off by default, which means not only are the exceptions compiled out, but it also keeps `iostream` and such from being included anywhere. This is to help remind us to hide messages behind this flag, as it will fail to compile otherwise. 

I also added `ManifoldParams()`, which allows us to control  the verbosity of the Boolean operations at runtime (assuming our debug flag is set). I chose to add a new flag instead of using `NDEBUG` because that flag does not play well with Thrust and makes things run too slowly. 

I've also changed the constructors that take an input `Mesh` to return an empty `Manifold` if the input data is bad. I also added `Status()` to the API to return an enum of the problem with the input `Mesh` if the result was empty. I also replaced some other small user-input asserts with data sanitation so that the program doesn't have to abort. 

I also added a bunch of tests about input validation and fixed a few bugs those revealed.


fire:
Amazing!

elalish:
@pca006132 I'm still trying to figure out why I'm getting a seg fault in the boolean when I have exceptions turned off. I suspect my macros are misbehaving. I also saw [this](https://google.github.io/styleguide/cppguide.html#Preprocessor_Macros) in Google's C++ style guide. They don't like macros in header files, which is exactly how I've been doing my asserts. I hadn't considered that would escape into our public API. Can you think of a better way?

elalish:
Okay, I think I've made sure our `ASSERT` macro is no longer part of our public header. Still pretty sure I'm not doing this all properly yet...

pca006132:
> I'm still trying to figure out why I'm getting a seg fault in the boolean when I have exceptions turned off.

Do you mean when the assertion is evaluated to false (which will throw an exception when exception is enabled), the program is aborted? When exception is disabled, throw is not a nop, but calls `abort`. https://gcc.gnu.org/onlinedocs/libstdc++/manual/using_exceptions.html

Sorry I have been relatively busy recently, I will try to find some time to look into this.

src/utilities/include/optional_assert.h
```diff
@@ -0,0 +1,39 @@
...
+    std::ostringstream output;
+    output << "Error in file: " << file << " (" << line << "): \'" << cond
+           << "\' is false: " << msg;
+    throw Ex(output.str());
+  }
+}
+#define ASSERT(condition, EX, msg) \
+  Assert<EX>(condition, __FILE__, __LINE__, #condition, msg);
+#else
+#define ASSERT(condition, EX, msg)
```
elalish, line 38: @pca006132 Replying to you here to give context. When `MANIFOLD_DEBUG` is not defined, the `ASSERT` macro (which replaced `ALWAYS_ASSERT`) should compile out completely so there are no `throw` statements left. However, it now seg faults, so I think my macro is expanding in a way I don't expect.

elalish: Nevermind, I found the problem. I had a side-effect inside one of my `ASSERT` calls :roll_eyes:

PR #171, opened 2022-08-02
re-enabled reordering

### LINKED ISSUE #114: feature request: lazy boolean operations
I guess it might be better to open an issue for this and discuss about the possible implementation for lazy boolean operations. The opt branch of my fork only implements the optimizations for union.

Idea: Apply boolean operations lazily to allow more optimizations, e.g. operation reordering or cheap batch processing for certain operations.

Operation reordering: 
1. Permutation for union and intersection. The idea is to work on manifolds that have smaller number of vertices/faces first. For larger manifolds, the time required for boolean operations is about linear to the number of vertices/faces, so work on manifolds that have smaller number of vertices/faces can reduce the time required for the operation.
2. De Morgan's Laws (?): `a - b - c <=> a - (b + c)`. Again, the idea is to minimize the number of required operations. For this case, `b` and `c` might be disjoint, if so they can be composed together cheaply. If not, the LHS is perhaps the cheapest way to do this.

Batch processing:
1. Use compose instead of union for a set of pairwise disjoint manifolds. Although checking for pairwise disjointness can take `O(n^2)` time with a simple implementation, it is still much cheaper in practice than doing union for up to something like 10000 manifolds.
2. Similarly, we can do batch processing for intersection by discarding disjoint manifolds.

Cache for manifolds that only differ in transformation: We can use a `shared_ptr` of the underlying CSG tree, and store the transformation separately. When we need to evaluate the CSG tree, e.g. when exporting the object or need to apply some complicated operations on it, we can cache the result and then apply transform to a copy of the result. This way, we can store objects that only differ in transformation (e.g. a grid of objects) cheaply, and allows us to defer the evaluation of the CSG tree as late as possible. This way, copying the manifold is also essentially free: we only need to copy the `shared_ptr` and the 4x3 matrix.

pca006132:
This reverts ecd9329 and make `BatchUnion` use `BatchBoolean` instead of `Compose`, as there are additional issues with `Compose`. Will finally close #114 after the triangulation problem is resolved.

Currently, the `Samples.Sponge4` test fails with

```
Triangulation failed! Precision = 1e-05
Error in file: /home/pca006132/code/manifold/src/polygon/src/polygon.cpp (1019): 'std::all_of(triangles.begin(), triangles.end(), [&vertPos, precision](const glm::ivec3 &tri) { return CCW(vertPos[tri[0]], vertPos[tri[1]], vertPos[tri[2]], precision) >= 0; })' is false: triangulation is not entirely CCW!
```

### DIFF
--- a/src/manifold/src/csg_tree.cpp
+++ b/src/manifold/src/csg_tree.cpp
@@ -259,29 +259,39 @@ std::shared_ptr<CsgLeafNode> CsgOpNode::ToLeafNode() const {
   if (children_.empty()) return nullptr;
   // turn the children into leaf nodes
   GetChildren();
-  Manifold::OpType op;
   switch (op_) {
     case CsgNodeType::UNION:
-      op = Manifold::OpType::ADD;
+      BatchUnion();
       break;
-    case CsgNodeType::DIFFERENCE:
-      op = Manifold::OpType::SUBTRACT;
+    case CsgNodeType::INTERSECTION: {
+      std::vector<std::shared_ptr<const Manifold::Impl>> impls;
+      for (auto &child : children_) {
+        impls.push_back(
+            std::dynamic_pointer_cast<CsgLeafNode>(child)->GetImpl());
+      }
+      BatchBoolean(Manifold::OpType::INTERSECT, impls);
+      children_.clear();
+      children_.push_back(std::make_shared<CsgLeafNode>(impls.front()));
       break;
-    case CsgNodeType::INTERSECTION:
-      op = Manifold::OpType::INTERSECT;
-      break;
-    default:
-      throw std::runtime_error("unreachable CSG operation");
+    };
+    case CsgNodeType::DIFFERENCE: {
+      // take the lhs out and treat the remaining nodes as the rhs, perform
+      // union optimization for them
+      auto lhs = std::dynamic_pointer_cast<CsgLeafNode>(children_.front());
+      children_.erase(children_.begin());
+      BatchUnion();
+      auto rhs = std::dynamic_pointer_cast<CsgLeafNode>(children_.front());
+      children_.clear();
+      Boolean3 boolean(*lhs->GetImpl(), *rhs->GetImpl(),
+                       Manifold::OpType::SUBTRACT);
+      children_.push_back(
+          std::make_shared<CsgLeafNode>(std::make_shared<Manifold::Impl>(
+              boolean.Result(Manifold::OpType::SUBTRACT))));
+    };
+    case CsgNodeType::LEAF:
+      // unreachable
       break;
   }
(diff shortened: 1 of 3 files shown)


pca006132:
Ah and interestingly the `Boolean.Close` is causing some problem. Will look into that.

pca006132:
It seems that it is stuck in an infinite loop in `FormLoop` in the 7th iteration.

elalish:
I upgraded our debug mode, so you should be able to run `./manifold_test --gtest_filter=Sponge4 -v` to get some more details; would probably be interesting to look at the difference between your branch and master. How much does it rearrange the three ops? Is it still three differences, or is it a difference of unions now?

elalish:
Well, I was hoping #179 would solve this problem, but it seems to still exist after merging. It's definitely a related issue though - I'll keeping looking into it.

elalish:
Okay, it looks like we've made progress with #180 and #182; we no longer have any infinite loops or bad triangulations. Now it's just that Sponge4 has 38k degenerate triangles, which is not great. Any thoughts as to why that might be?

pca006132:
Not sure if this is an issue with the f3d viewer or something, I see some weird lines in the triangulation which might explain the degenerate triangles:
image

elalish:
Okay, I've got a repro on master now, #183. I must say, this has been a great PR for testing!

pca006132:
Indeed. After this is merged we can try to enable usinf compose for union which also need some fixes. Also, I can start working on the fuzzing/property-based testing, which will probably expose more subtle bugs.

elalish:
It looks like this PR isn't breaking anything, but just exposing some degenerate triangle removal difficulties already present on master. That's fine, but before we start reordering the operations, can you think of a way that we can cache intermediate results that are reused? We don't have any great examples of this in the samples right now, but it's pretty common in design. I really don't want to force users to trigger intermediate evaluation manually.

pca006132:
I think intermediate results are already cached? Users only need to trigger intermediate evaluation manually if they want to debug or need to do some benchmarking.

Btw, should we modify the `NumDegenerateTris` values for `Boolean.Gyroid` and `Samples.Sponge4` to fix the broken tests? And then open an issue to track the degenerate triangle removal issue.

elalish:
Interesting; I was referring to [this](https://github.com/elalish/manifold/pull/131#discussion_r889637032), but I guess I'm not quite sure how common that behavior is. Maybe it's not a big deal; I should really write some more samples. 

Yes, let's go ahead and update the tests and merge this. :+1:

pca006132:
Oh, because we are not using compose for now, that behavior will not occur. For compose, it may be beneficial because we may be able to get more non-intersecting manifolds, but it is not beneficial when we do boolean directly.

elalish (approved):
Coverage is looking good :+1:

PR #173, opened 2022-08-03
Do nothing if the half edge array is empty.

fire:
Let's see if this fixes the bug.

### DIFF
--- a/src/manifold/src/edge_op.cpp
+++ b/src/manifold/src/edge_op.cpp
@@ -135,6 +135,10 @@ namespace manifold {
  * removal, by setting vertPos to NaN and halfedge to {-1, -1, -1, -1}.
  */
 void Manifold::Impl::SimplifyTopology() {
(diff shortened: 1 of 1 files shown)


elalish:
What bug was this fixing? Can you add a test case to demonstrate?

elalish (changes requested):
What bug is this fixing?

fire:
I don't have the test case on hard, it was with an empty manifold.

elalish:
Can you please add a test so we know what problem we're fixing? You should be able to copy this one and tweak slightly: https://github.com/elalish/manifold/blob/master/test/mesh_test.cpp#L608

fire:
I'm not able to work on this much because of siggraph 2022 this week.

elalish:
Got it, no prob. I'm curious, mostly :)

fire:
I wasn't able to get myself to work on this. Godot Engine 4 work has been tiring.

PR #175, opened 2022-08-04
Revert "use std if possible"

### LINKED ISSUE #172: Seg fault on CUDA
Ugh, I really wish our CI had a GPU. I just realized I've been neglecting to test on CUDA locally, and honestly I'm not sure for how long. However, today I noticed that `Manifold.SmoothSphere` gives: `for_each: failed to synchronize: cudaErrorIllegalAddress: an illegal memory access was encountered" thrown in the test body.` and then `Manifold.ManualSmooth` seg faults, but just on CUDA.

I went back to #157 and found the same problem there. I'm having some trouble getting things to build before that, probably because I need different flags or something. @pca006132 Can you repro?

elalish:
Fixes #172

This reverts commit 1e1fcf2dad4f0d7e57c10e61bf09dbb37356cfcf, from #153. This was the cause of CUDA failing to run our tests (weird memory errors and such). I noticed because there were some warnings in the compiler about not calling host functions from device functions regarding `std::swap` and such. I recall that the STL does not play well with device code, which pointed me to your PR, @pca006132. Sorry I forgot to test CUDA for you on that change. 

Interestingly, on CUDA I'm now seeing the infinite loop on `Boolean.Close` that you mentioned in #171, so I think that's not your PR's fault. I'll see if I can debug it here.

### DIFF
--- a/src/collider/include/collider.h
+++ b/src/collider/include/collider.h
@@ -36,7 +36,7 @@ class Collider {
   VecDH<Box> nodeBBox_;
   VecDH<int> nodeParent_;
   // even nodes are leaves, odd nodes are internal, root is 1
-  VecDH<std::pair<int, int>> internalChildren_;
+  VecDH<thrust::pair<int, int>> internalChildren_;
 
   int NumInternal() const { return internalChildren_.size(); };
   int NumLeaves() const { return NumInternal() + 1; };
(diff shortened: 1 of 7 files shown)


pca006132:
Interesting, I thought std functions are usually template functions and can be handled by CUDA as if they were CUDA code. Anyway this should be fine, I was just trying to reduce the dependency on thrust so we can potentially make it optional later, not a must for now.

PR #184, opened 2022-08-11
Changed MeshRelation API

### LINKED ISSUE #183: reordered sponge4 failure
I managed to repro the problem from #171 on master by manually reordering the sponge operations. Many different orders worked fine, but this one, which is most similar to the result of the other PR, did not:
```
Manifold tmp1 = hole.Rotate(90) + hole.Rotate(90).Rotate(0, 0, 90);
Manifold tmp2 = hole + tmp1;
result = result - tmp2;
```
Not only did it come up with 39k degenerate triangles, it also ended up with the wrong genus. It's pretty clear there's a problem:
Screenshot from 2022-08-09 09-35-51

elalish:
Fixes #183 

So, this fixes the main problem, which was that the meshIDs were wrong and thus were causing incorrect decimation (hence wrong genus and shape). Actually, in certain circumstances, they were wrong enough that it caused a crash. @pca006132 I ended up refactoring the UpdateMeshID stuff that you made to simplify it. See below for some details.


samples/src/menger_sponge.cpp
```diff
@@ -54,6 +54,12 @@ Manifold MengerSponge(int n) {
   result -= hole;
   hole = hole.Rotate(0, 0, 90);
   result -= hole;
+
+  // Alternative order causes degenerate triangles
+  // Manifold tmp1 = hole.Rotate(90) + hole.Rotate(90).Rotate(0, 0, 90);
+  // Manifold tmp2 = hole + tmp1;
+  // result = result - tmp2;
```
elalish, line 61: I'm leaving this here because it was the order that reproed the problem in #171 where all this was first noticed. It actually does still have 38k degenerate triangles, but the rest is correct now, so I'll look into that next.

src/manifold/include/manifold.h
```diff
@@ -118,7 +118,7 @@ class Manifold {
    */
   ///@{
   MeshRelation GetMeshRelation() const;
-  std::vector<int> GetMeshIDs() const;
+  int OriginalID() const;
```
elalish, line 121: Here's the first big API change. Again, I think I'm going in the right direction here because it simplifies our tests. If they want to know the total set, they can scan triBary as easily as we can.

src/utilities/include/structs.h
```diff
@@ -222,8 +222,13 @@ struct Curvature {
...
  */
 struct BaryRef {
-  /// The original mesh triangle index.
+  /// The unique ID of the mesh instance of this triangle. If .meshID and .tri
+  /// match for two triangles, then they are coplanar and came from the same
+  /// face.
   int meshID;
+  /// The OriginalID of the mesh this triangle came from. This ID is ideal for
+  /// reapplying properties like UV coordinates to the output mesh.
+  int originalID;
```
elalish, line 231: Here's the other API change: now we publicly store both the originalID (what meshID was before) and the meshID, which is now a mesh instance ID. The nice thing about exposing this, is that you can tell if triangles can be merged into the same polygon if their `meshID` and `tri` match, which is the same system used by our decimator.

.vscode/c_cpp_properties.json
```diff
@@ -3,19 +3,32 @@
         {
             "name": "Mac",
             "includePath": [
-                "${workspaceFolder}/**"
```
pca006132, line 6: I think it may be better to put `.vscode` into `.gitignore` as paths here are specific to each computer.

src/manifold/src/boolean_result.cpp
```diff
@@ -446,9 +444,10 @@ struct CreateBarycentric {
     const BaryRef oldRef = halfedgeRef.PQ == 0 ? triBaryP[tri] : triBaryQ[tri];
 
     faceRef[halfedgeR.face] = oldRef;
-    if (halfedgeRef.PQ == 0) {
-      // Mark the meshID as coming from P
-      faceRef[halfedgeR.face].meshID |= kFromP;
+
+    if (halfedgeRef.PQ == 1) {
+      // Mark the meshID as coming from Q
+      faceRef[halfedgeR.face].meshID += offsetQ;
```
pca006132, line 450: Just wondering what is the cause of the wrong meshID using bitwise or. Is it wrong because we have too many different meshes? Why adding `meshIDCounter_` is fine instead?

src/manifold/src/csg_tree.cpp
```diff
@@ -210,26 +210,14 @@ Manifold::Impl CsgLeafNode::Compose(
               combined.halfedge_.begin() + nextEdge,
               UpdateHalfedge({nextVert, nextEdge, nextTri}));
 
-    // Assign new IDs to triangles added in this iteration, to differentiate
```
pca006132, line 213: The `compose` function is applying transformations lazily, so different meshes may have the same meshID. Consider composing `A` and `A.Translate(...)`, we will add vertices in `A` two times, but in the second time we apply a transformation while copying the vertices. Not sure if it is sound to just perform one meshID update at the end.

src/manifold/src/boolean_result.cpp
elalish: I don't think that was a problem - the error was in the `UpdateMeshIDs` function, which was hitting its assertion, and that was because it on an earlier iteration it was creating an empty set of original IDs. I decided to refactor instead of debugging it and this simplified my use of an index vector.

src/manifold/src/csg_tree.cpp
elalish: This is an excellent point, I clearly didn't think this through. Will fix.

elalish: Fixed.

PR #189, opened 2022-09-01
Make MeshIO optional for tests

elalish:
Now that we have an SDF for generating complex manifolds, we no longer need to read files for our tests. This means I removed the `test/data` folder, but it also means the `MeshIO` dependency is optional. I put in a new compile flag: `MANIFOLD_EXPORT` which optionally links in `MeshIO` in order to allow the `-e` option for the tests to generate `.glb`s. 

This means we'll soon be able to remove all the custom building of Assimp since we won't need it for the CI. Basically coming back to #123.


elalish:
@pca006132 Any thoughts as to why this would make the WASM builds fail? I was going to do the actual removal of Assimp from third_party in a follow-on PR, but it seems like it's having some trouble building Assimp already, which is odd.

test/test_main.cpp
```diff
@@ -42,6 +42,11 @@ int main(int argc, char **argv) {
         print_usage();
         return 0;
       case 'e':
+#ifndef MANIFOLD_EXPORT
+        printf(
```
pca006132, line 46: maybe print to stderr?

pca006132:
> @pca006132 Any thoughts as to why this would make the WASM builds fail? I was going to do the actual removal of Assimp from third_party in a follow-on PR, but it seems like it's having some trouble building Assimp already, which is odd.

Because we removed the `data` directory but still ask emscripten to put the content inside that directory into the bundle. We can change the line https://github.com/elalish/manifold/blob/11e58ea4d27508f2f6468741d3f392ab295c8658/test/CMakeLists.txt#L40 to 
```
"-s ASSERTIONS=1 -s DEMANGLE_SUPPORT=1 --bind")
```

test/test_main.cpp
elalish: Fair, but I think since the tests will still run successfully it's more of a warning than an error.

elalish:
Excellent, thanks @pca006132!

PR #190, opened 2022-09-01
Removing export from python binding

### LINKED ISSUE #128: Separate meshIO from manifold bundles
`meshIO` is not really a core part of this Manifold library. It is the only part that depends on Assimp (a large, complicated dependency) and it is unlikely to fit into other libraries well, as they probably have their own mesh input/output anyway. I only included `meshIO` as a means to quickly test this library, but the functionality is lacking (no texture/property support for one thing). I don't want to focus on that, as it's a complicated area in its own right and others do it better. For C++, it's easy enough to not include that part, but for building bundles like WASM and Python, I think it's important to keep this separate. 

I'm fine with either removing `meshIO` from these bundles to focus only on our core library, or if we find it to be really useful, then making a separate bundle for `meshIO`. There is no dependency between `meshIO` and `manifold` since they communicate via `Mesh`, so the separation should be pretty clean.

elalish:
Fixes #128 

I found a nice little python library for import/export of various mesh formats: [meshio](https://github.com/nschloe/meshio). By removing this last dependency, we can separate our library completely from Assimp. However, this isn't yet functional because it seems the `to_mesh` function isn't working. I get the following Python error when I try `python3 run_all.py -e`:
```
TypeError: Unregistered type : glm::vec<3, int, (glm::qualifier)0>

The above exception was the direct cause of the following exception:

Traceback (most recent call last):
  File "run_all.py", line 38, in <module>
    cells = [("triangle", mesh.tri_verts)]
TypeError: Unable to convert function return value to a Python type! The signature was
	(self: pymanifold.Mesh) -> List[glm::vec<3, int, (glm::qualifier)0>]
```
I think we need to convert our `std::vector<glm::vec3>` into a two-dimensional python `list`, but I'm not sure how to do that. Any thoughts @pca006132?

UPDATE: working now.


pca006132:
We can use numpy array. https://pybind11.readthedocs.io/en/stable/advanced/pycpp/numpy.html

More specifically, it should be something similar to
```
https://github.com/elalish/manifold/blob/11e58ea4d27508f2f6468741d3f392ab295c8658/bindings/python/pymanifold.cpp#L188-L218
```
but we should flatten the buffer instead.

elalish:
@pca006132 Thanks, I think I've got all the python stuff working now, following your suggestion. I tried to install the python `meshio` package for nix (they seem to have it), but I think I'm just not familiar enough with the syntax. Would you mind giving it a look?

pca006132:
Try this:
```
python39.withPackages(ps: with ps; [meshio])
```

it seems that they don't have it for 3.8.

elalish:
Perfect, thanks!

elalish:
@rsaccon This improved our python bindings considerably, which should make it a lot easier to integrate with Blender. Are you still interested in giving that a go? Would love to see that connection made.

PR #194, opened 2022-09-06
remove assimp submodule

elalish:
Redo of #123 

Now that Assimp has been separated from our bindings and tests, it no longer needs to be a submodule at all. In fact it is now only a dependency of `meshIO`, which is no longer a part of the core library, but simply an optional test/debug helper. In order to be able to export test results as 3D models (e.g. GLB), you'll need to have `libassimp-dev` installed by your package manager and enable the `MANIFOLD_EXPORT` compile flag. 

I like this because the `Manifold` library is about computational geometry, not file I/O. This makes that boundary much more clear and also simplifies and speeds up our build process.


pca006132:
For nix, we need to remove
```
cp test/meshIO/libmeshIO.a $out/
```

Issue #196, opened 2022-09-06
Build multi-material three.js example


elalish:
I like the simplicity of @makc's first WASM [example](https://elalish.github.io/manifold/bindings/wasm/examples/index.html), so I plan to keep that (and add some explanation and link it from our readme). However, we could also use a more complex example to show how different objects with textures/materials can be combined into a [multi-material](https://github.com/elalish/manifold/pull/155#issuecomment-1186812460) result. 

I think as part of this I'll also integrate `<model-viewer>` as the viewer by exporting the three.js scene as a GLB blob. This work will also dovetail nicely with the other work going in to finishing #73.

pca006132:
I'm thinking about how to store the material information. Maybe we can use the registry I mentioned to store the `ID <-> material` association, and automatically apply the material when we convert the mesh object to threejs object?

makc:
ha, Ive never got back to uvs thing. @ me if you need any help with this, but like small help 👯

elalish:
As I started to work on an example for this, I realized that A) reapplying vertex properties is quite a pain and B) our `MeshRelation` actually loses some data, which will cause more duplication of verts than is necessary. After enough operations you could potentially end up with triangle soup, which is not great for rendering performance, but especially bad if you want to save to a file like GLB and then try to reconstruct the manifold. 

I'm now working on @pca006132's suggestion - basically store and update the optional `triProperties` and `properties` input vectors. I think I'll end up changing the new `MeshGL` type to output these vectors (for easy rendering) rather than the manifold triangles. 

The next step will be deciding on a data structure that contains the manifoldness information necessary to go back from `MeshGL` to `Mesh` - this could form a glTF extension that would enable lossless round-trip of manifold meshes with properties through glTF files. I think it can be pretty compact, as we only need to mark which sets of "open" verts/edges need to be merged.

PR #198, opened 2022-09-08
model-viewer WASM example

elalish:
This solves the first part of #196, which is making a new example that exports a GLTF in order to run `<model-viewer>`. Now we have a nice interactive example with PBR. I've also linked it from the README so it's easier to find. 

Next I'll work on the multi-material piece with textures.

FYI, the changes in index.html are almost entirely just auto-formatting.

### DIFF
--- a/.github/workflows/manifold.yml
+++ b/.github/workflows/manifold.yml
@@ -104,7 +104,7 @@ jobs:
         path: |
           build/bindings/wasm/manifold.js
           build/bindings/wasm/manifold.wasm
-        retention-days: 1
+        retention-days: 7
 
   build_windows:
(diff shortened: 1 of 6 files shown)


elalish:
@pca006132 This should give you what you need to for an editor 3D output: https://github.com/elalish/manifold/issues/73#issuecomment-1238412143

makc:
this is funny way of doing it, since \<model-viewer\> is three.js based, so in effect you include three.js twice

elalish:
Yeah, this was just the quick version. It's actually been on my list for awhile to publish a version of model-viewer that exposes three.js for just such an occasion.

Issue #197, opened 2022-09-08
JS Interface Modification


pca006132:
It seems that the code that embind generates is not very ergonomic:

- `std::vector<T>` is incompatible with javascript array, so we need a new type for each `std::vector<T>` and users have to use that.
- Seems to require overloading to generate functions with optional parameters, which is used extensively in our APIs.
- Unable to convert array to `glm::vector2`/`glm::vector3`.
- Seems hard to get function pointer working (for `Warp`).

I propose the following changes to our JS interface:
- Change the naming convention to lower camel case.
- Make the embind generated bindings internal, and provide a public API implemented with js that performs automatic type conversion mentioned above. (do have some performance penalty)
- Instead of passing a callback to warp, the API for js will be `std::vector<glm::vec3> -> std::vector<glm::vec3>`, i.e. give the user the entire buffer for modification, which will likely be faster and don't have to deal with callbacks.

elalish:
Agreed, that sounds like a better approach. I look forward to reviewing the changes. As always, we'll know we're on the right track if we can make clean usage examples. :+1:

pca006132:
There is another thing that I'm not sure whether to put it in the js binding or just in the js editor. Our code requires precise reference counting to free resources, and because our API are functional, users have to insert `delete` calls everywhere to avoid memory leak. For example:
```
b = a.translate(1, 1, 1)
b.delete()
```
This will leak resources as `a` is not being deleted.

I'm thinking about using something like an arena to manage this for the users, at least in our js editor. The idea is that each API call will register the object into the arena and user don't have to delete it manually. When the script finishes, we will delete every object in the arena. This approach will cause higher memory usage than inserting delete calls, but is much simpler than requiring users to manage the resources by themselves. The problem is I'm not sure whether this should be put into the binding or not.

An alternative way would be to write our own language, which can solve this optimally and allows operator overloading. I'm happy to do that but not sure if users will want another language :P

elalish:
Yeah, as a JS developer I can say we're very much not in the habit of thinking about memory management. Still, there is precedent with the required Three.js `dispose()` calls for GPU memory management. 

Do you think there's a way to make some kind of scope-wrapper that would do what you're describing but in the context of a given user-function? So if I make a JS function that makes a bunch of manifolds and combines them into a single output manifold, the function-wrapper automatically causes those temporary resources to be freed when the function returns, like would happen in C++? 

[...]

PR #201, opened 2022-09-11
updated js binding

pca006132:
Partially addresses #197, except that `warp` is not yet implemented. The API is now very permissive, e.g. translate can be called with `translate(1, 2, 3)`, `translate([1, 2, 3])` or `translate({x: 1, y: 2, z: 3})`, and most of the APIs now have optional parameters, matching the behavior in the C++ API. Some three.js helper functions are moved to the JS API or provided as a method in the `Manifold` class, such as `manifold.getGeometry()` will now return a three.js geometry directly.

For warp, I'm thinking about whether the proposed API (giving user access to the buffer) is a good way to implement it, because the vector class is different from list class in js, so users may find it not very natural to use. Performing automatic conversion is easy, but probably very slow because we need to do a lot of copying that way.

### DIFF
--- a/bindings/wasm/bindings.cpp
+++ b/bindings/wasm/bindings.cpp
@@ -29,7 +29,43 @@ void Subtract(Manifold& a, Manifold& b) { a -= b; }
 Manifold Intersection(Manifold& a, Manifold& b) { return a ^ b; }
 void Intersect(Manifold& a, Manifold& b) { a ^= b; }
 
+std::vector<SimplePolygon> ToPolygon(
+    std::vector<std::vector<glm::vec2>>& polygons) {
+  std::vector<SimplePolygon> simplePolygons(polygons.size());
+  for (int i = 0; i < polygons.size(); i++) {
+    std::vector<PolyVert> vertices(polygons[i].size());
+    for (int j = 0; j < polygons[i].size(); j++) {
+      vertices[j] = {polygons[i][j], j};
+    }
+    simplePolygons[i] = {vertices};
+  }
+  return simplePolygons;
+}
+
+Manifold Extrude(std::vector<std::vector<glm::vec2>>& polygons, float height,
+                 int nDivisions, float twistDegrees, glm::vec2 scaleTop) {
+  return Manifold::Extrude(ToPolygon(polygons), height, nDivisions,
+                           twistDegrees, scaleTop);
+}
+
+Manifold Revolve(std::vector<std::vector<glm::vec2>>& polygons,
+                 int circularSegments) {
+  return Manifold::Revolve(ToPolygon(polygons), circularSegments);
+}
+
+Manifold Transform(Manifold &manifold, std::vector<float> &mat) {
+  glm::mat4x3 matrix;
+  for (int i = 0; i < 4; i++)
+    for (int j = 0; j < 3; j++)
+      matrix[i][j] = mat[i*3+j];
+  return manifold.Transform(matrix);
+}
+
 EMSCRIPTEN_BINDINGS(whatever) {
+  value_object<glm::vec2>("vec2")
+      .field("x", &glm::vec2::x)
+      .field("y", &glm::vec2::y);
+
   value_object<glm::ivec3>("ivec3")
       .field("0", &glm::ivec3::x)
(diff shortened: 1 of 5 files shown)


bindings/wasm/bindings.js
```diff
@@ -0,0 +1,143 @@
...
+      temp.fromBufferAttribute(n, i);
+      mesh.vertNormal.push_back(temp);
+    }
+    for (let i = 0; i < x.count; i += 3) {
+      mesh.triVerts.push_back(x.array.subarray(i, i + 3));
+    }
+    return mesh;
+  };
+
+  function mesh2geometry(mesh) {
```
elalish, line 45: I'm not sure it's a good idea to codify these I/O functions in the bindings yet. Note my other comment about the two versions I already had to make. And it's going to get a lot more complicated when I add in texture support.

bindings/wasm/bindings.cpp
```diff
@@ -29,7 +29,43 @@ void Subtract(Manifold& a, Manifold& b) { a -= b; }
...
+Manifold Revolve(std::vector<std::vector<glm::vec2>>& polygons,
+                 int circularSegments) {
+  return Manifold::Revolve(ToPolygon(polygons), circularSegments);
+}
+
+Manifold Transform(Manifold &manifold, std::vector<float> &mat) {
+  glm::mat4x3 matrix;
+  for (int i = 0; i < 4; i++)
+    for (int j = 0; j < 3; j++)
+      matrix[i][j] = mat[i*3+j];
```
elalish, line 60: A comment might be nice about the matrix being column-major. :+1:

bindings/wasm/bindings.js
elalish: I'd also prefer to keep anything `THREE`-specific separate from these other generic bindings.

elalish (changes requested):
For `Warp`, all we're really doing is changing (replacing) the vertices. They can already get the vertices via `to_mesh`; perhaps `Warp` can just take an identical-length vector of verts (like `from_mesh` does) and simply copy them over the existing verts. Then all the functional stuff can live in user code.

bindings/wasm/examples/model-viewer.html
```diff
@@ -26,31 +26,31 @@
...
+      //   .revolve([[[0.1, 0.1], [0.2, 0.1], [0.1, 0.2]]], 64)
+      //   .translate(0, 0, -0.1);
+      const manifold_2 = new Module.Manifold(Module.geometry2mesh(icosahedron));
 
       const csg = function (operation) {
         mesh.geometry?.dispose();
-        mesh.geometry = mesh2geometry(
-          Module[operation](manifold_1, manifold_2).GetMesh()
-        );
+        mesh.geometry = Module[operation](manifold_1, manifold_2).getGeometry();
```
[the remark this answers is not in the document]
pca006132: Oh, I thought that the functions are the same as the old one. Will use your version of the function instead.

pca006132:
For documentation, I'm not sure what documentation generator do we use for js. Maybe I can write a markdown file directly?

bindings/wasm/bindings.js
```diff
@@ -0,0 +1,99 @@
...
+      return {x: vec[0][0], y: vec[0][1], z: vec[0][2]};
+    if (typeof (vec[0]) == 'number')
+      // default to 0
+      return {x: vec[0] || 0, y: vec[1] || 0, z: vec[2] || 0};
+    return vec[0];
+  }
+
+  Module.Manifold.prototype.getGeometry = function() {
+    let mesh = this.getMesh();
+    let geometry = mesh2geometry(mesh);
```
elalish, line 33: I think `getGeometry` should also be removed? Here's the downside of JS - no build errors.

elalish (approved):
Yeah, I have the same question for the Python binding docs. Honestly, the best docs are probably a `.d.ts` file; then you have all the types for a TypeScript project to use, and VSCode will even pick up and display the documentation comments you put in. I was hoping emscripten might be able to generate it, but it looks like not (although [this](https://github.com/giniedp/webidl2ts) looks close). Anyway, I can easily add a `.d.ts` file later.

PR #204, opened 2022-09-18
updated python exporter from meshio to trimesh

elalish:
I just found a much better and more up-to-date mesh exporter for Python: [trimesh](https://github.com/mikedh/trimesh). It supports GLB and 3MF, in addition to lots of other useful ops with an eye toward manifoldness. I also noticed this from their features section:
>Boolean operations on meshes (intersection, union, difference) using OpenSCAD or Blender as a back end. Note that mesh booleans in general are usually slow and unreliable

This makes me think our library is be something they might be interested in, since it solves those problems.

### DIFF
--- a/.github/workflows/manifold.yml
+++ b/.github/workflows/manifold.yml
@@ -25,7 +25,7 @@ jobs:
       run: |
         apt-get -y update
         DEBIAN_FRONTEND=noninteractive apt install -y libomp-dev libassimp-dev git libtbb-dev pkg-config libpython3-dev python3 python3-distutils python3-pip lcov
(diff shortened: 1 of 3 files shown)


pca006132:
> This makes me think our library is be something they might be interested in, since it solves those problems.

Maybe we can open an issue in their repo and see if they are interested in this? Considering we also have a python binding, integration should be pretty straight forward.

elalish:
Agreed; I was thinking let's push out v2.0 first and ideally publish an initial python version to pypi first to make it even easier to experiment with. Does that sound good?

PR #206, opened 2022-09-20
cache intermediate results (#199)

### LINKED ISSUE #199: Cache intermediate results automatically
Regarding this [comment](https://github.com/elalish/manifold/pull/171#issuecomment-1214282589) @pca006132, I did a little test with `./manifold_test --gtest_filter=Samples.Frame -v` before and after removing the two cache-forcing [lines](https://github.com/elalish/manifold/blob/master/samples/src/rounded_frame.cpp#L35). I'm seeing at least double the number of Boolean operations when I remove those lines. Is there a way to make the `CsgNode` aware of when it's being used more than once so it can skip flattening and allow reuse instead? Maybe some kind of ref-counter?

pca006132:
Should fix #199, the number of boolean operations required by `Samples.Frame` is now 5.


elalish (approved):
Thank you! This is a key improvement.

PR #208, opened 2022-09-20
added monaco editor to model-viewer example

pca006132:
Added code editor and typescript declaration as mentioned in #73. The typescript declaration file is now in `bindings.ts` (need some more documentation). It can now compile to js and run directly.

image

### DIFF
--- a/bindings/wasm/bindings.d.ts
+++ b/bindings/wasm/bindings.d.ts
@@ -1,0 +1,39 @@
+type Matrix1x3 = [number, number, number];
+type Matrix4x3 = [Matrix1x3, Matrix1x3, Matrix1x3, Matrix1x3];
+type Vec3 = [number, number, number];
+type Vec3Fun = (v: Vec3) => Manifold;
+type Vec2 = [number, number];
+type SimplePolygon = Vec2[];
+type Polygons = SimplePolygon[];
+
+declare class Manifold {
+  transform(matrix: Matrix4x3): Manifold;
+
+  translate: Vec3Fun;
+  rotate: Vec3Fun;
+  scale: Vec3Fun;
+  add(other: Manifold): Manifold;
+  subtract(other: Manifold): Manifold;
+  intersect(other: Manifold): Manifold;
+  refine(n: number): Manifold;
+}
+
+declare function cube(size: Vec3, center?: boolean): Manifold;
+declare function cube(size?: number, center?: boolean): Manifold;
+
+declare function cylinder(
+    height: number, radiusLow: number, radiusHigh?: number,
+    circularSegments?: number, center?: boolean): Manifold;
+
+declare function sphere(radius: number, circularSegments?: number): Manifold;
+
+declare function extrude(
+    polygon: Polygons, height: number, nDivisions?: number,
+    twistDegrees?: number, scaleTop?: Vec2): Manifold;
+
+declare function revolve(polygon: Polygons, circularSegments?: number): Manifold;
+
+declare function union(a: Manifold, b: Manifold): Manifold;
+declare function difference(a: Manifold, b: Manifold): Manifold;
+declare function intersection(a: Manifold, b: Manifold): Manifold;
+
--- a/bindings/wasm/bindings.d.ts
+++ b/bindings/wasm/bindings.d.ts
@@ -1,17 +1,14 @@
-type Matrix1x3 = [number, number, number];
-type Matrix4x3 = [Matrix1x3, Matrix1x3, Matrix1x3, Matrix1x3];
 type Vec3 = [number, number, number];
-type Vec3Fun = (v: Vec3) => Manifold;
+type Matrix4x3 = [Vec3, Vec3, Vec3, Vec3];
 type Vec2 = [number, number];
 type SimplePolygon = Vec2[];
 type Polygons = SimplePolygon[];
 
 declare class Manifold {
   transform(matrix: Matrix4x3): Manifold;
-
-  translate: Vec3Fun;
-  rotate: Vec3Fun;
-  scale: Vec3Fun;
+  translate(v: Vec3): Manifold;
+  rotate(v: Vec3): Manifold;
+  scale(v: Vec3): Manifold;
   add(other: Manifold): Manifold;
   subtract(other: Manifold): Manifold;
   intersect(other: Manifold): Manifold;
@@ -19,7 +16,6 @@ declare class Manifold {
 }
 
 declare function cube(size: Vec3, center?: boolean): Manifold;
-declare function cube(size?: number, center?: boolean): Manifold;
 
 declare function cylinder(
     height: number, radiusLow: number, radiusHigh?: number,
--- a/bindings/wasm/examples/editor.html
+++ b/bindings/wasm/examples/editor.html
@@ -55,7 +55,7 @@
 <script src="https://cdnjs.cloudflare.com/ajax/libs/monaco-editor/0.34.0/min/vs/loader.min.js" integrity="sha512-6bIYsGqvLpAiEBXPdRQeFf5cueeBECtAKJjIHer3BhBZNTV3WLcLA8Tm3pDfxUwTMIS+kAZwTUvJ1IrMdX8C5w==" crossorigin="anonymous" referrerpolicy="no-referrer"></script>
 <link rel="stylesheet" href="https://cdnjs.cloudflare.com/ajax/libs/monaco-editor/0.34.0/min/vs/editor/editor.main.min.css" integrity="sha512-iQEIc0rsSDujsfjtD+lfyJ1W23Bh/lbgriubKDAym6VlEIDRj9rrbSIyJRyshOrl8s0yRcQ0+gyrZfSLyjJGWQ==" crossorigin="anonymous" referrerpolicy="no-referrer" />
 <script>
-  var editor = undefined;
+  let editor = undefined;
   var Module = {
     onRuntimeInitialized: function () {
       Module.setup();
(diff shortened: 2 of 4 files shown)


bindings/wasm/bindings.ts
```diff
@@ -0,0 +1,43 @@
+type Matrix1x3 = [Number, Number, Number];
```
elalish, line 1: I believe the standard is to use the `.d.ts` file extension.

pca006132:
It supports autocomplete and error diagnostics out of the box. Not sure if we need to print error information with an output panel or something. Currently it will leak memory, I will add memory management functions later to deal with that. I can finally get a good sleep (thought about this in 1am, cannot fall asleep, finished it in 3am :P)

Btw remember to update model-viewer for the pan function! It will be very useful for me to port examples to js.

bindings/wasm/bindings.ts
```diff
@@ -0,0 +1,43 @@
...
+type Matrix4x3 = [Matrix1x3, Matrix1x3, Matrix1x3, Matrix1x3];
+type Vec3 = [Number, Number, Number];
```
elalish, line 3: I believe you want the lower-case version of `number`: https://stackoverflow.com/a/42654496

bindings/wasm/examples/model-viewer.html
```diff
@@ -159,8 +165,25 @@
     simplified.computeVertexNormals();
     return simplified;
   }
+
+  require.config({ paths: { vs: 'https://cdnjs.cloudflare.com/ajax/libs/monaco-editor/0.34.0/min/vs' } });
```
elalish, line 169: Why did you end up going with monaco instead of code-mirror? I ask because I never saw a response to https://github.com/elalish/manifold/issues/73#issuecomment-1238438742

bindings/wasm/examples/model-viewer.html
```diff
@@ -59,25 +64,26 @@
...
-        mesh.geometry = mesh2geometry(Module[operation](manifold_1, manifold_2).getMesh());
-        push2MV(mesh);
-      };
-
-      document.querySelector('select').onchange = function (event) {
-        csg(event.target.value);
-      };
-
-      csg('difference');
+      document.querySelector('#compile').onclick = function (e) {
```
elalish, line 67: Can we copy this into a new html page instead of editing the MV example? I'd like to keep that as a minimal example page, since this editor will probably grow a bit in complexity.

bindings/wasm/bindings.ts
```diff
@@ -0,0 +1,43 @@
...
+type Vec3Fun =
+    ((v: Vec3) => Manifold)&((x: Number, y?: Number, z?: Number) => Manifold);
+type Point = [Number, Number];
```
elalish, line 6: `Vec2`?

pca006132:
Oh and remember to deploy the typescript file as well, I am not sure how to modify the deploy script.

bindings/wasm/bindings.ts
```diff
@@ -0,0 +1,43 @@
...
+  scale: Vec3Fun;
+  add(other: Manifold): Manifold;
+  subtract(other: Manifold): Manifold;
+  intersect(other: Manifold): Manifold;
+  refine(n: Number): Manifold;
+}
+
+declare function cube(size: Vec3, center?: boolean): Manifold;
+declare function cube(
+    x: Number, y?: Number, z?: Number, center?: boolean): Manifold;
```
elalish, line 24: `size: Vec3`?

bindings/wasm/bindings.ts
```diff
@@ -0,0 +1,43 @@
+type Matrix1x3 = [Number, Number, Number];
+type Matrix4x3 = [Matrix1x3, Matrix1x3, Matrix1x3, Matrix1x3];
+type Vec3 = [Number, Number, Number];
+type Vec3Fun =
+    ((v: Vec3) => Manifold)&((x: Number, y?: Number, z?: Number) => Manifold);
+type Point = [Number, Number];
+type SimplePolygon = Point[];
+type Polygon = SimplePolygon[];
```
elalish, line 8: `Polygons`, to match C++ name.

bindings/wasm/bindings.ts
```diff
@@ -0,0 +1,43 @@
...
+    polygon: Polygon, height: Number, nDivisions?: Number,
+    twistDegrees?: Number, scaleTop?: [Number, Number]): Manifold;
+
+declare function revolve(polygon: Polygon, circularSegments?: Number): Manifold;
+
+declare function union(a: Manifold, b: Manifold): Manifold;
+declare function difference(a: Manifold, b: Manifold): Manifold;
+declare function intersection(a: Manifold, b: Manifold): Manifold;
+
+declare function geometry2mesh(geometry: any): Manifold;
```
elalish, line 42: Should probably make a separate declaration file for three.js-specific bridge API.

bindings/wasm/bindings.ts
```diff
@@ -0,0 +1,43 @@
+type Matrix1x3 = [Number, Number, Number];
+type Matrix4x3 = [Matrix1x3, Matrix1x3, Matrix1x3, Matrix1x3];
+type Vec3 = [Number, Number, Number];
+type Vec3Fun =
+    ((v: Vec3) => Manifold)&((x: Number, y?: Number, z?: Number) => Manifold);
```
elalish, line 5: I don't quite understand this; what does this do that a `Vec3` input would not?

bindings/wasm/bindings.ts
```diff
@@ -0,0 +1,43 @@
...
+
+declare function cylinder(
+    height: Number, radiusLow: Number, radiusHigh?: Number,
+    circularSegments?: Number, center?: boolean): Manifold;
+
+declare function sphere(radius: Number, circularSegments?: Number): Manifold;
+
+declare function extrude(
+    polygon: Polygon, height: Number, nDivisions?: Number,
+    twistDegrees?: Number, scaleTop?: [Number, Number]): Manifold;
```
elalish, line 34: `scaleTop: Vec2`?

elalish (changes requested):
This is exciting! Thank you, and get some sleep! Pan should already be enabled; I released v2.0 last week and I believe we're automatically pulling the latest.

bindings/wasm/examples/model-viewer.html
pca006132: Because monaco supports typescript and language server out of the box.

bindings/wasm/bindings.ts
pca006132: Ah this is interesting, I did not check this because it did not complain about the examples. Will fix.

bindings/wasm/bindings.ts
pca006132: Nice catch, the binding doesn't allow this to be optional right now. Will fix.

bindings/wasm/bindings.ts
pca006132: It seems that I remembered the wrong thing, I somehow thought that OpenSCAD allows both `translate(x, y, z)` and `translate([x, y, z])`, and this is used to allow this. Will remove this.

bindings/wasm/bindings.d.ts
```diff
@@ -0,0 +1,39 @@
...
+  rotate: Vec3Fun;
+  scale: Vec3Fun;
+  add(other: Manifold): Manifold;
+  subtract(other: Manifold): Manifold;
+  intersect(other: Manifold): Manifold;
+  refine(n: number): Manifold;
+}
+
+declare function cube(size: Vec3, center?: boolean): Manifold;
+declare function cube(size?: number, center?: boolean): Manifold;
```
elalish, line 22: Is this something we can do in the C++ API? I sort of thought it would be just `cube(size?: Vec3, center?: boolean): Manifold`

bindings/wasm/examples/editor.html
```diff
@@ -0,0 +1,192 @@
...
+</body>
+
+<script src="https://unpkg.com/three@0.144.0/build/three.js"></script>
+<script src="https://unpkg.com/three@0.144.0/examples/js/utils/BufferGeometryUtils.js"></script>
+<script src="https://unpkg.com/three@0.144.0/examples/js/exporters/GLTFExporter.js"></script>
+<script src="https://cdnjs.cloudflare.com/ajax/libs/monaco-editor/0.34.0/min/vs/loader.min.js" integrity="sha512-6bIYsGqvLpAiEBXPdRQeFf5cueeBECtAKJjIHer3BhBZNTV3WLcLA8Tm3pDfxUwTMIS+kAZwTUvJ1IrMdX8C5w==" crossorigin="anonymous" referrerpolicy="no-referrer"></script>
+<link rel="stylesheet" href="https://cdnjs.cloudflare.com/ajax/libs/monaco-editor/0.34.0/min/vs/editor/editor.main.min.css" integrity="sha512-iQEIc0rsSDujsfjtD+lfyJ1W23Bh/lbgriubKDAym6VlEIDRj9rrbSIyJRyshOrl8s0yRcQ0+gyrZfSLyjJGWQ==" crossorigin="anonymous" referrerpolicy="no-referrer" />
+<script>
+  var editor = undefined;
+  var Module = {
```
elalish, line 59: `var` is old-school; generally we try to just use `const` and `let` now.

bindings/wasm/examples/editor.html
```diff
@@ -0,0 +1,192 @@
...
+                'union', 'difference', 'intersection', 'THREE', 'geometry2mesh', 'show', content);
+              f(Module.cube, Module.cylinder, Module.sphere, Module.extrude,
+                Module.revolve, Module.union, Module.difference,
+                Module.intersection, THREE, 
+                geometry => new Module.Manifold(geometry2mesh(simplify(geometry))),
+                function(m) {
+                  mesh.geometry?.dispose();
+                  mesh.geometry = mesh2geometry(m.getMesh());
+                  push2MV(mesh);
+                })
```
elalish, line 82: I've never seen this pattern before; looks cool. Can you give me a little walk-through of what it's doing?

elalish (changes requested):
Looking really good, just some nits.

bindings/wasm/examples/editor.html
pca006132: Yeah I kind of messed up the styles when copying the example from monaco repo.

bindings/wasm/examples/editor.html
```diff
@@ -0,0 +1,192 @@
...
+      Module.setup();
+      const mesh = new THREE.Mesh(undefined, new THREE.MeshStandardMaterial({
+        color: 'yellow',
+        metalness: 1,
+        roughness: 0.2
+      }));
+      document.querySelector('#compile').onclick = function (e) {
+        monaco.languages.typescript.getTypeScriptWorker().then(worker => {
+          worker(editor.getModel().uri).then(w => {
+            w.getEmitOutput(editor.getModel().uri.toString()).then(output => {
```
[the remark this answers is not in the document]
pca006132: Yes, but I'm not sure about whether all browsers (commonly used ones) support async. I remember (a few years ago) we still need a transpiler to compile ES6 code to older JS code or something.

bindings/wasm/examples/editor.html
pca006132: This function constructor is to execute strings as javascript code, with access to normal top level variables and parameters you pass into the function (names defined in the constructor and values defined when you actually call the function.

bindings/wasm/bindings.d.ts
elalish: Ah, right. Yeah, let's stick with self-consistency for now. Hopefully it won't make too much difference when we eventually try to translate from OpenSCAD files.

bindings/wasm/examples/editor.html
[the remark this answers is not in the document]
elalish: Oh, that's a good point; I'm used to my TS compiler that takes care of all that for me. We can leave it.

bindings/wasm/examples/editor.html
elalish: Got it, thanks for the explanation. I see the `show` function now. Very nice!

bindings/wasm/examples/editor.html
pca006132: For the `Module` one, I'm not sure what is the appropriate way to do it: the wasm js file is loaded after this script, which will redefine `Module`, so we cannot change this to `let`. Perhaps we should just move this script section after loading wasm js files?

bindings/wasm/examples/editor.html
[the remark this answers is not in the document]
pca006132: ah, already changed this, you can just rebase it to remove the latest commit :)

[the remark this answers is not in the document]
elalish: eh, should be fine: https://caniuse.com/?search=await IE is dead anyway.

bindings/wasm/examples/editor.html
elalish: That's fine, or if this is a legit use of `var`, we can just leave it. I was mostly curious.

elalish (approved):
Awesome, I'm excited to see this!

Issue #209, opened 2022-09-21
Editor enhancements


elalish:
@pca006132, your new [editor](https://elalish.github.io/manifold/bindings/wasm/examples/editor.html) is amazing! I wanted to open this issue to have a place to discuss changes/improvements we want to add.

First, I plan to do a UX pass on it, add a download button, ideally make the `Compile` button get enabled only when the script has changed. 

Second, I think I'd like to hide `THREE` and even the `show` function from the script window itself. Maybe make the script return a manifold instead? Then again, I'd like to support creating OpenSCAD-like rigid animations eventually, which will require separate manifolds, really more of a scene hierarchy. Probably out of scope for the near term.

It looks like Monaco isn't picking up our `.d.ts` definitions yet - hopefully an easy fix. Also, do we need an error console of some kind for Monaco?

pca006132:
1. UX changes: Great, perhaps we can also make the page a bit more modern and put it as part of the home page for the project? Perhaps something like the model viewer home page.
2. Yes, I am fine with that. For animation, do you mean using something like the `$t` variable?
3. For ts file, I think we did not modify the deploy script, so the `bindings.d.ts` was not copied to the correct destination. I think perhaps we need a console for both typescript compiler error and javascript execution error, so users don't have to open developer tools to check the error.

elalish:
Okay, that all sounds good. I'll work on UX and deploy. Do you want to look at the consoles? We might be able to pipe the output from both TS and JS into the same box. 

Yes, the `$t` style animation is what I'm thinking of. It should be possible to export those to glTF and show them in MV. I'll have to think through how to make an API for that though.

I also want to add a drop-down to select some starter examples; I'm planning to translate my samples to TS for this.

pca006132:
I found that the typescript transpiler will try its best to compile whatever code to javascript, even in the presence of syntax error. I guess it is just trying to parse it into an AST, remove those type information, and serialize the AST back to string as javascript. So the only useful output from typescript is the diagnostics, but they are already highlighted in the editor so not sure if it is really useful to put them in a console.

For javascript exceptions, I guess we can just catch whatever exception thrown inside the function (and conversion to threejs mesh), turn them into strings based on the type of the exception (or maybe just call toString). I guess the most complex part would be to come up with a nice looking UI.

PR #210, opened 2022-09-22
Editor UX updates

elalish:
Addresses some of #209 

Should fix deployment of types and a bunch of UX updates.


pca006132 (approved):
LGTM. What do you think about making this a bit more responsive so mobile users can also view the page?

```css
@media only screen and (max-width: 600px) {
  .container {
    flex-direction: column;
    width: 100%;
  }
  .col {
    width: 100%;
  }
}
```

PR #212, opened 2022-09-23
Editor updates

### LINKED ISSUE #209: Editor enhancements

elalish:
Fixes #209 

Well, it doesn't address animation or error logging yet, but those can be separate issues. I changed the `add` and such operations to be functional like the rest of the API. There are not changes to `bindings.js`, just formatting. I also moved us away from unpkg because it's been super slow for me lately. I also copied our docs into our `.d.ts` file, which Monaco surfaces nicely.

### DIFF
--- a/bindings/wasm/bindings.d.ts
+++ b/bindings/wasm/bindings.d.ts
@@ -5,29 +5,140 @@ type SimplePolygon = Vec2[];
 type Polygons = SimplePolygon[];
 
 declare class Manifold {
-  transform(matrix: Matrix4x3): Manifold;
+  /**
+   * Transform this Manifold in space. The first three columns form a 3x3 matrix
+   * transform and the last is a translation vector. This operation can be
+   * chained. Transforms are combined and applied lazily.
+   *
+   * @param m The affine transform matrix to apply to all the vertices.
+   */
+  transform(m: Matrix4x3): Manifold;
+
+  /** 
+   * Move this Manifold in space. This operation can be chained. Transforms are
+   * combined and applied lazily.
+   *
+   * @param v The vector to add to every vertex.
+   */
   translate(v: Vec3): Manifold;
+
+  /**
+   * Applys an Euler angle rotation to the manifold, first about the X axis, then
+   * Y, then Z, in degrees. We use degrees so that we can minimize rounding error,
+   * and elimiate it completely for any multiples of 90 degrees. Addtionally, more
+   * efficient code paths are used to update the manifold when the transforms only
+   * rotate by multiples of 90 degrees. This operation can be chained. Transforms
+   * are combined and applied lazily.
+   *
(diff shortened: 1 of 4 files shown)


bindings/wasm/bindings.d.ts
```diff
@@ -5,29 +5,140 @@ type SimplePolygon = Vec2[];
 type Polygons = SimplePolygon[];
 
 declare class Manifold {
-  transform(matrix: Matrix4x3): Manifold;
+  /**
+   * Transform this Manifold in space. The first three columns form a 3x3 matrix
```
pca006132, line 9: Maybe we should mention that the matrix is row major? (I was confused about this previously)

bindings/wasm/bindings.d.ts
```diff
@@ -5,29 +5,140 @@ type SimplePolygon = Vec2[];
...
+* there are a circle of vertices on all three of the axis planes. Default is
+* calculated by the static Defaults.
+*/
 declare function sphere(radius: number, circularSegments?: number): Manifold;
 
+/**
+ * Constructs a manifold from a set of polygons by extruding them along the
+ * Z-axis.
+ *
+ * @param crossSection A set of non-overlapping polygons to extrude.
```
pca006132, line 115: Just found out that we may have this error in both our js and c++ documentation. I think this should be polygon with holes instead of a set of non-overlapping polygons?

pca006132 (changes requested):

pca006132:
And maybe we should modify the readme to link to the editor page as well.

bindings/wasm/bindings.d.ts
elalish: I wouldn't call it an error, but there may be a clearer way to say it. I define overlapping [here](https://github.com/elalish/manifold/wiki/Manifold-Library#overlaps), so holes are not considered overlaps. I used to have polygon-with-holes, but that makes it sound like there should only be one outer contour.

elalish:
I'm not quite ready to link to it yet. I'd like to kick the tires a bit more by writing some examples so I know the rest of the API is functional.

PR #213, opened 2022-09-24
free memory for js editor

pca006132:
This should free up most of the memory for the js editor. I tested using the menger sponge example (n=4) several times, and it seems that the memory usage is pretty stable. Debugging this is pretty hard as it seems that the heap for wasm will not be freed by the GC (it is retained until you reload the tab). The only way is to use leak sanitizer but it is pretty slow. It seems that there is still some minor leak somewhere (when I run the menger sponge example with n=3 for several times, the memory usage will slowly increase), but considering the memory usage for complicated example is pretty stable, I think minor leak should not be a big deal.

Code for the example:
<details>

```js
function vec2add(a: Vec2, b: Vec2): Vec2 {
  return [a[0] + b[0], a[1] + b[1]];
}

// union a set of objects together
// we did not expose the compose interface so we have to do this manually
function compose(objs: Manifold[]): Manifold {
  let result = objs.pop();
  for (let obj of objs) {
    result = union(result, obj);
  }
  return result;
}

function fractal(
    holes: Manifold[], hole: Manifold, w: number, position: Vec2, depth: number,
    maxDepth: number) {
  w /= 3;
  holes.push(
      hole.scale([w, w, 1.0]).translate([position[0], position[1], 0.0]));
  if (depth == maxDepth) return;
  let offsets: Vec2[] = [
    [-w, -w], [-w, 0.0], [-w, w], [0.0, w], [w, w], [w, 0.0], [w, -w], [0.0, -w]
  ];
  for (let offset of offsets)
    fractal(holes, hole, w, vec2add(position, offset), depth + 1, maxDepth);
}

function mengerSponge(n: number) {
  let result = cube([1, 1, 1], true);
  let holes: Manifold[] = [];
  fractal(holes, result, 1.0, [0.0, 0.0], 1, n);

  let hole = compose(holes);

  result = difference(result, hole);
  hole = hole.rotate([90, 0, 0]);
  result = difference(result, hole);
  hole = hole.rotate([0, 0, 90]);
  result = difference(result, hole);

  return result;
}

const result = mengerSponge(4);
```

</details>

### DIFF
--- a/bindings/wasm/examples/editor.html
+++ b/bindings/wasm/examples/editor.html
@@ -129,6 +129,42 @@
   var Module = {
     onRuntimeInitialized: function () {
       Module.setup();
+      // Setup memory management, such that users don't have to care about
+      // calling `delete` manually.
+      // Note that this only fixes memory leak across different runs: the memory
+      // will only be freed when the compilation finishes.
+
+      let manifoldRegistry = [];
+      for (const name of
+               ['add', 'subtract', 'intersect', 'refine', 'transform', 'translate',
+                'rotate', 'scale']) {
+        const originalFn = Module.Manifold.prototype[name];
+        Module.Manifold.prototype["_" + name] = originalFn;
+        Module.Manifold.prototype[name] = function(...args) {
+          const result = this["_" + name](...args);
+          manifoldRegistry.push(result);
+          return result;
+        }
+      }
+
+      for (const name
+               of ['cube', 'cylinder', 'sphere', 'extrude', 'revolve', 'union',
+                   'difference', 'intersection']) {
+        const originalFn = Module[name];
+        Module[name] = function(...args) {
+          const result = originalFn(...args);
+          manifoldRegistry.push(result);
+          return result;
+        }
+      }
+
+      Module.cleanup = function() {
+        for (const obj of manifoldRegistry) {
+          obj.delete();
+        }
+        manifoldRegistry = [];
+      }
+
       runButton.onclick = async function (e) {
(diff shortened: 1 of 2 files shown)


bindings/wasm/examples/editor.html
```diff
@@ -129,6 +129,42 @@
...
+          manifoldRegistry.push(result);
+          return result;
+        }
+      }
+
+      Module.cleanup = function() {
+        for (const obj of manifoldRegistry) {
+          obj.delete();
+        }
+        manifoldRegistry = [];
```
elalish, line 165: This looks great. Should we build this into the JS bindings themselves instead? Seems like anyone using this library would benefit from a cleanup utility. Or does it make assumptions that aren't necessarily valid? Anyway, this is great for now, but food for thought.

elalish (approved):
Awesome!

bindings/wasm/examples/editor.html
pca006132: This assumes that the user will not call `delete` manually for the objects. For the editor, we did not add the delete method to typescript definition and did not document about it, so I assume that users will not call delete on it.

elalish: Yeah, that sounds good. Still, there is the issue of large chains of boolean operations potentially needing to clean up memory intermittently to keep from running out before completion. We may want to think through how to do that before exposing a memory-management API.

elalish:
I'm going to add a drop-down in the editor for examples. Let's remember to put yours in there!

PR #217, opened 2022-09-24
use compose for disjoint union

pca006132:
Last mystery in #114. It turns out that the problem is due to misunderstanding of the bounding box API: `Box::Transform` is *not* a safe over-approximation of the bounding box of the manifold after the transform. In fact it might be smaller than that if the transform is not axis aligned (I should have just RTFM at that time). The simple fix is to apply the transform first and then compute the bounding box, this will also make sure that the bounding box is tight.

In the disjointness check, I enlarged the bounding box by the tolerance of the object, which can hopefully overapproximate the behavior of a perturbation of the vertices by at most the tolerance. However, I did not really add tests that can check this, so I am not sure if this is useful.

Note: This patch is not very important right now, maybe we can merge this after the next release, if we are not certain about this. I think the most important part of this patch is the potential performance improvement it brings and shows that our compose implementation should probably be fine.

### DIFF
--- a/src/manifold/src/csg_tree.cpp
+++ b/src/manifold/src/csg_tree.cpp
@@ -142,7 +142,8 @@ Manifold::Impl CsgLeafNode::Compose(
   int numBary = 0;
   for (auto &node : nodes) {
     float nodeOldScale = node->pImpl_->bBox_.Scale();
-    float nodeNewScale = node->pImpl_->bBox_.Transform(node->transform_).Scale();
+    float nodeNewScale =
+        node->pImpl_->bBox_.Transform(node->transform_).Scale();
     float nodePrecision = node->pImpl_->precision_;
     nodePrecision *= glm::max(1.0f, nodeNewScale / nodeOldScale);
     nodePrecision = glm::max(nodePrecision, kTolerance * nodeNewScale);
@@ -344,24 +345,15 @@ void CsgOpNode::BatchUnion() const {
   constexpr int kMaxUnionSize = 1000;
   auto &children_ = impl_->children_;
   while (children_.size() > 1) {
-    int start;
-    if (children_.size() > kMaxUnionSize) {
-      start = children_.size() - kMaxUnionSize;
-    } else {
-      start = 0;
-    }
+    const int start = (children_.size() > kMaxUnionSize)
+                          ? (children_.size() - kMaxUnionSize)
+                          : 0;
     VecDH<Box> boxes;
     boxes.reserve(children_.size() - start);
     for (int i = start; i < children_.size(); i++) {
       boxes.push_back(std::dynamic_pointer_cast<CsgLeafNode>(children_[i])
                           ->GetImpl()
                           ->bBox_);
-      const float precision =
-          std::dynamic_pointer_cast<CsgLeafNode>(children_[i])
-              ->GetImpl()
-              ->precision_;
-      boxes.back().max += glm::vec3(precision);
-      boxes.back().min -= glm::vec3(precision);
     }
     const Box *boxesD = boxes.cptrD();
     // partition the children into a set of disjoint sets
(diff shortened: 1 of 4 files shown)


src/manifold/src/csg_tree.cpp
```diff
@@ -340,14 +336,76 @@ void CsgOpNode::BatchBoolean(
...
+  constexpr int kMaxUnionSize = 1000;
   auto &children_ = impl_->children_;
-  for (auto &child : children_) {
-    impls.push_back(std::dynamic_pointer_cast<CsgLeafNode>(child)->GetImpl());
+  while (children_.size() > 1) {
+    int start;
+    if (children_.size() > kMaxUnionSize) {
+      start = children_.size() - kMaxUnionSize;
+    } else {
+      start = 0;
```
elalish, line 351: Perhaps a ternary to make start const?

src/manifold/src/csg_tree.cpp
```diff
@@ -340,14 +336,76 @@ void CsgOpNode::BatchBoolean(
...
+              ->precision_;
+      boxes.back().max += glm::vec3(precision);
+      boxes.back().min -= glm::vec3(precision);
+    }
+    const Box *boxesD = boxes.cptrD();
+    // partition the children into a set of disjoint sets
+    // each set contains a set of children that are pairwise disjoint
+    std::vector<VecDH<size_t>> disjointSets;
+    for (size_t i = 0; i < boxes.size(); i++) {
+      auto lambda = [boxesD, i](const VecDH<size_t> &set) {
```
elalish, line 371: I think CUDA doesn't like your lambda; should work if you make it a functor instead. Or possibly just add `__host__ __device__`?

src/manifold/src/csg_tree.cpp
```diff
@@ -340,14 +336,76 @@ void CsgOpNode::BatchBoolean(
  * `BatchBoolean` instead of using `Compose` for non-intersecting manifolds.
  */
 void CsgOpNode::BatchUnion() const {
-  std::vector<std::shared_ptr<const Manifold::Impl>> impls;
+  // INVARIANT: children_ is a vector of leaf nodes
+  // this kMaxUnionSize is a heuristic to avoid the pairwise disjoint check
+  // with O(n^2) complexity to take too long.
+  // If the number of children exceeded this limit, we will operate on chunks
+  // with size kMaxUnionSize.
```
elalish, line 343: If performance becomes an issue here, this is exactly what `collider` is for. It does bounding box overlap checks in nlogn, in parallel.

src/manifold/src/csg_tree.cpp
pca006132: Yeah I forgot to add the `__host__ __device__` again.

src/manifold/src/csg_tree.cpp
pca006132: Yes, I think we talked about it in the previous PR: https://github.com/elalish/manifold/pull/131#issuecomment-1146846350

elalish:
If the tests pass, I'll be happy to merge this. If we get some complicated failures, agreed that we shouldn't hold the release for it.

If you're now recomputing the bounding box, is there any source of rounding error? I wouldn't expect this to be a place where tolerance would be required.

pca006132:
> If you're now recomputing the bounding box, is there any source of rounding error? I wouldn't expect this to be a place where tolerance would be required.

Say we have ε = 0.1, one cube with two corners in (0, 0, 0) and (1, 1, 1), another cube in (1.05, 1, 1), (2.05, 2, 2). If we don't consider tolerance, the bounding box are not overlapping, but I guess we still need union to merge them into one cube?

src/manifold/src/csg_tree.cpp
elalish: Ah, right. Is there a reason you need to append boxes? Generally the purpose of the collider is to take your everything vector and find all the collisions at once (hence the parallelism).

elalish:
> Say we have ε = 0.1, one cube with two corners in (0, 0, 0) and (1, 1, 1), another cube in (1.05, 1, 1), (2.05, 2, 2). If we don't consider tolerance, the bounding box are not overlapping, but I guess we still need union to merge them into one cube?

Actually, no. There is no tolerance in the Boolean ops (they only use symbolic perturbation), so those would remain disjoint regardless. Only in the triangulation and degenerate removal does the tolerance play a role. I think you're safer to do compose without tolerance.

pca006132:
OK, I will remove that tolerance check.

src/manifold/src/csg_tree.cpp
pca006132: The problem is that we need to partition the set of boxes into sets of boxes such that each set contains disjoint boxes. The target is to minimize the number of boxes. Boxes that collided means that they have to be in different sets. Maybe I can think about how to do it efficiently later.

elalish (approved):
Good catch on the bounding boxes!

Issue #215, opened 2022-09-24
Bounding CSG tree depth


pca006132:
For the simplest implementation of compose, i.e.
```js
function compose(objs: Manifold[]): Manifold {
  let result = objs.pop();
  for (let obj of objs) {
    result = union(result, obj);
  }
  return result;
}
```

The CSG tree will be a linked list. This is problematic because evaluating this will involve a very deep recursion, excessive copying etc. Previously I thought this is not an issue due to tree flattening, but #199 restricted the cases that we can perform such flattening, and JavaScript binding will not decrement the reference count automatically, so the tree will not be flattened. Also, that flattening will not bound the height, it just makes it unlikely to have a very deep tree in the normal case.

I'm thinking about implementing this by storing weak references to the parent of the CSG tree node, so we can propagate the depth of the nodes when we update some of the element (by evaluating it).

elalish:
Is there a way to make the JS binding decrement the reference count automatically? That feels like the most elegant solution here, and to other problems as well. Not sure about the feasibility though...

pca006132:
I don't think it is possible to make JS binding decrement the reference count automatically, if we don't implement our own compiler/interpreter. Thinking about it, I think we probably don't need to store weak references, just store a depth would be fine.

pca006132:
It seems that wasm is really bad at recursion, for some reason. For the following example:
```typescript
function combine1(objs: Manifold[]): Manifold {
    let a = [];
    let b = objs;
    while (b.length > 1) {
        for (let i = 0; i + 1 < b.length; i += 2)
            a.push(union(b[i], b[i+1]));
        if (b.length % 2 == 1)
            a.push(b[b.length - 1]);
        b = a;
        a = [];
    }
    return b[0];
}

function combine2(objs: Manifold[]): Manifold {
    let a = objs.pop();
    for (let x of objs)
        a = a.add(x);
    return a;
}

let spheres = [];
for (let i = 0; i < 10; i++) {
    for (let j = 0; j < 10; j++) {
        const s = sphere(1.1/2, 50).translate([i, j, 0]);
        spheres.push(s);
    }
}

const result = combine1(spheres);
```

For `combine1`, it runs for ~800ms, and `combine2` runs for ~4000ms. For comparison, combine2 in Python just runs in 0.2s (although it is parallel).

pca006132:
Weird, it is still slow on wasm even after bounding the tree depth. No idea what is going on with wasm.

pca006132:
[...]

PR #220, opened 2022-09-25
expose batch boolean API to wasm

pca006132:
See #215 for background. Due to the problem with reference counting not decrementing automatically in js, we have to recursively evaluate the csg tree and do a lot of copying, which wasm seems to be pretty slow. This patch expose the batch boolean interface to users directly, so users don't have to write their own batch boolean function and can get much better performance in the case of wasm. (~4000ms vs ~500ms).

### DIFF
--- a/bindings/wasm/bindings.d.ts
+++ b/bindings/wasm/bindings.d.ts
@@ -14,7 +14,7 @@ declare class Manifold {
    */
   transform(m: Matrix3x4): Manifold;
 
-  /** 
+  /**
    * Move this Manifold in space. This operation can be chained. Transforms are
    * combined and applied lazily.
    *
@@ -144,3 +144,6 @@ declare function union(a: Manifold, b: Manifold): Manifold;
 declare function difference(a: Manifold, b: Manifold): Manifold;
 declare function intersection(a: Manifold, b: Manifold): Manifold;
 
+declare function union(manifolds: Manifold[]): Manifold;
+declare function difference(manifolds: Manifold[]): Manifold;
+declare function intersection(manifolds: Manifold[]): Manifold;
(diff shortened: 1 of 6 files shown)


bindings/wasm/bindings.d.ts
```diff
@@ -144,3 +144,6 @@ declare function union(a: Manifold, b: Manifold): Manifold;
 declare function difference(a: Manifold, b: Manifold): Manifold;
 declare function intersection(a: Manifold, b: Manifold): Manifold;
 
+declare function union(manifolds: Manifold[]): Manifold;
```
elalish, line 147: I was already thinking the a, b union, difference, intersection were feeling pretty duplicate to the add, subtract functions anyway. Shall we just remove them in favor of these vector versions?

bindings/wasm/bindings.js
```diff
@@ -95,4 +95,21 @@ Module.setup = function () {
...
+      for (const m of args)
+        v.push_back(m);
+      const result = Module['_' + name + 'N'](v);
+      v.delete();
+    return result;
+    }
+  }
+
+  Module.union = batchbool('union');
+  Module.difference = batchbool('difference');
```
elalish, line 113: How do you batch a difference operation? There's a negative and a positive object, so which is which with n?

elalish (commented):
I think this is a good idea, but I'm not sure what a batch difference is?

bindings/wasm/bindings.d.ts
pca006132: Yeah we can remove that.

bindings/wasm/bindings.js
pca006132: The same as openscad, the first one is positive and the remaining ones are negative.

elalish (approved):
Sounds good.

Issue #219, opened 2022-09-25
Add file dropdown to editor


elalish:
I like the simplicity of our editor and I want to keep it minimal, but two features it desperately needs to become a real product are code saving and examples. I believe I can get both with a single extra dropdown menu.

The initial value will be `Intro`, containing the current TS script. Clicking will show a two-part menu: read-only Examples (menger sponge, rounded frame, etc) and Saved, which starts with only `newScript`. When any read-only script is changed, it is immediately saved to `localStorage` with name `newScript1`, which is added to the menu under Saved. Once it has a name, any changes automatically update that entry. Next to each item under Saved are two icons: a trash can to delete it, and a pencil to rename it. Selecting a saved script simply loads it from the associated `localStorage` entry.

PR #221, opened 2022-09-27
WASM warp

### LINKED ISSUE #197: JS Interface Modification

elalish:
Fixes #197 (well, it's mostly fixed by earlier PRs)

This works! In the editor I tried this, which clearly created a squashed ball:
```
const ball = sphere(60, 100);
const func = (v: Vec3) => {
    v[2] /= 2;
};
const result = ball.warp(func);
```

### DIFF
--- a/bindings/wasm/bindings.cpp
+++ b/bindings/wasm/bindings.cpp
@@ -12,6 +12,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+#include <emscripten.h>
 #include <emscripten/bind.h>
 
 using namespace emscripten;
@@ -72,6 +73,13 @@ Manifold Transform(Manifold& manifold, std::vector<float>& mat) {
   return manifold.Transform(matrix);
 }
 
+Manifold Warp(Manifold& manifold, void* funcPtr) {
+  void (*f)(glm::vec3&) = reinterpret_cast<void (*)(glm::vec3&)>(funcPtr);
+  Manifold out = manifold.Warp(f);
+  EM_ASM({removeFunction($0)}, f);
+  return out;
+}
+
 EMSCRIPTEN_BINDINGS(whatever) {
   value_object<glm::vec2>("vec2")
       .field("x", &glm::vec2::x)
@@ -116,6 +124,7 @@ EMSCRIPTEN_BINDINGS(whatever) {
       .function("intersect", &Intersection)
       .function("getMesh", &Manifold::GetMesh)
       .function("refine", &Manifold::Refine)
+      .function("_Warp", &Warp, allow_raw_pointers())
       .function("_Transform", &Transform)
       .function("_Translate", &Manifold::Translate)
       .function("_Rotate", &Manifold::Rotate)
--- a/bindings/wasm/bindings.cpp
+++ b/bindings/wasm/bindings.cpp
@@ -73,7 +73,7 @@ Manifold Transform(Manifold& manifold, std::vector<float>& mat) {
   return manifold.Transform(matrix);
 }
 
-Manifold Warp(Manifold& manifold, void* funcPtr) {
+Manifold Warp(Manifold& manifold, uintptr_t funcPtr) {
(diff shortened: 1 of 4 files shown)


bindings/wasm/bindings.cpp
```diff
@@ -72,6 +73,13 @@ Manifold Transform(Manifold& manifold, std::vector<float>& mat) {
   return manifold.Transform(matrix);
 }
 
+Manifold Warp(Manifold& manifold, void* funcPtr) {
+  void (*f)(glm::vec3&) = reinterpret_cast<void (*)(glm::vec3&)>(funcPtr);
```
elalish, line 77: This part feels the most wrong to me. I'm slightly hazy on the distinction between a reference and a pointer, but I feel like the JS wants a pointer and the C++ is giving a reference to the `glm::vec3` and I'm not really sure how to resolve that. And I'm not even sure that's the error I'm running into.

bindings/wasm/bindings.cpp
```diff
@@ -72,6 +73,13 @@ Manifold Transform(Manifold& manifold, std::vector<float>& mat) {
   return manifold.Transform(matrix);
 }
 
+Manifold Warp(Manifold& manifold, void* funcPtr) {
```
elalish, line 76: I think my error message is related to this answer: https://stackoverflow.com/a/27364643

bindings/wasm/bindings.cpp
```diff
@@ -72,6 +73,13 @@ Manifold Transform(Manifold& manifold, std::vector<float>& mat) {
...
+  void (*f)(glm::vec3&) = reinterpret_cast<void (*)(glm::vec3&)>(funcPtr);
+  Manifold out = manifold.Warp(f);
+  EM_ASM({removeFunction($0)}, f);
```
elalish, line 79: I was following this example: https://github.com/emscripten-core/emscripten/tree/main/test/interop

bindings/wasm/bindings.cpp
elalish: Okay, not sure why I didn't just try what the suggested first...

bindings/wasm/bindings.cpp
elalish: Well, it's working now, so apparently that distinction is pretty moot. I'm honestly amazed this works.

pca006132: Well, references are basically implemented as pointers under the hood, but with a bit more restrictions (cannot be null, cannot have `T&&&&&&` while you can have `T********`, for example).

bindings/wasm/bindings.cpp
pca006132: I kind of feel like that the `removeFunction` call should be called in the binding code? So in case we want to reuse some warp with js functions, we can do so with minimal overhead. This is not really C++ code anyway.

pca006132 (commented):
Can we add an example and make sure that it works before merging this? Just to make sure that everything is working correctly.

bindings/wasm/bindings.cpp
elalish: Good call, I was following that example a little too blindly.

elalish:
I tested the example at the top. We don't have a great home for JS examples at the moment, so I'm going to add this along with yours when I do #219, which is coming soon.

PR #222, opened 2022-09-30
Editor save

### LINKED ISSUE #219: Add file dropdown to editor

elalish:
Fixes #219 

This is also a major rearrangement of the editor code to make it a bit more organized. I still need to add more examples and fix the GLB output so that it respects our Z-up coordinate system. This now persists changes between refreshes, making this an actual useful app!


elalish:
@pca006132 This isn't the easiest change to review, so I'm just going to merge it so we can try it out live. We can file bugs on any problems we run across.

Issue #223, opened 2022-09-30
`addFunction` not exported


pca006132:
For some reason `addFunction` is not being exported in our GitHub pages. Just try the `warp` example and it will report

```
ReferenceError: addFunction is not defined
```

Probably related: https://github.com/emscripten-core/emscripten/issues/17456

I guess this also says that we should probably add the js binding to wasm test cases.

elalish:
Good find! Agreed.

elalish:
I just verified this was indeed fixed by #224.

PR #224, opened 2022-10-01
add missing js bindings

pca006132:
Added most of the missing js bindings, there are still two missing bindings for now.

1. `status`: not sure if we should just convert this to an integer or something. This is actually exported but not yet added to the typescript as the type signature is subject to changes. Or maybe we should check this status in our js binding directly without exposing it to the users. (just throw an error when non-zero?)
2. construct manifold from a mesh.

Also, when debugging SDF, I found that the offset computation for warp is incorrect: it should be +4 and +8, as f32 occupies 4 bytes.

### DIFF
--- a/bindings/wasm/CMakeLists.txt
+++ b/bindings/wasm/CMakeLists.txt
@@ -22,7 +22,7 @@ set_source_files_properties(bindings.cpp OBJECT_DEPENDS
 target_link_libraries(manifoldjs manifold sdf)
 target_compile_options(manifoldjs PRIVATE ${MANIFOLD_FLAGS} -fexceptions)
 target_link_options(manifoldjs PUBLIC --pre-js ${CMAKE_CURRENT_SOURCE_DIR}/bindings.js --bind -s ALLOW_TABLE_GROWTH=1
-  -sEXPORTED_RUNTIME_METHODS=addFunction)
+  -sEXPORTED_RUNTIME_METHODS=addFunction,removeFunction)
 
 target_compile_features(manifoldjs PUBLIC cxx_std_14)
 set_target_properties(manifoldjs PROPERTIES OUTPUT_NAME "manifold")
--- a/bindings/wasm/CMakeLists.txt
+++ b/bindings/wasm/CMakeLists.txt
@@ -19,9 +19,10 @@ add_executable(manifoldjs bindings.cpp)
 # make sure that we recompile the wasm when bindings.js is being modified
 set_source_files_properties(bindings.cpp OBJECT_DEPENDS
   ${CMAKE_CURRENT_SOURCE_DIR}/bindings.js)
-target_link_libraries(manifoldjs manifold)
+target_link_libraries(manifoldjs manifold sdf)
 target_compile_options(manifoldjs PRIVATE ${MANIFOLD_FLAGS} -fexceptions)
-target_link_options(manifoldjs PUBLIC --pre-js ${CMAKE_CURRENT_SOURCE_DIR}/bindings.js --bind -s ALLOW_TABLE_GROWTH=1)
+target_link_options(manifoldjs PUBLIC --pre-js ${CMAKE_CURRENT_SOURCE_DIR}/bindings.js --bind -s ALLOW_TABLE_GROWTH=1
+  -sEXPORTED_RUNTIME_METHODS=addFunction)
 
 target_compile_features(manifoldjs PUBLIC cxx_std_14)
 set_target_properties(manifoldjs PROPERTIES OUTPUT_NAME "manifold")
--- a/bindings/wasm/bindings.js
+++ b/bindings/wasm/bindings.js
@@ -10,6 +10,14 @@ Module.setup = function () {
     return vec;
   }
 
+  function fromVec(vec, f = x => x) {
+    let result = [];
+    const size = vec.size();
+    for (let i = 0; i < size; i++)
+      result.push(f(vec.get(i)));
+    return result;
+  }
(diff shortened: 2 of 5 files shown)


pca006132:
For adding unit tests, it seems that we can just put a `test.mjs` next to the output `manifold.js`, with something like this:
```js
import Module from "./manifold.js";

Module.onRuntimeInitialized = function() {
  Module.setup();

  const {cylinder, sphere, union} = Module;

  function roundedFrame(edgeLength, radius, circularSegments = 0) {
    let edge = cylinder(edgeLength, radius, -1, circularSegments);
    let corner = sphere(radius, circularSegments);

    let edge1 = union(corner, edge);
    edge1 = edge1.rotate([-90, 0, 0]).translate([
      -edgeLength / 2, -edgeLength / 2, 0
    ]);

    let edge2 = edge1.rotate([0, 0, 180]);
    edge2 = union(edge2, edge1);
    edge2 =
        union(edge2, edge.translate([-edgeLength / 2, -edgeLength / 2, 0]));

    let edge4 = edge2.rotate([0, 0, 90]);
    edge4 = union(edge4, edge2);

    let frame = edge4.translate([0, 0, -edgeLength / 2]);
    frame = union(frame, frame.rotate([180, 0, 0]));

    return frame;
  }

  const result = roundedFrame(10, 1);
  // whatever assertion...
}
```

and run with Node.js. Wondering whether we should put the examples in a dedicated folder so we can share it in the `editor.js` and `test.mjs`. (perhaps we need typescript instead of mjs)

bindings/wasm/bindings.js
```diff
@@ -31,13 +39,13 @@ Module.setup = function () {
   Module.Manifold.prototype.warp = function (func) {
     const wasmFuncPtr = addFunction(function (vec3Ptr) {
       const x = getValue(vec3Ptr, 'float');
-      const y = getValue(vec3Ptr + 1, 'float');
-      const z = getValue(vec3Ptr + 2, 'float');
+      const y = getValue(vec3Ptr + 4, 'float');
+      const z = getValue(vec3Ptr + 8, 'float');
```
elalish, line 43: :100:

bindings/wasm/examples/editor.js
```diff
@@ -280,8 +285,10 @@ var Module = {
       const output = await worker.getEmitOutput(editor.getModel().uri.toString());
       const content = output.outputFiles[0].text + 'push2MV(result);';
       const exposedFunctions = [
-        'cube', 'cylinder', 'sphere', 'extrude', 'revolve',
-        'union', 'difference', 'intersection',
+        'cube', 'cylinder', 'sphere', 'tetrahedron', 'extrude', 'revolve',
+        'union', 'difference', 'intersection', 'compose', 'levelSet',
+        'setMinCircularAngle', 'setMinCircularEdgeLength', 'setCircularSegments',
+        'getCircularSegments'
```
elalish, line 291: We could make some separate lists: `constructors`, `staticDefaults`, `memberFunctions` and concatenate them here so we don't have to remember to update them in multiple places.

bindings/wasm/CMakeLists.txt
```diff
@@ -19,9 +19,10 @@ add_executable(manifoldjs bindings.cpp)
 # make sure that we recompile the wasm when bindings.js is being modified
 set_source_files_properties(bindings.cpp OBJECT_DEPENDS
   ${CMAKE_CURRENT_SOURCE_DIR}/bindings.js)
-target_link_libraries(manifoldjs manifold)
+target_link_libraries(manifoldjs manifold sdf)
 target_compile_options(manifoldjs PRIVATE ${MANIFOLD_FLAGS} -fexceptions)
-target_link_options(manifoldjs PUBLIC --pre-js ${CMAKE_CURRENT_SOURCE_DIR}/bindings.js --bind -s ALLOW_TABLE_GROWTH=1)
+target_link_options(manifoldjs PUBLIC --pre-js ${CMAKE_CURRENT_SOURCE_DIR}/bindings.js --bind -s ALLOW_TABLE_GROWTH=1
+  -sEXPORTED_RUNTIME_METHODS=addFunction)
```
elalish, line 25: Do you need `removeFunction` here too?

elalish (approved):
Awesome, thank you! I like the idea of sharing the examples with the NodeJS tests - do you want to do that here, or in a follow-on PR?

elalish:
Regarding 1 & 2, they pretty much go together since you'll get an error status only when you construct from an input mesh. I don't feel the need to add these to our editor just yet, but they are an important part of the bindings for anyone else using our library. I do like the idea of converting our status into a JS exception; that feels more ergonomic and no one is going to ask us to compile JS without exceptions :smile:

elalish:
Regarding needing TS in order to share examples between NodeJS and the editor; I've been thinking about this too. Really the main advantage TS is giving us in the editor is autocompletion/docs with our `.d.ts` file, but I wonder if Monaco will also show that in JS mode? If so, it might be easier to just do everything in JS, especially since we compile and run at the same time anyway.

bindings/wasm/examples/editor.js
pca006132: yes, will do

bindings/wasm/CMakeLists.txt
pca006132: indeed, I forgot about this

pca006132:
(and this should fix #223, although there is no way for me to test it for now)

elalish (approved):
Great, thank you!

PR #226, opened 2022-10-04
fixed csg_tree bugs

pca006132:
1. Due to the shared cache between op nodes, the children list can have only 1 child and the original code will access a dangling pointer.
2. Users may supply empty/singleton lists to BatchBoolean, which CsgOpNode cannot handle the empty case correctly.

The new test is for bug 1. This bug is found when I port the old keyboard code (in python) to typescript and trying to clean it up.


elalish (approved):
This is why I'm glad we're writing these examples. Thank you!

PR #228, opened 2022-10-05
updated build info

### LINKED ISSUE #225: Build process needs better documentation.
Moving this from JSCAD:
@elalish I discovered that the Bytecode Alliance released wasmtime-dotnet 1.0 5 days ago, so I was going to try Manifold for speed. But I cannot get it to build. I am confused, you seem only to build it with Github CI?
I just wanted to get Python and WASM versions under Linux.

You seem to be only using Github CI to build. I need to build locally and have never used Github CI.
You say the yaml file contains "canonical build instructions", but I have no idea what tool to use to execute the yaml file.
I tried using CMAKE directly and got no where, how do you build this when you're doing development?

elalish:
Fixes #225 

Does this look complete?

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -20,15 +20,13 @@ set(CMAKE_EXPORT_COMPILE_COMMANDS ON)
 
 option(MANIFOLD_EXPORT off)
 option(MANIFOLD_DEBUG off)
-option(MANIFOLD_USE_CUDA on)
-set(MANIFOLD_PAR "TBB" CACHE STRING "Parallel backend, either \"TBB\" or \"OpenMP\" or \"NONE\"")
+option(MANIFOLD_USE_CUDA off)
+set(MANIFOLD_PAR "NONE" CACHE STRING "Parallel backend, either \"TBB\" or \"OpenMP\" or \"NONE\"")
 
 if(EMSCRIPTEN)
-    message("Building for Emscripten")
-    set(MANIFOLD_FLAGS -fexceptions)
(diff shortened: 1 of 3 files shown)


CMakeLists.txt
```diff
@@ -20,15 +20,13 @@ set(CMAKE_EXPORT_COMPILE_COMMANDS ON)
 
 option(MANIFOLD_EXPORT off)
 option(MANIFOLD_DEBUG off)
-option(MANIFOLD_USE_CUDA on)
-set(MANIFOLD_PAR "TBB" CACHE STRING "Parallel backend, either \"TBB\" or \"OpenMP\" or \"NONE\"")
```
elalish, line 24: I changed these defaults for minimal dependency. @pca006132 Does that look reasonable to you?

pca006132 (approved):
In addition to this, I'm thinking if we should enable debug info for Release build by default. Seems that the only problem is slightly larger binary size but we should be fine with that.

elalish:
> In addition to this, I'm thinking if we should enable debug info for Release build by default. Seems that the only problem is slightly larger binary size but we should be fine with that.

I'd rather not, as it pulls in a lot of STL code: strings, exceptions, stdio, etc. Besides, I'd like to believe our users should not have much need to debug our code.

elalish:
@googlebot ping

pca006132:
Google bot seems dead. Btw, remember to put the editor page to the readme!

elalish:
Google bot is fixed; I'll link to the editor soon, I promise :grin:

PR #229, opened 2022-10-06
added mac to CI

elalish:
I didn't realize Github's CI does MacOS as well; might as well add it.

### DIFF
--- a/.github/workflows/manifold.yml
+++ b/.github/workflows/manifold.yml
@@ -150,6 +150,25 @@ jobs:
         cd build/test
         ./Release/manifold_test.exe
 
+  build_mac:
+    runs-on: macos-latest
+    if: github.event.pull_request.draft == false
+    steps:
+    - uses: actions/checkout@v3
+      with:
+        submodules: true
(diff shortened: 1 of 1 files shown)


pca006132:
Maybe we should install tbb using homebrew to test multithread as well?

elalish:
> Maybe we should install tbb using homebrew to test multithread as well?

I suppose we could, but I think it might make the CI take quite a bit longer. We can always add it later. Any idea how much that speeds it up?

pca006132:
It will probably not speed things up, action only give us 2 threads. I think it will be useful because end users will probably want to compile it with tbb, and we should make sure that they can do this.

elalish:
Right, not on the CI, but how much speed-up does TBB give on a nice macbook pro? Anyway, do you know the magic commands to install tbb with homebrew on the CI?

elalish:
@google-cla status

pca006132:
I guess it should be similar to normal Linux computers? Macbook pro also have many cores. I guess the command should be
```
brew install tbb
```

PR #231, opened 2022-10-06
Editor examples

elalish:
@pca006132 I've added your examples from your [comment](https://github.com/elalish/manifold/issues/197#issuecomment-1256891587). They're separated now in a way they should be usable for Node.js tests too.

### DIFF
--- a/bindings/wasm/examples/examples.js
+++ b/bindings/wasm/examples/examples.js
@@ -1,0 +1,37 @@
+const examples = new Map();
+
+function appendExample(name, func) {
+  const whole = func.toString();
+  const lines = whole.split('\n');
+  lines.splice(0, 1);// remove first line
+  lines.splice(-2, 2);// remove last two lines
+  for (const line of lines) {
+    line.substring(2);// remove first two leading spaces
+  }
+  const body = '\n' + lines.join('\n');
+  examples.set(name, body);
+}
+
+function Intro() {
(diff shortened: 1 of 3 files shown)


bindings/wasm/examples/examples.js
```diff
@@ -0,0 +1,170 @@
+const examples = new Map();
+
+function appendExample(name, func) {
+  const whole = func.toString();
```
pca006132, line 4: Good to know that toString will keep the comments!

bindings/wasm/examples/examples.js
```diff
@@ -0,0 +1,170 @@
...
+  // Type e.g. "box." to see the Manifold API.
+  // Use console.log() to print output (lower-right).
+  // This editor defines Z as up and units of mm.
+  const box = cube([100, 100, 100], true);
+  const ball = sphere(60, 100);
+  // You must name your final output "result".
+  const result = box.subtract(ball);
+  return result;
+}
+appendExample('Intro', Intro);
```
pca006132, line 27: I think that the normal way to do this would be to put all these functions as object fields, and iterate over those fields? I guess that would be a bit cleaner. To add spaces, we can just require these function names to be in camel case and split them when the case changes from lower case to upper case.

elalish: Good call, I'll do that.

PR #232, opened 2022-10-08
Replaced warp example with heart

elalish:
This is one of my early thingiverse contributions, which I originally made in Matlab. So much better to be able to create shapes like this directly in ManifoldCAD! Plus it's a much more interesting example of warp. 

I ended up needing a lot of console logging to debug my math, which surfaced a bunch of problems with the console, so those are fixed here. Likewise with a bunch of saving/renaming issues. Nothing like using the product to find the bugs...

### DIFF
--- a/bindings/wasm/examples/editor.css
+++ b/bindings/wasm/examples/editor.css
@@ -81,7 +81,7 @@ model-viewer {
   flex: 1;
   min-height: 0;
   max-height: 25vh;
-  overflow: scroll;
+  overflow-y: scroll;
   background-color: #eee;
   white-space: pre-line;
   font-size: 14px;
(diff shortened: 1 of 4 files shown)


elalish:
I also ran into some trouble with infinite loops while debugging my heart; those are annoying - we really need a way to cancel. I suppose all this execution should really be in a worker anyway. Still, to keep from getting in a really annoying state when you've had to manually kill an unresponsive page I now record whether the last operation finished - if not I disable auto-run on load to keep from jumping right into unresponsiveness again.

pca006132:
Agreed, I think we should just put the wasm execution inside a web worker.

pca006132 (approved):

PR #234, opened 2022-10-08
fixes collider data race

### LINKED ISSUE #227: weird cuda problem
I'm running the tests with my new work laptop with a RTX3050TI, with Driver Version: 515.48.07, CUDA Version: 11.7, nvcc version cuda_11.6.r11.6/compiler.31057947_0

The tests will fail with
```
C++ exception with description "Error in file: /home/pca006132/code/manifold/src/manifold/src/boolean_result.cpp (185): 'edgePos.size() % 2 == 0' is false: Non-manifold edge! Not an even number of points." thrown in the test body.
[  FAILED  ] Samples.GyroidModule (381 ms)
```

when MANIFOLD_DEBUG is enabled, and memory corruption when it is disabled. I am not sure if this is a problem due to driver version and compiler version mismatch, a problem in my driver installation (I'm using optimus prime offloading), a bug in thrust, or a bug in our code. I will have a look at this after this week to see what's wrong with this.

Also, for the compute capability, should we only compile for `sm_61` or should we include more common architecture? Not sure what is the usual way of doing this.

pca006132:
This should fix #227. The problem with the old code is that when we traverse up the tree, we will depend on values on another branch, which may be set by another thread, but we don't have any synchronization. The weird thing is that I think the old code should cause issue for single threaded workload as well: Consider the case where `A -> B, A -> C`, thread 1 updated the value of `B` and then go to update the value of `A`, now it will increment the counter for `A`, i.e. telling the thread that updates `C` to stop at this node because `A` is already updated. However, for the single thread case, the update sequence might be something like `B, A, C`, and in that case we will not propagate the updates in `C` to `A` because the counter is already set... But somehow we never encounter this issue before.

This patch fixes this issue by always performing synchronization (at the end of each kernel launch), and do not perform early return. The kernel is run for `log2(NumNodes())` time as I assume that the tree is always balanced (correct me if I am wrong) but the depth of each leave may differ by 1 (typical balanced binary tree). There is actually still write-write data race, but the writes should write the same value because the children value they see should be the same, and in practice we will not have out-of-thin-air values so it should be fine I guess. One potential optimization is to perform early exit when we found that the box value is not being updated, but in order to do the optimization we have to shrink the set of states as well, which is a bit complicated and is left for later.

### DIFF
--- a/src/utilities/include/utils.h
+++ b/src/utilities/include/utils.h
@@ -106,11 +106,14 @@ thrust::counting_iterator<T> countAt(T i) {
 template <typename T>
 __host__ __device__ T AtomicAdd(T& target, T add) {
 #ifdef __CUDA_ARCH__
+  // required for synchronization
+  __threadfence();
   return atomicAdd(&target, add);
 #else
   std::atomic<T>& tar = reinterpret_cast<std::atomic<T>&>(target);
   T old_val = tar.load();
-  while (!tar.compare_exchange_weak(old_val, old_val + add))
+  while (!tar.compare_exchange_weak(old_val, old_val + add,
+                                    std::memory_order_seq_cst))
     ;
   return old_val;
 #endif
@@ -119,10 +122,12 @@ __host__ __device__ T AtomicAdd(T& target, T add) {
 template <>
 inline __host__ __device__ int AtomicAdd(int& target, int add) {
 #ifdef __CUDA_ARCH__
+  // required for synchronization
+  __threadfence();
   return atomicAdd(&target, add);
 #else
   std::atomic<int>& tar = reinterpret_cast<std::atomic<int>&>(target);
-  int old_val = tar.fetch_add(add);
+  int old_val = tar.fetch_add(add, std::memory_order_seq_cst);
   return old_val;
 #endif
 }


src/collider/src/collider.cpp
```diff
@@ -310,12 +309,17 @@ void Collider::UpdateBoxes(const VecDH<Box>& leafBB) {
   auto policy = autoPolicy(NumInternal());
   copy(policy, leafBB.cbegin(), leafBB.cend(), leaves.begin());
   // create global counters
-  VecDH<int> counter(NumInternal(), 0);
+  VecDH<int> states(NumLeaves(), 0);
+  sequence(policy, states.begin(), states.end());
+  for_each(policy, states.begin(), states.end(),
+           [] __host__ __device__(int& v) { v *= 2; });
+  int count = std::ceil(std::log2(nodeBBox_.size()));
```
elalish, line 316: Hmm, you might want to have a look at the paper this is based on; it's a radix tree, not a balanced binary tree. It's been awhile, but if I recall the depth is limited to 32 (or maybe 64?). I'm too tired to give this a proper look tonight; I want to make sure I understand this; both my wrong assumption and what you're trying.

pca006132: Good catch, it is indeed not a balanced binary tree and its depth is at most 64. Actually I found that my explanation is wrong (I misunderstood the counter usage last night), will update it.

pca006132:
Actually the original algorithm is fine if synchronization is added. The counter tracks whether both of the children of an internal node is updated, the only problem is it only perform synchronization for the counter, but the value it read might still be stalled.

pca006132:
OK I have updated the patch, instead of using log2(N) or fixed number of iterations, it will now loop until every node is updated. It will also perform stream compaction which can improve the performance a bit.

src/utilities/include/utils.h
```diff
@@ -106,11 +106,14 @@ thrust::counting_iterator<T> countAt(T i) {
 template <typename T>
 __host__ __device__ T AtomicAdd(T& target, T add) {
 #ifdef __CUDA_ARCH__
+  // required for synchronization
+  __threadfence();
```
elalish, line 110: Okay, so the idea of putting the fence in our `AtomicAdd` function directly is for ease of use such that they behave more consistently between CPU and GPU, right? That seems reasonable; are there any other atomic functions that should also have it? Have you noticed any performance change from this?

elalish (approved):
Thanks for the explanation in your bug report; the error makes sense to me now. I guess the reason we didn't run into this before was simply that it was pretty robust. Usually if your input box was wrong, it would be infinite, which would only make a box more conservative, which would reduce performance, but not actually change the answer generally.

src/utilities/include/utils.h
pca006132: Yes, this is to make them behave a bit more consistently. No, I did not notice any significant performance change from this.

PR #252, opened 2022-10-20
updated getMesh JS binding

elalish:
Now `getMesh` returns a mesh of the same form that can be used to produce a new manifold with `smooth`.


bindings/wasm/bindings.cpp
```diff
@@ -169,9 +169,9 @@ EMSCRIPTEN_BINDINGS(whatever) {
       .function("add", &Union)
       .function("subtract", &Difference)
       .function("intersect", &Intersection)
-      .function("getMesh", &Manifold::GetMesh)
+      .function("_GetMesh", &Manifold::GetMesh)
       .function("refine", &Manifold::Refine)
-      .function("_Warp", &Warp, allow_raw_pointers())
+      .function("_Warp", &Warp)
```
pca006132, line 174: `allow_raw_pointers` is not needed now? Is it documented in emscripten documentation?

pca006132 (approved):
Other than that LGTM

bindings/wasm/bindings.cpp
elalish: Actually, I noticed you didn't add it when you made `levelSet`, which uses the same thing; I believe you changed the pointer type from `void *`. So I figured I wouldn't need it here anymore either, and sure enough, it works.

pca006132: Oh OK, good to know

PR #258, opened 2022-10-24
Manifold.getMesh() optional direct heap speed-up for JS/WASM

stewartoallen:
Ignore mode 1 and 2. only using mode 0 for now. the other two modes should be useful, but blow up with mem access issues when benchmarking. I suspect malloc is failing. this should use whatever heap increase mechanism is used elsewhere.

oh, I threw in access to `isManifold()` which is missing in the JS bindings.


stewartoallen:
It looks like the mem pointer cast I'm using is ok for the WASM build, but breaks in the normal build.

stewartoallen:
the purpose of the optional return array factories is so that existing arrays (sub-arrays of SharedArrayBuffers in my case) can be used instead of creating new ones.

pca006132:
I will review it later today, thanks!

elalish:
Referencing #255 here since that seems to be where this PR got started. Nice perf improvement! Agreed that the JS could use a better connection to the C++. Out of curiosity, why is a new C++ API needed? Could this technique work as well by accessing the pointer that backs each `std::vector` with `.data()`?

pca006132:
Regarding the pointer cast, in WASM you have 32 bit pointers while x86-64 have 64 bit pointers. The typical way to do this is to use `size_t`. I think this approach is faster than using `data()` of `std::vector` because it can eliminate 1 copy.

I wonder if people using C++ may also have this kind of requirement as well. Ideally we should allow people using the C++ API to write the mesh data to their buffer directly with some layout flattening, e.g. using a lambda. The annoying thing with WASM is that we cannot just write to some js buffer, the buffer has to reside in WASM heap, so it has to be copied out for use, so the fastest way for WASM will require direct access to the private vectors.

stewartoallen:
Any advice on an alternative to malloc/free that allows or follows heap resizing? I'm not familiar with emscripten's automatic bindings or other helpers.

pca006132:
No, I think none of us have much experience with emscripten :(
And what is the problem with malloc/free?

stewartoallen:
when I run benchmarking in a loop, especially with modes 1 and 2 that alloc space on the heap for compacted indexes or unrolled vertices, it eventually throws a memory access out of bounds error. where is the code style guide? I'll have an updated PR today.
Screen Shot 2022-10-24 at 10 33 14 AM

pca006132:
For the memory access error, I think you divided the pointer by 4 and then send it back to free? I guess that is the problem.

For style guide, we don't really have one currently but we use clang-format to do C++ code formatting. The main problem with this PR is that it modifies the `manifold.cpp`, and we want to avoid having port specific things there. For me, I would prefer making this available to C++ as well as mentioned above and use that mechanism for js binding. Not sure how @elalish thinks.

stewartoallen:
you are correct on the memory pointer. face palm.

if you can think of a way this can be coded usefully for both C++ and WASM, I'll give it a go. I won't really be able to test or build tests for it. to call my C++ skills rusty is to insult rusty stuff everywhere.

bindings/wasm/bindings.js
```diff
@@ -162,6 +170,49 @@ Module.setup = function () {
...
+        break;
+      case 2:
+        // in some cases, we want the unrolled vertex array (non-indexed)
+        // doing this work in WASM is substantially faster
+        vertex.set(fmem.subarray(mempos + 6, mempos + 6 + eR * 9));
+        break;
+    }
+    this._FreeMeshDirect(mempos * 4);
+
+    return { vertex, index };
```
elalish, line 213: I would guess that this part: that the JS function returns a flat array, is what is giving the majority of the speedup. Could we try bench-marking just that without the rest of the C++ changes?

src/manifold/src/manifold.cpp
```diff
@@ -138,6 +138,56 @@ Mesh Manifold::GetMesh() const {
...
+    if (mode == 1) {
+        for (int i = 0; i < nt * 3; i++) {
+            x[rp++] = impl.halfedge_[i].startVert;
+        }
+    } else if (mode == 2) {
+        for (int i = 0; i < nt * 3; i++) {
+            const glm::vec3 v = impl.vertPos_[ impl.halfedge_[i].startVert ];
+            f[rp++] = v.x;
+            f[rp++] = v.y;
+            f[rp++] = v.z;
```
elalish: Since you're getting 100x speed increase, rather than 2x, it makes me think this is less about having one copy rather than two, and more about using a fast C++-style copy rather than a slow JS array-of-arrays copy. If that's the case, then it would be nice to avoid all this old-school C-pointer stuff and new API, and just use two fast copies. Just trying to balance maintainability against performance.

bindings/wasm/bindings.cpp
```diff
@@ -170,6 +170,8 @@ EMSCRIPTEN_BINDINGS(whatever) {
       .function("subtract", &Difference)
       .function("intersect", &Intersection)
       .function("_GetMesh", &Manifold::GetMesh)
+      .function("_GetMeshDirect", &Manifold::GetMeshDirect)
+      .function("_FreeMeshDirect", &Manifold::FreeMeshDirect)
```
elalish, line 174: Would you mind adding the example you're using for bench-marking? You might make a copy of three.html and modify it. I'd like to see how usage of the JS side of the API looks. Also, if you can include a summary of your bench-mark results in this PR (most of them are in the discussion at the moment), that would be very helpful!

stewartoallen: there are several speed-ups with this approach. it avoids the construction of an intermediate nested object representation of flat arrays and all the accessor calls that go with it. it also avoids the need to re-construct the flat arrays from the intermediate object representation. this is the bulk of the speedup. not the copy. it also reduces memory pressure by avoiding the cleanup of the intermediate structures. when you take all of that into account, it's more like a 250x speedup. I wasn't including the re-encoding of arrays in the benchmark.

I posted the benchmark code in the discussion [here](https://github.com/elalish/manifold/discussions/255#discussioncomment-3946541).

src/manifold/src/manifold.cpp
stewartoallen: this is mode 2 which produces a flattened vertex array in WASM. I threw that in because when you want the vertices in this format, it is a lot faster than doing it in JS. mode 1, based on benchmarking, is the one I most care about. it still has old school pointer stuff, though. I think it's unavoidable when you just want to hand JS the location of the arrays on the heap for fast copy-out. Emscripten is hiding stuff like this somewhere in its bindings so that the rest of the code looks neat.

bindings/wasm/bindings.js
stewartoallen: that's mode 0 (just the heap pointers) vs mode 1 (heap pointer to flat vertex array and malloc'd new flat index array). you can't get either of these without the C++ helper function to return the pointers to the heap locations.

I, too, would prefer a more elegant-looking solution.

stewartoallen:
another approach to the API is to make an optional "`direct`" parameter only on the JS `getMesh()` call that could return flattened arrays instead of an object structure. the rationale here is that anyone using the C++ API could already access the data directly and doesn't really need `getMeshDirect()`

stewartoallen:
This latest commit cleans things up a bit and removes the extra API endpoint. I would like to write a test case for this, but am completely lost on how do to this for the JS side.

bindings/wasm/bindings.cpp
elalish: Okay, would you mind just updating the ManifoldCAD output [function](https://github.com/elalish/manifold/blob/master/bindings/wasm/examples/worker.js#L149)? That'll make it very easy for us to bench-mark a variety of uses, and of course we'll want to use your fast version here anyway.

elalish:
Indeed, we were just talking about adding JS tests (we don't have any yet 😱); I'll set up a framework for that soon.

bindings/wasm/bindings.cpp
stewartoallen: will do as soon as I get this PR into its final form.

src/manifold/src/manifold.cpp
```diff
@@ -138,6 +138,52 @@ Mesh Manifold::GetMesh() const {
...
+  int rp = 0;
+  // fixed header
+  x[rp++] = (size_t)impl.vertPos_.begin();
+  x[rp++] = (size_t)impl.vertPos_.end();
+  x[rp++] = nv;
+  x[rp++] = (size_t)impl.halfedge_.begin();
+  x[rp++] = (size_t)impl.halfedge_.end();
+  x[rp++] = nt;
+  x[rp++] = (size_t)impl.vertNormal_.begin();
+  x[rp++] = (size_t)impl.vertNormal_.end();
```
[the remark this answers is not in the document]
elalish: How about avoiding malloc and free using some modern constructs?
```
struct MeshGL {
  std::array<float> vertPos;
  std::array<uint> triVerts;
  std::array<float> vertNormal;
}
```

elalish (changes requested):

elalish:
Thank you, I think this is definitely taking things in the right direction.

bindings/wasm/bindings.js
```diff
@@ -143,7 +151,58 @@ Module.setup = function () {
...
+      const nE = head[7]; // normals vector end
+      const vertex = vertexFactory(mode === 2 ? eR * 9 : vR * 3);
+      const index = mode != 2 ? indexFactory(eR * 3) : undefined;
+      // helper does not unroll vertices and normals. use mode 0 or 1 and do manually
+      const normal = normalFactory && mode !== 2 ? normalFactory(vertex.length) : undefined;
+      switch (mode) {
+        case 0:
+          vertex.set(fmem.subarray(vS / 4, vE / 4));
+          // 30x slower than for let
+          // index.set(indx.filter( (v,i) => i % 4 === 0));
```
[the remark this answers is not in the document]
pca006132: I guess he means `index.set(indx.filter(...))` is 30x slower than a for loop. This is probably due to javascript creating an array for `indx.filter(...)` and then copying it to `index` in the first case.

src/manifold/src/manifold.cpp
```diff
@@ -138,6 +138,52 @@ Mesh Manifold::GetMesh() const {
...
+    for (int i = 0; i < nt * 3; i++) {
+      x[rp++] = impl.halfedge_[i].startVert;
+    }
+  } else if (mode == 2) {
+    // return unrolled and flattened vertex array after header
+    for (int i = 0; i < nt * 3; i++) {
+      const glm::vec3 v = impl.vertPos_[impl.halfedge_[i].startVert];
+      f[rp++] = v.x;
+      f[rp++] = v.y;
+      f[rp++] = v.z;
```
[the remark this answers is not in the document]
stewartoallen: will updated the PR later today and remove the modes (which were primarily instructive for benchmarking)

bindings/wasm/bindings.js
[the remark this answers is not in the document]
stewartoallen: yes. there is a lot of this sort of thing in the bindings, too

src/manifold/src/manifold.cpp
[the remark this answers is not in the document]
elalish: Yes, the WASM side will need to call `delete()`, which will then invoke the default destructor of `std::array` to free the memory. I believe there will always be a copy somewhere - part of why we don't expose our internal data structures is because on some platforms they reside on the GPU. We just need to make that copy efficient. At the end of the day these buffer copies (at least in C++) are dramatically faster than the operations this library is doing anyway. For this reason I don't want to worry about things like conditionally including normals - I think it's more API complexity than it's worth. 

I think copying our buffers to the right form in the C++ function and then using a `subarray` to pull a `std::array.data()` pointer into a `Float32Array` in JS should give us most of the speedup we need. If I'm wrong though, please show me the bench marks!

bindings/wasm/bindings.js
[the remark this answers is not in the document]
elalish: Got it, thanks. If you see any more optimization opportunities in these JS bindings, let's fix them! In the next PR...

src/manifold/src/manifold.cpp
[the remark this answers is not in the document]
stewartoallen: Understood wrt GPU and internal data. I hope this can be approached iteratively as this may require C++ skills not currently in my inventory and the gains are important to CAM simulation work.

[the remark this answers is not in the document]
stewartoallen: I would very much like to be able to construct a Manifold.Mesh directly from arrays, as well. But it was not clear to me how to do that.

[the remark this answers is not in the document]
hrgdavor: > I would very much like to be able to construct a Manifold.Mesh directly from arrays, as well. But it was not clear to me how to do that.

that could be very useful :) ... I am not sure exactly, but sound like a nice option to have

src/manifold/src/manifold.cpp
```diff
@@ -138,6 +138,29 @@ Mesh Manifold::GetMesh() const {
...
+  float* f = (float*)u;
+
+  int rp = 0;
+  // header
+  u[rp++] = NumVert();
+  u[rp++] = NumTri();
+  u[rp++] = (size_t)impl.vertPos_.begin();
+  u[rp++] = (size_t)impl.vertPos_.end();
+  u[rp++] = (size_t)impl.vertNormal_.begin();
+  u[rp++] = (size_t)impl.vertNormal_.end();
```
elalish, line 155: I don't think we can safely pass around pointers to our internal data structures. First, they need to be pointers to const or someone could reek all kinds of havoc. Second, what if this impl goes out of scope before this pointer gets read? 

We'll need to copy the vert buffers just like the triangle buffer, but at least we can copy them into the right shape:
```
struct MeshGL {
  int numVert, numTri;
  std::unique_ptr<float[]> vertPos, vertNormal;
  std::unique_ptr<uint32_t[]> triVerts;
};

MeshGL out;
out.vertPos = std::make_unique<float[]>(3*numVert);
for(int i = 0; i < numVert; ++i) {
  glm::vec3 v = impl.vertPos_[i];
  out.vertPos[3*i] = v.x;
  out.vertPos[3*i + 1] = v.y;
  out.vertPos[3*i + 2] = v.z;
} 
```
And use `out.vertPos.get()` for the pointer in your `subarray` call in JS. Embind [supports](https://emscripten.org/docs/porting/connecting_cpp_and_javascript/embind.html#smart-pointers) `std::unique_ptr` out of the box.

elalish (changes requested):
Thanks for the cleanup; definitely easier to read now.

src/manifold/src/manifold.cpp
stewartoallen: the JS + WASM VMs run in the same thread, so there is no chance the heap changes under you before the data is copied out. at least, this is true for the way the code is currently written. I'll try the approach above and see how that performs. I understand the desire for parity across targets. some performance gains won't necessarily translate.

stewartoallen: [trying and failing](https://github.com/stewartoallen/manifold/tree/direct-mesh-data-access-proper) to do this the right way ... not understanding how to complete the abstraction / binding / wrapping

src/manifold/src/manifold.cpp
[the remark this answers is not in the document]
elalish: Agreed, once we settle on some structure like `MeshGL`, we should enable it for both input and output.

src/manifold/src/manifold.cpp
elalish: Agreed this would probably be safe for the WASM API, but our C++ API is even more important (that's the only one you can use currently to get GPU speedups). And yes, the Embind stuff is a bit convoluted. We're all figuring it out together - go ahead and push your attempt so we can look it over and give pointers.

elalish:
Great discussions! Led to #263 and #269, but enough has changed it's probably time to start with a fresh PR for the updated JS API.

PR #262, opened 2022-10-26
Adding NodeJS tests for WASM module

elalish:
In order to get this working in Node I needed to modularize our WASM bundle, which I think is a good practice regardless. I tried going all the way to an ES6 module, but couldn't make it work. UPDATE: turns out I wasn't crazy, [ES6 modules from Emscipten don't yet work in NodeJS](https://github.com/emscripten-core/emscripten/issues/11792). So we'll just stick with CommonJS for now until they fix that (I don't feel like dealing with workarounds, since this works just fine).

### DIFF
--- a/bindings/wasm/test/test.js
+++ b/bindings/wasm/test/test.js
@@ -13,8 +13,9 @@
 // limitations under the License.
 
 const expect = require('chai').expect;
+const wasm = require('../manifold.js');
 const examples = require('../examples/examples');
-
+const THREE = require('three');
 
 // manifold member functions that returns a new manifold
 const memberFunctions = [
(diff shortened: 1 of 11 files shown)


bindings/wasm/test/test.js
```diff
@@ -0,0 +1,105 @@
...
+  function runExample(name) {
+    try {
+      const content = examples.functionBodies.get(name) + '\nreturn result;\n';;
+      const f = new Function(...exposedFunctions, content);
+      const manifold = f(...exposedFunctions.map(name => Module[name]));
+      const prop = manifold.getProperties();
+      const genus = manifold.genus();
+      return { ...prop, genus };
+    } catch (error) {
+      console.log(error.toString());
```
pca006132, line 84: Just wondering do we need to catch the error. I think it is also fine to just let the error propagate and catch by the unit test framework?

pca006132 (approved):

bindings/wasm/test/test.js
elalish: Done.

PR #263, opened 2022-10-27
Manifold.getMesh() optimized for GL input/output

stewartoallen:
This is an update to https://github.com/elalish/manifold/pull/258 using C++ conventions and Emscripten bindings


stewartoallen:
I'm not as worried about the JS bindings yet. Just want to get the C++ to compile.

src/utilities/include/public.h
```diff
@@ -167,6 +167,13 @@ struct Mesh {
   std::vector<glm::vec4> halfedgeTangent;
 };
 
+struct MeshGL {
+  int numVert, numTri;
+  std::unique_ptr<float[]> vertPos;
+  std::unique_ptr<float[]> vertNormal;
+  std::unique_ptr<uint32_t[]> triVerts;
```
elalish, line 174: I see this compile error, which in the CI, which seems pretty straight-forward: `manifold/manifold/src/utilities/include/public.h:174:3: note: 'std::unique_ptr' is defined in header '<memory>'; did you forget to '#include <memory>'?`

stewartoallen: including that doesn't change the extremely long list of compiler errors and warnings.

elalish: Okay, that was the only one I saw on the CI; as usual we'll have to work through them one a time. If your compiler isn't giving you very useful debug output, you might try looking at our CI output - sometimes its compiler gives more targeted info. Anyway, please push with whatever fixes you make until our CI gives an error you don't know what to do with and then we'll help. I'm traveling, so it may be a bit random when I have time to look at this.

stewartoallen: haven't had a lot of time this week. mostly I'm at my limits with C++ since it's been 20+ years since I last touched it. none of the errors spark any sort of helpful insights. most are related to unique_ptr in some way.

elalish: Agreed that emscripten's error messages leave something to be desired. I think it's basically ticked off because you can't copy a unique_ptr, and embind is trying to make a setter for that member of the struct, which isn't possible (or needed). I have something almost working, I think, but need to get some sleep.

elalish:
Closing in favor of #269, which was built on this.

PR #266, opened 2022-11-01
Updated JS binding type definitions

rafern:
The type definitions for the JS binding don't match the actual usage of the library, which creates a lot of issues when using the library in a Typescript project


rafern:
Closing and making a new pull request later. There are more things that need to be changed

PR #267, opened 2022-11-02
WASM bindings: add missing smooth function to the ManifoldStatic interface

jirihon:
Add missing smooth function to the `ManifoldStatic` interface.


elalish (approved):
Thanks!

PR #268, opened 2022-11-02
Updated JS bindings

rafern:
This is a continuation of #266.

The type definitions for the JS bindings don't match the actual usage of the library, which creates a lot of issues when using the library in a Typescript project. To fix this:
- Added setup function
- Changed Module definition to be a function that returns a Promise with ManifoldStatic
- Added Manifold constructor
- Optional Mesh properties no longer throw an exception when they aren't included

The Manifold.isManifold method was also exposed, and the Manifold constructor was changed to use a JS-style Mesh object (Vec3 arrays, etc...) instead of serialized Meshes (Vector_vec3 arrays, etc...). The old constructor which used serialized meshes was kept as a new function (ManifoldFromMeshVec) in case the user wants to reuse serialized Meshes.

The Manifold constructor now also checks for errors and throws an error with the appropriate code and message.

Other changes:
- Fixed broken MeshRelation deserialization (the `vertBary` field was broken)

This changes the current API, so I'm keeping this as a draft PR as there's probably going to be some disagreements. Let me know if you want anything changed


bindings/wasm/bindings.js
```diff
@@ -190,6 +191,63 @@ Module.setup = function () {
...
+    return manifold;
+  }
+
+  Module.Manifold = function Manifold(mesh) {
+    const meshVec = mesh2vec(mesh);
+    const manifold = Module.ManifoldFromMeshVec(meshVec);
+    disposeMesh(meshVec);
+    return manifold;
+  }
+  Module.Manifold.prototype = Object.create(ManifoldCtor.prototype);
```
elalish, line 249: Ah, is this how you build the constructor? I tried, but did not figure that out. Thank you!

bindings/wasm/bindings.cpp
```diff
@@ -178,6 +178,7 @@ EMSCRIPTEN_BINDINGS(whatever) {
       .function("_Scale", &Manifold::Scale)
       .function("_Decompose", &Manifold::Decompose)
       .function("isEmpty", &Manifold::IsEmpty)
+      .function("isManifold", &Manifold::IsEmpty)
```
elalish, line 181: Let's skip this one; we discussed elsewhere removing it from the C++ API too; it's a bit of a misnomer at this point anyway, and it was only ever for internal testing.

bindings/wasm/bindings.js
```diff
@@ -190,6 +191,63 @@ Module.setup = function () {
     };
   };
 
+  Module.ManifoldError = function ManifoldError(code, ...args) {
+    let message = 'Unknown error';
+    switch (code) {
```
elalish, line 196: 💯

bindings/wasm/manifold.d.ts
```diff
@@ -424,6 +462,11 @@ declare interface ManifoldStatic {
...
+  ManifoldFromMeshVec: typeof ManifoldFromMeshVec;
+  Manifold: typeof Manifold;
+  Vector_vec3: typeof Vector_vec3;
+  Vector_vec4: typeof Vector_vec4;
+  setup: function(): void;
 }
 
-declare const Module: ManifoldStatic;
\ No newline at end of file
+declare function Module(): Promise<ManifoldStatic>;
```
elalish, line 472: 👍

bindings/wasm/manifold.d.ts
```diff
@@ -408,6 +438,14 @@ declare function setCircularSegments(segments: number): void;
 declare function getCircularSegments(radius: number): number;
 ///@}
 
+/**
+ * Create a Manifold from a serialized Mesh object (MeshVec). Unlike the
+ * constructor, this method does not dispose the Mesh after using it.
+ *
+ * @param meshVec The serialized Mesh object to convert into a Manifold.
+ */
+declare function ManifoldFromMeshVec(meshVec: MeshVec): Manifold;
```
elalish, line 447: Can we discuss this one a bit? I think we may need multiple ways to access a `Mesh`: fast flat buffers and easier to read, but slower vectors of arrays/structs. That's at the core of #263, and it would be great to get your opinion on that. Basically we need to settle on some consistency, and I clearly didn't think it through in the first version (for either our C++ or JS API).

elalish: The issue is, I don't think either of these `Mesh` structures fit the fast bill yet; we really should have something that can be copied straight into OpenGL/WebGL, and then no more than one other, more ergonomic structure.

elalish (commented):
Thanks, this is looking great. We need to get out a v2.0 on npm hopefully soonish, but I'd like to first figure out some better `Mesh` input/output APIs. Feedback welcome!

jirihon:
Thank you @elalish, @rafern and @stewartoallen for working towards better JS/TypeScript bindings. Manifold is already incredible C++ library. However, combined with fast JS bindings, it could reach really broad audience and transform the industry. It would unlock the whole new world of possibilities for web CAD apps.

bindings/wasm/bindings.cpp
rafern: Reverted. Also, I didn't notice this at the time, but I forgot to change the function reference 😬

bindings/wasm/manifold.d.ts
rafern: Not all WebGL engines have the same way of setting mesh data, but there are a lot of similarities between all of them. It usually boils down to whether they interleave vertex attributes, or provide a contiguous buffer for each vertex attribute. For example, to make a mesh in ThreeJS, you can give buffers to the engine, with one buffer per vertex attribute (contiguous). In Playcanvas, you have the choice between one buffer per attribute VS interleaved vertex attributes (in a single vertex buffer). Wonderland Engine is a bit of an oddball, since you can only set mesh data with vertex attribute accessors, because the low-level API for directly using buffers has been deprecated (as of 0.9.0 IIRC, the previous API used interleaved buffers).

I have never used raw WebGL, but from what I can tell, you can configure whether to use interleaved, or non-interleaved vertex buffers with gl.vertexAttribPointer, so it shouldn't matter for raw WebGL (as long as it's Float32Arrays, then you're golden).

In my opinion, I think stewartoallen's system is good, but the fast path for getMesh should be in a separate method instead of an argument to getMesh, and the buffers should be wrapped in a JS class with getter methods. The getters aren't necessary, but they would make working with the library easier. Something like this:

```JS
class FastMesh {
    constructor(positions, normals, indices) {
        // note that the flat buffers are still accessible. this is intentional
        this.positions = positions;
        this.indices = indices;
        this.normals = normals;
    }

    get size() {
        return this.index.length;
    }

    getVertex(index) {
        const offset = this.indices[index] * 3;
        return [
            this.positions.subarray(offset, offset + 3),
            this.indices.subarray(offset, offset + 3),
        ];
    }
}
```

Another thing to note is that direct mesh data will only ever be used in applications where there are no extra vertex attributes (like CAD programs, but even then it won't work if you want nice looking normals as explained later). In a game which needs to use CSG for, for example, terrain generation, the extra vertex attributes (which would be UVs in this case) need to be restored, so it's best to keep the buffers as contiguous buffers (non-interleaved/one vertex attribute type per buffer).

Yet another problem is that, to turn a mesh into a manifold, the mesh needs to be simplified, which effectively smooths out the normals. To restore the normals, the mesh needs to be desimplified (is that a word?), which means that you need to add new vertices to the buffer, which also makes using the direct meshes harder since Float32Arrays can't be resized (you need to make a new ArrayBuffer with a bigger size, and copy the old buffer to the new buffer, kind of like how C++ handles vectors but less efficient, since there is no capacity, only size)

(sorry for the wall of text)

elalish:
Thank you, @jirihon, that's the goal! And don't forget @pca006132, who has been hugely instrumental in getting this library to where it is now, especially the JS API.

bindings/wasm/manifold.d.ts
elalish: Agreed, non-interleaved buffers is the way to go here. I also really like your concept of building the ergonomics as a helper method of the same flat-buffer object. I think that can simply take the place of our existing `Mesh` and simplify everything. Since we've already made a breaking change and need to release, we might as well change this too. I would adjust your class slightly:
```
class Mesh {
  constructor(triVerts, vertPos, vertNormal) {
    this.triVerts = triVerts;
    this.vertPos = vertPos;
    this.vertNormal = vertNormal;
  }

  get numTri() {
    return this.triVerts.length;
  }

  get numVert() {
    return this.vertPos.length;
  }

  verts(tri) {
    return this.triVerts.subarray(3 * tri, 3 * (tri + 1));
  }

  position(vert) {
    return this.vertPos.subarray(3 * vert, 3 * (vert + 1));
  }

  normal(vert) {
    return this.vertNormal.subarray(3 * vert, 3 * (vert + 1));
  }
}
```
Regarding simplifying meshes into manifolds, I think what you're really referring to is a vertex with multiple properties. These need to be split into duplicate vertices for rendering, but that breaks manifoldness (and cannot always be undone). The `triProperties` and `MeshRelation` are designed to facilitate this conversion, but I really need to create some examples for that.

elalish (approved):
I think this PR looks great. I want to make further API changes, but that will likely be easier after this is merged.

elalish:
@rafern would you mind pushing some trivial change? I unmarked it from draft, but our CI still doesn't seem to want to run it.

rafern:
@elalish Done. Looking forward to the MeshRelation example. I've been trying to wrap my head around it lately, but I'm still unsure how to reverse the simplification stage

PR #269, opened 2022-11-04
Add GetMeshGL

elalish:
Adjustment to #263. @stewartoallen I've got the C++ stuff fixed up here for you. I've removed any update to the JS API for this PR. I'm thinking of going along the direction from this [thread](https://github.com/elalish/manifold/pull/268#discussion_r1014505864). While `GetMesh` and `GetMeshGL` make sense in the C++ API, I think for JS we only need the single `Mesh` API and it can be driven by `GetMeshGL`. Do you want to take a stab at it? I'm fine with this being a breaking change for the npm library - it'll be part of v2.0.


bindings/wasm/bindings.js
```diff
@@ -13,7 +13,7 @@
 // limitations under the License.
 
 var _ManifoldInitialized = false;
-Module.setup = function () {
+Module.setup = function() {
```
elalish, line 16: Ignore all these formatting changes; apparently I didn't have my VSCode set up to auto-format JS files. We should really add that check to our CI (I think we only check C files currently).

stewartoallen: will get back into it this weekend and take a shot at integrating JS

PR #261, opened 2022-10-25
fix various code documentation typos

luzpaz:
Found via `codespell -q 3 -S *.pdf,./**/third_party -L boxs,inout,intstruct`

PR #270, opened 2022-11-05
Fix spelling

elalish:
Alternative to #261.

### DIFF
--- a/bindings/wasm/manifold.d.ts
+++ b/bindings/wasm/manifold.d.ts
@@ -17,7 +17,7 @@ type Vec3 = [number, number, number];
 type Vec4 = [number, number, number, number];
 type Matrix3x4 = [Vec3, Vec3, Vec3, Vec3];
 type SimplePolygon = Vec2[];
-type Polygons = SimplePolygon | SimplePolygon[];
+type Polygons = SimplePolygon|SimplePolygon[];
 type Mesh = {
   vertPos: Vec3[],
   triVerts: Vec3[],
@@ -42,8 +42,8 @@ interface Vector<T> {
   set(idx: number, value: T): void;
   size(): number;
 }
-class Vector_vec3 implements Vector<SerializedVec3>{};
-class Vector_vec4 implements Vector<SerializedVec4>{};
+type Vector_vec3 = Vector<SerializedVec3>;
+type Vector_vec4 = Vector<SerializedVec4>;
 type MeshVec = {
   vertPos: Vector_vec3,
   triVerts: Vector_vec3,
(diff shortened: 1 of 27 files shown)


bindings/wasm/manifold.d.ts
```diff
@@ -42,8 +42,8 @@ interface Vector<T> {
   set(idx: number, value: T): void;
   size(): number;
 }
-class Vector_vec3 implements Vector<SerializedVec3>{};
-class Vector_vec4 implements Vector<SerializedVec4>{};
+type Vector_vec3 = Vector<SerializedVec3>;
+type Vector_vec4 = Vector<SerializedVec4>;
```
elalish, line 46: @rafern, I noticed TS didn't like a few of your changes here, so I updated them to I think what you were intending? Mind reviewing this file?

bindings/wasm/manifold.d.ts
```diff
@@ -465,9 +465,7 @@ declare interface ManifoldStatic {
   getCircularSegments: typeof getCircularSegments;
   ManifoldFromMeshVec: typeof ManifoldFromMeshVec;
   Manifold: typeof Manifold;
-  Vector_vec3: typeof Vector_vec3;
-  Vector_vec4: typeof Vector_vec4;
```
elalish, line 469: @rafern I don't think these are necessary, but maybe I'm missing something?

bindings/wasm/manifold.d.ts
rafern: Yup, that is what I intended. I never got an error on my end because I've been using esbuild instead of tsc, need to fix that on monday

bindings/wasm/manifold.d.ts
rafern: I added them because they're not just plain JS objects that need to follow that interface; they need to be created via a constructor. Since all the classes and methods that are only available after importing were put in ManifoldStatic, I also added those.

rafern (commented):
I'd keep the vector constructors in the ManifoldStatic interface, unless the mesh API is going to be replaced, in which case they aren't needed anymore

bindings/wasm/manifold.d.ts
elalish: Okay, well TS doesn't like them, and I will likely remove this anyway as you said, so good enough for now.

elalish:
@rafern @stewartoallen @pca006132 I'm having a problem that I don't think is actually related to this PR at all; I've been developing mostly on Ubuntu (gcc) and emscripten has been building just fine. I'm now on MacOS (clang) and emscripten appears to build fine, but the output is not a module anymore (I get `Uncaught TypeError: Module is not a function at worker.js:37:1` when I try to run the editor locally). I thought there might be a problem with CMake not getting the `MOULARIZE` flag in properly, but that doesn't seem to be it, though I can see that `manifold.js` definitely does not have the proper wrapper anymore. 

Anyway, I tested with master and have the same problem there, so I don't think it's this PR. Has anyone else run into a problem like this? This laptop is all I have with me and I'd really like to be able to test my changes.

pca006132:
What is your MacOS version and emscripten version? Perhaps I can try to ask my friends with a macbook to try to reproduce this and let me debug.

elalish:
@pca006132 Thank you! MacOS: 12.6.1, emscripten: 3.1.24 (from homebrew), clang: 16.0.0

pca006132:
I can look at it at the end of this week, for now maybe you can try nix or docker. nix support for darwin is actually pretty good iirc.

elalish:
Wow, okay nevermind, I must have been tired. The build is working just fine, I had simply forgotten to copy the right files around. I've updated our CMake script to automate this so it's easier to just serve from `buildWASM/bindings/wasm/examples`. Then the only trick is to disable the network cache so you actually get an updated file when you build.

PR #271, opened 2022-11-07
JS bindings: convert mesh.triVerts to array of arrays

jirihon:
Convert the items of `mesh.triVerts` to plain arrays to match the `Vec3` type which is `[number, number, number]`.

### DIFF
--- a/bindings/wasm/bindings.js
+++ b/bindings/wasm/bindings.js
@@ -157,8 +157,9 @@ Module.setup = function() {
     const oldHalfedgeTangent = result.halfedgeTangent;
     const conversion1 = v => ['x', 'y', 'z'].map(f => v[f]);
     const conversion2 = v => ['x', 'y', 'z', 'w'].map(f => v[f]);
+    const conversion3 = v => [v[0], v[1], v[2]];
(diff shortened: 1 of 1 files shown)


bindings/wasm/bindings.js
```diff
@@ -157,8 +157,9 @@ Module.setup = function() {
     const oldHalfedgeTangent = result.halfedgeTangent;
     const conversion1 = v => ['x', 'y', 'z'].map(f => v[f]);
     const conversion2 = v => ['x', 'y', 'z', 'w'].map(f => v[f]);
+    const conversion3 = v => [v[0], v[1], v[2]];
```
elalish, line 160: Hmm, I must be missing something; this looks like a no-op, doesn't it? What does this change?

jirihon: This is intended to convert elements of type `{0: number, 1: number, 2: number}` to elements of type `[number, number, number]`. If this patch is not applied, the elements of `result.triVerts` would be generic javascript objects. With this patch, the elements will by javascript arrays. I understand the function looks like no-op, as indexing object has the same syntax as indexing array.

Issue #218, opened 2022-09-25
Visual debug


elalish:
One of the most powerful features of OpenSCAD is its visual debug, where you put `#` in front of a line to see that object rendered in the final output and so view its position even if it's a "negative" object. This is key to learning the language of CAD, especially since it is easy to transform something wrong and especially in the case of a difference, for it to simply not affect the output at all and so be effectively lost.

I would like to enable this feature in our editor, but I believe it needs to be implemented within our library itself. I think this is a good idea anyway, since this feature will likely be useful to any users of our library that have a UX. @pca006132, I think this would fit most naturally into the `csg_tree` class, so I'd like your opinion on it.

As far as API, I'm thinking add two methods to the `Manifold` class: `void Mark()` and `std::vector<Mesh> GetMarkedMeshes() const`. The idea is that any manifold that you `Mark` saves that marked condition in the `csg_node`. Then when you call `GetMarkedMeshes` on a downstream product, it walks the CSG tree doing only operations upstream of the marked nodes. It applies all downstream transformations and outputs all the marked meshes (there may be several copies). Then an editor displays those meshes in a different color (say, red, semi-transparent) on top of the regular output mesh for script debugging purposes.

pca006132:
Sounds good, but I'm not sure how this will interact with operations such as smooth and warp, which skipping downstream boolean operations may not produce the same output.

elalish:
True, since warp is generally nonlinear, inserting an intersection will effectively change the interpolation. I doubt that's a huge deal in practice; warp is only going to be useful if the mesh is already sufficiently refined to make the interpolation fairly accurate. Smooth is no problem because it's basically only available as the first operation; we don't have a good way to push tangents through operations yet. 

The nice thing is it's just for debug, so it really doesn't need to be perfect so long as it's useful.

PR #280, opened 2022-11-11
Fix memory leak

### LINKED ISSUE #245: Memory leak problem
Hi, I was using this library to do following operations using the python bindings provided by this repository.

I think the memory leaks when the mesh triangularization fails.. Is there a way to fix this?

pca006132:
Fix #245. Not yet tested to see if it works on the particular example, but I think this should solve the root issue. The `meshids` is a rough estimate of the number of distinct mesh ids in the mesh, so we can initialize the hash table with large enough size and avoid many collision which will slow down the performance (especially when using CUDA because of warp divergent).


pca006132:
btw @elalish I am not sure why the step for hashtable is set to 127 by default. It seems that if we are using linear probing we should use 1?

src/utilities/include/hashtable.h
```diff
@@ -0,0 +1,124 @@
...
+  HashTableD(VecDH<Uint64>& keys, VecDH<V>& values, VecDH<uint32_t>& used,
+             uint32_t step = 1)
+      : step_{step}, keys_{keys}, values_{values}, used_{used} {}
+
+  __host__ __device__ int Size() const { return keys_.size(); }
+
+  __host__ __device__ bool Full() const { return used_[0] * 2 > Size(); }
+
+  __host__ __device__ void Insert(Uint64 key, const V& val) {
+    uint32_t idx = hash64bit(key) & (Size() - 1);
```
elalish, line 64: Since I can't see the diff, is there anything else you changed in this hash table? I can see you added this hash function, which may be good for your new uses, but I think will not be for its original SDF purpose. 

The reason I used a step of 127 is the same reason I used a trivial hash function: this is designed to be a locality-sensitive hash function, and therefor fast for a GPU cache. The threads in a warp are sequential Morton codes, so they are nearby in space, and when they make hash table entries (near the surface) they also hash to nearby (often sequential) locations. This keeps the cache in good shape, but if I collide, then stepping by 1 is nearly always another collision. Better to step far enough to get off the local surface, since we'll probably need a new cache line anyway (and the other threads can probably reuse it). 

I haven't done a proper performance comparison, but I imagine it'd be noticeable for a large SDF running on CUDA.

elalish (changes requested):
It's late and I haven't managed to read through this properly yet, but the main actionable thing is to make the hash function take another input so it can be used in its original form too. Thanks!

src/utilities/include/hashtable.h
pca006132: I separated keys and values into two vectors, because for the case of storing u64 key and u32 value, using a single vector with `pair<u64, u32>` will waste 32 bit of memory per element due to padding requirement. For the hash function for SDF, I just run the tests and don't see any performance regression, I am not sure how it will perform for larger SDF instances. I guess maybe I can try to come up with some larger instances and check?

pca006132: I wanted to use template to allow users to choose the hash function, but I somehow cannot get the compiler to accept it.

src/manifold/src/impl.cpp
```diff
@@ -655,25 +661,24 @@ void Manifold::Impl::IncrementMeshIDs(int start, int length) {
   ASSERT(start >= 0 && length >= 0 && start + length <= triBary.size(),
          logicErr, "out of bounds");
   const auto policy = autoPolicy(length);
-  // Use double the space since the Boolean has P and Q instances.
-  VecDH<int> includesMeshID(2 * Manifold::Impl::meshIDCounter_, 0);
+  HashTable<uint32_t> meshidTable(std::max(16u, meshids * 2));
```
elalish, line 664: Is this hash table really going to be big enough to warrant trying to use a GPU-accelerated one? Would a regular STL hash table be simpler?

src/manifold/src/impl.cpp
```diff
@@ -655,25 +661,24 @@ void Manifold::Impl::IncrementMeshIDs(int start, int length) {
...
-  const int numMeshIDs = includesMeshID[includesMeshID.size() - 1];
+  while (1) {
+    for_each(policy, begin, end, MarkMeshID({meshidTable.D()}));
+    if (!meshidTable.Full()) break;
+    meshidTable = HashTable<uint32_t>(meshidTable.Size() * 2);
+  }
+  inclusive_scan(
+      autoPolicy(meshidTable.Size()), meshidTable.GetValueStore().begin(),
+      meshidTable.GetValueStore().end(), meshidTable.GetValueStore().begin());
+  const int numMeshIDs = meshidTable.GetValueStore().back();
```
elalish, line 676: If you just want the number of hash table entires, can you just call `Entries()`?

src/utilities/include/hashtable.h
elalish: If you try GyroidModule at a higher resolution that should run it long enough for a stress test pretty easily.

src/manifold/src/impl.cpp
elalish: Also, shouldn't this be `min` instead of `max`? Or else aren't we allocating just as much memory as the vector was?

elalish (changes requested):

src/manifold/src/impl.cpp
pca006132: The hash table itself is not large enough, but we can have many vertices that we can benefit from using GPU acceleration. 

The max is used to avoid allocating a hash table that is too small at the start, which may cause slowness in the GPU implementation due to warp divergence. Although I am not sure if this is really needed, I added it while debugging meshids (it was set to some weird values previously but the bug is fixed)

src/manifold/src/impl.cpp
pca006132: Ah true!

pca006132: Oh it doesn't work, because every value in the map will be 1.

src/utilities/include/hashtable.h
pca006132: OK, it seems that the hash function has quite significant impact on the performance. I tested the Gyroid test `edgeLength=0.05`. The run using *random hash* took 1900ms while the run using the identity function took ~1500ms. I tried using different step size but it doesn't seem to matter in this case. I now added the generic hash function in the template parameter (it seems that the compiler is happy when we use c++17, which is already the case).

pca006132: I guess the problem is that we cannot avoid encountering primary clustering especially when we use locality sensitive hashing. Step size of 1, 127, or anything will not give us better running time because we can fall into another cluster after stepping through that many entries.

src/manifold/src/impl.cpp
pca006132: The problem in the original issue is that, the vector being allocated has size the global mesh id counter, which can be very large even when the mesh object itself only consists of a few mesh ids. A hash table is used so we can ignore those meshids that are not present in the current mesh.
For performance, I did not measure it against the ols version. It may be a bit slower in general, but as the global counter keep increasing, it will eventually become faster than the original one and uses less memory.

elalish: Ah, thank you, I missed what `meshids` was doing. Now that this is a hash table, is this scan to remove orphaned meshIDs useful anymore? I'm thinking in context of #282 and #218, that we could hit two birds with one stone by saving the transforms associated with each meshID. However, especially for #218, it would be important to keep them all rather than removing ones that don't contribute triangles.

elalish (commented):
Thanks, this is looking just about ready. Time to update the branch?

test/mesh_test.cpp
```diff
@@ -586,7 +586,7 @@ TEST(Manifold, Transform) {
 TEST(Manifold, MeshRelation) {
   const float period = glm::two_pi<float>();
 
-  Mesh gyroidMesh = LevelSet(Gyroid(), {glm::vec3(0), glm::vec3(period)}, 0.5);
+  Mesh gyroidMesh = LevelSet(Gyroid(), {glm::vec3(0), glm::vec3(period)}, 0.05);
```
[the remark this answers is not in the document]
pca006132: ah yes, I should not have committed this.

src/manifold/src/impl.cpp
pca006132: We can remove the scan by using the hash table index as some unique ID for each mesh id, although I am not sure how much of a difference can this make. And this will require a slightly different API. (get raw index or something)
This is probably not related to saving transforms because the hash table is just temporary, we need another thing to store them.

elalish: Oh right, I forgot we were also using the scan to assign new ids. Thanks!

elalish (approved):
Thanks!

Issue #91, opened 2022-04-05
Improve Godot Engine's boolean operator engine (CSG) with `elalish/manifold`


fire:
- [*] Propose idea
- [ ] Ask for comments
- [ ] Investigate code details
- [ ] Write a Godot Engine proposal

elalish:
I think this test and the following one demonstrate all of that: https://github.com/elalish/manifold/blob/master/test/mesh_test.cpp#L698

And of course the docs: https://elalish.github.io/manifold/classmanifold_1_1_manifold.html

fire:
Is it difficult to disable assimp?

elalish:
Assimp is only linked for building `meshIO`. It's separated for exactly this reason; I figured most users would use their own input/output code.

Calinou:
> Tech lead of Godot Engine is ok with requiring manifold geometry inputs for CSG

For reference, the current custom CSG implementation already has this as a requirement:

> Any mesh can be used for CSG, this makes it easier to implement some types of custom shapes. > Even multiple materials will be properly supported. There are some restrictions for geometry, though:
>
> - It must be closed
> - It must not self-intersect
> - Is must not contain internal faces
> - Every edge must connect to only two other faces
>
> —[Godot gets CSG support](https://godotengine.org/article/godot-gets-csg-support)

fire:
@elalish Can you remove boost graph dependency?

https://github.com/haasdo95/graphlite was the closest I can find.

fire:
I was able to extract https://github.com/V-Sekai/godot/tree/csg-manifold and did a build system trick to enable `.cu`,  but the boost dependency was too much after the 8th Boost folder...

elalish:
Interesting, I'm still relatively new to the world of C++ libraries and didn't imagine Boost would be a big problem dependency-wise. I'm only using it for a connected components algorithm - figured better not to reinvent the wheel. Still, that's a generic algorithm and I'm sure it's available elsewhere, or we could just write our own. Thoughts?

fire:
Can you evaluate https://github.com/haasdo95/graphlite? I wasn't able to find many substitutes.

What operations are you using?

elalish:
All I use is connected_components (in two places): https://github.com/elalish/manifold/blob/3e92f47c89b48aa384c41f27e2c7b1a63194c1c5/manifold/src/impl.cu#L427

[...]

PR #290, opened 2022-11-27
Property output

elalish:
The upshot of discussions from #274, #196, and #91 is that the existing `MeshRelation` isn't the best API for reapplying mesh properties (UVs, normals, etc). Part of that is because it requires users to write interpolation functions, but a more fundamental issue is that the barycentric data structure actually loses some information: when an edge with shared property verts is cut by a Boolean op, the new vert no longer shares its properties on each side of the retained edge. Over time, this can lead to a lot of unnecessary duplication of property verts for rendering. 

Instead, I'm now looking to make the new `MeshGL` struct focus on handling the property input and output. By adding a bit of merge information, it's able to take the non-manifold GL structures and allow exact reconstruction of the manifold (no geometric checks needed). This can also be saved in a glTF as a sparse accessor on the indices, so this could easily become a manifoldness extension to glTF.


hrgdavor:
This sounds amazing. Looking fwd to some kind of blog post or explainer when it is implemented :)

src/utilities/include/public.h
```diff
@@ -150,17 +150,36 @@ using Polygons = std::vector<SimplePolygon>;
 
 /**
  * An alternative to Mesh for output suitable for pushing into graphics
- * libraries directly.
+ * libraries directly. This may not be manifold since the verts are duplicated
+ * along property boundaries that do not match. The additional merge vectors
+ * store this missing information, allowing the manifold to be reconstructed.
```
elalish, line 155: This is the core of the new API, but it's still incomplete. For now the mesh/material/triangle reference still lives in the `MeshRelation`; I'm planning to build a different version of that information here and probably deprecate `MeshRelation` entirely.

elalish:
@pca006132 I could use a hand if you can repro: our CI is consistently erroring out with `Floating point exception`, but everything is running fine for me locally on my mac. It sounds like the most likely culprit might be an integer divide by zero, but I can't find any. It's probably a simple error, but I'm baffled as to why it would work for me and not the CI.

pca006132:
```
[ RUN      ] Boolean.Split

Thread 1 "manifold_test" received signal SIGFPE, Arithmetic exception.
0x00000000004cee78 in operator() (__closure=0x7fffffff5540) at /home/pca006132/code/manifold/src/manifold/src/edge_op.cpp:471
471	     const int newProp = prop.size() / numProp;
(gdb) p numProp
$1 = 0
```

Not sure why it works on your mac.

elalish:
Perfect, thanks @pca006132!

elalish:
@pca006132 I looked at the code, and that `/ numProp` is inside a check of `triProperties.size() > 0`, so having `numProp == 0` shouldn't be possible in there. I switched it to `properties.size() > 0` and now Linux is happy, but mac and windows are segfaulting. I'm pretty sure there must be some uninitialized data or writing outside of bounds somewhere. Sadly valgrind doesn't exist on the mac and `leaks` isn't showing anything. If you have a moment for code review / debug, that'd be awesome, since I still can't repro locally.

pca006132:
Yeah there are some invalid access. I tried adding some assertions and got some failures:
```diff
diff --git a/src/manifold/src/edge_op.cpp b/src/manifold/src/edge_op.cpp
index e046a9f..6516eca 100644
--- a/src/manifold/src/edge_op.cpp
+++ b/src/manifold/src/edge_op.cpp
@@ -143,6 +143,11 @@ void Manifold::Impl::SimplifyTopology() {
   sequence(policy, idx.begin(), idx.end());
   sort_by_key(policy, halfedge.begin(), halfedge.end(), idx.begin());
 
+  int numtri = NumTri();
+  ASSERT(all_of(policy, meshRelation_.triBary.begin(), meshRelation_.triBary.end(),
+        [=]__host__ __device__(const BaryRef& ref) {return ref.tri < numtri;}),
+      logicErr, "tri out of range");
+
   VecDH<int> flaggedEdges(halfedge_.size());
 
   int numFlagged =
@@ -461,7 +466,7 @@ void Manifold::Impl::RecursiveEdgeSwap(const int edge) {
     triBary[tri1].vertBary[perm1[0]] = newBary;
     triBary[tri0].vertBary[perm0[2]] = newBary;
     // Update properties if applicable
-    if (meshRelation_.triProperties.size() > 0) {
+    if (NumProp() > 0) {
       VecDH<glm::ivec3>& triProp = meshRelation_.triProperties;
       VecDH<float>& prop = meshRelation_.properties;
       triProp[tri0] = triProp[tri1];
diff --git a/src/manifold/src/impl.cpp b/src/manifold/src/impl.cpp
index 92b6ba7..a7c12ef 100644
--- a/src/manifold/src/impl.cpp
+++ b/src/manifold/src/impl.cpp
@@ -336,6 +336,7 @@ Manifold::Impl::Impl(const MeshGL& meshGL,
       mesh.halfedgeTangent[i][j] = meshGL.halfedgeTangent[4 * i + j];
   }
 
+  ASSERT(propertyTolerance.size() == numProp, userErr, "invalid number of properties");
   *this = Impl(mesh, triProperties, properties, propertyTolerance);
 }
```

Both of the assertions are sometimes invalid. For the first one regarding `BaryRef::tri`, I think that is triggering the segmentation fault you observed. The second assertion is based on the initialization of `numProp = propertyTolerance.size()` so I think the two should match. Here are the list of failures:

```
[ RUN      ] Manifold.MeshRelationRefine
unknown file: Failure
C++ exception with description "Error in file: /home/pca006132/code/manifold/src/manifold/src/impl.cpp (339): 'propertyTolerance.size() == numProp' is false: invalid number of properties" thrown in the test body.
[  FAILED  ] Manifold.MeshRelationRefine (0 ms)
[ RUN      ] Boolean.MeshRelation
unknown file: Failure
C++ exception with description "Error in file: /home/pca006132/code/manifold/src/manifold/src/impl.cpp (339): 'propertyTolerance.size() == numProp' is false: invalid number of properties" thrown in the test body.
[ RUN      ] SDF.Surface
unknown file: Failure
C++ exception with description "Error in file: /home/pca006132/code/manifold/src/manifold/src/edge_op.cpp (147): 'all_of(policy, meshRelation_.triBary.begin(), meshRelation_.triBary.end(), [=]__host__ __device__(const BaryRef& ref) {return ref.tri < numtri;})' is false: tri out of range" thrown in the test body.
[ RUN      ] Samples.TetPuzzle
unknown file: Failure
C++ exception with description "Error in file: /home/pca006132/code/manifold/src/manifold/src/edge_op.cpp (147): 'all_of(policy, meshRelation_.triBary.begin(), meshRelation_.triBary.end(), [=]__host__ __device__(const BaryRef& ref) {return ref.tri < numtri;})' is false: tri out of range" thrown in the test body.
[ RUN      ] Samples.Bracelet
unknown file: Failure
C++ exception with description "Error in file: /home/pca006132/code/manifold/src/manifold/src/edge_op.cpp (147): 'all_of(policy, meshRelation_.triBary.begin(), meshRelation_.triBary.end(), [=]__host__ __device__(const BaryRef& ref) {return ref.tri < numtri;})' is false: tri out of range" thrown in the test body.
[ RUN      ] Samples.GyroidModule
unknown file: Failure
C++ exception with description "Error in file: /home/pca006132/code/manifold/src/manifold/src/edge_op.cpp (147): 'all_of(policy, meshRelation_.triBary.begin(), meshRelation_.triBary.end(), [=]__host__ __device__(const BaryRef& ref) {return ref.tri < numtri;})' is false: tri out of range" thrown in the test body.
```

elalish:
@pca006132 Perfect, thanks! Fixed that issue and now my tests are actually running the code I care about - and seg-faulting, so now I can debug locally. I thought it was a bit fishy they passed so easily before...

fire:
How far is this from user testing?

elalish:
@fire I think it's pretty close; I'm down to one test failure that I think I know how to address. Then I need to update the API for WASM. I'm also going to have to remove the barycentric stuff from `MeshRelation`; part of why this took so much longer than I thought it would is that this process led to me realize how broken my barycentric solution actually was, so I couldn't just reuse its logic.

elalish:
Okay, I think I finally have this working, just need to update the WASM bindings for the new API.

src/manifold/src/impl.h
```diff
@@ -76,6 +80,11 @@ struct Manifold::Impl {
   int NumVert() const { return vertPos_.size(); }
   int NumEdge() const { return halfedge_.size() / 2; }
   int NumTri() const { return halfedge_.size() / 3; }
+  int NumProp() const { return meshRelation_.numProp; }
+  int NumPropVert() const {
+    return NumProp() == 0 ? NumVert()
```
pca006132, line 85: Should we return `NumVert()` when there is no property or 0 as none of the vertices have any property? Actually I prefer having something like `Error(NoProperty)`/`Ok(n)`, although this isn't really popular in C++..

elalish: I use `NumProp()` to determine if there aren't any properties, rather than `NumPropVert()`. I recall it simplified some code somewhere.

pca006132: OK thanks. Btw I will be a bit slow to review recently, so feel free to merge it first if you want, I will try to go over it later when I have more time. (plus this PR is huge and I am unfamiliar with the usecases, so will take more time to understand it)

elalish:
@geoffder I see a cbind compile error on the CI I don't understand. Would you mind taking a look?

geoffder:
https://github.com/elalish/manifold/blob/6a590657f06f8d891b89f79e832d03b7acc6754a/bindings/c/include/manifoldc.h#L190 `from_` and `to_vert` declarations should return `uint32_t` instead of `int` to match the definitions.

elalish:
@geoffder Ah, thank you! I updated those but somehow missed `manifoldc.h`. Fixed.

src/manifold/src/impl.h
elalish: Thanks, and no problem. Yeah, this did turn out huge, since it involved a wholesale replacement of the barycentric part of `MeshRelation` when I realized how broken it was. The good news is I feel like the testing is a lot tighter now - this took forever because of how many errors were caught in those sanity checks. So I feel pretty confident about it algorithmically, though it could probably use further cleanup. As a bonus it seems to run a bit faster now too.

Issue #286, opened 2022-11-16
Imprecise barycentric coordinates


rafern:
Barycentric coordinates in a MeshRelation object get warped after doing a boolean operation. I made a project that restores vertex attributes like UVs from a MeshRelation. If I do an operation that does nothing to the manifold (like translating by 0,0,0) and convert the manifold back to a mesh with restored vertex attributes, then the barycentric coordinates aren't warped:

![original](https://user-images.githubusercontent.com/15365765/202147164-5d94af28-e834-4ef2-9a80-24ec41c6837b.png)

However, if I do the same thing with an operation that modifies the manifold, then the barycentric coordinates are warped, resulting in UV artifacts:

![bary-warped](https://user-images.githubusercontent.com/15365765/202147403-f41e80ab-a9fa-4a28-9d18-e661dfae7709.png)

This only happens for high UV values. In the screenshots, the U values are in the 0-1 range, but the V values are in the 18-19 range (due to repeating textures), suggesting that these issues are being created due to low precision

elalish:
I think we'll need a more complete repro to see what's going on. I'm working on an example for restoring UVs right now, so we'll see if I run into this. That doesn't really look like rounding error to me, especially the duplicated `0.1`.

rafern:
I think you're right, something else is going on. I found a seam at the beginning where UV values are in the 0-1 range, and the vertex normals also have seams when there is a seam in the UVs:

![bary-bug](https://user-images.githubusercontent.com/15365765/202222714-f69a25ef-5446-42dc-bcc9-33bf044c6817.gif)

![bary-bug-uv-texture](https://user-images.githubusercontent.com/15365765/202222745-b7cf9f7a-f3ba-4d64-8cd5-38856093ef0c.gif)

The seams also appear when I'm doing an operation that does nothing, and using the MeshRelation to get the UVs and normals back.

Hopefully I'm not wasting your time because this is a bug on my end, but if you need a project to debug this, or don't want to start from scratch for the UVs example, the project is now open-source at [playkostudios/manifold-wle](https://github.com/playkostudios/manifold-wle) and [playkostudios/manifold-wle-example](https://github.com/playkostudios/manifold-wle-example). The project is written for Wonderland Engine instead of Three.js though.

elalish:
[...]

PR #291, opened 2022-11-28
Add missing triProperties, properties and propertyTolerance JS bindings

rafern:
Turns out these properties were missing in the Manifold constructor for the JS bindings.

Related issue: #286

Also, I just saw #290. ~~Feel free to merge both PRs into one~~ Actually, it looks like the other PR might completely replace this one, since it moves the missing properties to the MeshGL class? Keeping this open just in case.

### DIFF
--- a/bindings/wasm/bindings.cpp
+++ b/bindings/wasm/bindings.cpp
@@ -91,7 +91,8 @@ MeshGL MeshJS2GL(const val& mesh) {
   return out;
 }
 
-Manifold FromMeshJS(const val& mesh, const val& inTriProperties, const val& inProperties, const val& inPropertyTolerance) {
+Manifold FromMeshJS(const val& mesh, const val& inTriProperties,
+                    const val& inProperties, const val& inPropertyTolerance) {
   const MeshGL meshGL = MeshJS2GL(mesh);
   const int numTri = meshGL.NumTri();
   std::vector<glm::ivec3> triProperties;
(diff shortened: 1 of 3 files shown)


elalish:
Thanks! Yes, #290 is probably going to supercede this, but thanks for pointing it out and verifying it fixes the problem.

elalish:
@rafern I finally landed #290. This PR could certainly still be merged, but I'm considering removing `triProperties` from the API completely and saying properties go through `MeshGL` and `Mesh` is for geometry-only use. Honestly I'm even considering deprecating `Mesh` entirely, but maybe the C++ folks would still appreciate the `glm::vec3`s for convenience. I'm open to feedback here. Let me know how the new `MeshGL` API is for ergonomics.

elalish:
Okay, as of #309, these have been removed from the public API, so this one is not necessary anymore. What do you think of the new properties example?

rafern:
Sorry, I forgot to respond to your message last week.

I won't know for sure until I update my project to the new API, but the new API looks good. My only gripe is handling vertex normals, but it looks like you are already working on this so it should be fine (thanks btw, these are some good changes to the library!). Adding a transforms vector that is tied to a meshID is definitely a good idea, since there are other vertex properties, like tangents, that also need to be transformed.

The example is a different use-case than mine (in my use-case a single input manifold can have multiple materials, where each material is in a sub-mesh, but faceID should be able to handle that), although I recon most people will be using this library for CAD where there is a single material per input manifold like in the example, so it also seems fine.

elalish:
Thanks, your feedback will be valuable as I update this API. Honestly, I think usage of multiple materials will actually be pretty common, which is why I'm putting so much effort into this API. I plan to handle the sub-mesh case by taking all the `MeshGL` outputs as inputs as well, so as to make a lossless round-trip, which is what the `ReserveIDs()` API is about. I even have an idea for a glTF extension so we can pass multi-material manifolds losslessly through that format.

PR #292, opened 2022-12-11
C bindings to facilitate static FFI bindings

geoffder:
As [discussed](https://github.com/elalish/manifold/discussions/284), I've written C bindings to facilitate bindings through FFI from static compiled languages. I've been writing bindings to Manifold from [OCaml](https://github.com/geoffder/OManifold) using these bindings to make sure that they work as intended. They aren't polished and fully documented yet, but the API is covered and integration with the garbage collector seems correct.

The changes mostly live under [bindings/c](https://github.com/geoffder/manifold/tree/cbind/bindings/c), but the Manifold build/source/API has been touched in a few places, namely:

- sub-libraries are compiled as `OBJECT` and the `fPIC` flag is used when building the `c` bindings to allow more painless relocation of the archives/objects that is required for statically bundling them into executables in binding host languages (ran into this building the OCaml bindings)
- GPU `ExecutionPolicy` is avoided when the sdf parameter to `LevelSet` is not a static lambda (results in runtime failure)
- Optional control of `ExecutionPolicy` has been added to `LevelSet` to allow languages with thread/interpreter locks to avoid parallel access to the sdf closure
- `Decompose` has been broken up to allow the `c` bindings to know how many `Manifold` allocations will be needed in advance without needing to repeat the expensive part of the method.


elalish:
Thanks! Hopefully the CLA is no big deal? Any thoughts on what it'll take to get it passing our CI? This is great, but my main concern is it's a lot of code that'll need to be updated as other developments happen. What do you think is the best way to go about this? For instance, #290 is steadily growing in scope.

geoffder:
Ah it appears `std::optional` which I used to allow optional control over `ExecutionPolicy` for `LevelSet` wasn't available until C++17, so that is causing some failures.

elalish:
I don't have any problem with pushing us up to C++17. If it helps to update the Ubuntu version of our CI, that's fine too.

pca006132:
I think we were already using C++17 somewhere in the code, but the public API is C++14, so changing the version should not require updating Ubuntu. I think it would be better if we can reduce the size of this C binding, I think it can be reduced considering most of them are merely converting the fields to the C equivalent one, perhaps we can do some template magic here. (haven't look at it in details yet)

pca006132:
After reading this a bit more carefully, it seems that it is hard to simplify this by using templates. For the API design I guess we can discuss about it later when there are other users of this C binding? I am not using the binding so it is hard for me to think about the API design.

geoffder:
> After reading this a bit more carefully, it seems that it is hard to simplify this by using templates. 

I was using a template for the array conversion helper initially for example, but then I rewrote without it since there were only a couple instantiations anyway.

> For the API design I guess we can discuss about it later when there are other users of this C binding? I am not using the binding so it is hard for me to think about the API design.

For an idea of how these are used you can look at my C stub generation [descriptions](https://github.com/geoffder/OManifold/blob/main/c/manifold_c_function_descriptions.ml), and one of the [modules](https://github.com/geoffder/OManifold/blob/main/lib/manifold.ml) using them. Each language will have it's own quirks in interfacing with C of course, but I think that a lot of this pattern is preserved.

```ocaml
type t = C.Types.Manifold.t Ctypes_static.ptr

let size = C.Funcs.manifold_size () |> size_to_int
let destruct t = C.Funcs.destruct_manifold t

let alloc () =
  let finalise = Mem.finaliser C.Types.Manifold.t destruct in
  let buf = Mem.allocate_buf ~finalise size in
  buf, Ctypes_static.(Ctypes.coerce (ptr void) (ptr C.Types.Manifold.t) buf)

let add a b =
  let buf, t = alloc () in
  let _ = C.Funcs.manifold_union buf a b in
  t
```

Taking pre-allocated memory as an argument makes the process of ensuring it is registered with the host language properly fairly simple.

For a bit more complicate of an example using the additional machinery around `Decompose` that I added:

```ocaml
let decompose t =
  let comps_buf, comps = Components.alloc () in
  let _ = C.Funcs.manifold_get_components comps_buf t in
  let len = size_to_int @@ C.Funcs.manifold_components_length comps in
  let bufs = Ctypes.(CArray.make (ptr void) len)
  and ts = ref [] in
  for i = 0 to len - 1 do
    let buf, man = alloc () in
    Ctypes.CArray.set bufs i buf;
    ts := man :: !ts
  done;
  let _ = C.Funcs.manifold_decompose (Ctypes.CArray.start bufs) t comps in
  !ts
```

For a non-ocaml version of following this kind of C API binding scheme, there are the [olm python bindings](https://gitlab.matrix.org/matrix-org/olm/-/blob/master/python/olm/account.py), which as I mentioned in the discussion, is the library I used as an example. Minus the object-orientation, I used these official python bindings as a guide when I wrote mine in [ocaml](https://github.com/geoffder/olm-ml).

More people working on binding through these from other languages would certainly help to catch any rough edges that remain. They had to get a handful of fixes as I worked through them myself, like switching all arrays to pointers (including the dbl ptrs for the opaque types).

geoffder:
- After figuring out how to point emscripten to the correct overload of `Decompose`, the WASM bindings are building again. 
- There is one test failing on Mac and I'm not sure why, I haven't touched anything in Polygon
- Despite bumping ubuntu to 22.04, the update to using the variables in the new `material.h` (`pbrmaterial.h` deprecated in assimp) is still causing a breakage. The nix builds fine, so obviously ubuntu is still lagging. Maybe some template conditionals are in order for compatibility?

elalish:
Nice work! The Assimp stuff is not core to library, so feel free to hack in whatever's necessary to make it function. 

The `Boolean.Close` failure on mac is indeed interesting. I agree that's unlikely to be your fault, though it might be nice to make a small PR with only the C++17 bump to see if it repros there. In any case, I don't think it's your fault. You can change it to `TEST(Boolean, Close_DISABLED)` to get the CI passing. That'll remind me to look into it.

geoffder:
I tried adding the `_DISABLED` suffix, but it still runs and fails on the test, so I probably haven't executed on your suggestion properly. 

I opened https://github.com/elalish/manifold/pull/293 to test whether the std version bump is responsible, but it passed.

test/mesh_test.cpp
```diff
@@ -1053,7 +1053,8 @@ TEST(Boolean, Subtract) {
   first.GetMesh();
 }
 
-TEST(Boolean, Close) {
+// FIXME: test is failing on Mac CI (passing on others)
+TEST(Boolean, Close_DISABLED) {
```
elalish, line 1057: Sorry, meant `DISABLED_Close` - I always get that wrong...

test/meshIO/src/meshIO.cpp
```diff
@@ -19,11 +19,19 @@
...
-#include "assimp/pbrmaterial.h"
+#include "assimp/material.h"
 #include "assimp/postprocess.h"
 #include "assimp/scene.h"
 #include "optional_assert.h"
 
+#ifndef AI_MATKEY_ROUGHNESS_FACTOR
+#include "assimp/pbrmaterial.h"
+#define AI_MATKEY_METALLIC_FACTOR \
+  AI_MATKEY_GLTF_PBRMETALLICROUGHNESS_METALLIC_FACTOR
```
elalish, line 30: 👍

bindings/c/box.cpp
```diff
@@ -0,0 +1,97 @@
...
+#include <conv.hh>
+#include <vector>
+
+#include "types.h"
+
+ManifoldBox *manifold_box(void *mem, float x1, float y1, float z1, float x2,
+                          float y2, float z2) {
+  auto p1 = glm::vec3(x1, y1, z1);
+  auto p2 = glm::vec3(x2, y2, z2);
+  auto box = new (mem) Box(p1, p2);
```
elalish, line 14: What does this syntax mean? Is it just constructing a box at the pointer location `mem` and assuming there's enough space allocated there?

bindings/c/include/conv.hh
```diff
@@ -0,0 +1,46 @@
+#pragma once
+#include <manifold.h>
+#include <meshIO.h>
+#include <public.h>
+#include <sdf.h>
+#include <types.h>
+
+#include <vector>
+
+#include "glm/ext/matrix_float4x3.hpp"
```
elalish, line 10: What is this for?

bindings/c/include/conv.hh
```diff
@@ -0,0 +1,46 @@
+#pragma once
```
elalish, line 1: Why the `.hh` extension?

bindings/c/include/types.h
```diff
@@ -0,0 +1,78 @@
...
+typedef struct ManifoldVec2 {
+  float x;
+  float y;
+} ManifoldVec2;
+
+typedef struct ManifoldVec3 {
+  float x;
+  float y;
+  float z;
+} ManifoldVec3;
```
elalish, line 30: I take it this is because `glm` isn't a c library?

bindings/c/meshio.cpp
```diff
@@ -0,0 +1,72 @@
+#include "meshIO.h"
```
elalish, line 1: Do you even want to make bindings for this? MeshIO is nothing special, just a testing convenience.

bindings/c/testc.c
```diff
@@ -0,0 +1,91 @@
...
+  ManifoldVec3 res = {0.5 * x + 30, 0.5 * y + 30, z + 30};
+  return res;
+}
+
+const float radius = 15;
+const float xscale = 3;
+const float yscale = 1;
+const float zscale = 1;
+const float bb = radius * 2;
+float sdf_ellipsoid(float x, float y, float z) {
```
elalish, line 18: Thanks for writing tests! See my other comment about the CI.

bindings/c/testc.c
```diff
@@ -0,0 +1,91 @@
...
+  size_t bary_len = manifold_mesh_relation_barycentric_length(rel);
+  ManifoldVec3 *barycentric = manifold_mesh_relation_barycentric(
+      malloc(sizeof(ManifoldVec3) * bary_len), rel);
+  size_t print_until = (bary_len > 10 ? 10 : bary_len);
+  for (int i = 0; i < print_until; ++i) {
+    ManifoldVec3 v = barycentric[i];
+    printf("(%f %f %f)\n", v.x, v.y, v.z);
+  }
+  ManifoldVec3 last = barycentric[bary_len - 1];
+  printf("(%f %f %f)\n", last.x, last.y, last.z);
```
elalish, line 77: Should we make some `expect` calls to show when it fails? Does `GTest` have a C version?

bindings/c/manifoldc.cpp
```diff
@@ -0,0 +1,611 @@
...
+namespace {
+ManifoldMesh *level_set(void *mem, float (*sdf)(float, float, float),
+                        ManifoldBox *bounds, float edge_length, float level,
+                        bool seq) {
+  // typing with std::function rather than auto compiles when CUDA is on,
+  // passing it into GPU (and crashing) is avoided dynamically in `sdf.h`
+  std::function<float(glm::vec3)> fun = [sdf](glm::vec3 v) {
+    return (sdf(v.x, v.y, v.z));
+  };
+  auto pol = seq ? std::make_optional(Seq) : std::nullopt;
```
elalish, line 36: Is there a way we can void this bool parameter by making `LevelSet` choose the `Seq` policy when the template parameter is not a Functor? Curious of @pca006132 knows any magic here.

bindings/c/manifoldc.cpp
```diff
@@ -0,0 +1,611 @@
...
+  return to_c(new (mem) Manifold(warped));
+}
+
+ManifoldMesh *manifold_level_set(void *mem, float (*sdf)(float, float, float),
+                                 ManifoldBox *bounds, float edge_length,
+                                 float level) {
+  return level_set(mem, sdf, bounds, edge_length, level, false);
+}
+
+ManifoldMesh *manifold_level_set_seq(void *mem,
```
elalish, line 152: It'd be nice to not have two functions here; see my other comment above. But I can't tell if there's a way around it.

elalish (changes requested):
This is looking pretty good to me; mostly questions. What do you think, @pca006132?

src/polygon/CMakeLists.txt
```diff
@@ -15,7 +15,7 @@
 project (polygon)
 
 file(GLOB_RECURSE SOURCE_FILES CONFIGURE_DEPENDS *.cpp)
-add_library(${PROJECT_NAME} ${SOURCE_FILES})
+add_library(${PROJECT_NAME} OBJECT ${SOURCE_FILES})
```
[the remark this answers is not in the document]
geoffder: Since it is a sublibrary not used on it's own, this lets it compile separately still, but get linked into the main `a.`/`.so` being built. I was having some weird issues with relocating the many archives/objects when bundling for the ocaml library, and doing this to get rid of the extra library artifacts fixed it. The [entry](https://cmake.org/cmake/help/latest/command/add_library.html#object-libraries) in the cmake manual about it is most of what I know about OBJECT.

src/sdf/include/sdf.h
```diff
@@ -279,6 +281,20 @@ struct BuildTris {
...
+
+template <typename Func>
+void for_each_n_wrapper(ExecutionPolicy policy, int morton,
+                        ComputeVerts<Func> cv) {
+  for_each_n(policy, countAt(0), morton, cv);
+}
+template <>
+void for_each_n_wrapper<std::function<float(glm::vec3)>>(
+    ExecutionPolicy policy, int morton,
+    ComputeVerts<std::function<float(glm::vec3)>> cv) {
```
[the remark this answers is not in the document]
geoffder: Yeap! This means that dynamic lambdas can be used when CUDA is enabled without a runtime crash (something that the bindings can't avoid).

src/sdf/include/sdf.h
```diff
@@ -321,7 +338,15 @@ inline Mesh LevelSet(Func sdf, Box bounds, float edgeLength, float level = 0) {
...
   const Uint64 maxMorton = MortonCode(glm::ivec4(gridSize + 1, 1));
-  const auto policy = autoPolicy(maxMorton);
+
+  // Parallel policies violate will crash language runtimes with runtime locks
+  // that expect to not be called back by unregistered threads. This allows
+  // bindings use LevelSet despite being compiled with MANIFOLD_PAR
+  // active (CUDA is already avoided when Func is a function ptr).
+  const auto pol =
+      (!policy.has_value() || (policy.value() == ParUnseq && !CudaEnabled()))
+          ? autoPolicy(maxMorton)
```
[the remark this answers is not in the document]
geoffder: The template trick avoids CUDA, but languages that have locks on their runtime (like ocaml and python) will not allow their closures to be run in parallel (which `LevelSet` will do by handing the sdf function over to thrust when the policy is parallel. So the policy parameter here was a way of allowing the caller to specify sequential computation. In the C bindings, the overloaded `LevelSet` with policy param is used to provide a `manifold_level_set_seq` alternative to the original one.

As far as the check for `!CudaEnabled()`, I figured that theoretically the caller could be using a static lambda (not `std::function`) from C++ while not having CUDA disabled, so this guard prevent them from setting `ParUnseq` as their policy.

bindings/c/testc.c
geoffder: I'm not totally sure, I hadn't thought farther about the tests other than what I needed to do along the way to make sure things were working as I added them. :sweat_smile:

Happy to hear suggestions on how the testing should be done, laid out. I'm pretty unfamiliar with how it is usually done in C/C++.

bindings/c/meshio.cpp
geoffder: It doesn't do a lot and I know there isn't much in that file, but I figured that generally people are going to want to write their models into some formats, and `assimp` bindings don't necessarily exist already. Should spare some duplication of effort if people don't have to write their own off-ramps. Outputting to ascii stl or 3mf is quick enough to do on your own, but the more efficient binary formats that assimp can provide are nice to have. Speaking of, I haven't read more into it, but the default that is output right now for stl is ascii, would be nice to have the option to write binstl from MeshIO. 

Is there a particular reason why you want to leave mesh output as a just for testing convenience rather than providing something as part of `Manifold`?

bindings/c/manifoldc.cpp
geoffder: `ParSeq` is fine for `std::function` though generally, it is only a problem from particular host languages that might be using the C bindings, so I wanted to provide a way to lock in sequential usage of the closure without adding another separate named method to the API.

bindings/c/manifoldc.cpp
geoffder: I commented on the other one, but I think it is either this, or making the caller to provide a "boolean" (int) flag specifying that execution should be forced sequential. Unfortunately the factors that require avoiding parallel execution of the closure are external and not encoded into the type system as the distinction between static and dynamic lambdas are.

src/manifold/src/constructors.cpp
```diff
@@ -374,16 +369,15 @@ std::vector<Manifold> Manifold::Decompose() const {
   }
   std::vector<int> components;
   const int numLabel = ConnectedComponents(components, graph);
+  return {components, numLabel};
+}
 
-  if (numLabel == 1) {
-    std::vector<Manifold> meshes(1);
-    meshes[0] = *this;
-    return meshes;
```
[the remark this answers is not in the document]
geoffder: Oh I don't think I meant to get rid of that, sorry. I'll add that back in now!

.github/workflows/manifold.yml
```diff
@@ -35,7 +35,7 @@ jobs:
         git apply thrust.diff
         mkdir build
         cd build
-        cmake -DCMAKE_BUILD_TYPE=Release -DBUILD_SHARED_LIBS=ON -DMANIFOLD_DEBUG=ON -DMANIFOLD_PAR=${{matrix.parallel_backend}} -DMANIFOLD_USE_CUDA=${{matrix.cuda_support}} .. && make
+        cmake -DCMAKE_BUILD_TYPE=Release -DBUILD_SHARED_LIBS=ON -DMANIFOLD_DEBUG=ON -DMANIFOLD_PYBIND=ON -DMANIFOLD_CBIND=ON -DMANIFOLD_EXPORT=ON -DMANIFOLD_PAR=${{matrix.parallel_backend}} -DMANIFOLD_USE_CUDA=${{matrix.cuda_support}} .. && make
```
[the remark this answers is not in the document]
geoffder: Having them only turned on for one build (maybe one with Par and CUDA enabled) and not failing CI sounds good to me. I just flipped it on everywhere to make sure they were working.

bindings/c/box.cpp
```diff
@@ -0,0 +1,97 @@
...
+  auto v = box.Center();
+  return {v.x, v.y, v.z};
+}
+
+float manifold_box_scale(ManifoldBox *b) {
+  auto box = *from_c(b);
+  return box.Scale();
+}
+
+int manifold_box_contains_pt(ManifoldBox *b, float x, float y, float z) {
```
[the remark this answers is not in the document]
geoffder: C doesn't really do bool

bindings/c/include/conv.hh
geoffder: I think they were auto added my my editor, I should be able to remove them.

bindings/c/box.cpp
geoffder: Yeap that's right

bindings/c/include/conv.hh
geoffder: No reason other than I thought C++ headers were supposed to have `.hh` extensions.

bindings/c/include/types.h
geoffder: Yes it isn't, there is a `cglm` counterpart library. This way there isn't any dependency on `glm` to be able to understand the structs (at least that was my assumption). Not sure if there is a better way to handle the passing of vector arrays back and forth, I might have taken an overly cautious route.

src/utilities/include/public.h
```diff
@@ -255,6 +255,15 @@ struct Curvature {
   std::vector<float> vertMeanCurvature, vertGaussianCurvature;
 };
 
+/**
+ * Connected component indices and number of discrete components. Intermediate
+ * step for Manifold.Decompose(), completed by Manifold.GetComponents()
+ */
+struct Components {
+  std::vector<int> indices;
+  int discrete;
```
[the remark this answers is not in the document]
geoffder: Done!

bindings/c/include/conv.hh
geoffder: Renamed: https://github.com/elalish/manifold/pull/292/commits/1f77ef77f14fd76ace5d0608a683637fb557c0f2

src/manifold/src/constructors.cpp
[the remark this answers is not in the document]
geoffder: Added back in here: https://github.com/elalish/manifold/pull/292/commits/20a2fb5663a6ff935ee18299944a106abee2a42f

test/mesh_test.cpp
geoffder: Ok that did the trick, test is being skipped so the mac build is "passing". :+1:

bindings/c/include/conv.hh
geoffder: Went through and pruned a bunch of auto-included things. The C++ LSP running in my emacs must be doing it.

bindings/c/meshio.cpp
elalish: Yes, the reason is that formats are complex and lots of code to maintain and other projects are already better at it than mine. I need some boundaries, or else a hell of a lot more collaborators! 

I refuse to support STL though: STL cannot represent a manifold mesh in any kind of reliable way, so using this library and outputting STL is self-defeating. Pretty much any other format will do, 3MF being probably ideal. 

Anyway, if you want to maintain it, you can, but be warned: someone's going to want UVs and then atlasing and eventually you'll have to write Blender all over.

bindings/c/testc.c
elalish: GTest is the framework we use for the rest of our C++ tests and it looks like it'll be fine for your C too: https://stackoverflow.com/questions/5335268/is-google-test-ok-for-testing-c-code

It's pretty simple, I don't think you'll have any trouble adapting it from our existing tests.

bindings/c/manifoldc.cpp
elalish: Ah, okay I see now.

bindings/c/manifoldc.cpp
elalish: Okay this makes sense now; can you add a comment on these bindings so people know they should really only link to one or the other depending on their language?

bindings/c/include/types.h
elalish: Up to you as the user. I find glm to be an excellent library that makes my life much easier. If you want to include cglm for these bindings that's fine, but I don't have a strong opinion.

bindings/c/meshio.cpp
geoffder: I didn't mean to suggest re-writing what assimp is providing and increase your responsibilities, just to do what is basically done now -- letting assimp do it. My decision to use MeshIO basically just that the interface with assimp is already being provided by it, so it is simple to provide it along with the C bindings. If assimp bindings (or some native/bound serialization library) exists in the language that the C bindings are being used from, then there is no problem, otherwise using Manifold will also mean coming up with those solutions on their own. Does that make sense?

bindings/c/manifoldc.cpp
geoffder: Yeap I can do that.

bindings/c/meshio.cpp
elalish: Maybe what I really mean is, `Manifold` can handle any kind of crazy vertex properties you want, but I'm afraid someone will just look at my crappy little MeshIO function that only outputs vertex colors and say, "oh I guess this library doesn't work on my meshes" and move on. There are just so many ways to deal with mesh data; I'm doing my best not to make any assumptions, but once you output to a file you've necessarily picked one.

bindings/c/include/types.h
geoffder: Okay. Since there isn't actually any number crunching happening and it's just a struct of floats I think this way works well enough then.

src/sdf/include/sdf.h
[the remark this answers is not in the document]
elalish: > As far as the check for `!CudaEnabled()`, I figured that theoretically the caller could be using a static lambda (not `std::function`) from C++ while not having CUDA disabled, so this guard prevent them from setting `ParUnseq` as their policy.

I'm pretty sure this particular logic is in `autoPolicy` already; would be great if you can look it over to verify.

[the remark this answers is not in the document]
elalish: Oh, I see it now, thanks.

bindings/c/manifoldc.cpp
geoffder: Added here: https://github.com/elalish/manifold/pull/292/commits/e25964bad67db377c3c193f787d020ccb62d1fb5

bindings/c/meshio.cpp
geoffder: I don't think you have to worry about that too much! If people want more, they can always do it themselves with the data available from `Mesh`, I just think it's a nice batteries included solution to have output functions readily available, even if they aren't state of the art.

bindings/c/testc.c
geoffder: Ok I'll give that a shot soon. How much additional coverage beyond what I have in there do you think I should add? Right now it just has a couple basics, plus some of the trickier parts like the sdf.

pca006132:
Sorry I am not able to review this thoroughly as I just caught the flu, but it seems that the bindings are nicely written and will not affect the main code (except the weird macos error), I think this looks good to me.

geoffder:
Started up a test file for the C bindings: https://github.com/elalish/manifold/pull/292/commits/5dd830936867902e6519e5d5b6895554cdf64570

bindings/c/testc.c
elalish: What you have looks good; again since you've signed up to maintain these C bindings, these tests are mostly to make your own life easier. You're welcome to add more as you see fit.

.github/workflows/manifold.yml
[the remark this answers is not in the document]
elalish: Okay, I think this is the only part left to do for this PR, yes? I'm happy to merge once this is done.

bindings/c/testc.c
geoffder: Roger. I'll add some more of what I had in `testc.c` to the new `manifoldc_test.cpp` so I can get rid of that lone executable, then populate further as needed. 👍

.github/workflows/manifold.yml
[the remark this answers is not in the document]
geoffder: I've broken out the building/testing of the C bindings into a separate job so that it can be marked as not required easily in your branch protection rules.

src/utilities/include/public.h
```diff
@@ -255,6 +255,15 @@ struct Curvature {
   std::vector<float> vertMeanCurvature, vertGaussianCurvature;
 };
 
+/**
+ * Connected component indices and number of discrete components. Intermediate
+ * step for Manifold.Decompose(), completed by Manifold.GetComponents()
+ */
+struct Components {
+  std::vector<int> indices;
```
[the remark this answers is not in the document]
geoffder: The discrete (renamed `numComponents`) is just the number of discrete mesh components. So the new name is more clear.

elalish (approved):
Thanks, this looks good to me; I look forward to seeing what you do with it! Progress has been a bit slow on #290, but it's getting there and there will be a couple of follow-ons to finish updating the API. Once all that lands, you'll probably want to look into updating these bindings accordingly.

geoffder:
Ok sounds good. :pray: Thanks for providing the library and being open to this addition!

PR #293, opened 2022-12-14
Bump std to 17 to test for Mac Union regression

geoffder:
As [suggested](https://github.com/elalish/manifold/pull/292#issuecomment-1350679574) by @elalish, this PR contains just the std17 version bump that is part of the C binding PR (to check for regression in `TEST(Boolean, Close)` on the Mac CI.


geoffder:
Checks have passed, so the test failure mentioned seems to be due to something else.

elalish:
Thanks for checking!

PR #295, opened 2022-12-21
Fix incorrect value for components length + test

geoffder:
I didn't realize I missed the correct tests on my side (in the ocaml bindings) when I made some changes to the decomposition components code. `manifold_components_length` has been fixed to return the value of `numComponents` as it should, and a test has been added for compose/decompose.


elalish:
Thanks!

PR #296, opened 2022-12-22
fix css issue

pca006132:
Fixed line number column margin.

Before:
image

After:
image

Notice that the first line in the original html page is offset by 1 line.


elalish:
This was bugging me, but I couldn't find it. Well done!

Issue #297, opened 2022-12-23
Triangulation failure on wasm


pca006132:
It seems that the behavior on wasm is slightly different from native, which causes a weird triangulation bug that only occur on wasm but not on native. I've been testing this to find the cause of non-manifoldness and weird genus for my model when written in typescript.

```c++
TEST(Boolean, UnionWarp) {
  glm::ivec3 keys[] = {
    {1, -1, -1},
    {1, -1, 1},
    {1, 1, -1},
    {1, 1, 1},
    {-1, -1, -1},
    {-1, -1, 1},
    {-1, 1, -1},
    {-1, 1, 1}};
  glm::vec3 val1[] = {
    {-31.15,28.135,1.353},
    {-28.426,27.432,8.842},
    {-36.195,31.394,-0.573},
    {-34.85,30.694,3.128},

    {-37.452,27.928,4.975},
    {-32.32,27.304,11.08},
    {-41.304,31.061,2.323},
    {-38.753,30.44,5.341},
  };
  glm::vec3 val2[] = {
    {-37.452,27.928,4.975},
    {-32.32,27.304,11.08},
    {-41.304,31.061,2.323},
    {-38.753,30.44,5.341},

    {-50.49,27.203,15.861},
    {-45.357,26.579,21.966},
    {-54.385,29.619,13.085},
    {-51.834,28.998,16.102},
  };
  Manifold solid1 = Manifold::Cube(glm::vec3(2.0), true).Warp([&](glm::vec3 &v) {
    glm::ivec3 rounded = glm::ivec3(v);
    for (int i = 0; i < 8; i++)
      if (rounded == keys[i])
        for (int j : {0, 1, 2})
          v[j] = val1[i][j];
  });
  Manifold solid2 = Manifold::Cube(glm::vec3(2.0), true).Warp([&](glm::vec3 &v) {
    glm::ivec3 rounded = glm::ivec3(v);
    for (int i = 0; i < 8; i++)
      if (rounded == keys[i])
        for (int j : {0, 1, 2})
          v[j] = val2[i][j];
  });
  EXPECT_TRUE(solid1.IsManifold());
  EXPECT_TRUE(solid1.MatchesTriNormals());
  EXPECT_TRUE(solid2.IsManifold());
  EXPECT_TRUE(solid2.MatchesTriNormals());

  EXPECT_TRUE((solid1+solid2).IsManifold());
  EXPECT_EQ((solid1+solid2).Genus(), 0);
}
```

[...]

PR #299, opened 2022-12-27
Fix triangulation bugs

elalish:
Fixes #297 as well as the disabled Boolean.Close test that was failing on the Mac CI.

I haven't actually tried to fix this yet, just make small tests to check which compilers are failing, since these work fine for me locally.

PR #300, opened 2022-12-27
Fix triangulation

pca006132:
Fixes triangulation bug (https://github.com/elalish/manifold/issues/297)

Make priority queue respect insertion order.

### DIFF
--- a/src/polygon/src/polygon.cpp
+++ b/src/polygon/src/polygon.cpp
@@ -329,6 +329,7 @@ class Monotones {
     int index;
     VertItr left, right;
     PairItr eastPair, westPair;
+    int insertionID = 0;
 
     bool Processed() const { return index < 0; }
     void SetSkip() { index = -2; }
@@ -344,7 +345,10 @@ class Monotones {
     bool IsPast(const VertItr other, float precision) const {
       return pos.y > other->pos.y + precision;
     }
-    bool operator<(const VertAdj &other) const { return pos.y < other.pos.y; }
+    bool operator<(const VertAdj &other) const {
+      return pos.y < other.pos.y ||
+             (pos.y == other.pos.y && insertionID < other.insertionID);
+    }
   };
 
   /**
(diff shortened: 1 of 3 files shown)


elalish:
Excellent, thank you! I believe this is the more important fix, but #299 will also be useful for general robustness. I'll review this more closely later today.

pca006132:
Ah, priority queue actually accepts `<` rather than `<=`, but somehow it violates FIFO for equal elements... Will try to debug this when I wake up.

pca006132:
OK stl priority queue doesn't care about FIFO, I guess we should add an insertion count to make it stable.

src/polygon/src/polygon.cpp
```diff
@@ -344,7 +345,10 @@ class Monotones {
     bool IsPast(const VertItr other, float precision) const {
       return pos.y > other->pos.y + precision;
     }
-    bool operator<(const VertAdj &other) const { return pos.y < other.pos.y; }
+    bool operator<(const VertAdj &other) const {
+      return pos.y < other.pos.y ||
+             (pos.y == other.pos.y && insertionID < other.insertionID);
```
elalish, line 350: So why is it important to make the priority queue stable? Is this about correctness or just determinism? Why was the new polygon test failing in the first place?

pca006132: It can at least improve determinism. I am not sure if it is more correct as I haven't yet look into the details of the triangulation algorithm. I was comparing the run of wasm and native and found that the wasm one got unstable priority queue, and changing that fixes this particular test.

pca006132: I did some other tests, it seems that the stability of the priority queue is not the fundamental issue, I guess I should try to understand the triangulation algorithm and see how make some minimal example.

elalish: That makes sense. Yeah, my triangulator is a bit convoluted - believe it or not I've refactored it multiple times to try to simplify it. Would be great to have another set of eyes on it. Let me know if you have any questions.

PR #307, opened 2023-01-06
weird bug for python binding

pca006132:
For some reason the ``union_failure.py` will cause collider error. I have no idea why and no idea why the fix works... Unable to reproduce with c++.


elalish (approved):
Weird, thanks for the work-around!

Issue #282, opened 2022-11-14
Incomplete mesh relations API


rafern:
There is no transformation matrix in a MeshRelation object, which is needed for correctly restoring vertex normals. The only alternative at the moment is to separately track each transformation, and apply them to the vertex normals at the end.

There is also no flag for knowing whether a triangle needs to be flipped or not (for example, subtracted triangles are flipped, but not if there is a double-subtraction/quadruple-subtraction/etc...). The only alternatives are:
- compare the face normal of a triangle to the original face normal (with transformations applied), and use a dot product to check whether the triangle was flipped
- manually track which triangles need to be flipped by walking through each subtraction separately

elalish:
That's a great point - I hadn't considered normals requiring a transform matrix. Perhaps we should have a private static vector of transforms for each meshID? Then we could have a function like `static glm::mat4x3 Transform(int meshID)`? What do you think, @pca006132? I recall you fixed some of my earlier static usage for thread safety. 

I was thinking you'd tell if it was flipped by the order of the verts, but I see now that'll only work for retained triangles, but not for intersected triangles. The face normal dot product method should actually be safe because I retain the original face normals rather than recalculating them (once you have the transforms from above). Does that seem reasonable, or is additional data necessary?

pca006132:
Do we need a vector of transforms or is 1 transformation matrix for each mesh id sufficient? Not sure if storing them globally would be a good idea, as it would cause "leak" when users keep creating new meshes. I think a better way would be to store them in impl.

One problem is that this will not handle warp appropriately, users using this should be aware of this limitation.

rafern:
I was under the assumption that the library already stored which faces were flipped and that it would be better to reuse those flags, but I can't find any flags like these in the code, so it's definitely more reasonable to use the face normal method.

For the transform matrices, I agree with @pca006132 if the only use is to correct vertex normals/other attributes, but if you want to let users inverse/decompose the matrix for some reason, then it's probably better to store all transformation matrices in a vector because of non-uniform scaling.

pca006132:
I'm thinking if it would make sense to store the transformations in https://github.com/elalish/manifold/blob/master/src/manifold/src/csg_tree.h#L88

Because we are already tracking them in the csg tree.

elalish:
[...]

PR #309, opened 2023-01-25
Multi-material example

### LINKED ISSUE #196: Build multi-material three.js example

### LINKED ISSUE #286: Imprecise barycentric coordinates

elalish:
Fixes #196 
Fixes #286 

This is a follow-on to #290 to demonstrate that the new property APIs work end-to-end by connecting them to glTF output. I've expanded `MeshGL` with a different style of `MeshRelation` data. I'm thinking I'll remove the old `MeshRelation` and `triProperties` input entirely now, basically saying `Mesh` is only for pure, simple geometry, while `MeshGL` is for anything requiring mesh properties and relationships.


elalish:
Oops, I jumped the gun a bit; I still need to add the triangle reference into `MeshGL` so the tests can pass. Shouldn't be hard, hopefully I'll get to that tomorrow.

The next PR will be addressing #282 by adding a transforms vector that corresponds to the `meshID` vector. I think I'll even add some optional parameters to `GetMeshGL()` to specify which properties are normals so I can update them automatically.

elalish:
@geoffder FYI, I removed the `MeshRelation` API and removed them from your cbindings too, but you'll probably want to add in the other API changes.

geoffder:
Is `Mesh` on it's way to be a purely internal class? I see that `LevelSet` still returns one, and there remains a `Mesh` constructor for `Manifold`, yet `Smooth` has been switched to take `MeshGL`.

elalish:
I'm considering it. Definitely I'll update things like `LevelSet` to return one. The other constructor was superfluous since `MeshGL` already has a constructor that takes a `Mesh`, so really both are supported anyway.

geoffder:
Huh, giving a `Mesh` to `Smooth` still compiles? Pardon my ignorance, but does C++ implicitly find the single argument constructor of `MeshGL` and just use that? Never would have expected that.

elalish:
Yup, indeed it does. I think it's basically the same system that let's you pass `2` (an int) when the function wants a float.

geoffder:
Interesting that it does the same on it's own for user defined types with no "class/inheritance relationship". Maybe just surprising to me due to my minimal C++ experience.

geoffder:
@elalish Opened a PR with the new meshgl accessors: https://github.com/elalish/manifold/pull/311. Let me know if I've missed something!

pca006132:
I won't be able to review this soon, I am currently on vacation for 7 days in Japan without my computer.

elalish:
@pca006132 No problem, enjoy your vacation!

elalish:
Good news: `meshID` is no longer needed as part of the public API - it is now internal-only. Likewise, no need for `triProperties` and such in the public API as we can use `MeshGL` for this input instead. I'm keeping the explicit `Mesh` constructors for now simply because they avoid some computation. I'd recommend that `Mesh` remain a C++-only construct though. For bindings to other libraries, better to directly create a `Manifold` and use `MeshGL` as the only I/O struct.

PR #310, opened 2023-01-25
Remove manifold_meshgl_vert_normal from header (impl gone)

geoffder:
Caught a straggling ghost function in the header when updating my bindings.


elalish (approved):
Thanks!

PR #311, opened 2023-01-26
Add new MeshGL accessors (replacing MeshRelation)

geoffder:
Is there anything I'm new that should be exposed that I'm missing here?


elalish (approved):
I'm still making some API changes; I'll let you know on the next PR.

PR #312, opened 2023-01-28
Deduplicate vector content memcpy-ing

geoffder:
With the latest changes, I realized that there were now enough repeats (> 10) of the same pattern of copying a vectors `.data()`, so I replaced them with a template.


elalish:
oh, sorry, I missed this; you can just update it to merge into master instead.

PR #313, opened 2023-01-28
Remove Mesh from interface + memcpy depuplication

geoffder:
In addition to the cleanup in the closed PR, I've removed Mesh from the interface in the spirit of your intention to keep it as a C++ only implementation detail.

### DIFF
--- a/bindings/c/manifoldc.cpp
+++ b/bindings/c/manifoldc.cpp
@@ -14,9 +14,9 @@
 using namespace manifold;
 
 namespace {
-ManifoldMesh *level_set(void *mem, float (*sdf)(float, float, float),
-                        ManifoldBox *bounds, float edge_length, float level,
-                        bool seq) {
+ManifoldMeshGL *level_set(void *mem, float (*sdf)(float, float, float),
+                          ManifoldBox *bounds, float edge_length, float level,
+                          bool seq) {
   // typing with std::function rather than auto compiles when CUDA is on,
   // passing it into GPU (and crashing) is avoided dynamically in `sdf.h`
   std::function<float(glm::vec3)> fun = [sdf](glm::vec3 v) {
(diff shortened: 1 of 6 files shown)


bindings/c/manifoldc.cpp
```diff
@@ -132,16 +132,16 @@ ManifoldManifold *manifold_warp(void *mem, ManifoldManifold *m,
   return to_c(new (mem) Manifold(warped));
 }
 
-ManifoldMesh *manifold_level_set(void *mem, float (*sdf)(float, float, float),
-                                 ManifoldBox *bounds, float edge_length,
-                                 float level) {
+ManifoldMeshGL *manifold_level_set(void *mem, float (*sdf)(float, float, float),
```
elalish, line 135: You might consider returning a `Manifold` instead. I return a `Mesh` so there isn't a dependency, but here I'd guess the whole point would be to do further operations. It'll be fastest to push the `Mesh` directly into the `Manifold` constructor without the intermediate `MeshGL` step in this case.

elalish (approved):
Thanks, less code is always nice!

bindings/c/manifoldc.cpp
geoffder: In the higher level bindings I do provide a direct to manifold function (by including the manifold construction), but I suppose I could also expose one directly in the C. I think it still makes sense to have a mesh returning version in the case the user plans to go straight to pulling out / using properties.

geoffder: Thinking about it a bit more, providing a straight to manifold version would mean two more, rather than one more function since there already has to be a forced sequential version. Since it is so simple to provide the higher level direct manifold function in the FFI bindings, I think the MeshGL version should be sufficient for the C side.

Issue #302, opened 2022-12-29
Mesh to manifold may run indefinitely


chpark1111:
Hi! It' me again with an another issue..

When I convert the Mesh to Manifold, it takes too much time. I found that this is coming from the stable_sort_by_key function in  CreateHalfedges function.

It handles meshes with number of vertices lower than 1000 well, but if it gets about 1500, it takes too long time (it does not end).

Is there any way to reduce the time of Mesh to Manifold conversion without lowering the number of vertices? Also, in my experience this time complexity is definitely not O(N). Because vertices of 1500 mesh didn't end until a whole day. Can you explain about this too?

Or where do you think the time bottleneck is happening?

And I tried to enlarge the precision_, but it didn't work..

pca006132:
Are you using this on wasm or native? And if you are using it on native, did you enable multithreading?

I tried generating a sphere with 4096 circular segments, resulting in a mesh with 4194306 vertices, and the conversion of that mesh into manifold object takes about 8945962us (~9s) for native without parallelization and 4978475us (~5s) with TBB backend, so I can't reproduce your issue. Can you provide your mesh?

chpark1111:
Oh my, I'm sorry. I think I used a degenerated mesh.. (Maybe not watertight mesh) I'll close the issue!

hrgdavor:
@chpark1111 silly question, maybe it is mentioned elsewhere.
 
I am wondering if JS side can give manifold mesh to WASM with proper info or it has to go through mesh->manifold converter ?

I am looking to play a bit with manifold in near future, so I am curious about such details.

elalish:
@chpark1111 I'd still like to see you mesh! It sounds like you hit an infinite loop, probably in the edge collapse step. This would be a great test case to look at.

@hrgdavor I'm not quite sure what your question is. What is "proper info"? I would consider `Manifold(mesh)` to be both giving a manifold mesh and a mesh->manifold converter - not sure what distinction you're trying to draw.

chpark1111:
@elalish I think that's the exact answer I wanted to get!! Thanks! :)
I'll provide you one here. If this contains too much vertices to look at, I can give you small number of meshes that gets an infinite loop.

And just for a suggestion, can't we detect does infinite loop cases and make the program output an error?
[mesh.txt](https://github.com/elalish/manifold/files/10320445/mesh.txt)

pca006132:
Yeah we should detect that.

hrgdavor:
[...]

PR #315, opened 2023-02-01
Transform relations

### LINKED ISSUE #282: Incomplete mesh relations API

elalish:
Related #282 

This adds transformation feedback to to the `MeshGL` output, as well as taking these transforms and original IDs as input. The `ReserveIDs` function allows you to make your own IDs to pass in for when your input mesh has multiple materials, for instance. 

I think this is moving the API in the right direction, because the tests are getting cleaner while checking more things. I still need to add feedback about if the reference is inverted to complete the normal transforms for #282.


elalish:
@geoffder I updated cbind enough it would build, but there are some API updates here, which I presume is causing the CI failure.

elalish:
Possibly related to #302: I noticed the ubuntu OMP builds have started to flake out sometimes, appearing to infinite-loop on `Samples.Bracelet`, but only about 50% of the time. I've marked them not required for now, but I need to come back and investigate.

geoffder:
@elalish The [CI error](https://github.com/elalish/manifold/actions/runs/4116754328/jobs/7107206929#step:7:14) 
```
 [ RUN      ] CBIND.level_set
unknown file: Failure
C++ exception with description "Error in file: /__w/manifold/manifold/test/meshIO/src/meshIO.cpp (207): 'validChannels' is false: Invalid colorChannels." thrown in the test body.
[  FAILED  ] CBIND.level_set (1849 ms)
```
doesn't reproduce on my machine compiled with the current master. My first naive inclination was to try using the `MeshGL` extracted from a `Manifold` constructed from the level set mesh, rather than directly exporting it, thinking that the process may fill in whatever was missing (I have not looked into it) but it doesn't appear to hit the same error for me. Not sure what would be different for me vs the CI given that the exception is coming from within Manifold code (not something external like assimp which may have version mismatch).

On the topic of parallel backends, I found that testing hung out on `Samples.Knot42` when built with `OMP`, and `Samples.Bracelet` when built with `TBB`.

elalish:
@geoffder I ran into this in #317 and solved it there. Turns out it was because I changed the API to return a `uint_32t` instead of an `int` - apparently C++ prefers to convert `int` to `uint` during a compare rather than the other way around. :/

geoffder:
@elalish Ahh interesting, good catch! I wonder if that behaviour changed between compiler versions then since I didn't hit the bug.

PR #317, opened 2023-02-08
added normals update

### LINKED ISSUE #282: Incomplete mesh relations API

elalish:
Fixes #282 

Instead of exposing front/back side as yet another output of `MeshGL`, I decided it would be more ergonomic to just add a parameter to `GetMeshGL` so that it can apply the normal transforms automatically.

### DIFF
--- a/src/manifold/include/manifold.h
+++ b/src/manifold/include/manifold.h
@@ -100,7 +100,7 @@ class Manifold {
    */
   ///@{
   Mesh GetMesh() const;
-  MeshGL GetMeshGL() const;
+  MeshGL GetMeshGL(glm::ivec3 normalIdx = glm::ivec3(0)) const;
   bool IsEmpty() const;
   enum class Error {
     NO_ERROR,
--- a/src/manifold/src/boolean_result.cpp
+++ b/src/manifold/src/boolean_result.cpp
@@ -412,7 +412,7 @@ struct MapTriRef {
 };
 
 VecDH<TriRef> UpdateReference(Manifold::Impl &outR, const Manifold::Impl &inP,
-                              const Manifold::Impl &inQ,
+                              const Manifold::Impl &inQ, bool invertQ,
                               ExecutionPolicy policy) {
   VecDH<TriRef> refPQ = outR.meshRelation_.triRef;
   const int offsetQ = Manifold::Impl::meshIDCounter_;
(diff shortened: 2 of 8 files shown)


rafern:
The API looks good. As far as I know normals are the only vertex attribute that needs to be flipped, so it should be fine. Tangents can still be transformed manually by the user without flipping (btw, `MeshGL.transform` isn't exposed on the JS API yet).

However, I tried to use it on my PC but it fails to build with the latest emsdk (3.1.31) due to a deprecated API: [build-log.txt](https://github.com/elalish/manifold/files/10684174/build-log.txt)

elalish:
Thanks for the reminder, I'll get the js API updated in this PR, but I just got sick, so it will have to wait. 

Thanks for debugging the emscripten issue, you're not the first to report that. If you figure out how to fix it, a PR would be most welcome. Otherwise I'll take a look at it once I get this one finished.

rafern:
Hope you get better soon. I've added some bindings if you want to use them when you return, but haven't had the time to test them yet, since a lot of my library would have to be rewritten to support the new API. Patch (I had to zip it because Github isn't letting me upload patch files for some reason):

[js-bindings.patch.zip](https://github.com/elalish/manifold/files/10696522/js-bindings.patch.zip)

Typed arrays can't have a size specified as part of their type, so I added new SealedFloat32Array/SealedUint32Array types. I also added a new matrix method to the Mesh class, which is a helper similar to verts, but for the transform field. It's a bad name, but transform is already taken.

Also, it turns out the build failures were because of a stupid mistake I did. I forgot to apply the patches because I skipped straight to the WASM build instructions, instead of also reading the regular build instructions.

elalish:
@rafern How about I just merge this and you open a PR with your updates? I have some other things I need to fix in parallel anyway.

PR #320, opened 2023-02-13
Added missing JS bindings (from discussion in #317)

rafern:
- Add missing type definitions
- Exposes MeshJS.transform to JS bindings
- Added getter for transforms, called `matrix`. I feel like this is a bad name, any suggestions?

Marking this as a draft since I haven't had the time to test this yet. Will update this PR once I port gypsum and test it with the new API.


bindings/wasm/bindings.js
```diff
@@ -142,6 +143,7 @@ Module.setup = function () {
       this.originalID = originalID;
       this.faceID = faceID;
       this.halfedgeTangent = halfedgeTangent;
+      this.transform = transform;
```
elalish, line 146: It occurs to me I haven't really followed my own naming convention here - I should switch `originalID` to `runOriginalID` and `transform` to `runTransform`, since I generally name with what I'm indexing from and then to.

bindings/wasm/bindings.js
```diff
@@ -167,6 +173,10 @@ Module.setup = function () {
     tangent(halfedge) {
       return this.halfedgeTangent.subarray(4 * halfedge, 4 * (halfedge + 1));
     }
+
+    matrix(index) {
```
elalish, line 177: This can become `transform` due to the above renaming.

bindings/wasm/manifold.d.ts
```diff
@@ -49,12 +69,15 @@ declare class Mesh {
...
   faceID?: Uint32Array;
   halfedgeTangent?: Float32Array;
+  transform?: Float32Array;
   get numTri(): number;
   get numVert(): number;
-  verts(tri: number): Uint32Array<3>;
-  position(vert: number): Float32Array<3>;
+  get numMatrix(): number;
+  verts(tri: number): SealedUint32Array<3>;
+  position(vert: number): SealedFloat32Array<3>;
```
elalish, line 77: Remind me of the advantage of this over what we had previously?

bindings/wasm/manifold.d.ts
```diff
@@ -40,6 +48,18 @@ type Curvature = {
 };
 
 declare class Mesh {
+  constructor(options: {
+    numProp?: number,
+    triVerts?: Uint32Array,
+    vertProperties?: Float32Array,
```
elalish, line 54: I believe these first three are required.

elalish (changes requested):
Thanks! Let's fix the naming here at least. If you want to fix it back in the C++ too, you're welcome to, otherwise I'll do that as a follow-up.

bindings/wasm/manifold.d.ts
rafern: Float32Array doesn't have any type arguments, so you can't specify the size of the array this way; this creates a compile error. If it was a regular array you could have used, for example, a tuple by doing `[x: number, y: number, z: number]` instead. This is a typed array though, so you can't use the tuple syntax. Instead, I extended the typed array to have a fixed length by doing something similar to what was done here: https://stackoverflow.com/a/66628633

rafern:
All done, let me know if you have any more changes. The C++ is already changed and it doesn't throw any compile errors, but I haven't done any testing with real code. I'll probably port gypsum this weekend, but if you want to merge it now that's OK too, and I can test later

elalish:
@rafern So, our CI only runs on non-draft PRs, so I marked yours as ready for review, but Github has a bug where it caches the stale value of "draft" and so still won't run the CI. Can you push some trivial commit to prod it into action?

rafern:
@elalish Should be good now

bindings/wasm/manifold.d.ts
elalish: Very nice, thanks for the explanation.

bindings/wasm/manifold.d.ts
```diff
@@ -40,21 +48,36 @@ type Curvature = {
...
   get numVert(): number;
-  verts(tri: number): Uint32Array<3>;
-  position(vert: number): Float32Array<3>;
+  get numMatrix(): number;
+  verts(tri: number): SealedUint32Array<3>;
+  position(vert: number): SealedFloat32Array<3>;
   extras(vert: number): Float32Array;
-  tangent(halfedge: number): Float32Array<4>;
+  tangent(halfedge: number): SealedFloat32Array<4>;
+  matrix(index: number): SealedFloat32Array<12>;
```
elalish, line 80: Looks like this and `numMatrix` above need their names updated. Nit: I'd call the input `run` instead of `index`.

elalish (changes requested):
Excellent, pretty much there. If you look at the one failing test, it looks like something in cbind hasn't been updated with the rename. Should be an easy fix, but if not, we can take care of it after.

rafern:
Yeah, working on it. I forgot that there were other bindings that needed to be updated. I'll ping you once it's done

rafern:
@elalish Done. I haven't updated the Python bindings though, since I'm not familiar with how that works, but it should be fine for now since there aren't any tests

bindings/wasm/manifold.d.ts
elalish: Did you miss this comment by chance?

rafern: I did, my bad. It should be fixed now

elalish (approved):
Excellent, thank you!

rafern:
@elalish I was supposed to do this 3 weekends ago (sorry), but I just wanted to let you know that I finally ported gypsum to the newer version of Manifold, and got to test the WASM bindings. Everything seems to work fine, and normals get transformed as expected:
image

Tangents need to be transformed manually, but that's OK because there's the `runTangent` field:
image

Thanks for your work on the new API!

elalish:
@rafern Excellent, great to know it's working! Would you be interested in adding gypsum to our list of [users](https://github.com/elalish/manifold/discussions/340)? Also, considering the recent addition of `CrossSection` for 2D, would you be interested in adding the JS bindings? I really appreciate your help and would be happy to add you as a collaborator if you'd like to stay involved.

rafern:
@elalish Yeah, sure, I'll add to the list of users. I can work on the bindings, but I won't make any promises on a timeline as those haven't gone very well so far 😅

elalish:
@rafern No problem! Just let us know when you get started. I just filed an issue for it: https://github.com/elalish/manifold/issues/360

PR #322, opened 2023-02-15
convert to ES6 module

elalish:
Emscripten fixed ES6 modules a few months ago so they can now work in both Node.js and browser contexts. This switches us over and updates our examples in preparation for our npm v2.0 release. FYI @prideout, since I know you've been wanting this. 

This also switches manifoldcad.org over from using three.js and GLTFExporter to using gltf-transform for creating our GLBs. I was planning to anyway, but my hand was forced because using a three.js exporter requires an import map, which doesn't seem to be defined for a worker yet.


elalish:
@pca006132 I have manifoldcad almost working again, but I get an error: `exportGLB` not defined. It clearly is, and weirder yet I really haven't changed anything about where it's defined or how it's called. I'm guessing I just made some kind of silly syntax error - let me know if you notice anything.

pca006132:
You need to add it to `exposedFunctions` or it won't be exposed.

elalish:
> You need to add it to `exposedFunctions` or it won't be exposed.

Okay, but then why did it work before? The rest of those functions are mapped to WASM calls - does this one need to be mapped somehow too?

pca006132:
My guess is that the symbol visibility is different when using ES6 module, although I am not really sure about this. Yes we will need to map it, I think the following code should work although I haven't yet tried:
```javascript
    const f = new Function('exportGLB', ...exposedFunctions, content);
    f(exportGLB, ...exposedFunctions.map(name => wasm[name]));
```

elalish:
Okay, this all appears to be working now. The diff isn't really that big (try hiding whitespace changes) - I'm having some trouble with my JS formatter picking different styles on different computers with VScode, which I haven't tracked down yet. We should really get our code format CI check working on the other languages...

I've now removed all dependence on three.js (except for the actual three.js example). I'm also using top-level await for loading the WASM module, which is much cleaner.

PR #323, opened 2023-02-16
Fix deploy

elalish:
I think because the OMP runs are sometimes timing out, it makes the workflow as failed (the red X) even though they aren't required. I believe this was causing our deploy the fail because it wouldn't grab the artifacts from a "failed" run.

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -26,6 +26,7 @@ jobs:
       uses: dawidd6/action-download-artifact@v2
       with:
         workflow: manifold.yml
+        workflow_conclusion: completed
         pr: ${{github.event.pull_request.number}}
         name: wasm
         path: ./public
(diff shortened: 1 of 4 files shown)


.github/workflows/deploy.yml
```diff
@@ -26,6 +26,7 @@ jobs:
       uses: dawidd6/action-download-artifact@v2
       with:
         workflow: manifold.yml
+        workflow_conclusion: completed
         pr: ${{github.event.pull_request.number}}
```
lukasz-mitka, line 30: @elalish how does this work? you trigger on push, but here look for `github.event.pull_request`

Came here from https://github.com/dawidd6/action-download-artifact/issues/224, maybe it's related.

elalish: Huh, good point. I think I had figured every pull request has an associated push, but maybe I'm wrong?

PR #331, opened 2023-02-21
hashtable atomic load store fix

pca006132:
Note that I intentionally leave this behind master for GitHub action to check.

### DIFF
--- a/src/utilities/include/hashtable.h
+++ b/src/utilities/include/hashtable.h
@@ -31,8 +31,30 @@ __host__ __device__ T AtomicCAS(T& target, T compare, T val) {
   return atomicCAS(&target, compare, val);
 #else
   std::atomic<T>& tar = reinterpret_cast<std::atomic<T>&>(target);
-  tar.compare_exchange_strong(compare, val);
+  tar.compare_exchange_strong(compare, val, std::memory_order_acq_rel);
   return compare;
+#endif
+}
+
+template <typename T>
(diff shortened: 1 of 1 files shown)


pca006132:
It seems that there is also another problem: because the use count is incremented after setting the keys, the hashtable can be already full before the use count indicates it is full, causing infinite loop. A simple fix is to check for full for every iteration, though it may be better to just increment the use count before entering the loop and check for full.

pca006132:
No idea why is there undefined reference to those atomic functions, I have already added `-latomic` to the linker flags...

elalish:
CMake says it found `atomic_load_8`, but then the build fails looking for atomic store. Maybe they're in different places?

pca006132:
Unlikely, I will get a Ubuntu container to debug later. Or you can try to debug this if you have time, just look at the symbols libatomic provides on Ubuntu and perhaps try to google it and see if others encountered similar issues...

pca006132:
Oh and I actually thought that GCC can compile these into instructions instead of function calls. Maybe I should make them acquire release for better performance and avoid such a pain in compiling them.

pca006132:
hmmm, we don't need atomic load/store for values, just keys is fine.

pca006132:
Seems working, will clean this up tmr and should be ready to merge at that time.

elalish:
Perfect, thanks! And remember to pull master.

elalish (approved):
Thanks!

PR #332, opened 2023-02-21
Add gl-matrix to editor

### LINKED ISSUE #238: Vector library for JS
I think we have a pretty good generic JS binding API now, but for ManifoldCAD, which is three.js-based, I think it would be quite a bit more ergonomic to have basic vector 2-4 support like GLM gives us in C++. Three.js has this built in (`Vector2`, `Matrix4`, etc), so it would be nice to bridge across to that so they can be used in our editor rather than bare arrays.

### LINKED ISSUE #324: Integrate glMatrix into manifold.js
It's pretty clear if you're using `Manifold`, you're also doing matrix operations on 2, 3, and 4 -element vectors. Currently we have our own types for these without any useful functions, just fixed-size arrays. I'm thinking it would be better to integrate a mature package like [`glMatrix`](https://www.npmjs.com/package/gl-matrix) and use their types in our API instead, to avoid a bunch of manual conversion. It's an extra 15kb (minified and gzipped) compared to >200kb for our library currently, so the size difference seems a worthwhile trade.

What does everyone think of this? (particularly @pca006132).

elalish:
Fixes #238 
Fixes #324 

This does simplify some of the examples nicely, and it feels pretty easy to use.

Also, I switched manifold.js from using 3x4 to 4x4 matrices, since that seems to be what all the other libraries use (we're still only using the affine portion).

### DIFF
--- a/bindings/wasm/manifold.d.ts
+++ b/bindings/wasm/manifold.d.ts
@@ -20,10 +20,9 @@ interface SealedFloat32Array<N extends number> extends Float32Array {
   length: N;
 }
 
-type Vec2 = [number, number];
-type Vec3 = [number, number, number];
-type Vec4 = [number, number, number, number];
-type Matrix3x4 = [Vec3, Vec3, Vec3, Vec3];
+type Vec2 = number[2];
+type Vec3 = number[3];
+type Mat4 = number[16];
 type SimplePolygon = Vec2[];
 type Polygons = SimplePolygon|SimplePolygon[];
 type Box = {
@@ -56,9 +55,9 @@ declare class Mesh {
     mergeToVert?: Uint32Array,
     runIndex?: Uint32Array,
     runOriginalID?: Uint32Array,
+    runTransform?: Float32Array,
     faceID?: Uint32Array,
-    halfedgeTangent?: Float32Array,
-    runTransform?: Float32Array
+    halfedgeTangent?: Float32Array
   });
   numProp: number;
   vertProperties: Float32Array;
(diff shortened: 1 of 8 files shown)


bindings/wasm/manifold.d.ts
```diff
@@ -20,10 +20,9 @@ interface SealedFloat32Array<N extends number> extends Float32Array {
   length: N;
 }
 
-type Vec2 = [number, number];
-type Vec3 = [number, number, number];
-type Vec4 = [number, number, number, number];
-type Matrix3x4 = [Vec3, Vec3, Vec3, Vec3];
+type Vec2 = number[2];
+type Vec3 = number[3];
+type Mat4 = number[16];
```
elalish, line 25: Oops, pretty sure this is wrong.

pca006132: Maybe we should do this: https://stackoverflow.com/a/59906630

It is interesting that typescript did not put this in their type definitions even though this should be a rather common usecase...

pca006132: And maybe we should explicitly say that the matrix is column-major? Otherwise users may be confused about this.

pca006132 (commented):
LGTM apart from the array issue.

bindings/wasm/manifold.d.ts
elalish: I went back mostly to what we had before with tuple types. They seem to work the best and give pretty good type safety. The only drawback is I had to use `vec2.fromValues(1, 1)` instead of `[1, 2]` whenever I wanted to save the variable, but that doesn't seem too bad - and that's just because I can't cast it in our JS tests like users can in the editor.

Issue #216, opened 2022-09-24
Comparison against other libraries


pca006132:
We can add some charts in our README to compare the performance of manifold to other libraries. We have some data comparing against OpenSCAD but it was a bit outdated (we got some performance improvement after that and did not care about tracking the data).

Some libraries to compare against:
- OpenSCAD
- CGAL
- JSCAD

I can write some code to generate the OpenSCAD and JSCAD version of the same js/python script. However, I have no idea about CGAL.

elalish:
Sounds good. OpenSCAD uses CGAL, so they aren't entirely independent. I assume [this](https://elalish.blogspot.com/2022/03/manifold-performance.html) is what you're referring to as outdated (also not terribly precise).  In addition we have a CGAL comparison [contributed](https://github.com/elalish/manifold/blob/master/extras/perf_test_cgal.cpp) to our library.

pca006132:
OK so is this the only blocker for release 2.0.0?

elalish:
This feels more nice-to-have than release-blocker to me. For the release, I'd really like to publish npm and pypi packages, which means I'd like to translate and test our examples so we know those bindings are working. Would you be interested in making a modified colab based on [this](https://colab.research.google.com/drive/1pcndwqeY8vker3bLKQNJKr3B-7-SYenE?usp=sharing) to test our python examples?

pca006132:
Yes I can have a look at this.

elalish:
Also, your [comment](https://github.com/elalish/manifold/issues/197#issuecomment-1256891587) on missing APIs in our bindings is a pretty high priority too.

pca006132:
> This feels more nice-to-have than release-blocker to me. For the release, I'd really like to publish npm and pypi packages, which means I'd like to translate and test our examples so we know those bindings are working. Would you be interested in making a modified colab based on [this](https://colab.research.google.com/drive/1pcndwqeY8vker3bLKQNJKr3B-7-SYenE?usp=sharing) to test our python examples?

Do we need pyrender for preview? It seems that trimesh already provides this nicely. Here is an [example](https://colab.research.google.com/drive/1J8_pWrIOvUx5wAUn8H-4GN35lSHm3LAl?usp=sharing).

elalish:
Oh, nice; I didn't realize that. Sure, that sounds fine for now.

pca006132:
probably not needed for now

PR #333, opened 2023-02-23
added maze benchmark

pca006132:
I was talking to my friend and realized we need some updated benchmarks to tell others how fast we are :P

I ported the maze benchmark from https://gist.github.com/ochafik/2db96400e3c1f73558fcede990b8a355
I didn't bother to run the benchmark with OpenSCAD on my computer, it took 1167ms on my computer to run it with manifold, and 15s when I increase `fn` to 128, which is a lot faster than the 5 minutes with fast-csg enabled.

We should probably deal with #216 as well, put some nice graphs or something.


elalish:
Good idea! The best pretty graphs I have are [here](https://elalish.blogspot.com/2022/03/manifold-performance.html), but they're pretty out of date now. We should probably put some up on our README. 

By the way, how's our Python API these days? I haven't touched it lately and our API has changed a fair bit. This benchmark looks like it could be the start of some automated testing for it. It would be nice to finally publish it to pypi.

elalish (approved):

pca006132:
The Python API is actually very stable, but lacks some advanced functionalities we recently added, we can add them back before publishing to pypi. There is always automated testing for it: https://github.com/elalish/manifold/blob/master/.github/workflows/manifold.yml#L48, and I'm still using the Python API as I like operator overloading :)

elalish:
Excellent, I forgot you already added that testing. Might be nice to convert some of our JS examples to python to test our more advanced APIs once they're added. Do you want to try updating it for pypi release along with npm v2, or will that take awhile still?

pca006132:
No, I am a bit busy now and can only squeeze a bit of time to do side projects, so that will take me a while.

elalish:
Okay, no prob - thanks for letting me know!

PR #334, opened 2023-02-23
add only() to editor

elalish:
Follow-on to #330 - now we have OpenSCAD `#` functionality with `show()` and `*` functionality with `only()`. Ghosting out the final output gives nice context to the object shown with `only()`.


bindings/wasm/examples/editor.d.ts
```diff
@@ -13,6 +13,7 @@
 // limitations under the License.
 
 declare function show(manifold: Manifold): Manifold;
+declare function only(manifold: Manifold): Manifold;
```
pca006132, line 16: Should we put the documentation for `show` and `only` here as well?

pca006132:
Btw, wondering what would happen if there are multiple `only` within 1 model. The last one wins or it will be `only(union(...))`?

elalish:
It'll include each `only` and `show` mesh separately. I think that's pretty close to OpenSCAD's behavior too.

bindings/wasm/examples/editor.d.ts
elalish: Good call, done.

PR #336, opened 2023-02-23
Convert manifoldCAD.org to a PWA

elalish:
This should allow us to be installable now.


elalish:
Oh, actually it needs a service worker too, stay tuned.

PR #341, opened 2023-02-27
2d-operations via Clipper2

### LINKED ISSUE #127: Flesh out 2D subsystem
We currently use `Polygons` as input to several `Manifold` constructors, and they could easily be output as well (think `Slice`), but we have no way to operate on them in 2D. I would like to think of `Polygon` as the 2D equivalent to `Mesh`, and I'd like to create a `CrossSection` class that is the 2D equivalent to `Manifold`. 

A `CrossSection` will always be geometrically-valid (non-overlapping), as any input `Polygons` will be automatically fixed up according to winding number. Clipper lib seems like a good candidate to support useful 2D operations. I propose we adopt the "strictly positive" fill rule; this will allow Boolean-style operations simply by adding Polygons with the proper winding direction. This class will be totally independent of the `Manifold` class; they can simply interact through exported `Polygons`.

geoffder:
Fixes #127

Minimal proof of concept of integrating Clipper2 in order to flesh out Manifold's 2d capabilities.


src/clippoly/src/clippoly.cpp
```diff
@@ -0,0 +1,271 @@
...
+  for (int i = 0; i < n; ++i) {
+    circle[0].push_back(
+        C2::PointD(radius * cosd(dPhi * i), radius * sind(dPhi * i)));
+  }
+  return Clippoly(circle, true);
+}
+
+Clippoly Clippoly::Boolean(const Clippoly& second, OpType op) const {
+  auto ct = cliptype_of_op(op);
+  auto res = C2::BooleanOp(ct, C2::FillRule::NonZero, paths_, second.paths_,
```
elalish, line 138: Does clipper have a strictly-positive fill rule? I generally prefer that to non-zero.

src/clippoly/src/clippoly.cpp
```diff
@@ -0,0 +1,271 @@
...
+  auto ps = C2::TranslatePaths(paths_, v.x, v.y);
+  return Clippoly(ps, true);
+}
+
+Clippoly Clippoly::Scale(glm::vec2 scale) {
+  auto scaled = C2::PathsD();
+  for (auto path : paths_) {
+    auto s = C2::PathD();
+    for (auto p : path) {
+      s.push_back(C2::PointD(p.x * scale.x, p.y * scale.y));
```
elalish, line 217: Huh, they have a translate but no scale? What about rotate?

src/clippoly/src/clippoly.cpp
```diff
@@ -0,0 +1,271 @@
...
+  return Clippoly(scaled, false);
+}
+
+Clippoly Clippoly::Simplify(double epsilon) {
+  auto ps = SimplifyPaths(paths_, epsilon, false);
+  return Clippoly(ps, false);
+}
+
+Clippoly Clippoly::RamerDouglasPeucker(double epsilon) {
+  auto ps = C2::RamerDouglasPeucker(paths_, epsilon);
```
elalish, line 230: What is this?

src/clippoly/src/clippoly.cpp
```diff
@@ -0,0 +1,271 @@
...
+
+Clippoly Clippoly::StripDuplicates() {
+  auto ps = C2::StripDuplicates(paths_, true);
+  return Clippoly(ps, false);
+}
+
+Clippoly Clippoly::TrimCollinear() {
+  auto trimmed = C2::PathsD();
+  for (auto p : paths_) {
+    trimmed.push_back(C2::TrimCollinear(p, false));
```
elalish, line 247: A bunch of these simplification methods sound very similar. Mind adding some comments describing the differences?

src/clippoly/src/clippoly.cpp
```diff
@@ -0,0 +1,271 @@
...
+    diffed.insert(diffed.end(), ss.begin(), ss.end());
+  }
+  auto u = Union(diffed, C2::FillRule::NonZero);
+  return Clippoly(u, true);
+}
+
+Polygons Clippoly::ToPolygons() const {
+  auto polys = Polygons();
+  auto paths =
+      clean_ ? paths_ : C2::Union(paths_, C2::FillRule::NonZero, precision_);
```
elalish, line 284: clean_ appears to always be false? I suppose this would be for future optimization? Alternatively, we could just clean them all on construction.

src/clippoly/include/clippoly.h
```diff
@@ -0,0 +1,82 @@
...
+class Clippoly {
+ public:
+  Clippoly();
+  ~Clippoly();
+  Clippoly(const Clippoly& other);
+  Clippoly& operator=(const Clippoly& other);
+  Clippoly(Clippoly&&) noexcept;
+  Clippoly& operator=(Clippoly&&) noexcept;
+  Clippoly(std::vector<glm::vec2> contour);
+  Clippoly(std::vector<std::vector<glm::vec2>> contours);
```
elalish, line 35: Might as well shorthand this to `Clippoly(const Polygons& polygons);`

What do you think of renaming `Clippoly` to `CrossSection`? Then `Polygons` is akin to `Mesh` while `CrossSection` is akin to `Manifold`.

src/manifold/include/manifold.h
```diff
@@ -65,8 +66,13 @@ class Manifold {
   static Manifold Extrude(Polygons crossSection, float height,
                           int nDivisions = 0, float twistDegrees = 0.0f,
                           glm::vec2 scaleTop = glm::vec2(1.0f));
+  static Manifold Extrude(Clippoly crossSection, float height,
```
elalish, line 69: I think we can remove the `Polygon` versions - better to force the input to not have overlaps.

src/clippoly/include/clippoly.h
elalish: Oh right, I forgot about the `idx` that's part of `Polygon`. Maybe that should be removed - it's only really used internally anyway. Does Clipper provide any way to append data to carry along with the edges?

src/clippoly/include/clippoly.h
```diff
@@ -0,0 +1,87 @@
...
+
+  enum class JoinType { Square, Round, Miter };
+  Clippoly Offset(double delta, JoinType jt, double miter_limit = 2.0,
+                  double arc_tolerance = 0.0);
+
+  // Minkowski
+  // NOTE: OpenSCAD does not use Minkowski as is (from Clipper1). Also, they
+  // allow a list patterns (see applyMinkowski). Should read through that and
+  // write these methods such that they can take Clippoly rather than a single
+  // path pattern.
```
elalish, line 72: You can Minkowski a polygon with holes onto a polygon with holes? I can't even imagine that. I honestly never found much use for Minkowski anyway, especially if we already have offset. What's the practical application here?

src/clippoly/src/clippoly.cpp
```diff
@@ -0,0 +1,295 @@
...
+#include "clipper2/clipper.core.h"
+#include "clipper2/clipper.engine.h"
+#include "clipper2/clipper.offset.h"
+#include "glm/ext/vector_float2.hpp"
+#include "public.h"
+
+using namespace manifold;
+
+namespace {
+const int precision_ = 8;
```
elalish, line 30: Is this absolute or relative?

elalish (commented):
Thank you, this is excellent - I'm also really glad to see how much functionality is here with not that much code. @pca006132 what do you think regarding API? Anything we should consider regarding our various bindings?

src/clippoly/src/clippoly.cpp
geoffder: They have scale, but it is mainly for converting between double and int64. The non-conversion one takes only a single factor for xy and the conversion scale takes a mutable int reference for as an error reporter. I just decided to skip that by doing it straight here.

src/clippoly/src/clippoly.cpp
geoffder: It's a path simplification algorithm meant to remove extraneous segments (particularly those that can be introduced by offsetting).

src/clippoly/src/clippoly.cpp
geoffder: And no, no rotate.

src/clippoly/src/clippoly.cpp
geoffder: RMD commented on above is an existing algorithm, and simplify is Clipper2's own take on similar. From what I recall they work at the similarity of consecutive segments (short, similar angle, can be replaced by one segment). Collinear just removes points that are fully collinear without getting into those types of calculations.

src/clippoly/src/clippoly.cpp
geoffder: It does, but OpenSCAD uses non-zero, so I have been favouring that.

src/clippoly/src/clippoly.cpp
geoffder: My idea with clean_ is to score a little laziness without much effort. Going through clipping sets clean_ to true, and controlled construction like square can start as clean_, while construction from arbitrary contours is unclean, since it can contain intersections. I'm a little unsure as to which non-clipping operations should be marked as unclean though.

src/clippoly/include/clippoly.h
geoffder: Not on edges specifically, but it does support including Z coordinates on points that aren't used, but carried around.

src/clippoly/src/clippoly.cpp
geoffder: Number of decimal places (they use int64 internally, so this controls the conversion)

src/clippoly/include/clippoly.h
geoffder: CrossSection sounds good. Yea I wasn't sure about whether the indices were getting used or not so I just went along with it 😅.

src/clippoly/include/clippoly.h
geoffder: I've never really used minkowski either to be honest, but yea in OpenSCAD minkowski takes a list of polygons (which can have holes) rather than just a pattern like clipper2 does. 

The sticky part API wise is that it a pattern must be given as a singular path rather than a CrossSection, we'd have to have shape constructors outside of CrossSection as well for convenience.

src/manifold/include/manifold.h
geoffder: Yea, nice to have the abstraction along the way to prevent avoidable errors.

src/clippoly/include/clippoly.h
geoffder: There could be a `Contour` class that wraps Clipper2 `PathD` in a similar way to `CrossSection`, but it's of course more duplication (just like there is in Clipper2 between `PathD` and `PathsD`).

I've cooked up a scheme that I think works pretty nicely in my OCaml Clipper2 bindings, where the single and plural paths are wrapped in an [abstract type](https://github.com/geoffder/ocaml-clipper2/blob/6d62f2d385cbd124c672560bdbb6672876c21326/lib/clipper.ml#L20) that allows the same functions to operate on both. In OOP land though, the user experience would be fairly seemless with similar methods on Contour vs CrossSection.

src/clippoly/src/clippoly.cpp
elalish: Okay, since this is absolute precision like our own `Precision` in Manifold, we should use a similar bounding box computation to determine the right value. Otherwise if someone starts using meters and gets tiny numbers, everything will just collapse to a point.

src/clippoly/include/clippoly.h
elalish: OpenSCAD is great, but I don't feel the need to follow their every move. I don't see much value in Minkowski and I don't like the idea of complicating everything to support it (since this seems to be the only place a simple polygon is required). Let's leave it out for now and see if someone comes to us with a powerful use case.

src/clippoly/src/clippoly.cpp
elalish: I think for the sake of simplicity and to match Manifold (the equivalent process is `SimplifyTopology()`), we should just clean on construction from arbitrary contours. Then we can just say all `CrossSections` are clean of self-intersections and there's no need for extra state.

src/clippoly/src/clippoly.cpp
elalish: Let's add `Rotate` - make sure to use our `sind` and `cosd` functions so the 90-degree increments are exact.

src/clippoly/src/clippoly.cpp
elalish: We don't follow *everything* OpenSCAD does - note our sphere is quite a bit different (and better) than theirs. In the event of self-intersection, non-zero gives xor-like polygons that share verts. It also discards winding direction information, since it's symmetric. The 3MF spec uses strictly-positive for this reason, so I think we should too.

src/clippoly/src/clippoly.cpp
elalish: I don't love having five functions that do basically the same thing. Let's be opinionated in our API and pick one. Since Manifold is based on inexact geometry, there's no point in exact checks anyway.

src/clippoly/src/clippoly.cpp
geoffder: I should also mention that 8 is the maximum value.

src/clippoly/src/clippoly.cpp
geoffder: Hmm, you're right that a preemptive redundant union before other booleans probably isn't the end of the world. I still wonder if intersections can be introduced or not by the simplifications (particularly with high epsilon values), if it's possible for CrossSections to become unclean, I'd favour marking them rather than unioning every time. Not too much work to add back anyway if it does seem necessary though.

src/clippoly/src/clippoly.cpp
geoffder: Simplify might be enough on its own then. I think that might be designed to catch the others anyway, I'll look at the source again.

src/clippoly/include/clippoly.h
elalish: Okay, I've cleaned up our polygons API in #343. Honestly it's a nice simplification anyway, but should also help here. Good to know regarding Z values, but let's ignore them for now.

src/clippoly/src/clippoly.cpp
elalish: Oh, interesting. Any idea why that is? Do they expect us to scale things into a range closer to 1 before input?

src/clippoly/src/clippoly.cpp
elalish: If push comes to shove, we can always add an optional construction parameter to specify the desired fill rule.

geoffder:
No sorry, I haven't actually gone digging about why it stops at 8 with 10 digits of headroom in int64 😅. Clipper used to not even have the double interface at all. If more flexibility than 8 is required, we'd need to use the int64 side and do the precision scaling in CrossSection.

src/clippoly/src/clippoly.cpp
geoffder: No sorry, I haven't actually gone digging about why it stops at 8 with 10 digits of headroom in int64 😅. Clipper used to not even have the double interface at all. If more flexibility than 8 is required, we'd need to use the int64 side and do the precision scaling in CrossSection.

elalish: Okay, good enough for now, but maybe file an issue so we don't forget when this gets merged.

src/cross_section/CMakeLists.txt
```diff
@@ -0,0 +1,28 @@
+# Copyright 2020 The Manifold Authors.
```
elalish, line 1: Let's update these dates to 2023.

elalish (commented):
Since all the comments are now marked outdated due to the rename, would you mind resolving the ones you've finished so we know what's left to do?

src/cross_section/CMakeLists.txt
elalish: There's a few more of them...

geoffder: Oh right... I'll do a replace. For some reason I interpreted as just on this new one. :sweat_smile:

geoffder:
Now that usage of CrossSection is forced, 11 of the tests are failing.

src/collider/src/collider.cpp
```diff
@@ -1,4 +1,4 @@
-// Copyright 2021 The Manifold Authors.
+// Copyright 2023 The Manifold Authors.
```
elalish, line 1: Sorry, not what I meant. Would you mind reverting that replace all commit? I just mean you created several files, and each of them had a copied data that was old - I commented on one, but there were ~3?

elalish (changes requested):
Looks like the tests are doing their job! Do you feel comfortable debugging? The simplest thing is probably just to dump out the input polygons and `ToPolygon` result and see what changed. Let me know if you get stumped.

src/collider/src/collider.cpp
geoffder: Will do, I should have clarified first.

geoffder:
> Looks like the tests are doing their job! Do you feel comfortable debugging? The simplest thing is probably just to dump out the input polygons and `ToPolygon` result and see what changed. Let me know if you get stumped.

Ok, I'll try to get some in later and report on what's happening to the inputs. I'm suspecting that at least some of them are due to the precision rounding step, but there could of course be deeper things going on with the clipping. The outputs are still manifold and the exported model glbs seem fine, but obviously some vertices are moving around to throw off the geometry checks.

src/cross_section/src/cross_section.cpp
```diff
@@ -0,0 +1,269 @@
...
+  auto ps = C2::PathsD{(pathd_of_contour(contour))};
+  paths_ = C2::Union(ps, C2::FillRule::Positive);
+}
+
+CrossSection::CrossSection(const Polygons& contours) {
+  auto ps = C2::PathsD(contours.size());
+  for (auto ctr : contours) {
+    ps.push_back(pathd_of_contour(ctr));
+  }
+  paths_ = C2::Union(ps, C2::FillRule::Positive);
```
elalish, line 94: I suppose the quick test is to check if removing this line fixes the tests. Still Revolve and Extrude are getting the wrong genus (!) and some of the tests are off by a lot in volume. I think there's a real issue somewhere.

geoffder: Coming back with slightly fresher eyes, I caught that I missed a couple fixes earlier with how I was incorrectly using the `std::vector` constructor to pre-allocate. With those switched to `reserve`, the empty padding on the front is gone, and now only two failures remain.

Boolean.Coplanar and Boolean-CoplanarProp both have lose a small number of vertices/triangles.

geoffder:
All tests now passing on my machine :+1:.

geoffder:
I haven't converted Manifold (and the rest of the codebase) over yet, but I lifted the circular quality static globals into a class `Quality` that can be shared by `CrossSection` and `Manifold`. If that sounds reasonable to you, I can remove the originals from Manifold and replace the usages with `Quality` (or whatever you'd like it renamed as).

elalish:
> I haven't converted Manifold (and the rest of the codebase) over yet, but I lifted the circular quality static globals into a class `Quality` that can be shared by `CrossSection` and `Manifold`. If that sounds reasonable to you, I can remove the originals from Manifold and replace the usages with `Quality` (or whatever you'd like it renamed as).

Yes, that sounds like a great idea. Also, how do you want to approach this PR? We'll want to update various bindings and examples to make use of what you've built. Would you like to keep working here, or get this PR merged and look at follow-ons? I'm not too picky, but I'm working on getting a release out, and I think it's best for this to come after.

src/cross_section/src/cross_section.cpp
elalish: Good finds! And it gives me further confidence in our testing regimen. Speaking of, the codecov report does give some pointers now and then about where some extra tests would be helpful. As you can probably tell, I tend to prefer end-to-end tests over unit tests - test more with less code whenever possible!

geoffder:
> Yes, that sounds like a great idea.

Roger, I'll change things over then.

> Also, how do you want to approach this PR? We'll want to update various bindings and examples to make use of what you've built. Would you like to keep working here, or get this PR merged and look at follow-ons? I'm not too picky, but I'm working on getting a release out, and I think it's best for this to come after.

I think merging first, then updating bindings works fine. Since Polygons has a constructor in CrossSection and everything is working without "knowing" that CrossSections exists, the bindings haven't been broken and from here the changes they need are additive.

src/cross_section/include/quality.h
```diff
@@ -8,6 +8,15 @@ class Quality {
   static float circularEdgeLength_;
 
  public:
+  /** @name Defaults
```
elalish, line 11: Let's move this file to utilities, since it's shared.

src/cross_section/include/cross_section.h
```diff
@@ -63,6 +67,9 @@ class CrossSection {
   CrossSection Offset(double delta, JoinType jt, double miter_limit = 2.0,
                       double arc_tolerance = 0.0);
 
+  // Geometry
+  double Area();
```
elalish, line 71: 👍

src/cross_section/include/cross_section.h
```diff
@@ -36,7 +36,9 @@ class CrossSection {
 
   // Shapes
   static CrossSection Square(glm::vec2 dims, bool center = false);
-  static CrossSection Circle(float radius, int circularSegments);
+  static CrossSection Circle(float radius, int circularSegments = 0);
+  static CrossSection Ellipse(float radiusX, float radiusY,
```
elalish, line 40: In keeping with the Manifold/OpenSCAD API, I think we can skip this, since it's just `Circle().Scale()`.

src/cross_section/include/cross_section.h
```diff
@@ -54,6 +56,8 @@ class CrossSection {
   CrossSection Translate(glm::vec2 v);
   CrossSection Rotate(float degrees);
   CrossSection Scale(glm::vec2 s);
+  CrossSection Mirror(glm::vec2 ax, bool reverse = true);
```
elalish, line 59: Let's merge this code into `Scale` - a negative scale should automatically mirror. This is how `Scale` works for `Manifold` too: it can invert the winding.

src/cross_section/include/cross_section.h
```diff
@@ -54,6 +56,8 @@ class CrossSection {
...
+  CrossSection Rewind();
```
elalish, line 60: Is this necessary?

samples/src/knot.cpp
```diff
@@ -44,7 +44,7 @@ Manifold TorusKnot(int p, int q, float majorRadius, float minorRadius,
   p /= kLoops;
   q /= kLoops;
   int n = circularSegments > 2 ? circularSegments
-                               : Manifold::GetCircularSegments(threadRadius);
+                               : Quality::GetCircularSegments(threadRadius);
```
elalish, line 47: This is a good opportunity to switch to using your new `Circle()` method.

elalish (changes requested):

src/cross_section/include/cross_section.h
geoffder: Negative scaling can't mirror across arbitrary vectors without changing the scaling though can it? Or is my brain not working right now 🤔.

src/cross_section/include/cross_section.h
geoffder: Sure, makes sense.

src/cross_section/include/quality.h
geoffder: That is where I tried to have it initially, but I was having troubling getting it to compile as expected. I didn't explore changing it up much, but I think the problem might have stemmed from the utilities library being an interface, so the headers are included differently than usual. 

Do you have an idea of what I was doing wrong, or what would need to change to move the quality source and header over without issue?

src/cross_section/include/cross_section.h
geoffder: I thought it would be helpful to include when I was testing Mirror, before I added the reverse parameter. Mirroring reverses the paths, so outlines are now holes and vice-versa. This is something that doesn't come up in OpenSCAD, as I imagine they counter the reversal of the paths for you. I figured it's conceivable that someone might feed contours into CrossSection and find that they are the wrong winding, so having Rewind provides an easy way to remedy that without having to mess with iterators or loops to fix their input vectors.

To bleed across reviews, besides mirroring on arbitrary vectors/axes, this is another reason why I think Mirror is better than negative scaling. To be honest, I think I'd remove the reverse parameter since I'm not sure people would expect or want their cross section to invert/dissapear on mirroring.

samples/src/knot.cpp
geoffder: Oh I thought I got all of the `Manifold::GetCircularSegments` uses.. weird, must have messed up my regex. Do you find any others or is that the last hanger on? I got rid of the Manifold ones so I figured the compiler would catch me for missing any.

geoffder: Ah good point. Cylinder as well.

src/cross_section/include/quality.h
elalish: Actually, it would be nice to condense it into just the header. Want to try one of the solutions outlined [here](https://stackoverflow.com/questions/18860895/how-to-initialize-static-members-in-the-header)?

src/cross_section/include/cross_section.h
elalish: Yes, but you can accomplish it with `Rotate().Scale().Rotate()`. If people feel it's important, we should add it to `Manifold` too - let's save that discussion for later.

src/cross_section/include/cross_section.h
elalish: Yeah, I hear you, but my main design goal is to mimic the `Manifold` API, and we don't have an invert function. If it becomes important, we can add it to both at once down the road.

src/cross_section/include/cross_section.h
geoffder: I'm not really a fan of that as a workaround to be honest. Is there anything to stop us adding axis mirroring to Manifold instead of removing it here? Does applying the 3d version of this formula to each point and reversing the faces of a manifold to compensate something that trips over the properties that manifold tracks that makes it a more complicated operation than it is for plain meshes?

src/cross_section/include/quality.h
geoffder: I'll give it a shot during my next session 👍

src/cross_section/include/cross_section.h
elalish: No, it's the same. Honestly I think you're probably right, just figured it might be better in its own PR.

src/cross_section/include/cross_section.h
pca006132: I’m wondering if we should automatically check for winding direction and help users fix their cross section before creating a solid from it.

src/cross_section/src/cross_section.cpp
```diff
@@ -0,0 +1,318 @@
...
+CrossSection CrossSection::Mirror(glm::vec2 ax, bool reverse) {
+  auto mirrored = C2::PathsD();
+  mirrored.reserve(paths_.size());
+  for (auto path : paths_) {
+    auto sz = path.size();
+    auto m = C2::PathD(sz);
+    for (int i = 0; i < sz; ++i) {
+      auto v = v2_of_pd(path[i]);
+      int idx = reverse ? sz - 1 - i : i;
+      m[idx] = v2_to_pd(ax * (2 * glm::dot(v, ax) / glm::dot(ax, ax)) - v);
```
pca006132, line 269: Perhaps we should have a check to make sure that `ax` length is non-zero and throw an exception/different return value when it is? So people writing bindings or using it as the backend don't have to write too much code for checking.

src/cross_section/include/cross_section.h
elalish: This is equivalent to a having a manifold with the triangles wound CW instead of CCW. I wouldn't want users to accidentally use the wrong winding and have us secretly fixing it as best we can. I'm open to throwing an exception if they're enabled when the output is empty. 

Alternative: we had mentioned earlier the possibility of specifying an optional fill rule in the constructor - non-zero would take care of this. Then at least it's explicit.

geoffder: Since contours can be an arbitrary number of polygons including nesting and overlaps, I don't think the intended winding is easily knowable. If the input is a single contour, or a data type that is understood to only represent a single outline with zero or more holes, then doing that could make sense (maybe with a usually on optional parameter in case they intended it for some reason, like to do clipping with the opposite winding polygon with another to achieve with inverted mean ii ng or something 😂.

elalish:
> Just curious: Why do we name this as cross section instead of polygon? Other than this I don't think I have anything to comment on, this looks pretty good to me!

We still have `Polygons` which is just a vector of vectors. I think of `Polygons` as the 2D version of `Mesh` and `CrossSection` as the 2D version of `Manifold`. With polygons you have to ask yourself if they are self-intersected or not, while a cross section is meant to imply geometric sanity. Someday I hope to get there with `Manifold` too. And certainly its usage so far (in Extrude and Revolve) is as a cross-section.

src/cross_section/include/cross_section.h
geoffder: > I'm open to throwing an exception if they're enabled when the output is empty.

As long as it can be turned off since exceptions aren't great for FFI.

> Alternative: we had mentioned earlier the possibility of specifying an optional fill rule in the constructor - non-zero would take care of this. Then at least it's explicit.

Unfortunately carrying around a fillrule from construction time wouldn't really work, since there would be CrossSections with mismatching fillrules getting out into booleans with eachother. It would need to be an optional on every method/constructor that does a boolean.

elalish: No, the specified fill rule would only affect the initial cleanup on construction. That guarantees every CrossSection has no overlaps, so the fill rule no longer matters for any operations thereafter.

src/cross_section/src/cross_section.cpp
geoffder: Would returning the same CrossSection back be appropriate (no axis to mirror over)? Or an empty CrossSection? Speaking of, should probably add a method to check if it's empty.

src/cross_section/include/cross_section.h
geoffder: Oh gotcha 👍.

src/cross_section/include/cross_section.h
geoffder: Should we follow up with another PR to add Mirror and Invert then? Or just add them to this one?

elalish: Let's do those in a follow-up since they affect the Manifold API.

geoffder: Ok, shall I rename Rewind to Invert then?

elalish: I think I'd rather remove it entirely in favor of the fill rule argument discussed below. I think they're basically two ways of addressing the same issue.

geoffder: Ok, sure. The norm I'd already to flip meshes if they don't agree with manifolds convention when constructing from a mesh anyway I guess (I do coming from OCADml meshes, since the winding is done such that it makes non-inverted polyhedrons for OpenSCAD), so with the addition of a mirror that does the flip, Manifolds shouldn't end up inverted unless negative scaling is performed.

src/cross_section/include/cross_section.h
geoffder: I think I still had my wires crossed for a bit with the lazy cleaning mindset (which is *sort of* like what I have in my other Clipper bindings). This is definitely a win for the clean on construction scheme.

src/cross_section/src/cross_section.cpp
geoffder: I've changed it to return an empty CrossSection for the moment (coinciding with the addition of IsEmpty). Please let me know if you guys think that it should be the unchanged input instead.

src/cross_section/include/cross_section.h
geoffder: Rewind is removed, and FillRule has been added as an optional (default Positive) to the contour constructors.

src/cross_section/include/quality.h
geoffder: I've moved it into `public.h`, the `inline static` variable declaration worked. My LSP insists on warning me that it is a C++17 extension, but I added a definition to the utilities CMakeLists to suppress the warning for compiling purposes.

src/cross_section/include/cross_section.h
```diff
@@ -31,14 +31,17 @@ class CrossSection {
...
   CrossSection& operator=(CrossSection&&) noexcept;
-  CrossSection(const SimplePolygon& contour);
-  CrossSection(const Polygons& contours);
+
+  // Construction from arbitrary contours
+  enum class FillRule { EvenOdd, NonZero, Positive, Negative };
+  CrossSection(const SimplePolygon& contour,
+               FillRule fillrule = FillRule::Positive);
+  CrossSection(const Polygons& contours,
+               FillRule fillrule = FillRule::Positive);
```
elalish, line 40: 💯

src/cross_section/src/cross_section.cpp
```diff
@@ -257,36 +266,24 @@ CrossSection CrossSection::Scale(glm::vec2 scale) {
   return CrossSection(scaled);
 }
 
-CrossSection CrossSection::Mirror(glm::vec2 ax, bool reverse) {
+CrossSection CrossSection::Mirror(glm::vec2 ax) {
+  if (glm::length(ax) == 0.) {
+    return CrossSection();
```
elalish, line 271: 👍

src/cross_section/src/cross_section.cpp
```diff
@@ -300,6 +297,7 @@ CrossSection CrossSection::Offset(double delta, JoinType jointype,
 }
 
 double CrossSection::Area() { return C2::Area(paths_); }
+bool CrossSection::IsEmpty() { return paths_.empty(); }
```
elalish, line 300: If paths contains several empty path contours, but no actual points, does this return true or false?

src/utilities/include/public.h
```diff
@@ -423,6 +423,83 @@ struct Box {
   }
 };
 
+class Quality {
+ private:
+  inline static int circularSegments_ = 0;
+  inline static float circularAngle_ = 10.0f;
+  inline static float circularEdgeLength_ = 1.0f;
```
elalish, line 430: 💯

elalish (commented):
This is looking great! What else would you like to do before we merge?

src/cross_section/src/cross_section.cpp
geoffder: Ah it would be false in that case. I'm unsure whether that is a possible case (that clipper would return such a thing from union), but I could add a check for that in anyway.

elalish: Better yet, add a test - then you'll know if you need to handle the case or not.

samples/src/knot.cpp
geoffder: I've converted all of the manual circle creations that I found :+1:

src/cross_section/src/cross_section.cpp
geoffder:
```
TEST(CrossSection, Empty) {
  Polygons polys(2);
  auto e = CrossSection(polys);
  EXPECT_TRUE(e.IsEmpty());
}
```
Was just coming back to confirm that this passes, so we don't have to worry about it.

geoffder:
> This is looking great! What else would you like to do before we merge?

I can't think of anything at the moment that *needs* to be added in the first go. Probably a good place to call it?

geoffder:
> > This is looking great! What else would you like to do before we merge?
> 
> I can't think of anything at the moment that _needs_ to be added in the first go. Probably a good place to call it?

Actually, the last potential piece would be wrapping `Clipper2Lib::Rect` as well. Giving us a 2d bounding box and access to "rect clipping" which is a fast special case of intersection.

elalish (approved):
Excellent work! Ready to merge?

geoffder:
Looking at it again I've remembered that I haven't added Transform yet for the 2d types. That would bring them another step closer to the 3d counterparts and make it ready I think.

geoffder:
Ok, `Transform` is in (along with a test like the one in for Manifold) :+1:

elalish (approved):
Looks great, thanks!

elalish:
@geoffder I would still like to see the equivalent `Mirror` function in `Manifold`, not to mention exposing `CrossSection` in our various bindings. Are you interested in taking some of that on? Also, do you want to add yourself to our `AUTHORS` file? I can also add you as a collaborator on Github if you'd like to stay involved. I appreciate your contributions!

geoffder:
@elalish 
I wrote out the new C bindings on the weekend, just need to test them out via the ocaml bindings to make sure that I didn't miss something and that they work as expected. I can add the Manifold mirroring to my list as well.

> Also, do you want to add yourself to our AUTHORS file? I can also add you as a collaborator on Github if you'd like to stay involved. I appreciate your contributions!

Sure I'll add myself on there soon, thanks for the offer :pray:. Continuing to collaborate and stay involved sounds good to me.

PR #342, opened 2023-02-28
Fix ES6 .d.ts exports

rafern:
The new ES6 module can't be used with Typescript, this should fix it

### DIFF
--- a/bindings/wasm/manifold.d.ts
+++ b/bindings/wasm/manifold.d.ts
@@ -12,18 +12,18 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-interface SealedUint32Array<N extends number> extends Uint32Array {
+export interface SealedUint32Array<N extends number> extends Uint32Array {
   length: N;
 }
 
-interface SealedFloat32Array<N extends number> extends Float32Array {
+export interface SealedFloat32Array<N extends number> extends Float32Array {
   length: N;
 }
 
-type Vec2 = [number, number];
-type Vec3 = [number, number, number];
+export type Vec2 = [number, number];
+export type Vec3 = [number, number, number];
 // 4x4 matrix stored in column-major order
-type Mat4 = [
+export type Mat4 = [
   number,
   number,
   number,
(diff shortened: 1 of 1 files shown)


elalish (approved):
Perfect timing, thank you!

bindings/wasm/manifold.d.ts
```diff
@@ -97,7 +97,7 @@ declare class Mesh {
   transform(run: number): Mat4;
 }
 
-declare class Manifold {
+export class Manifold {
```
elalish, line 100: Maybe I was too hasty here - Doesn't this mean you're exporting `Mesh` and `Manifold` before the `Module` loads? I think that's why they're part of `ManifoldStatic`?

bindings/wasm/manifold.d.ts
```diff
@@ -299,7 +299,7 @@ declare class Manifold {
  * @param size The X, Y, and Z dimensions of the box.
  * @param center Set to true to shift the center to the origin.
  */
-declare function cube(size?: Vec3|number, center?: boolean): Manifold;
```
elalish, line 302: This makes sense as a change for our library, but it broke manifoldCAD.org which also loads the functions into the global namespace. I'm curious what the cleanest way to fix that is in our editor.d.ts file?

bindings/wasm/manifold.d.ts
elalish: nevermind, I'm pretty sure I'm wrong.

bindings/wasm/manifold.d.ts
elalish: See #344 - would be nice to find a way to improve this.

bindings/wasm/manifold.d.ts
rafern: You're actually right. It doesn't actually affect the initialization (if you do `new Manifold` you will get a runtime error), but the IDE will think that doing `new Manifold` instead of `new module.Manifold` is valid, which is a mistake. I came up with a solution and will open a new PR

PR #344, opened 2023-02-28
ugly fix for editor TS

elalish:
Follow-on to #342, since that broke all the declarations in our manifoldcad.org editor. I couldn't find a decent solution yet, so  I did the ugly but functional thing, which is to copy the old version of the declarations into the editor's .d.ts file and not import the ES6 one. I tried fixing it according to https://stackoverflow.com/a/66948535, but couldn't get it to work.

### DIFF
--- a/bindings/wasm/examples/editor.d.ts
+++ b/bindings/wasm/examples/editor.d.ts
@@ -12,6 +12,480 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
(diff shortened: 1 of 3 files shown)


elalish:
FYI @rafern I think your change is good, but since we're importing our module inside our `worker.js`, we need a way to "unpack" the module exports from the `manifold.d.ts` file, but I can't find a way to get Monaco to do that.

PR #346, opened 2023-02-28
Properly encapsulate ES6 type defs. in module

rafern:
This should fix the weird discrepancy between the type definitions and the actual runtime; doing `new Manifold` instead of `new module.Manifold` will now actually show an IDE error.

I'm pretty sure this is correct now, but I'll keep this as a draft so I can take a better look into this tomorrow.


rafern:
@elalish This should be ready now. I tried to make it work with Monaco, but I can't figure out how; the type definitions are split into multiple files, because we need a public and private version of the Mesh and Manifold classes, but for some reason Monaco doesn't like to cross-reference type definitions. I also can't make it more usable for Monaco by merging the type definitions into a single file, because there is no way to do private type definitions in a `.d.ts` without multiple files; if I were to use a namespace instead of a separate file, the namespace is auto-exported, meaning you could do, for example, `new ManifoldStatic.Manifold`, which will result in a runtime error.

Seems like there's 2 choices at the moment:
- Keep using `declare` as before, for every function and class inside `ManifoldStatic`. This should make the type definition usable by Monaco, but also allow invalid usage in Typescript (doing `new Manifold` instead of `new module.Manifold`, calling `cube()` instead of `module.cube()`, etc...)
- Keep this version of the type definitions, but have a separate type definition file for the editor. This is more work, but should give correct type definitions everywhere

I also updated the cmake file to copy the new type definition files to the editor's folder. These aren't being used right now, but it should still be useful if a solution is found in the future.

bindings/wasm/manifold.d.ts
```diff
@@ -12,474 +12,32 @@
...
+export * from './manifold-global-types';
 
-/**
- * Returns the first of n sequential new unique mesh IDs for marking sets of
- * triangles that can be looked up after further operations. Assign to
- * Mesh.runOriginalID vector.
- */
-function reserveIDs(count: number): number;
+export type Manifold = T.Manifold;
+export type Mesh = T.Mesh;
```
elalish, line 19: So, do we actually want these? I thought this was what you removed so that `new Manifold()` would correctly error?

bindings/wasm/manifold.d.ts
```diff
@@ -12,474 +12,32 @@
...
- * must be specified.
- */
-///@{
-function setMinCircularAngle(angle: number): void;
-function setMinCircularEdgeLength(length: number): void;
-function setCircularSegments(segments: number): void;
-function getCircularSegments(radius: number): number;
-///@}
+import * as T from './manifold-encapsulated-types';
+export * from './manifold-global-types';
```
elalish, line 16: Good idea on separating these. I definitely don't want to go back to the `declare` version. The first priority is getting the types right for npm users - the editor is our own problem. Maybe we could hack it a little though - I'm considering some string concatenation and replacing `export` with `declare` during import to the editor.

bindings/wasm/manifold.d.ts
rafern: The way it was done before, both the type and the value are exported. In the new way, Manifold and Mesh are re-exported only as a type (which is why it has `export type Manifold` instead of `export class Manifold`), so if you try to do `new Manifold` you will now get an error. It's only a type AND value when used from a ManifoldStatic instance (which is why it's using `Manifold: typeof T.Manifold` in ManifoldStatic)

elalish: Ah, thank you for the explanation!

elalish (approved):
Thanks for the help!

bindings/wasm/manifold.d.ts
rafern: I'm not sure if that would work, because Typescript really hates bundling type definitions, but it's worth a try. Maybe you could use something like dts-bundle?

PR #347, opened 2023-03-01
prod deployment again

elalish:
Trying to prod deployment again, a la #345 - hoping this isn't just the definition of insanity...


pca006132:
I think we can click the 'rerun all jobs' button in the action page. Not sure if it helps with deployment though.

elalish:
Yeah, I should try that - I will if this one fails too. The trouble is I'm not sure whether the upload our download step is failing, and they're part of different jobs.

PR #348, opened 2023-03-02
clean up editor.d.ts

elalish:
Reverts the ugly solution from #344, using the refactor from #347. FYI @rafern.


rafern:
Oh wow, I'm surprised this works, and it's much cleaner than before. I thought it wouldn't work because concatenating the type definition files doesn't encapsulate the types, but you don't actually want to encapsulate them, since you can directly call functions like `cube` in the editor.

PR #352, opened 2023-03-07
Added docs, refactored for single OpType

elalish:
@geoffder This should add CrossSection to our [docs](https://manifoldcad.org/docs/html/modules.html). I also removed the `Xor` op type (since it creates corner-touching polygons, which we don't want to encourage, and since we don't have a Manifold equivalent) and combined the OpType definitions.


geoffder:
Ok, I'll adjust the C bindings accordingly 👍. I didn't really think of that downside.

PR #353, opened 2023-03-07
Add CrossSection and Rect to C bindings

geoffder:
`ManifoldCrossSection` and `ManifoldRect` types and their associated methods have been added, along with switching over `extrude` and `revolve` to take `ManifoldCrossSection*` rather than `ManifoldPolygons*`


elalish (approved):
LGTM, thanks!

PR #354, opened 2023-03-07
Add Mirror transformation method to Manifold

geoffder:
At first I thought that additional lazy tracking of inversion would be required, but it seems like the `invert` check with `glm::determinant` in `Impl::Transform` is already taking care of it? The added test is a simple 3d adapted version of the one for `CrossSection`.

### DIFF
--- a/src/manifold/src/manifold.cpp
+++ b/src/manifold/src/manifold.cpp
@@ -518,6 +518,30 @@ Manifold Manifold::Transform(const glm::mat4x3& m) const {
 }
 
 /**
+ * Mirror this Manifold over the plane described by the unit form of the given
+ * normal vector. If the length of the normal is zero, an empty Manifold is
+ * returned. This operation can be chained. Transforms are combined and applied
+ * lazily.
+ *
+ * @param normal The normal vector of the plane to be mirrored over
+ */
+Manifold Manifold::Mirror(glm::vec3 normal) const {
+  if (glm::length(normal) == 0.) {
+    return Manifold();
+  }
+  auto n = glm::normalize(normal);
+  glm::mat4x3 m(1.0f - 2.0f * n.x * n.x, -2.0f * n.x * n.y,
+                -2.0f * n.x * n.z,  //
+                -2.0f * n.x * n.y, 1.0f - 2.0f * n.y * n.y,
+                -2.0f * n.y * n.z,  //
+                -2.0f * n.x * n.z, -2.0f * n.y * n.z,
+                1.0f - 2.0f * n.z * n.z  //
+                ,
+                0.0f, 0.0f, 0.0f);
+  return Manifold(pNode_->Transform(m));
+}
+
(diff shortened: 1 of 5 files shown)


src/manifold/src/manifold.cpp
```diff
@@ -517,6 +517,30 @@ Manifold Manifold::Transform(const glm::mat4x3& m) const {
...
+  }
+  auto n = glm::normalize(normal);
+  glm::mat4x3 m(1.0f - 2.0f * n.x * n.x, -2.0f * n.x * n.y,
+                -2.0f * n.x * n.z,  //
+                -2.0f * n.x * n.y, 1.0f - 2.0f * n.y * n.y,
+                -2.0f * n.y * n.z,  //
+                -2.0f * n.x * n.z, -2.0f * n.y * n.z,
+                1.0f - 2.0f * n.z * n.z  //
+                ,
+                0.0f, 0.0f, 0.0f);
```
elalish, line 540: Could this be rewritten as identity - outerProduct? Like `glm::mat4x3(glm::mat3(1) - glm::outerProduct(normal, normal))`.

test/manifold_test.cpp
```diff
@@ -490,3 +490,18 @@ TEST(Manifold, FaceIDRoundTrip) {
...
+  auto b = a.Translate({2.5, 2.5, 2.5});
+  auto result = a + b + b.Mirror({1, 1, 0});
+
+#ifdef MANIFOLD_EXPORT
+  if (options.exportModels)
+    ExportMesh("manifold_mirror_union.glb", result.GetMesh(), {});
+#endif
+
+  auto vol_a = a.GetProperties().volume;
+  EXPECT_FLOAT_EQ(vol_a + 1.75 * vol_a, result.GetProperties().volume);
```
elalish, line 505: 2.75 * vol_a?

geoffder: Yes, that would be clearer. :sweat_smile:

src/manifold/src/manifold.cpp
geoffder: I just did it this "manual" way because that is what I have used in the past based on other libraries and math notes. That does look like it describes the same operation though.

geoffder: Actually, no seems like not. Your equation worked with `{1, 1, 0}` as in current test, but a normal of `{1, 1, 1}` it gives this broken output:
image
When the other way gives this:
image
Which appears correct.

elalish: Oh, I think I missed the factor of 2: `glm::mat4x3(glm::mat3(1) - 2.0 * glm::outerProduct(normal, normal))`

geoffder: Ah yes, that works (with a normalized normal).

elalish (approved):
Thanks!

PR #355, opened 2023-03-08
Lazy transformations for CrossSection

geoffder:
After reading through the transformation code more for Manifold (csg_tree and impl), I decided try bringing CrossSection more in line with the lazy transformation semantics expressed there. Only copying and applying the transformations when needed cuts out some of the repeated iteration and point type conversion that pervaded the wrapper. 

Similar to the CSG nodes and `GetImpl`, the private `GetPaths` applies the transformation whenever `paths_` is needed, and updates the object so that the same operation won't be repeated.

### DIFF
--- a/src/cross_section/include/cross_section.h
+++ b/src/cross_section/include/cross_section.h
@@ -16,6 +16,8 @@
 
 #include <clipper2/clipper.h>
 
+#include <memory>
+
 #include "clipper2/clipper.core.h"
 #include "clipper2/clipper.offset.h"
 #include "glm/ext/matrix_float3x2.hpp"
@@ -97,7 +99,7 @@ class CrossSection {
   ///@}
 
  private:
-  mutable C2::PathsD paths_;
+  mutable std::shared_ptr<const C2::PathsD> paths_;
   mutable glm::mat3x2 transform_ = glm::mat3x2(1.0f);
   CrossSection(C2::PathsD paths);
   C2::PathsD GetPaths() const;
--- a/src/cross_section/include/cross_section.h
+++ b/src/cross_section/include/cross_section.h
@@ -18,6 +18,7 @@
 
 #include "clipper2/clipper.core.h"
 #include "clipper2/clipper.offset.h"
+#include "glm/ext/matrix_float3x2.hpp"
 #include "glm/ext/vector_float2.hpp"
 #include "public.h"
 
@@ -97,7 +98,7 @@ class CrossSection {
 
  private:
   mutable C2::PathsD paths_;
-  mutable glm::mat3x2 transform_;
+  mutable glm::mat3x2 transform_ = glm::mat3x2(1.0f);
   CrossSection(C2::PathsD paths);
   C2::PathsD GetPaths() const;
 };
--- a/src/cross_section/include/cross_section.h
+++ b/src/cross_section/include/cross_section.h
@@ -96,8 +96,10 @@ class CrossSection {
   ///@}
 
  private:
-  C2::PathsD paths_;
+  mutable C2::PathsD paths_;
+  mutable glm::mat3x2 transform_;
   CrossSection(C2::PathsD paths);
+  C2::PathsD GetPaths() const;
 };
 /** @} */
 
--- a/src/cross_section/src/cross_section.cpp
+++ b/src/cross_section/src/cross_section.cpp
@@ -118,7 +118,10 @@ CrossSection::CrossSection() { paths_ = C2::PathsD(); }
 CrossSection::~CrossSection() = default;
 CrossSection::CrossSection(CrossSection&&) noexcept = default;
 CrossSection& CrossSection::operator=(CrossSection&&) noexcept = default;
-CrossSection::CrossSection(const CrossSection& other) { paths_ = other.paths_; }
+CrossSection::CrossSection(const CrossSection& other) {
+  paths_ = other.paths_;
+  transform_ = other.transform_;
+}
 CrossSection::CrossSection(C2::PathsD ps) { paths_ = ps; }
 
 CrossSection::CrossSection(const SimplePolygon& contour, FillRule fillrule) {
--- a/src/cross_section/src/cross_section.cpp
+++ b/src/cross_section/src/cross_section.cpp
@@ -95,12 +95,9 @@ C2::PathD pathd_of_contour(const SimplePolygon& ctr) {
 
 C2::PathsD transform(const C2::PathsD ps, const glm::mat3x2 m) {
   if (m == glm::mat3x2(1.0f)) {
-    // return ps;
-    return C2::PathsD(ps);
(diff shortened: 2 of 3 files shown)


pca006132:
Did you implement shared cache when users copy the path? If not, we will need to do n transformations instead of 1 when the path is copied n times, for example for other boolean operation or something.
(Although the transformation is probably cheap enough and not warrant such a cache)

Otherwise this looks good to me.

geoffder:
It could be that I'm not totally familiar with when copies are happening due to my C++ inexperience, but I thought that in this setup there wouldn't be multiple repeated transformation copies of the vector for separate booleans on the same parent manifold since `transform_` is reset to the identity. 

I also could be wrong in my assumption that `transformed.paths_ = paths_` doesn't copy the *elements* of the vector (my intention was for `transform` to be when the copy into a fresh `PathsD` is performed, severing the mutable link with the parent `CrossSection` if there was one). That's something I realized I should experiment with / read about, but forgot.

pca006132:
Ah, no, I mean the copy constructor.
https://github.com/elalish/manifold/blob/f92667e67e4bc2e9315228d782af881c43a47b34/src/cross_section/src/cross_section.cpp#L121-L124

The new `CrossSection` object will have a new `paths_` vector that is not yet transformed, and the `paths_` field is not shared between the copies, so calls to `GetPaths` on each copy will perform the same transformations repeatedly. For `CsgNode`, we have

https://github.com/elalish/manifold/blob/83131c672e90fd3019c60c900c4de1bdd840b1f1/src/manifold/src/csg_tree.h#L58

the `std::shared_ptr` implements a shared cache.

geoffder:
Oh right. Well it wouldn't be too much effort to convert `paths_` to a shared ptr to cover that case as well I think. I was also thinking about whether these mutables should be protected by a lock but you've beat me to bringing it up https://github.com/elalish/manifold/issues/356.

geoffder:
Thinking about this a bit more, is there a point of having a copy constructor for these objects that don't expose any mutable state to the outside anyway?

geoffder:
Looking at `GetImpl()`, the `pImpl_` pointer appears to be replaced by a new pointer, rather than updated. So if any copies share the original `pImpl_` pointer, they aren't actually updated, right? Am I getting lost in the indirection and looking at the wrong place?
```cpp
std::shared_ptr<const Manifold::Impl> CsgLeafNode::GetImpl() const {
  if (transform_ == glm::mat4x3(1.0f)) return pImpl_;
  pImpl_ =
      std::make_shared<const Manifold::Impl>(pImpl_->Transform(transform_));
  transform_ = glm::mat4x3(1.0f);
  return pImpl_;
}
```

elalish:
@geoffder I'll defer to @pca006132 who understands this part of the code better than I do, though I'd guess he's asleep now. It might help to make a little example and throw in print statements where transforms are actually applied to verts. The results would probably make for nice documentation as well.

geoffder:
In the meantime, here is a bit more thinking out loud, plus an experiment.

The `mutable std::shared_ptr<const Manifold::Impl> pImpl_;` pointer in `CsgLeafNode` is actually `const`, so updating what it is pointing at with `*pImpl_ = ...` is not actually allowed (which from what I understand, would be how transformation application would be shared between copies). I tried out making the equivalent `paths_` in CrossSection non-const for this experiment and it does not work as is. I made a branch to experiment with a shared_ptr version [here](https://github.com/geoffder/manifold/blob/e941335693308b952c526c66da99087d29a41722/src/cross_section/src/cross_section.cpp#L147). The comments re-iterating some of what I have already said in here for context.

Running this with the fresh ptr and const data version based on `GetImpl`, this [test](https://github.com/geoffder/manifold/blob/54de031dd5931d919e71cd857854f169ae78db74/test/cross_section_test.cpp#L77) gives the following output:

```
making a...
making b...
copying b...
extruding a and b...
Applied transform in GetPaths()
Applied transform in GetPaths()
extruding b_copy...
Applied transform in GetPaths()
extruding the original b again...
```
So, as expected the same transformation is applied on the as yet un-transformed `b_copy` before it is extruded (meshes produced by the extrusions are identical). This behaviour is that same as the non-`shared_ptr` version of this PR, so there isn't anything funny happening, it just similarly does not avoid duplication of transformation effort between copies. Then again, calling back to my previous comment, I'm not sure how much of a problem that is with the functional style (nice) const interface.

pca006132:
Oh, nice, I forgot to add a `*` for the `pImpl_`. And yes, the transform matrix also need to be shared between different meshes. I guess we can leave it here for now and fix them together with another PR. Thanks for spotting the bugs!

pca006132:
I found that the old me is partially correct: the cache for `CsgLeafNode` is meant to be immutable, the benefit of the cache is to reduce the number of copying rather than reduce the number of transformation application. I implemented the transform cache just to find that due to additional copying, the performance is worse. This may also be important here: we should avoid copying the path as much as possible, and we can just perform the transform on-the-fly while copying at the end. I will document this later.

But for the `CsgOpNode`, there is just the concurrency issue.

geoffder:
I finally went and re-read the reference page for vectors and was reminded (or TIL'd, I'm not sure) that the assignment operator for does indeed copy the contents. Something that was not obvious to me coming from languages where that is not the case :man_facepalming:. I've updated `paths_` to follow the immutable (contents) shared_ptr scheme accordingly.

geoffder:
A (final?) note on the topic of the copy constructor again, I don't know that there is necessarily a better solution than what we are doing currently.

```cpp
auto a = CrossSection::Circle(10).Scale({2, 0});
auto b = CrossSection(a);
auto ex_a = a.Extrude(5);
auto ex_b = b.Rotate(90).Extrude(5);
```
Even if the additional machinery was implemented to allow the transformation of **a** to be propagated to **b**, the same number of transformations is going to occur. Alternatively, if the copy constructor was changed to preemptively apply the transformations in **a** to cut off the possibility of re-doing the same `transform_`, if there are any further transforms to either or both, there could end up being **more** applications than the current case.

The solution is really to just not use the copy constructor, since there is nothing to be gained from it (just do what you want to **a**, don't make a copy of it to do things to).

geoffder:
Rebased thinking I was going to use the `ConcurrentSharedPtr` (and I still could change it anyway), but then I remembered that the contents of the `shared_ptr` are const anyway, so there aren't any race conditions to protect against. Update is only done by replacing the whole thing. So if thread safety is to be added, it must be at the class member level (a private mutex member to protect the mutable `paths_` and `transform_` members, rather than one bundled with the pointer itself in `ConcurrentSharedPtr`). Right?

pca006132 (approved):

pca006132:
Yes, you don't need `ConcurrentSharedPtr` if you directly replaces the pointer.

elalish (approved):
Thanks!

Issue #356, opened 2023-03-08
Fix shared cache for `CsgNode`


pca006132:
Just thought of this when talking about #355, the current shared cache for `CsgNode` is not thread-safe. Just need to add a lock around the `std::shared_ptr`.

Edit: And make it actually caching stuff.

PR #357, opened 2023-03-09
Fix cache concurrency issue (#356)

pca006132:
Fixed the cache issue and added a simple helper class so we can reuse the logic later.


geoffder:
Seems like a nice and minimal drop-in solution to make the transformation updates safer, shouldn't add too much overhead either especially since acquiring the lock won't even be happening often.

elalish (approved):
Thanks!

PR #366, opened 2023-03-13
Make tests optional + fix option declarations

geoffder:
This breaks out the MeshIO from the tests, and makes building the tests optional. In cases where building/re-building the tests is not wanted/needed, this can shorten build times by over 25% on my machine. A major motivator for this is building 3rd party bindings using the C FFI bindings (that include an export function for convenience).

Also included here are descriptors for the option declarations, which I noticed do not actually set the default value as expected if the description argument is omitted (defaults to OFF, rather than ON for `_TEST` and `_PYBIND` for example -- on my machine anyway).


geoffder:
I suspect these windows and mac failures are due to `PYBIND11_FINDPYTHON` actually being "correctly" set to `ON`. I'll give changing it explicitly to `OFF` a try tomorrow.

pca006132:
Interesting, I did not pay much attention to the Mac and Windows CI. I think we did not install python headers for those jobs? It would be nice if we can install them and do a test for python binding.
We can also simplify the CI commands for Linux so it does not override the default values.

geoffder:
Setting `-DPYBIND11_FINDPYTHON=OFF` fixed the windows build, but interestingly mac is still failing.

geoffder:
Investigating the error being hit in the mac build, I found an [issue](https://github.com/sirfz/tesserocr/issues/298) that makes it look like it is due to newer python version deprecations in `PyFrameObject`. I'm trying to test this out by specifying python 3.10 (though I don't know my way around mac/brew).

I wonder if this is something that has already been remedied in newer pybind11 commits (eventually the default python will increment in the other systems).

geoffder:
The mac CI is succeeding with `PYBIND11_FINDPYTHON=ON` now, following update of the pybind11 submodule (https://github.com/elalish/manifold/pull/367). From reading some of the commit history since the previous stable version, I'm hopeful that the issues that I believe are stemming from python 3.11 here will be resolved.

geoffder:
:tada: the mac is building

CMakeLists.txt
```diff
@@ -83,7 +84,14 @@ if(CODE_COVERAGE AND NOT MSVC)
 endif()
 
 add_subdirectory(src)
-add_subdirectory(samples)
-add_subdirectory(test)
-add_subdirectory(extras)
 add_subdirectory(bindings)
+
+if(MANIFOLD_EXPORT)
+  add_subdirectory(meshIO)
```
elalish, line 90: I'm still a bit concerned about promoting meshIO out of the test directory. I really can't stand behind it - export is decent, but import is very shaky. Formats are a lot of work - we need to careful about the promises we make. What do you think our goal should be here?

elalish (commented):
Thanks, this is looking great overall; let's discuss meshIO - I'm still not convinced it belongs in our C bindings either.

hrgdavor:
>let's discuss meshIO - I'm still not convinced it belongs in our C bindings either.

@elalish I am only watching from the side-lines, so I do not know many details. What is the recommendation for most efficient import of meshes (closest to internals for fast, and also reliable import that does not need fixing like STL)?

just a thought as an outsider, .. if meshIO is useful, and used for tests, but is not up-to snuff to be part of manifold, maybe separate it in a lib that has more limited support for issues than manifold. This would make it available to those eager to use it regardless. You would avoid the burden of supporting it too much.

CMakeLists.txt
elalish: Yeah, that's a good point, particularly about driving users into the arms of STL. @hrgdavor I'd say 3MF is probably the best, though I'm considering an extension for glTF that would make it equally good. Maybe what this really means is we should put enough effort into `meshIO` that it does a decent job. 

Let's go ahead with this change and take it as an action to improve the I/O.

elalish (approved):
Thank you!

PR #368, opened 2023-03-14
Simplify the CsgOpNodes as we build them, rather than in GetChildren

ochafik:
Early simplification (while building boolean op nodes) makes it easier to flatten the tree and avoids... stack overflows in GetChildren.

large_scene_test.cpp in this PR will just segfault without the flattening changes (and could potentially form the basis for some large scene benchmark so I thought it worth to check it in).

Found this crash while testing https://github.com/openscad/openscad/pull/4533 (segfaulted on issue2342.scad)

### DIFF
--- a/extras/large_scene_test.cpp
+++ b/extras/large_scene_test.cpp
@@ -23,11 +23,7 @@ using namespace manifold;
   Build & execute with the following command:
 
   ( mkdir -p build && cd build && \
-    cmake -DCMAKE_BUILD_TYPE=Release -DBUILD_SHARED_LIBS=ON \
-      -DMANIFOLD_PAR=TBB \
-      -DTHRUST_MULTICONFIG_ENABLE_SYSTEM_TBB=1 \
-      -DTHRUST_HOST_SYSTEM=THRUST_HOST_SYSTEM_TBB \
-      .. && \
+    cmake -DCMAKE_BUILD_TYPE=Release -DMANIFOLD_PAR=TBB .. && \
     make -j && \
     time ./extras/largeSceneTest 50 )
 */
--- a/extras/large_scene_test.cpp
+++ b/extras/large_scene_test.cpp
@@ -1,0 +1,58 @@
+// Copyright 2020 The Manifold Authors.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#include <chrono>
+#include <iostream>
+
+#include "manifold.h"
+
+using namespace manifold;
+
+/*
+  Build & execute with the following command:
+
+  ( mkdir -p build && cd build && \
+    cmake -DCMAKE_BUILD_TYPE=Release -DBUILD_SHARED_LIBS=ON \
+      -DMANIFOLD_PAR=TBB \
+      -DTHRUST_MULTICONFIG_ENABLE_SYSTEM_TBB=1 \
+      -DTHRUST_HOST_SYSTEM=THRUST_HOST_SYSTEM_TBB \
+      .. && \
+    make -j && \
+    time ./extras/largeSceneTest 50 )
+*/
+int main(int argc, char **argv) {
+  int n = 20;  // Crashes at n=50
+  if (argc == 2) n = atoi(argv[1]);
+
+  std::cout << "n = " << n << std::endl;
+
+  auto start = std::chrono::high_resolution_clock::now();
+  Manifold scene;
+
+  for (int i = 0; i < n; ++i) {
+    for (int j = 0; j < n; ++j) {
+      for (int k = 0; k < n; ++k) {
+        if (i == 0 && j == 0 && k == 0) continue;
+
+        Manifold sphere = Manifold::Sphere(1).Translate(glm::vec3(i, j, k));
+        scene = scene.Boolean(sphere, OpType::Add);
+      }
+    }
+  }
+  scene.NumTri();
+  auto end = std::chrono::high_resolution_clock::now();
+  std::chrono::duration<double> elapsed = end - start;
+  std::cout << "nTri = " << scene.NumTri() << ", time = " << elapsed.count()
+            << " sec" << std::endl;
+}
--- a/extras/large_scene_test.cpp
+++ b/extras/large_scene_test.cpp
@@ -32,7 +32,7 @@ using namespace manifold;
     time ./extras/largeSceneTest 50 )
 */
 int main(int argc, char **argv) {
-  int n = 20;  // Crashes at n=50
+  int n = 20;
   if (argc == 2) n = atoi(argv[1]);
 
   std::cout << "n = " << n << std::endl;
--- a/src/manifold/src/csg_tree.cpp
+++ b/src/manifold/src/csg_tree.cpp
@@ -257,7 +257,7 @@ std::shared_ptr<CsgNode> CsgOpNode::Boolean(
   auto handleOperand = [&](const std::shared_ptr<CsgNode> &operand) {
     if (auto opNode = std::dynamic_pointer_cast<CsgOpNode>(operand)) {
       if (opNode->IsOp(op) && opNode->impl_.UseCount() == 1) {
-        for (auto &child : opNode->GetChildren(/* finalize= */ false)) {
+        for (auto &child : opNode->GetChildren(/* forceToLeafNodes= */ false)) {
           children.push_back(child);
         }
         return;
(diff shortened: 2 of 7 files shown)


src/manifold/src/csg_tree.h
```diff
@@ -22,13 +22,16 @@ enum class CsgNodeType { Union, Intersection, Difference, Leaf };
 
 class CsgLeafNode;
 
-class CsgNode {
+class CsgNode : public std::enable_shared_from_this<CsgNode> {
```
pca006132, line 25: Thanks! I never know such a thing exists, I wanted to do something similar and have to do it the dumb way.

src/manifold/src/csg_tree.cpp
```diff
@@ -237,8 +248,36 @@ CsgOpNode::CsgOpNode(std::vector<std::shared_ptr<CsgNode>> &&children,
...
-  GetChildren(false);
+}
+
+std::shared_ptr<CsgNode> CsgOpNode::Boolean(std::shared_ptr<CsgNode> second,
+                                            OpType op) {
+  std::vector<std::shared_ptr<CsgNode>> children;
+
+  auto handleOperand = [&](const std::shared_ptr<CsgNode> &operand) {
+    if (auto opNode = std::dynamic_pointer_cast<CsgOpNode>(operand)) {
+      if (opNode->IsOp(op) && opNode.use_count() == 1 &&
```
pca006132, line 259: Maybe we can also have some sort of max recursion level parameter, and force flattening the tree when the recursion level is reached, regardless of whether or not the child is shared? So we can avoid stack overflow in any situation.

pca006132 (commented):
Thanks. I thought that I have already fixed this issue previously but it seems that it is not.

src/manifold/src/csg_tree.cpp
ochafik: So tbh I'm not quite sure in which circumstances the nodes and/or their impl are shared. I've relaxed the test anyway as I didn't realize the `operand.use_count()==1` broke the intended fix (there's copies of shared pointers around, and not just because I mistakenly passed it by value in that method - now fixed). Checking that count seems a bit brittle.

I've left the other check (in the impl.UseCount()) to keeps some mysterious tests happy, though.

src/manifold/src/csg_tree.h
ochafik: Yeah it's a nice get out of jail free card 🤣

src/manifold/src/csg_tree.cpp
ochafik: Re/ your suggestion to force-flatten in extreme sharing cases, maybe we could explore as follow up? Not sure if geometry from OpenSCAD would trigger that case though.

pca006132: Sure, I don't think it will happen normally.

src/manifold/src/csg_tree.cpp
```diff
@@ -419,34 +457,16 @@ std::vector<std::shared_ptr<CsgNode>> &CsgOpNode::GetChildren(
     bool finalize) const {
   auto impl = impl_.GetGuard();
   auto &children_ = impl->children_;
-  if (children_.empty() || (impl->simplified_ && !finalize) || impl->flattened_)
-    return children_;
-  impl->simplified_ = true;
```
elalish, line 424: Happy to see this simplification, but what was the difference between `simplified` and `flattened`?

elalish (commented):
@pca006132, you know this code better than I, so I'll leave approval to you.

src/manifold/src/csg_tree.cpp
pca006132: `simplified` did not perform flattening if the nodes are shared, iirc.

pca006132:
This looks fine for me, but can you do a simple benchmark comparing with the current master? Wondering if there is any performance improvement/degradation for our test cases (`manifold_test` and the python examples, which are extracted from your gist of benchmarks)

src/manifold/src/csg_tree.cpp
ochafik: Flattening (also aliased here to finalization) was forcing leaf nodes, while simplification was flattening 1 level of op nodes nesting (adopting the children of same op children, and in the case of difference, of union op children past the first child).

Have now renamed finalize & flattened_ to force[d]ToLeafNodes[_] to disambiguate, and simplified this method further.

src/manifold/src/csg_tree.cpp
```diff
@@ -237,8 +248,29 @@ CsgOpNode::CsgOpNode(std::vector<std::shared_ptr<CsgNode>> &&children,
...
+}
+
+std::shared_ptr<CsgNode> CsgOpNode::Boolean(
+    const std::shared_ptr<CsgNode> &second, OpType op) {
+  std::vector<std::shared_ptr<CsgNode>> children;
+
+  auto handleOperand = [&](const std::shared_ptr<CsgNode> &operand) {
+    if (auto opNode = std::dynamic_pointer_cast<CsgOpNode>(operand)) {
+      if ((opNode->IsOp(op) ||
+           (op == OpType::Subtract && opNode->IsOp(OpType::Add))) &&
```
ochafik, line 260: Now doing `(a-(b+c)-(d+e))` -> `(a-b-c-d-e)` here to play better w/ CsgOpNode::ToLeafNode (which turns it to `(a-batch(b+c+d+e))`

ochafik:
> can you do a simple benchmark comparing with the current master?

@pca006132 performance looks identical for these tests and examples, here's the [script I used to benchmark](https://gist.github.com/ochafik/9ba170000e561a184c6fefe364da7b60) (uses [hyperfine](https://github.com/sharkdp/hyperfine))

extras/large_scene_test.cpp
```diff
@@ -0,0 +1,58 @@
...
+    cmake -DCMAKE_BUILD_TYPE=Release -DBUILD_SHARED_LIBS=ON \
+      -DMANIFOLD_PAR=TBB \
+      -DTHRUST_MULTICONFIG_ENABLE_SYSTEM_TBB=1 \
+      -DTHRUST_HOST_SYSTEM=THRUST_HOST_SYSTEM_TBB \
+      .. && \
+    make -j && \
+    time ./extras/largeSceneTest 50 )
+*/
+int main(int argc, char **argv) {
+  int n = 20;  // Crashes at n=50
```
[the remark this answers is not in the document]
ochafik: That was only before, edited this out, thanks!

pca006132:
I was testing this branch and I found it breaks my existing python code. The output stl is different from the current master. I will try to make a minimal example for it.

pca006132:
```c++
  auto a = Manifold::Cylinder(10, 10);
  auto b = Manifold::Cube({10, 10, 10});
  auto c = Manifold::Cylinder(1, 1);

  float volume1 = (a - c).GetProperties().volume;
  float volume2 = (a + b - c).GetProperties().volume;
  EXPECT_GT(volume2, volume1);
```

The assertion failed:
```
Expected: (volume2) > (volume1), actual: 2342.13 vs 3122.84
```
while this passed with the current master:
```
3341.423096 > 3122.840088
```

ochafik:
@pca006132 thanks for the repro! Pushed a fix and some tests based on your snippet

src/manifold/src/csg_tree.cpp
```diff
@@ -237,8 +248,42 @@ CsgOpNode::CsgOpNode(std::vector<std::shared_ptr<CsgNode>> &&children,
...
-  GetChildren(false);
+}
+
+std::shared_ptr<CsgNode> CsgOpNode::Boolean(
+    const std::shared_ptr<CsgNode> &second, OpType op) {
+  std::vector<std::shared_ptr<CsgNode>> children;
+
+  if (IsOp(op) && (impl_.UseCount() == 1)) {
+    auto impl = impl_.GetGuard();
+    std::copy(impl->children_.begin(), impl->children_.end(),
```
pca006132, line 259: This copy doesn't apply transform from the parent.

src/manifold/src/csg_tree.cpp
```diff
@@ -237,8 +248,42 @@ CsgOpNode::CsgOpNode(std::vector<std::shared_ptr<CsgNode>> &&children,
...
+      case OpType::Subtract:
+        return secondOp->IsOp(OpType::Add);
+      default:
+        return false;
+    }
+  };
+
+  if (secondOp && (secondOp->impl_.UseCount() == 1) && canInlineSecondOp()) {
+    auto secondImpl = secondOp->impl_.GetGuard();
+    std::copy(secondImpl->children_.begin(), secondImpl->children_.end(),
```
pca006132, line 280: same here

pca006132 (changes requested):
Minimal example that you can add to the test case:
```c++
  auto a = (Manifold::Cube({1, 1, 1}) + Manifold::Cube({1, 1, 1})).Translate({1, 0, 0});
  auto b = (Manifold::Cube({1, 1, 1}) + Manifold::Cube({1, 1, 1}));

  EXPECT_FLOAT_EQ((a + b).GetProperties().volume, 2);
```

Currently this evaluates to 1. Note that if you write it as
```c++
  EXPECT_FLOAT_EQ(((Manifold::Cube({1, 1, 1}) + Manifold::Cube({1, 1, 1})).Translate({1, 0, 0}) + (Manifold::Cube({1, 1, 1}) + Manifold::Cube({1, 1, 1}))).GetProperties().volume, 2);
```
it will *not* trigger the bug, as the use count is somehow set to 2, perhaps due to how C++ deal with destructors.

src/manifold/src/csg_tree.cpp
ochafik: Fixed both w/ a helper

pca006132 (approved):

test/manifold_test.cpp
```diff
@@ -505,3 +505,39 @@ TEST(Manifold, MirrorUnion) {
...
   EXPECT_TRUE(a.Mirror(glm::vec3(0)).IsEmpty());
 }
+
+TEST(Manifold, BooleanVolumes) {
+  glm::mat4 m = glm::translate(glm::mat4(1.0f), glm::vec3(1.0f));
+
+  // Define solids which volumes are easy to compute w/ bit arithmetics:
+  // m1, m2, m4 are unique, non intersecting "bits" (of volume 1, 2, 4)
+  // m3 = m1 + m2
+  // m7 = m1 + m2 + m3
```
elalish, line 515: Excellent tests, thank you! Technically these should go in `boolean_test.cpp` now that I've separated them.

elalish (approved):
Nice job working together on this!

ochafik:
Thanks a lot guys, appreciate the help & quick turnaround!

PR #369, opened 2023-03-15
SimplePolygon/Polygons accessors for C bindings

geoffder:
These were left out before because they were only used for input to extrude/revolve, thus there was no reason to ever want to get points back out of the vectors. Now we have useful 2d operations that can create new Polygons worth extracting.


elalish (approved):
LGTM, thanks!

Issue #289, opened 2022-11-22
Overlap removal


elalish:
This will be a large feature; I'm not sure when I'll have time to tackle it, but I want to write down my thoughts since I now have a sketch of the algorithm. 

Removing overlaps would have two major uses: as the final stage of a manifold fixing algorithm for non-manifold input meshes, and also as a cleanup after `Warp()` that would take care of produced intersections. Technically it could also replace the Batch Boolean algorithm, but that's not a good idea as it probably won't be as fast. 

In the paper this library's Boolean algorithm is based on, it makes clear that it only works for two input manifolds, not for removing the self-intersections of a single manifold. The reason is that the paper's method of linking edges can be done independently and still guarantee a manifold result, which is why the algorithm is parallelizable, but examples are given that break this guarantee in the single-manifold case.

However, the first half of the Boolean should still work basically the same. The collider will need to be updated so it can automatically skip intersections between a triangle and any edges that share a vertex with it. Then the calculation of intersection points and the winding numbers of vertices can proceed as it currently does.

The difference is there are a few new vertices that will not be found by these intersections of triangles with edges. These new verts will be the intersection of three non-neighboring triangles, which will be findable because they are also the intersection of three new edges, and this intersection exists if and only if the three new edges share a single set of three triangles as neighbors.

The bigger problem for manifoldness is that previously the new verts on a given partially-retained edge could be paired up arbitrarily (and independently) and still guarantee a manifold result. In the case of these new triple-intersections, that is no longer true - they depend on each other topologically. This means the assembly algorithm cannot be parallelized (that I know of), but by working serially it should be possible to start at an intersection and work our way along each seam, checking any previous neighbor pair-ups and basing the new pair-ups on those to maintain topological consistency. 

Additionally, another type of new edge can form, which has only one intersection vert. This is when a manifold is folded, producing an intersection curve that terminates on both ends instead of forming a loop. In the case when a vert opposite the intersected edge (the other vert of either neighboring triangle) is shared with the triangle it is intersecting, then this new edge will go from the intersection vert to this opposite vert.

pca006132:
Apart from Warp, Refine can also cause overlap when the surfaces are too close.

elalish:
[...]

PR #370, opened 2023-03-15
Add NumVert, NumContour, and Warp

geoffder:
A few more methods to `CrossSection` to bring it another step in line with `Manifold`.

- RE: `Warp`, do you guys think that it should go through a union to clean up any user induced shenanigans, or leave it be (similar to Manifold).

### DIFF
--- a/src/cross_section/src/cross_section.cpp
+++ b/src/cross_section/src/cross_section.cpp
@@ -348,7 +348,7 @@ int CrossSection::NumVert() const {
   return n;
 }
 
-int CrossSection::NumContours() const { return GetPaths().size(); }
+int CrossSection::NumContour() const { return GetPaths().size(); }
 
 Rect CrossSection::Bounds() const {
   auto r = C2::GetBounds(GetPaths());
--- a/src/cross_section/src/cross_section.cpp
+++ b/src/cross_section/src/cross_section.cpp
@@ -320,7 +320,7 @@ CrossSection CrossSection::Warp(
     }
     warped.push_back(s);
   }
-  return warped;
+  return CrossSection(C2::Union(warped, C2::FillRule::Positive, precision_));
 }
 
 CrossSection CrossSection::Simplify(double epsilon) const {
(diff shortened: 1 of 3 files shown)


test/cross_section_test.cpp
```diff
@@ -97,3 +97,17 @@ TEST(CrossSection, Transform) {
...
+  auto a = sq.Scale({2, 3}).Translate({4, 5});
+  auto b = sq.Warp([](glm::vec2 &v) {
+    v.x = v.x * 2 + 4;
+    v.y = v.y * 3 + 5;
+  });
+
+  EXPECT_EQ(sq.NumVert(), 4);
+  EXPECT_EQ(sq.NumContour(), 1);
+  Identical(Manifold::Extrude(a, 1.).GetMesh(),
+            Manifold::Extrude(b, 1.).GetMesh());
```
elalish, line 112: 👍

elalish (approved):
Yes, I think a union after Warp is a good idea. I plan to eventually manage something equivalent in Manifold (#289).

geoffder:
> Yes, I think a union after Warp is a good idea. I plan to eventually manage something equivalent in Manifold (#289).

Ok, I'll add that in then :+1:

src/cross_section/src/cross_section.cpp
```diff
@@ -305,6 +305,24 @@ CrossSection CrossSection::Transform(const glm::mat3x2& m) const {
   return transformed;
 }
 
+CrossSection CrossSection::Warp(
```
elalish, line 308: Also, either now or in a follow-on PR, would you mind adding docs to these functions? The CrossSection [docs](https://manifoldcad.org/docs/html/classmanifold_1_1_cross_section.html) are looking a little sparse.

elalish (approved):

src/cross_section/src/cross_section.cpp
geoffder: Sure, I'll make that my next PR.

Issue #365, opened 2023-03-12
Can't resolve 'module' error when importing npm module for the brower + wasm compile error


frading:
Hello,

As I mentioned in the [Users of Manifold discussion](https://github.com/elalish/manifold/discussions/340), I'm starting to integrate Manifold in [Polygonjs](https://github.com/polygonjs/polygonjs). But I'm quickly facing issues, as importing the library generates the following error:

```
ERROR in ./node_modules/manifold-3d/manifold.js 8:8370-8386
Module not found: Error: Can't resolve 'module' in '.../polygonjs/node_modules/manifold-3d'
resolve 'module' in '.../polygonjs/node_modules/manifold-3d'
Parsed request is a module
using description file: .../polygonjs/node_modules/manifold-3d/package.json (relative path: .)
Field 'browser' doesn't contain a valid alias configuration
```

It seems to come from `await import('module')` inside `node_modules/manifold-3d/manifold.js`. So I can temporarily work around it by removing the whole block inside `if (ENVIRONMENT_IS_NODE) {...}`. (as I'm building for the browser, not for node).

Then the app builds, but loading the wasm file generates a second error:

```
Aborted(CompileError: WebAssembly.instantiate(): expected magic word 00 61 73 6d, found 3c 21 44 4f @+0). Build with -sASSERTIONS for more info.
```

And this is unfortunately not an error I'm able to debug.

So this brings me 2 questions:

1 - Would it be possible to have a build in the npm module that is only for the browser and does not include nodejs specific instructions? Alternatively, could you suggest modifications to my webpack config that could solve this?

2 - Is there anything I should change to load the .wasm without the error above?

This issue may need to be split into 2 for clarity, feel free to let me know if so, I'll happily do it.

You can see my attempt to integrate manifold in the [integration-manifold branch of polygonjs](https://github.com/polygonjs/polygonjs/tree/integration-manifold), and specifically [this file](https://github.com/polygonjs/polygonjs/blob/integration-manifold/src/engine/nodes/sop/SDFBox.ts). You can also start it with `yarn && yarn start` and view any errors in the dev console when opening it on `http://localhost:8080/example.html?id=nodes/sop/SDFBox`.

Don't hesitate to let me know if you need more info from me

pca006132:
This is weird, we do have tests running with nodejs in our CI:

https://github.com/elalish/manifold/blob/dca9cce5299f8829da79fada0a345beb615ad82c/.github/workflows/manifold.yml#L132

[...]

PR #371, opened 2023-03-15
locateFile property in Module function

jirihon:
Allows to provide custom url of the manifold.wasm file. This is needed for Vite. Partial fix for #365


elalish (approved):
Thanks!

PR #374, opened 2023-03-16
Add documentation comments to CrossSection and Rect

geoffder:


src/cross_section/include/cross_section.h
```diff
@@ -34,53 +34,250 @@ class Rect;
  *  @{
  */
 
+/**
+ * Two-dimensional cross sections guaranteed to be without self-intersections,
+ * or overlaps between polygons (from construction onwards). This class makes
+ * use of the [Clipper2](http://www.angusj.com/clipper2/Docs/Overview.htm)
+ * library for polygon clipping (boolean) and offsetting operations.
+ */
```
elalish, line 42: In `Manifold` we put these comments in the cpp files instead of headers, which I think makes the code a bit more readable? What do you think?

geoffder: I *personally* found that a bit funny when I was getting to know the library (and ran into it again when I was looking for the constructor docstrings for example, which live in another file). The header is in one spot with all of the exposed bits, and doesn't have the source to scroll through as well. My outlook is shaped by OCaml though, where the norm is to document the `.mli` (the interface file with the expoaed type and function signatures). I'm not as familiar with the general C++ norms.

I can move them over if you want.

geoffder: Comments on methods are moved over, those attached to classes/enums remain in the header.

elalish (approved):
Thanks!

PR #367, opened 2023-03-13
Update pybind11 to latest stable commit

geoffder:
While investigating the mac CI failures in PR https://github.com/elalish/manifold/pull/366, I noticed that pybind11 has a new stable release (which may include fixes for the python 3.11 issues that are currently being run into there).

PR #375, opened 2023-03-16
Add missing JS bindings

### LINKED ISSUE #373: .transform not available in wasm
Using the npm module, I get `.transform is not a function`.

And double checking in your online editor https://manifoldcad.org/, while I can see that `.translate`, `.rotate` and `.scale` work perfectly fine,  I also get an error when trying to use `.transform`. Even though it is mentioned in the [types](https://github.com/elalish/manifold/blob/master/bindings/wasm/manifold-encapsulated-types.d.ts#L206) 

edit:
to clarify what I tried in the editor, the following works:

```
const box = cube([100, 100, 100], true);
const ball = sphere(60, 100).translate([50,0,0]);
const result = box.subtract(ball);
```

But this generates an error:

```
const box = cube([100, 100, 100], true);
const ball = sphere(60, 100).transform([1, 0, 0, 0, 0, 1, 0, 0, 0, 0, 1, 0, 0, 0, 0, 1]);
const result = box.subtract(ball);
```

elalish:
Fixes #373 

Also adds `mirror` and `trimByPlane`.

### DIFF
--- a/bindings/wasm/bindings.cpp
+++ b/bindings/wasm/bindings.cpp
@@ -242,6 +242,7 @@ EMSCRIPTEN_BINDINGS(whatever) {
       .function("_Translate", &Manifold::Translate)
       .function("_Rotate", &Manifold::Rotate)
       .function("_Scale", &Manifold::Scale)
+      .function("_Mirror", &Manifold::Mirror)
       .function("_Decompose", select_overload<std::vector<Manifold>() const>(
                                   &Manifold::Decompose))
       .function("isEmpty", &Manifold::IsEmpty)
--- a/bindings/wasm/bindings.cpp
+++ b/bindings/wasm/bindings.cpp
@@ -239,7 +239,7 @@ EMSCRIPTEN_BINDINGS(whatever) {
       .function("_GetMeshJS", &GetMeshJS)
       .function("refine", &Manifold::Refine)
       .function("_Warp", &Warp)
-      .function("_Transform", &Transform)
+      .function("transform", &Transform)
       .function("_Translate", &Manifold::Translate)
       .function("_Rotate", &Manifold::Rotate)
       .function("_Scale", &Manifold::Scale)
--- a/bindings/wasm/bindings.cpp
+++ b/bindings/wasm/bindings.cpp
@@ -235,6 +235,7 @@ EMSCRIPTEN_BINDINGS(whatever) {
       .function("add", &Union)
       .function("subtract", &Difference)
       .function("intersect", &Intersection)
+      .function("_TrimByPlane", &Manifold::TrimByPlane)
       .function("_GetMeshJS", &GetMeshJS)
       .function("refine", &Manifold::Refine)
       .function("_Warp", &Warp)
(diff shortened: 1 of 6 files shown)


elalish:
@pca006132 Do you have any idea why the mac build is suddenly failing? It's complaining about python, but I've only changed the JS bindings...

bindings/python/third_party/pybind11
geoffder, line 1: Looks like you have not updated your pybind11 locally (may need to do something like merging in the changes that weren't applied from master or init again). This is why the mac CI failed.

geoffder (changes requested):

bindings/python/third_party/pybind11
elalish: It looks like the changes from https://github.com/elalish/manifold/pull/367 were somehow lost.

geoffder: https://github.com/elalish/manifold/tree/master/bindings/python/third_party They are still present on master, the issue is that git is a bit funny about actually applying these types of changes for you locally. For instance, when I do a submodule update on one branch and commit, then switch over to another, the commit change becomes visible as an unstaged changed.

elalish: Looks like it was removed here? https://github.com/elalish/manifold/pull/366/files#diff-723c1253d0c94075d3d7c80504b08943c02209f016433937ccf803a3ca6eb5feL198 I've already merged master - how would my local setup affect the CI?

pca006132: I think when you do the commit, you accidentally added `pybind11` directory to the list of staged changes, which is probably there because you merged master without doing a manual submodule update.  Submodules are really second class citizen in git...
Maybe you can try to do the update and see if there is an unstaged change to pybind11 now?

elalish: Yeah, okay I see I was confused. This is such a strange state. I swear changes to submodules used to just show up as a hash change in a hidden text file and this was all simple to fix in github. Now that seems to all be obfuscated and I can't figure out where to revert whatever difference these seems to be in my branch.

pca006132: Usually I just checkout to the master and cherry-pick the commits.

geoffder: @elalish 
> I can't figure out where to revert whatever difference there seems to be in my branch.

In this case, I think you should be able to navigate to the pybind11 submodule folder, and do a pull on the pybind11 stable branch.

elalish: Awesome, that worked, thanks @geoffder! Still baffled as to how this happened though. One thing: I notice that pybind11 is our only submodule that has a branch specified (stable). I wonder if that's an error?

geoffder: I'm not sure, I just noticed that the commit you were pinned on before was a previous stable branch commit (corresponding to a release). So I pulled from that branch to update. I didn't checkout stable though (just pulled the most recent commit from it).

geoffder (approved):

PR #376, opened 2023-03-16
Add CrossSection decomposition

geoffder:
Add ability to decompose a `CrossSection` into a set of non-intersecting independent parts (much like `Manifold::Decompose`). Implemented using the Clipper2 tree boolean operation that tracks the nesting relationships of paths, and followed by unpacking the results into vectors of contours (with the outlines at their heads).


elalish (approved):
Good idea! Thanks for the test.

PR #381, opened 2023-03-18
parallelize some csg operations

pca006132:
Just some trivial parallelization, results in ~5% improvement for meshes with many unions. Just thought of this and this makes me unable to fall asleep...
`BatchBoolean` is not optimized yet, not sure if I should do parallelization as it will mess up with the mesh size and may end up being slower.

### DIFF
--- a/src/manifold/src/csg_tree.cpp
+++ b/src/manifold/src/csg_tree.cpp
@@ -151,6 +151,9 @@ Manifold::Impl CsgLeafNode::Compose(
   int numVert = 0;
   int numEdge = 0;
   int numTri = 0;
+  std::vector<int> vertIndices;
+  std::vector<int> edgeIndices;
+  std::vector<int> triIndices;
   for (auto &node : nodes) {
     float nodeOldScale = node->pImpl_->bBox_.Scale();
     float nodeNewScale =
@@ -161,6 +164,9 @@ Manifold::Impl CsgLeafNode::Compose(
     if (!glm::isfinite(nodePrecision)) nodePrecision = -1;
     precision = glm::max(precision, nodePrecision);
 
+    vertIndices.push_back(numVert);
+    edgeIndices.push_back(numEdge * 2);
+    triIndices.push_back(numTri);
     numVert += node->pImpl_->NumVert();
     numEdge += node->pImpl_->NumEdge();
     numTri += node->pImpl_->NumTri();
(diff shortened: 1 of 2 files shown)


pca006132:
Btw @elalish part of the compose function has to be executed sequentially and will cause segfault when parallelized (the for loop that I left), not sure if that is a bug.

elalish (approved):
Thanks!

src/manifold/src/csg_tree.cpp
```diff
@@ -161,6 +164,9 @@ Manifold::Impl CsgLeafNode::Compose(
     if (!glm::isfinite(nodePrecision)) nodePrecision = -1;
     precision = glm::max(precision, nodePrecision);
 
+    vertIndices.push_back(numVert);
+    edgeIndices.push_back(numEdge * 2);
+    triIndices.push_back(numTri);
```
elalish, line 169: So this is the loop that segfaults when parallelized? I assume something is going wrong with the precision calculation? Does feel like a bug - I can't think of an immediate reason it couldn't be parallelized...

src/manifold/src/csg_tree.cpp
```diff
@@ -175,54 +181,66 @@ Manifold::Impl CsgLeafNode::Compose(
...
+        // important to add an offset so that each node instance gets
+        // unique meshIDs.
+        const int offset = i * Manifold::Impl::meshIDCounter_;
+        transform(policy, node->pImpl_->meshRelation_.triRef.begin(),
+                  node->pImpl_->meshRelation_.triRef.end(),
+                  combined.meshRelation_.triRef.begin() + triIndices[i],
+                  UpdateMeshIDs({offset}));
+      });
+
+  for (int i = 0; i < nodes.size(); i++) {
```
pca006132, line 237: @elalish this one

elalish: Oh weird, seems like that should work, as long as the offsets are calculated up font like `vertIndices`. Mind giving it a shot here so I can see the test failures on the CI?

pca006132: done, note that it may not segfault but may loop indefinitely.

elalish: Huh, it fails for parallel; I wonder if `std::map` doesn't play well with parallel insertions. Seems like that would be hard to manage...

pca006132: Oh ok, forgot that it is a map. This is normal then.

pca006132: removed that commit now

PR #384, opened 2023-03-21
update readme

elalish:
To include our usage, now that we have some!


pca006132:
LGTM. In addition I think we can mention the [OCaml binding](https://github.com/OCADml/OManifold) by @geoffder and add some figures for https://gist.github.com/ochafik/70a6b15e982b7ccd5a79ff9afd99dbcf#results as figures are more eye-catching than stating 100x faster.

PR #385, opened 2023-03-21
Improve handling of invalid shape construction.

### LINKED ISSUE #379: More sanity check
We don't do much validation for function inputs, and some inputs can cause a crash, such as cylinder with zero/negative radius. We should probably add this so downstream users will not get a segfault.

geoffder:
Fixes #379

- Add `Error::InvalidConstruction`, and add checks to `Manifold` constructors that return empty manifolds with this error when given invalid arguments (such as negative radius)
- Add similar checks to `CrossSection`, returning empty cross-sections
- As part of testing the behaviour of `Cube` with negative dimensions, I noticed the semantic is to return mirrored cubes. To match this in `CrossSection::Square`, I added a check that reverses the winding of the path to make sure the shape is still positive.

### DIFF
--- a/src/cross_section/src/cross_section.cpp
+++ b/src/cross_section/src/cross_section.cpp
@@ -231,36 +231,32 @@ C2::PathsD CrossSection::GetPaths() const {
 
 /**
  * Constructs a square with the given XY dimensions. By default it is
- * positioned in the first quadrant, touching the origin.
+ * positioned in the first quadrant, touching the origin. If any dimensions in
+ * size are negative, or if all are zero, an empty Manifold will be returned.
  *
  * @param size The X, and Y dimensions of the square.
  * @param center Set to true to shift the center to the origin.
  */
-CrossSection CrossSection::Square(const glm::vec2 dims, bool center) {
-  if (glm::length(dims) == 0.0f) {
+CrossSection CrossSection::Square(const glm::vec2 size, bool center) {
+  if (size.x < 0.0f || size.y < 0.0f || glm::length(size) == 0.0f) {
     return CrossSection();
   }
 
-  // reverse winding if dimensions are negative
-  const bool neg = dims.x * dims.y < 0.0f;
-  const int start = neg ? 3 : 0;
-  const int step = neg ? -1 : 1;
-
   auto p = C2::PathD(4);
   if (center) {
-    const auto w = dims.x / 2;
-    const auto h = dims.y / 2;
-    p[start + step * 0] = C2::PointD(w, h);
-    p[start + step * 1] = C2::PointD(-w, h);
-    p[start + step * 2] = C2::PointD(-w, -h);
-    p[start + step * 3] = C2::PointD(w, -h);
+    const auto w = size.x / 2;
+    const auto h = size.y / 2;
+    p[0] = C2::PointD(w, h);
+    p[1] = C2::PointD(-w, h);
+    p[2] = C2::PointD(-w, -h);
+    p[3] = C2::PointD(w, -h);
   } else {
-    const double x = dims.x;
-    const double y = dims.y;
-    p[start + step * 0] = C2::PointD(0.0, 0.0);
-    p[start + step * 1] = C2::PointD(x, 0.0);
-    p[start + step * 2] = C2::PointD(x, y);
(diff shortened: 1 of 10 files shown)


src/cross_section/src/cross_section.cpp
```diff
@@ -236,21 +237,30 @@ C2::PathsD CrossSection::GetPaths() const {
  * @param center Set to true to shift the center to the origin.
  */
 CrossSection CrossSection::Square(const glm::vec2 dims, bool center) {
+  if (glm::length(dims) == 0.0f) {
+    return CrossSection();
+  }
+
+  // reverse winding if dimensions are negative
```
elalish, line 244: Considering we're marking negative radius as invalid, should we just mark negative dimensions invalid too? Likewise for Manifold of course.

src/manifold/src/manifold.cpp
```diff
@@ -71,6 +71,12 @@ Manifold::Manifold(std::shared_ptr<CsgNode> pNode) : pNode_(pNode) {}
 Manifold::Manifold(std::shared_ptr<Impl> pImpl_)
     : pNode_(std::make_shared<CsgLeafNode>(pImpl_)) {}
 
+Manifold Manifold::Invalid() {
```
elalish, line 74: I think we have a `MarkFailure` method somewhere; could we use that instead?

test/cross_section_test.cpp
```diff
@@ -28,6 +28,24 @@
...
+  auto cb = Manifold::Cube({5, 5, 5});
+  auto neg_x_cb = Manifold::Cube({-5, 5, 5});
+  auto neg_y_cb = Manifold::Cube({5, -5, 5});
+  auto neg_xy_cb = Manifold::Cube({-5, -5, 5});
+
+  auto cubes = cb + neg_x_cb + neg_y_cb + neg_xy_cb;
+  auto ex = Manifold::Extrude(sq + neg_x_sq + neg_y_sq + neg_xy_sq, 5);
+
+  // ensure semantics for negative dimensions for square and cube agree
+  EXPECT_FLOAT_EQ((cubes - ex).GetProperties().volume, 0.);
```
elalish, line 46: 👍

src/cross_section/src/cross_section.cpp
geoffder: That would also make sense,  for reference, negative dimensions for squares and cubes results in empty shapes in OpenSCAD. On the other hand at least negative dimensions of squares/cubes have some intuitive sense/purpose (like negative scaling does),compared to negative radius which seems more obviously not useful (only interpretation other than failing is ignoring it).

src/manifold/src/manifold.cpp
geoffder: This one? https://github.com/elalish/manifold/blob/2c597d1eefea3febad4efe6683f3df0f44b7a9df/src/manifold/src/impl.h#L77 

It returns void, so still needs an impl created before applying it. Do you mean use it in the Invalid helper instead of setting the status directly? (I made the private helper just so the same lines wouldn't be repeated each time)

elalish: Ah, fair, that's not as useful as I was hoping. This is fine.

src/cross_section/src/cross_section.cpp
elalish: Yeah, let's make that change.

geoffder: Ok, I'll add a negative (on any dimension) check for square and cube then.

test/cross_section_test.cpp
geoffder: With negative dimensions on the way out, this test would just come down to extruded square vs cube I guess. 🤔

src/cross_section/src/cross_section.cpp
```diff
@@ -17,6 +17,7 @@
 #include <clipper2/clipper.h>
 
 #include <memory>
+#include <tuple>
```
elalish, line 20: What's this for?

elalish (approved):
Thanks! Would you mind updating our other bindings with this new enum value?

elalish (approved):
Great, thank you!

PR #389, opened 2023-03-23
update thrust to 2.1

elalish:
Thrust finally fixed the issue that required us to patch it, so figured it would be nice to update and simplify our build process. Of course it also has some [breaking changes](https://github.com/NVIDIA/thrust/releases/tag/2.0.0), though I *think* they shouldn't particularly affect us. However, they have added new submodules of their own, so I think some kind of submodule command will need to be run when you pull this change down to your machine before it'll work.


pca006132:
Writing CMake file is probably harder than writing C++... I think the problem is that we are not using their CMake file for building, instead we just specified the include path manually. When thrust added other dependencies, they expect us to run their CMake rules and set the include paths appropriately while we did not.

pca006132:
I have no idea why the windows build failed, for the wasm/js builds I think it is due to lack of thread API which is required by libcudacxx.

elalish:
Thank you! Okay, I fixed windows and simplified our yaml a bit. That thread API thing looks like a serious blocker though - hopefully Nvidia comes back with something there. 

Should we use their CMake file? I'm not even sure how to do that, but it sounds like the right way.

geoffder:
Using their CMake as in adding the subdirectory, and setting flags to avoid the stuff we don't need I assume? That's why we're doing with Clipper2 (and graphlite before it).

pca006132:
I tried implementing a very simple OMP backend with pthread a long time ago, but the performance is pretty poor. Maybe we can see if it builds when pthread is enabled.

CMakeLists.txt
```diff
@@ -29,7 +29,7 @@ set(MANIFOLD_FLAGS -O3)
 
 if(EMSCRIPTEN)
   message("Building for Emscripten")
-  set(MANIFOLD_FLAGS -fexceptions)
+  set(MANIFOLD_FLAGS -fexceptions -pthread)
```
elalish, line 32: According to [emscripten](https://emscripten.org/docs/porting/pthreads.html), this is all that's needed, but I'm getting the same compile error from Thrust.

elalish:
@pca006132 I've been looking into trying to simplify our CMake by using Thrust's CMake instead. I'm curious about our current CMake setup: we have a 'use CUDA' flag, and also a 'par' flag that can be CPP, OMP, or TBB. Reading Thrust's [docs](https://github.com/NVIDIA/thrust/blob/main/thrust/cmake/README.md), I feel like our dynamic 3-backend system is not really what they have in mind. How would you feel about simplifying it to always use CPP for host and compile for either CUDA, OMP, or TBB for device and just dynamically pick between host and device?

pca006132:
They actually allow users to choose between different backends, just probably a more advanced usage, e.g. they allow users to select between cuda streams. I don't want to go back to the host vs device way as it will make performance worse for operations that take a short time (and users may do many this kind of operations.)

elalish:
Well, advanced enough that their CMake doesn't support it, which makes me concerned they may not be testing it. I definitely like your dynamic approach to sending small operations through the serial code and large ops through the parallel code. My only question is, in the specific case where a CUDA GPU is available, how common is it that OMP or TBB have significantly better performance than both CPP and CUDA? That feels like a pretty narrow use case to me, and it would simplify things a lot to cap our number of dynamic backends at 2 instead of 3.

elalish:
Oh, interesting - I should give TBB a try. Thanks!

PR #393, opened 2023-03-28
Add JS bindings for triangulation

jirihon:
The triangulation algorithm available from JS allows you to create specific shapes efficiently, often without the need for boolean operations.

I was missing this very much as I frequently create custom polyhedrons, either from scratch, or based on modified geometry. This allows me to construct a manifold from a set of planar 3D polygons. I just project them to 2D (axis-aligned projection), triangulate and feed all the triangles to the manifold mesh constructor. Similar functionality is in OpenSCAD too [(polyhedron)](https://en.wikibooks.org/wiki/OpenSCAD_User_Manual/Primitive_Solids#polyhedron).

I believe the triangulation could be useful for many other users of the JS bindings and hope you'll be willing to merge this in some form.

### DIFF
--- a/bindings/wasm/bindings.cpp
+++ b/bindings/wasm/bindings.cpp
@@ -142,21 +142,6 @@ Manifold Extrude(std::vector<std::vector<glm::vec2>>& polygons, float height,
                            twistDegrees, scaleTop);
 }
 
-std::vector<glm::ivec3> TriangulateJS(
-    std::vector<std::vector<glm::vec2>>& polygons) {
-  CrossSection crossSection(polygons);
-  int idx = 0;
-  PolygonsIdx polygonsIndexed;
-  for (auto& poly : polygons) {
-    SimplePolygonIdx simpleIndexed;
-    for (const glm::vec2& polyVert : poly) {
-      simpleIndexed.push_back({polyVert, idx++});
-    }
-    polygonsIndexed.push_back(simpleIndexed);
-  }
-  return Triangulate(polygonsIndexed);
-}
-
 Manifold Revolve(std::vector<std::vector<glm::vec2>>& polygons,
                  int circularSegments) {
   return Manifold::Revolve(ToPolygon(polygons), circularSegments);
@@ -282,7 +267,7 @@ EMSCRIPTEN_BINDINGS(whatever) {
   function("tetrahedron", &Manifold::Tetrahedron);
   function("_Smooth", &Smooth);
   function("_Extrude", &Extrude);
-  function("_Triangulate", &TriangulateJS);
+  function("_Triangulate", &Triangulate);
   function("_Revolve", &Revolve);
   function("_LevelSet", &LevelSetJs);
 
--- a/bindings/wasm/bindings.cpp
+++ b/bindings/wasm/bindings.cpp
@@ -145,11 +145,9 @@ Manifold Extrude(std::vector<std::vector<glm::vec2>>& polygons, float height,
 std::vector<glm::ivec3> TriangulateJS(
     std::vector<std::vector<glm::vec2>>& polygons) {
   CrossSection crossSection(polygons);
-  auto _polygons = crossSection.ToPolygons();
-
   int idx = 0;
   PolygonsIdx polygonsIndexed;
-  for (auto& poly : _polygons) {
+  for (auto& poly : polygons) {
     SimplePolygonIdx simpleIndexed;
     for (const glm::vec2& polyVert : poly) {
       simpleIndexed.push_back({polyVert, idx++});
--- a/bindings/wasm/bindings.cpp
+++ b/bindings/wasm/bindings.cpp
@@ -20,6 +20,7 @@
 using namespace emscripten;
 
 #include <manifold.h>
+#include <polygon.h>
 #include <sdf.h>
 
 using namespace manifold;
@@ -139,6 +140,23 @@ Manifold Extrude(std::vector<std::vector<glm::vec2>>& polygons, float height,
                  int nDivisions, float twistDegrees, glm::vec2 scaleTop) {
   return Manifold::Extrude(ToPolygon(polygons), height, nDivisions,
                            twistDegrees, scaleTop);
+}
+
+std::vector<glm::ivec3> TriangulateJS(
+    std::vector<std::vector<glm::vec2>>& polygons) {
+  CrossSection crossSection(polygons);
+  auto _polygons = crossSection.ToPolygons();
+
+  int idx = 0;
+  PolygonsIdx polygonsIndexed;
+  for (auto& poly : _polygons) {
+    SimplePolygonIdx simpleIndexed;
+    for (const glm::vec2& polyVert : poly) {
+      simpleIndexed.push_back({polyVert, idx++});
+    }
+    polygonsIndexed.push_back(simpleIndexed);
+  }
+  return Triangulate(polygonsIndexed);
 }
 
 Manifold Revolve(std::vector<std::vector<glm::vec2>>& polygons,
@@ -266,6 +284,7 @@ EMSCRIPTEN_BINDINGS(whatever) {
   function("tetrahedron", &Manifold::Tetrahedron);
   function("_Smooth", &Smooth);
   function("_Extrude", &Extrude);
+  function("_Triangulate", &TriangulateJS);
   function("_Revolve", &Revolve);
   function("_LevelSet", &LevelSetJs);
 
--- a/bindings/wasm/CMakeLists.txt
+++ b/bindings/wasm/CMakeLists.txt
@@ -19,7 +19,7 @@ add_executable(manifoldjs bindings.cpp)
 # make sure that we recompile the wasm when bindings.js is being modified
 set_source_files_properties(bindings.cpp OBJECT_DEPENDS
   ${CMAKE_CURRENT_SOURCE_DIR}/bindings.js)
-target_link_libraries(manifoldjs manifold sdf cross_section)
+target_link_libraries(manifoldjs manifold sdf cross_section polygon)
 target_compile_options(manifoldjs PRIVATE ${MANIFOLD_FLAGS} -fexceptions)
 target_link_options(manifoldjs PUBLIC --pre-js ${CMAKE_CURRENT_SOURCE_DIR}/bindings.js --bind -sALLOW_TABLE_GROWTH=1
   -sEXPORTED_RUNTIME_METHODS=addFunction,removeFunction -sMODULARIZE=1 -sEXPORT_ES6=1)
--- a/bindings/wasm/bindings.js
+++ b/bindings/wasm/bindings.js
@@ -302,9 +302,9 @@ Module.setup = function() {
     return result;
   };
 
-  Module.triangulate = function(polygons) {
+  Module.triangulate = function(polygons, precision = -1) {
     const polygonsVec = polygons2vec(polygons);
-    const result = fromVec(Module._Triangulate(polygonsVec), (x) => [x[0], x[1], x[2]]);
+    const result = fromVec(Module._Triangulate(polygonsVec, precision), (x) => [x[0], x[1], x[2]]);
     disposePolygons(polygonsVec);
     return result;
   }
--- a/bindings/wasm/bindings.js
+++ b/bindings/wasm/bindings.js
@@ -302,6 +302,13 @@ Module.setup = function() {
     return result;
   };
 
+  Module.triangulate = function(polygons) {
+    const polygonsVec = polygons2vec(polygons);
+    const result = fromVec(Module._Triangulate(polygonsVec), (x) => [x[0], x[1], x[2]]);
+    disposePolygons(polygonsVec);
+    return result;
+  }
+
   Module.revolve = function(polygons, circularSegments = 0) {
     const polygonsVec = polygons2vec(polygons);
     const result = Module._Revolve(polygonsVec, circularSegments);
--- a/bindings/wasm/manifold-encapsulated-types.d.ts
+++ b/bindings/wasm/manifold-encapsulated-types.d.ts
@@ -109,11 +109,14 @@ export function extrude(
     twistDegrees?: number, scaleTop?: Vec2): Manifold;
 
 /**
- * Triangulates a set of polygons.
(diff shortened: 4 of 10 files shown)


hrgdavor:
I would definitely love to see this too :)

bindings/wasm/bindings.cpp
```diff
@@ -141,6 +142,21 @@ Manifold Extrude(std::vector<std::vector<glm::vec2>>& polygons, float height,
                            twistDegrees, scaleTop);
 }
 
+std::vector<glm::ivec3> TriangulateJS(
+    std::vector<std::vector<glm::vec2>>& polygons) {
+  CrossSection crossSection(polygons);
```
elalish, line 147: What's the crossSection for?

bindings/wasm/bindings.cpp
```diff
@@ -141,6 +142,21 @@ Manifold Extrude(std::vector<std::vector<glm::vec2>>& polygons, float height,
                            twistDegrees, scaleTop);
 }
 
+std::vector<glm::ivec3> TriangulateJS(
+    std::vector<std::vector<glm::vec2>>& polygons) {
```
elalish, line 146: Thanks, I was wondering how/when I should expose Triangulate as part of our public API. I'd like to keep our API consistent, so let's add this to the C++ and then expose it through JS, rather than writing it multiple times.

bindings/wasm/bindings.cpp
```diff
@@ -141,6 +142,21 @@ Manifold Extrude(std::vector<std::vector<glm::vec2>>& polygons, float height,
...
+  CrossSection crossSection(polygons);
+  int idx = 0;
+  PolygonsIdx polygonsIndexed;
+  for (auto& poly : polygons) {
+    SimplePolygonIdx simpleIndexed;
+    for (const glm::vec2& polyVert : poly) {
+      simpleIndexed.push_back({polyVert, idx++});
```
elalish, line 153: This implies the output indices refer to the polygon points in order, which I suppose is simple enough. I guess you have to do a mapping back to your polyhedron verts then? It would be helpful to include an example showing how to use this.

bindings/wasm/bindings.cpp
elalish: I would propose in `polygon.h`, change `Triangulate()` to `TriangulateIdx()` and add a `Triangulate()` function that matches your signature.

elalish (changes requested):
Thanks, good idea. I've basically been waiting for someone to ask for this, because I wasn't quite sure what the right API shape would be.

bindings/wasm/bindings.cpp
jirihon: Oh, I forgot to remove this line. Thanks for pointing out.

bindings/wasm/bindings.cpp
jirihon: You're right. I do backmapping of the verts. What is the right place for such an example?

bindings/wasm/bindings.cpp
jirihon: Good idea, I'll do the change in the following commit.

bindings/wasm/bindings.cpp
jirihon: Here is an example how I use the triangulation API:

```js
const projection = axis_aligned_projection(normal);
let idx = 0;
const vert_i_map: number[] = [];
const polygons_2d = polygons.map((polygon) =>
	polygon.map((vert_i) => {
		vert_i_map[idx++] = vert_i;
		return projection(vert_pos[vert_i]);
	}),
);
const result = manifold.triangulate(polygons_2d);
const tris = result.map((tri) => tri.map((vert_i) => vert_i_map[vert_i]));

function axis_aligned_projection(normal) {
	const abs_normal = new Vector3(Math.abs(normal.x), Math.abs(normal.y), Math.abs(normal.z));
	let x, y, z;

	if (abs_normal.z > abs_normal.x && abs_normal.z > abs_normal.y) {
		// x-y plane
		z = normal.z;
		x = 0;
		y = 1;
	} else if (abs_normal.y > abs_normal.x) {
		// z-x plane
		z = normal.y;
		x = 2;
		y = 0;
	} else {
		// y-z plane
		z = normal.x;
		x = 1;
		y = 2;
	}
	if (z < 0) {
		return (v) => [v[y], v[x]];
	} else {
		return (v) => [v[x], v[y]];
	}
}
```

elalish:
This looks ready to merge to me; let's just get the sample in there so people know how to use it.

bindings/wasm/bindings.cpp
jirihon: I'm having trouble coming up with a good example. If I make something simple, e.g. a modified cube with the top face a little bit smaller than the bottom face and with a square hole through both top and bottom face, it feels too simple to be included in the samples. Right now, samples are pretty complex awesome artistic objects. 

On the other hand, if I make something more sophisticated like a basic chair (without the need for booleans), it would be more practical to have some kind of general polyhedron constructor function than just a triangulation function. And for this polyhedron constructor, `TriangulateIdx` is more practical than `Triangulate`.

So after thinking about this more deeply I suggest two things:

1. Add a polyhedron constructor. Possibly in `constructors.cpp`? This could be also exposed in the JS bindings. There is, however, one limitation of the following implementation. The face triangles can't be backmapped to the input face polygons as we don't know how many triangles make up which face. We could easily return `tri2face` mapping from this function alongside the Manifold object (which would make it an outlier compared to the other constructors always returning just a manifold), or leave the users to use the triangulation directly if they need the mapping. I personally prefer the direct use of triangulation in this case as I'm able to prepare all the triangles together with normals and UVs in one go.

```cpp
Manifold Manifold::Polyhedron(
    const std::vector<glm::vec3> vertPos,
    const std::vector<std::vector<std::vector<int>>>& polygonsVec,
    const std::vector<glm::vec3> normals) {
  Mesh mesh;
  mesh.vertPos = vertPos;

  for (int i = 0; i < polygonsVec.size(); i++) {
    auto& polygons = polygonsVec[i];
    std::vector<glm::ivec3> tris;

    if (polygons.size() == 1 && polygons[0].size() == 3) {
      auto& p = polygons[0];
      tris.push_back({p[0], p[1], p[2]});
    } else if (polygons.size() == 1 && polygons[0].size() == 4) {
      auto& p = polygons[0];
      tris.push_back({p[0], p[1], p[3]});
      tris.push_back({p[1], p[2], p[3]});
    } else {
      auto& normal = normals[i];
      const glm::mat3x2 projection = GetAxisAlignedProjection(normal);
      auto polygons2d = To2DPolygon(polygons, projection, vertPos);
      tris = Triangulate(polygons2d);
    }
    for (auto& tri : tris) {
      mesh.triVerts.push_back(tri);
    }
  }
  return Manifold(mesh);
}
```

2. Add `TriangulateIdx` to the JS bindings, either as a replacement of `Triangulate`, or as an extra option. This would make it easier to implement custom polyhedron constructors in JS as it would not need remapping. The typescript signature could follow the C++ signature.

```typescript
export type PolyVert = [Vec2, number]
export type SimplePolygonIdx = PolyVert[];
export type PolygonsIdx = SimplePolygonIdx[];
export function triangulateIdx(polygons: PolygonsIdx, precision?: number): Vec3[];
```

I look forward to hearing what you think about this. I'm sorry that I can't react faster, but the timezone shift makes it hard :slightly_smiling_face:

geoffder: I do think that a polyhedron constructor like that would be a lot nice to have included (given that many people would end up using the triangulation function to do similar anyway).

jirihon: I've already tried to automatically calculate the normals and was stopped by the random order of the polygons. There is no guarantee that the outlining polygon will be first and the inner hole polygon will be second. So it can't be easily decided between the two possible directions of the normal. And wrong direction of the normal will reverse the polygon winding through the `GetAxisAlignedProjection` and make the manifold invalid.

jirihon: For comparison, OpenSCAD does not require normals in their [polyhedron constructor](https://en.wikibooks.org/wiki/OpenSCAD_User_Manual/Primitive_Solids#polyhedron), but they support simple polygons only - no holes.

geoffder: Hmm fair enough, I suppose I haven't come up against that yet myself. In OCADml I support polygons with holes by way of my [port](https://github.com/OCADml/OCADml/blob/main/lib/polyHoles.ml) of RonaldoCMPs [partitioning algorithm](https://github.com/RonaldoCMP/Polygon-stuffs/blob/master/polyHolePartition.scad) in OpenSCAD. 

This is of course done with polygons where the outline comes first, so the workflow for sweeps with holes for example is to slap on the partioned caps onto the open ends of the outer and inner sweeps.

elalish: @jirihon You've hit on exactly the dilemma that kept this out of our API so far. If the goal is polyhedra construction, then a polyhedron constructor is the way to go. However, I intentionally chose to base this library's API on triangulated rather than polygonal meshes (we use polygonal meshes internally) because:
1) If the verts of a polygon are not coplanar, its surface and enclosed volume are not well defined. Rounding error means they are almost never coplanar, which means checking this condition is pretty arbitrary.
2) If a polygon with holes degenerates to where the contours touch or aren't geometrically-valid, then even the topology is not well defined.
3) As you mention, polygon orientation isn't always clear. Providing normals is yet another way to create invalid input. Better would probably be to check face area and flip to force it positive. But again, we're just making ourselves even more dependent on unreliable geometric checks. 

This is not to say it's a hard no, but I'd like to understand what the polygon use cases look like. What makes it easier to construct a polyhedron chair than using Booleans? If an operation is fraught enough that we can't support it reliably, then it may be better to leave it out of our API and expose only a lower-level function (like Triangulate) so the users who want to do this have to think through the problems for themselves and be aware.

elalish: Right, this is why I want to understand the use cases of polygon meshes better. Right now my understanding is basically: 
1) Boolean operations naturally form polygonal meshes, even if the inputs are triangle meshes.
2) Triangulation is hard.
3) Might as well output polygonal meshes and make the downstream libraries figure it out.

If that's the only reason, then for the sake of pushing geometric reliability, I'm tempted to push back on this the same way as I push back on the use of STL files. I just don't think polygonal meshes are a reliable way to transmit information. But perhaps I'm missing some more fundamental use cases?

I intentionally made `polygon.cpp` very self-contained and light on dependencies, the hope being that libraries in the above situation would be able to incorporate it and thus output triangle meshes. But I need to understand from you, the authors of these libraries, if that's a reasonable solution. Because only you know the intricacies and assumptions inherent in your particular polygons.

geoffder: As I said, I was fine with taking care of triangulation myself, and did so. I don't think you have a moral imperative to support it or anything. My suggestion wouldn't be to allow polygonal meshes in manifold, so much as have a helper for what I imagine to be a common case (ones like mine where all faces are simple, and should pass coplanarity to something like `1e-6`) -- summarily rejecting inputs that cannot be transformed into triangular meshes via triangulation of each face independently (without additional information like normals). From that point I would think that it becomes the same as a mesh that was generated with strictly triangular faces to begin with.

As far as intentionally using polygonal meshes rather than triangular ones for an explicit purpose other than convenience I'm not sure. In my case, all faces are usually triangular other than the polygonal end caps of sweeps and a couple other special cases. But in each of them, post-hoc triangulation is as simple as doing it from the get go as they are all simple polygons. I've considered changing the mesh type to only have triangular faces in the past since I need triangular faces at the couple non-OpenSCAD polyhedron boundaries (Manifold and 😬STL😬 serialization), but haven't done so yet.

Anyway, I feel like a (picky) polyhedron helper would be a nice to have, but I also understand your point of view and have already done part of the leg work to align with it in practice.

hrgdavor: To me it looks like adding polygon triangualtion utility into manifold would make sense if a big performance boost can be achieved. How about only for 2d that is extruded ? ... I am following discussions, but have not found time to go further than git pull and running existing examples, so my comments may out of place from lack of insight.

elalish: We have `Manifold::Extrude()` already, so that should be covered.

jirihon: I have to admit, I got lost in the discussion :smile: I'll try to add something reasonable.

From my perspective, polyhedron constructor is just a convenient way to make _triangle_ meshes. I don't expect any special validity checks to happen on the input - it is my problem if I have it wrong.

In my project, I do a lot of derived geometry. Let's say I have a polyhedron as input and I need to push one of its faces. Or I want to "extrude" one of its faces. Or both at the same time. See the pictures below.

The first picture shows the input polyhedron. The front face is selected, so you can see what triangles it consists of.

Snímek obrazovky z 2023-04-04 13-38-43

The second picture shows the output after "extruding" the blue object from one of the other faces. The geometry of the blue object is calculated by moving the face points and using custom JS implementation of polyhedron constructor which uses manifold's triangulation. The input polyhedron is also modified. The affected points are shifted and the faces are re-triangulated. In this case, the re-triangulation is really needed, the input triangles would not work for the changed polyhedron.

Snímek obrazovky z 2023-04-04 13-38-36

You may wonder why I need manifold at all as this could be done with triangulation only. But I also need fast booleans, mainly for cutting the objects into pieces.

So this is my use case for working with polyhedrons (as an abstraction over triangle mesh) and I accept the risk of rounding errors. After all, who can find a perfectly planar polygon in the real world?

But to get back on the original topic:

I think it is enough for manifold to give access to the triangulation :triangular_ruler: So it is no problem to merge this PR without any further additions :smile:

I just can't think of any example showcasing the triangulation that wouldn't be trivial or too tedious to make. Please, do you have any ideas? And regarding tests, I think the triangulation is well tested right now.

elalish: Thanks for the explanation, I agree this PR can merge as is. That's great that you're working on face push/pull - honestly I was thinking this could be a good feature for Manifold. Internally we already have a concept of polygon faces: groups of neighboring triangles with the same `triRef.tri`. Would you have any interest in contributing this feature into Manifold?

elalish (approved):
Thanks!

PR #394, opened 2023-03-28
Add Merge function

elalish:
Fixes #363 
Fixes #398

This adds a `Merge` function to `MeshGL` that will heuristically fill in `mergeFromVert` and `mergeToVert` vectors, the purpose of which is to create manifold meshes from GL-style meshes that appear manifold but are not due to unshared verts where materials, UVs, normals, etc. have a boundary. 

This will not attempt to fix meshes that are more broken - I'll leave that to external auto-manifold tools for now, since that is more complicated and arbitrary.

PR #400, opened 2023-04-04
added assimp as submodule again

elalish:
Okay, so we've gone back and forth on this one a bit, but since we agreed to promote `meshIO` out of the test directory, it feels right to include assimp as a submodule again so we can build it everywhere. Partly this is because I'm on a Mac now and it's helpful to have that test export functionality. Also, I *think* I've got the CMake working better for it now.

### DIFF
--- a/meshIO/third_party/CMakeLists.txt
+++ b/meshIO/third_party/CMakeLists.txt
@@ -1,0 +1,27 @@
+# Copyright 2023 The Manifold Authors.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+# https://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# I feel like there has to be a better way - I hate using the CMake cache. 
+set(ASSIMP_BUILD_TESTS OFF CACHE INTERNAL "" FORCE)
+set(ASSIMP_WARNINGS_AS_ERRORS OFF CACHE INTERNAL "" FORCE)
+set(ASSIMP_INSTALL OFF CACHE INTERNAL "" FORCE)
+set(ASSIMP_BUILD_ASSIMP_VIEW OFF CACHE INTERNAL "" FORCE)
+set(ASSIMP_BUILD_ALL_EXPORTERS_BY_DEFAULT OFF CACHE INTERNAL "" FORCE)
+set(ASSIMP_BUILD_ALL_IMPORTERS_BY_DEFAULT OFF CACHE INTERNAL "" FORCE)
+set(ASSIMP_BUILD_GLTF_EXPORTER ON CACHE INTERNAL "" FORCE)
+set(ASSIMP_BUILD_GLTF_IMPORTER ON CACHE INTERNAL "" FORCE)
+
+add_subdirectory(assimp)
+
+target_compile_options(assimp PRIVATE -Wno-error -Wno-deprecated-declarations)


t-paul:
Is there a way of using / building against system libraries too? Most Linux distributions allow only very limited and specialized source dependencies on libraries. Recursively including a chain of libraries might make it difficult to keep applications like OpenSCAD as native distro builds.

elalish:
@t-paul Some advice here would be greatly appreciated! Do you know of a CMake way to set that up?

geoffder:
Not requiring the external dependency on assimp for other bindings will simplify things as well imo. I was going to say that I suppose this would remove the need for the versioning hack I added to meshIO in the past, but I suppose to maintain compatibility with t-paul's suggestion it will have to stay.

kintel:
Assimp comes with a CMake macro for this: https://github.com/assimp/assimp/blob/master/cmake-modules/Findassimp.cmake

..so using that should yields the Assimp offered by the environment.

Similarly for GoogleTest etc.

..and if following that strategy, the problem, is reduced to set up and environment, which should cleanly decouple the submodules from the rest of the library (i.e. make submodules optional, only needed for cases where it's tricky to build a suitable environment).

t-paul:
My `cmake` foo is not good enough for a future proof recommendation but there's always options to learn new stuff :grinning:. I had a similar topic with `lib3mf` which included `libz` and `zlib`, one of which is even forbidden by Debian `lintian` automated checks. In this case the way was adding flags like `USE_INCLUDED_ZLIB` but maybe there's better options in `cmake` for this.

I was eying a library package build we could use for the various official OpenSCAD builds but threw in the towel for now as `make install` only installed Clipper2. I don't know how much of a topic that is, but having distro packages of Manifold would be quite nice.

elalish:
@kintel That sounds like a very reasonable solution, but I'm not sure I understand it well enough to implement it in CMake. Is there any chance you could contribute here, or somehow spell out the details for those of us less well-versed?

kintel:
@elalish I can try to write up a draft PR.
The old saying that building software is as hard as writing software probably will come into play at some point :)

geoffder:
https://github.com/elalish/manifold/blob/a5f92e25227898bf16d6c787e793238869171437/test/manifoldc_test.cpp#L92
Woops, my bad! Changing this to an @elalish approved format should fix the CBIND CI.

kintel:
@elalish Perhaps you could help me getting Manifold to build in the meantime.
I'm on a stock Ubuntu 22.04:

* googletest is too old, fails due to the known issue `gtest-death-test.cc:1224:24: error: ‘dummy’ may be used uninitialized [-Werror=maybe-uninitialized]`. I updated to the latest version, and got rid of that issue.

* This fails to build
```
[1/1] Linking CXX shared module bindings/python/pymanifold.cpython-310-x86_64-linux-gnu.so
FAILED: bindings/python/pymanifold.cpython-310-x86_64-linux-gnu.so 
: && /usr/bin/c++ -fPIC   -flto -shared  -o bindings/python/pymanifold.cpython-310-x86_64-linux-gnu.so bindings/python/CMakeFiles/pymanifold.dir/pymanifold.cpp.o  src/manifold/libmanifold.a  src/third_party/clipper2/CPP/libClipper2.a  -lm && cd /home/kintel/code/manifold/build/bindings/python && /usr/bin/strip /home/kintel/code/manifold/build/bindings/python/pymanifold.cpython-310-x86_64-linux-gnu.so
/usr/bin/ld: src/manifold/libmanifold.a(csg_tree.cpp.o): warning: relocation against `_ZTVSt23_Sp_counted_ptr_inplaceIKN8manifold8Manifold4ImplESaIS2_ELN9__gnu_cxx12_Lock_policyE2EE' in read-only section `.text'
/usr/bin/ld: src/manifold/libmanifold.a(cross_section.cpp.o): relocation R_X86_64_PC32 against symbol `__libc_single_threaded@@GLIBC_2.32' can not be used when making a shared object; recompile with -fPIC
```

kintel:
I figured I could disable Python to make it build.

Here's a proof of concept for how I'd deal with dependencies:
https://github.com/elalish/manifold/pull/402

elalish:
I've seen that `-fPIC` thing come up before, seems to have something to do with building shared libs - this was an earlier related fix: https://github.com/elalish/manifold/pull/103/commits/4d84ebf1729154b10c61f8b9216e3650e8f5d071

pca006132:
@t-paul Just wondering which part of debian policy restricts using git submodules when there is a system dependency? I am not familiar with packaging (except for nixos, which I do know some basic stuff), so I am wondering how people make sure they can have reproducible builds on different distros so they don't have to debug various different library combinations...

elalish:
Thanks to @kintel's suggestion and the fact that homebrew now has an assimp package, I'm unblocked for testing with a much simpler fix: https://github.com/elalish/manifold/pull/394/commits/203a19dafb4d02086ba2fa1f822166d7d0f65500

I don't love doubling our CI compile times, so I'm going to close this one for now. I do tend to agree that linking to system packages is preferable to submodules: easier for downstream users to avoid duplication and compile time, but it also forces us to stay up to date. Yes, we'll have to deal with breaking changes here and there (see @geoffder's assimp versions fix), but we want pretty stable dependencies anyway and better to figure out those problems earlier than later. Especially once we move off Thrust, we won't have many dependencies to worry about.

If someone has an idea for how to install packages on our Windows CI, that would be very helpful in all this.

PR #404, opened 2023-04-06
Flip triangles in csg_tree

### LINKED ISSUE #403: mirror inverts triangles
Using the `.mirror()` function in ManifoldCAD (WASM/js) results in inverted triangles.

You can try it out using this sample code in the web editor:

```js
const generateProfile = (options = {}) => {
  const defaultOptions = {
    cutterThickness: 1.2,
    cutterHeight: 20,
    fillet: 5,
    sharpCutter: true
  }

  const settings = Object.assign({}, defaultOptions, options)

  let profile = [
    [0, 0],
    [settings.fillet, settings.fillet],
    [settings.cutterThickness, settings.fillet],
    [settings.cutterThickness, settings.sharpCutter ? settings.cutterHeight - settings.cutterThickness : settings.cutterHeight],
    [0, settings.cutterHeight]
  ]

  return profile
}

const result = compose([
  extrude(generateProfile(), 10),
  extrude(generateProfile(), 10).mirror([1, 0, 0]),
])
```

The mirrored extrusion is inverted:
image

pca006132:
Although we already have related logic in `impl.cpp`, the compose code will skip performing transformation on the original mesh, and apply transformation on the combined mesh directly, skipping the previous fix. The logic is now added to the compose function to fix the bug without much performance penalty. This also factors out some functions used in both `impl.cpp` and `csg_tree.cpp` (although we should probably give it a better name).

Fixes #403

### DIFF
--- a/src/manifold/src/csg_tree.cpp
+++ b/src/manifold/src/csg_tree.cpp
@@ -18,6 +18,7 @@
 
 #include "boolean3.h"
 #include "impl.h"
+#include "mesh_fixes.h"
 #include "par.h"
 
 namespace {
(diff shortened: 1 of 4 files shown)


pca006132:
@elalish is there a way to check if triangle orientations are correct? I think we should add this to our tests as this bug occurred at least twice.

elalish (approved):
Thanks! As for tests, the best check of triangle orientation is the sign of the volume. There's also our testing hook, `MatchesTriNormals()`, but that's more of an internal consistency check.

PR #406, opened 2023-04-11
Fix triangulator assertion

### LINKED ISSUE #395: Manifold - PerfTest (Debug) : List Iterators Incompatible error (Windows)
## This will happen running PerfTest in debug on windows.

perfTest_mblGMevvG0

devenv_EccVui8L7U

devenv_Nu7DlJ6hWs

devenv_OZdHjgqVtd

see polygon.cpp line 186
see polygon.cpp line 950

--------------------------------------

Apparently the west/east iterator from `vert`  is not coming from the same list it tries to compare to. 

https://stackoverflow.com/questions/31217529/c-list-iterators-incompatible

on Release the STL assert won't happen.

elalish:
Fixes #395 

Indeed, the assertion was just wrong - a one-line fix. I don't have a Windows machine to test on, but I'm confident this will fix it.


elalish:
Thanks for the patch @pca006132, I used that to repro the issue and test if this fixes it, which it does.

PR #408, opened 2023-04-12
added global coplanarity check

### LINKED ISSUE #388: Corrupted output for seemingly simple large models
Manifold seems to explode with large inputs that *seem* valid.

```js
$fn=1500;
sphere(100);
translate([50, 0, 0]) sphere(100);
```

image

Here's a repro case:

```bash
echo 'sphere(100, $fn=1500);' | openscad - -o a.stl --export-format=binstl
echo 'translate([50, 0, 0]) sphere(100, $fn=1500);' | openscad - -o b.stl --export-format=binstl
```

```c++
#include "manifold.h"
#include "meshIO.h"

using namespace manifold;

int main(int argc, char **argv) {
  auto forceCleanup = false; // true gives same output
  Manifold a = Manifold(ImportMesh("a.stl", forceCleanup));
  Manifold b = Manifold(ImportMesh("b.stl", forceCleanup));
  ExportMesh("out.stl", (a + b).GetMesh(), {});
}
```

elalish:
Fixes #388 

Planar faces are determined by testing neighboring triangles for coplanarity and then doing connected components on the result. The problem is little changes can add up: a finely tessellated sphere can potentially have every edge be coplanar within tolerance, resulting in a single huge planar face, which is obviously incorrect. This affects output `faceID`, but also the internal edge collapsing and can result in wildly incorrect imports. 

The fix is to do a second pass, since we already identify the largest triangle as the reference for a plane, we can simply check if any verts in the plane are out of that triangle's plane. If so, we simply discard the entire plane and all contained triangles are assumed to be not coplanar with each other. There are a few edge cases where this will miss actual planar faces, but I think they'll be rare, and it's much less of a problem geometrically to be conservative in this direction.

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -120,6 +120,7 @@
   "editor.detectIndentation": false,
   "cSpell.words": [
     "Bary",
+    "Coplanarity",
     "csaszar",
     "Gyroid",
     "halfedge",
--- a/src/manifold/src/impl.cpp
+++ b/src/manifold/src/impl.cpp
@@ -284,7 +284,7 @@ struct CheckCoplanarity {
   __host__ __device__ void operator()(int tri) {
     const int component = components[tri];
     const int referenceTri = comp2tri[component];
-    if (referenceTri == tri) return;
+    if (referenceTri < 0 || referenceTri == tri) return;
 
     const glm::vec3 origin = vertPos[halfedge[3 * referenceTri].startVert];
     const glm::vec3 normal = glm::normalize(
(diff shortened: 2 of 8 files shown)


elalish:
I also added a test that catches this problem, and in the process noticed that I wasn't passing `precision` in and out of the library properly yet, so that's fixed here. I also noticed that my cute idea of clamping our surface area calculation was resulting in quite a bit of error, so that's simplified and fixed too.

pca006132 (approved):

PR #419, opened 2023-04-21
Fix bug in `three.html` example.

NodeGuy:
The bug didn't change the output but made it harder for newcomers to learn the JavaScript API.


elalish (approved):
Thanks!

PR #414, opened 2023-04-19
Manifold extension for glTF

elalish:
This is the beginning of a JS glTF I/O library for Manifold, allowing lossless round-trip of multi-material models by defining a new glTF extension (which I plan to write up and submit to the Khronos).

PR #420, opened 2023-04-25
EXT_manifold for glTF

elalish:
Deals with most of #422

Replaces #414, avoiding breaking changes to `Mesh`.

This is the beginning of a JS glTF I/O library for Manifold, allowing lossless round-trip of multi-material models by defining a new glTF extension (which I plan to write up and submit to Khronos).


bindings/wasm/gltfIO.js
```diff
@@ -0,0 +1,208 @@
...
+                      .setType(Accessor.Type.SCALAR)
+                      .setArray(manifoldMesh.triVerts);
+
+  const mesh = doc.createMesh();
+  const numPrimitive = manifoldMesh.runIndex.length - 1;
+  for (let run = 0; run < numPrimitive; ++run) {
+    const primitive = doc.createPrimitive().setIndices(indices);
+    const material = materials[run];
+    if (material) {
+      primitive.setMaterial(material);
```
elalish, line 120: @donmccurdy I'm loving your library! However, I'm getting `Error: Cannot connect disconnected graphs.` at this line. I can't quite tell from your docs how I should go about moving a `Material` (along with the textures it references) from a file I read to a file I'm writing. Is there a reason it can't attach them to the requested doc automatically? I was looking for something like `Root.attach()`, but came up empty.

donmccurdy: There isn't currently a feature that transfers a resource on its own or with its dependencies, though that's an interesting idea. Currently the options would be:

(a) Merge documents, then clean up any unwanted resources. The prune() transform can help with getting rid of unused resources.

```js
documentA.merge(documentB);
documentA.getRoot().listScenes()[1].dispose();

// ... assign materials as needed ...

await documentA.transform(prune());
```

(b) Create a new material in the target document, then deep copy the original.

```js
document.createMaterial().copy(material, (ref) => ref.clone());
```

donmccurdy: /cc

- https://github.com/donmccurdy/glTF-Transform/issues/924

elalish: Thanks, that'll unblock me. Looking forward to the new feature though 😄

bindings/wasm/examples/model-viewer.html
```diff
@@ -110,30 +134,30 @@
...
+    disposeMesh(node.getMesh());
+    const materials = [];
+    for (const id of manifoldMesh.runOriginalID) {
+      materials.push(materialMap.get(id));
     }
+    const mesh = writeMesh(doc, manifoldMesh, attributes, materials);
+    node.setMesh(mesh);
+    doc.getRoot().listNodes().forEach((n) => console.log(n.listParents()));
+    doc.transform(prune());
+    doc.getRoot().listNodes().forEach((n) => console.log(n.listParents()));
```
elalish, line 158: @donmccurdy Here's the spot where I can see on the command line that two of the three nodes contain only the root as a parent, yet `prune()` doesn't remove them. Any thoughts?

elalish: Argh, thanks - however interestingly I get exactly the same results. After prune, I have nodes whose only parent is the Root, yet they're not marked as disposed (and prune still says it found nothing to dispose of).

donmccurdy: Hmm that's a surprise! By default it should remove unused nodes, and also empty leaf nodes. Would it be possible to share the .glb in which prune is failing?

bindings/wasm/gltfIO.js
donmccurdy: I'm still not sure what's going on here sorry, but I did publish an uncompressed version (no other relevant changes, just not minified) as v4.0.0-alpha.0 if you want to test from that:

```
npm install --save @gltf-transform/core@next @gltf-transform/extensions@next @gltf-transform/functions@next
```

bindings/wasm/examples/model-viewer.html
```diff
@@ -45,59 +45,83 @@
...
+  const wasm = await Module();
+  wasm.setup();
+
+  const attributes = ['POSITION', 'TEXCOORD_0'];
+  const materialMap = new Map();
+
+  async function readGLB(url, attributes, materialMap) {
+    const manifolds = [];
+    const docIn = await io.read(url);
+    docIn.transform(flatten());
```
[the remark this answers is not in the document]
donmccurdy: NVM, outdated comment.

bindings/wasm/gltfIO.js
elalish: Strange, when I update my import map to `https://cdn.jsdelivr.net/npm/@gltf-transform/functions@v4.0.0-alpha.0/+esm` I still get a minified version (maybe something jsdelivr is doing?). When I point directly to your file `https://cdn.jsdelivr.net/npm/@gltf-transform/functions@v4.0.0-alpha.0/dist/functions.modern.js` it doesn't seem to work as a module, since I get `caught TypeError: Failed to resolve module specifier "property-graph". Relative references must start with either "/", "./", or "../".`

Since I can't debug it locally, I'll work on getting a raw.githack working so you can see my repro.

donmccurdy: Hm, yeah it sounds like jsdelivr must be rewriting the file to resolve imports, and also minifying... that does complicate things. I could try running this PR locally if you could share steps for reproducing the issue here.

elalish: Perfect, that fixed it! Actually `@3.x` had the same problem, but `@3.2.1` works fine. Curios if that will break again the next time a new version is released? Or was that just something weird about v3.1.2? 

I suppose vitejs is an option, but the point of these examples is to be simple. It's kind of nice if all anyone needs is an import map. One question though: why do I need `/+esm` when your default file appears to be an ES6 module already? It'd be nice to not have to rely on those CDN transformations.

elalish: Perhaps a piece of the puzzle: when I try `https://cdn.jsdelivr.net/npm/@gltf-transform/core@3.2.1` I get `Failed to load module script: Expected a JavaScript module script but the server responded with a MIME type of "application/node". Strict MIME type checking is enforced for module scripts per HTML spec.` - I wonder if that's a bug with jsDelivr or if the extensions need to be `.mjs` or something?

donmccurdy: > Curious if that will break again the next time a new version is released?

Unfortunately, yes. The CDN is dumber than a bundler, it doesn't know what version(s) are already on the page, so it just resolves the newest compatible version of /core.

> Expected a JavaScript module script but the server responded with a MIME type of "application/node" ...

It looks like JSDelivr will serve the (`.cjs`) "main" entry by default, and the `/+esm`  tag is required to request something else. Maybe this new esm.run thing by JSDeliver would be the way to go? It isn't something I've used before, though:

https://www.jsdelivr.com/esm

elalish: Ah, interesting. Well this works, and I think should be reasonably future-proof since I'm guessing you at least restrict by major version:
```
    "imports": {
      "@gltf-transform/core": "https://cdn.jsdelivr.net/npm/@gltf-transform/core@3/+esm",
      "@gltf-transform/functions": "https://cdn.jsdelivr.net/npm/@gltf-transform/functions@3/+esm"
    }
```

elalish: Ugh, I guess I should just switch to TypeScript. Probably a good idea anyway, especially for manifoldCAD.org. I just need to bite the bullet.

donmccurdy: I do restrict to same major version, not sure why `@3/+esm` doesn't work. ☹️

elalish: Ah, I get it now - the `/+esm` thing is adding exact versions to your import statements and allowing them to automatically resolve. The reason `/dist/core.modern.js` isn't working is that I need to add all of your imports to my import map. I'm beginning to realize why bundling and tree-shaking is so necessary.

elalish:
@pca006132 I'm getting a strange build error on the nix JS that I don't understand. I'm disabling it for now, but would you mind taking a look at it?

pca006132:
Sure I will have a look at it

bindings/wasm/README.md
```diff
@@ -20,7 +20,24 @@ The most significant contribution here is a guaranteed-manifold [mesh Boolean](h
...
+First, follow the directions in the root README to get your C++ build environment set up and working for WASM. From this directory (`bindings/wasm/`) you can test the JS bindings by running:
+
+```
+npm install
+npm test
+```
+
+You can also test the manifoldCAD.org editor as well as our other example pages by serving from `bindings/wasm/examples/` with e.g. `npx http-server`.
+
+Note that the `emcmake` command automatically copies your WASM build into `examples/built/` - these are checked into our repo in order to make sharing repro cases much easier. Note that you can test manifoldCAD.org on anyone's branch by simply going to: `https://raw.githack.com/<user>/manifold/<branch>/bindings/wasm/examples/index.html` e.g. https://raw.githack.com/elalish/manifold/glTFextension/bindings/wasm/examples/index.html
```
elalish, line 36: This is inspired by three.js and I think will make working on the WASM a lot easier.

bindings/wasm/examples/manifold-gltf.js
```diff
@@ -0,0 +1,176 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+import {Extension, ExtensionProperty, PropertyType} from '@gltf-transform/core';
+
+const NAME = 'EXT_manifold';
+
+export class ManifoldPrimitive extends ExtensionProperty {
```
elalish, line 19: This file is designed to make it easy to read and write the EXT_manifold glTF extension regardless of whether someone is using Manifold.

bindings/wasm/examples/gltf-io.js
```diff
@@ -0,0 +1,208 @@
...
+    numProp,
+    triVerts,
+    vertProperties,
+    runIndex,
+    mergeFromVert,
+    mergeToVert
+  };
+}
+
+export function writeMesh(doc, manifoldMesh, attributes, materials) {
```
elalish, line 102: This file is designed for interop specifically between Manifold and glTF. It doesn't actually have Manifold as a dependency because it can just use generic objects that look like a `Mesh`, but it's still based on our API.

bindings/wasm/examples/model-viewer.html
```diff
@@ -45,59 +45,71 @@
...
-  const spaceTexture = doc.createTexture('space');
-  const moonMaterial = doc.createMaterial('moon').setBaseColorTexture(moonTexture);
-  const spaceMaterial = doc.createMaterial('moon').setBaseColorTexture(spaceTexture);
-  const moonPrimitive = doc.createPrimitive().setMaterial(moonMaterial)
-    .setIndices(moonIndices).setAttribute('POSITION', position).setAttribute('TEXCOORD_0', uv);
-  const spacePrimitive = doc.createPrimitive().setMaterial(spaceMaterial)
-    .setIndices(spaceIndices).setAttribute('POSITION', position).setAttribute('TEXCOORD_0', uv);
-  const mesh = doc.createMesh('result').addPrimitive(moonPrimitive).addPrimitive(spacePrimitive);
-  const node = doc.createNode().setMesh(mesh);
-  const scene = doc.createScene().addChild(node);
```
elalish, line 73: Much simpler example now that we can read and write files!

pca006132:
It seems that the bug is this: https://github.com/emscripten-core/emscripten/issues/19052 which is fixed in a new release of emscripten.  I am waiting for https://github.com/NixOS/nixpkgs/pull/229718 to land, and then we can just update the flakes (did not update it for so long!).

elalish:
Excellent, thank you!

PR #428, opened 2023-05-09
make-manifold

### LINKED ISSUE #422: Build a manifold extension to glTF
Considering that Manifold handles arbitrary mesh properties like normals and UVs, it would be nice to be able to round-trip this type of material data through a file format losslessly. This is not generally possible, as even indexed files like OBJ or glTF must be exported as non-manifold when they contain multiple materials (or even sharp corners) as GPUs require the vertex positions to be duplicated for any vertex properties that don't match. Relying on geometric checks (verts having identical positions) to reassemble the manifold will often work, but is unreliable in general. 

It appears that a glTF extension is possible to make that will encode this manifoldness data, and at very little size cost, by using their concept of sparse accessors. I'll write up this extension and propose it to Khronos (since I work with them for my day job anyway), but as far as Manifold is concerned, the work will be to make a reader and writer of this extension, proving that it works. 

I think it'll also be a good opportunity to show off #394 by making a small web tool to upload a GLB and download the manifold version of it (or show that we can't fix it automatically), which will hopefully help with problems like https://github.com/elalish/manifold/issues/91#issuecomment-1429030070

elalish:
Fixes #422

This introduces a tool to upload a GLB and download a manifold GLB with the new extension if it's possible to make the mesh water-tight. I'm running into some OOM problems with certain models, so there are probably some bugs to work out before this can be merged.

### DIFF
--- a/bindings/wasm/CMakeLists.txt
+++ b/bindings/wasm/CMakeLists.txt
@@ -20,7 +20,7 @@ add_executable(manifoldjs bindings.cpp)
 set_source_files_properties(bindings.cpp OBJECT_DEPENDS
   ${CMAKE_CURRENT_SOURCE_DIR}/bindings.js)
 target_link_libraries(manifoldjs manifold sdf cross_section polygon)
-target_compile_options(manifoldjs PRIVATE ${MANIFOLD_FLAGS} -fexceptions)
+target_compile_options(manifoldjs PRIVATE ${MANIFOLD_FLAGS})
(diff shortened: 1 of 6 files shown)


elalish:
Okay, there are definitely still some bugs, but this little test pages works reasonable well now. It'll be public at https://manifoldcad.org/make-manifold.html when this is merged. Don't get your hopes up though, most glTFs, even if they look fairly solid, are not actually manifold, nor even close enough for us to merge them automatically. 

The worst bug I see now is some complex scenes like the Khronos Buggy and Dragon throw an OOM somewhere in the WASM. I'm guessing this is an error in the `Merge` function, but we can deal with that in a further PR.

PR #429, opened 2023-05-10
add unionError test case

Loosetooth:
See [the discussion](https://github.com/elalish/manifold/discussions/423) for more details.

The test case is currently a bit convoluted. I'm open to simplifying it, but have not managed to find a simpler error case in manifoldcad.org, or C++.


elalish:
Would you mind signing our CLA? It's pretty innocuous, but sadly required.

Loosetooth:
It's signed.

Some ideas for simplifying the test case:

1. try removing path points and see if it still breaks.
2. move the path vector to another file, and import it in the test file. (Not sure how easy that is in C++)

If you have any other ideas let me know.

elalish (approved):
Thanks! I'll tweak it from here while I debug.

PR #436, opened 2023-05-20
Only link MeshIO when MANIFOLD_EXPORT is on

geoffder:
This makes linking MeshIO optional for the C bindings, rather than forcing the flag on.


geoffder:
Ah woops I missed the stuff in `conv.cpp`.

pca006132 (approved):

PR #438, opened 2023-05-22
Editor scene

### LINKED ISSUE #431: ManifoldSet API?
From #425, I wonder if the best solution would be to provide a wrapper class for a group of manifolds. Every operation and transformation will be applied to each of the element inside the set, and at the end the user can get each of the parts out.
Everything can be done now, but this can probably make it a bit easier for users. And perhaps make a simple function that put each mesh in a way such that the bounding boxes are not overlapping, and can be imported into a 3D printer for example.

Example:
```js
const parts = ManifoldSet([
    cube([1, 2, 3]),
    cube([4, 5, 6])
])
parts = parts.subtract(cylinder(10, 1))
export(parts.sorted())
```

elalish:
Fixes #431 

This adds `GLTFNode` and `GLTFMaterial` to ManifoldCAD.org, which gives us some pretty huge new capabilities. We can now easily create entire scene hierarchies and export them to GLB, all while avoiding duplication of the underlying mesh data if you reference a manifold several times. You can change material properties on any node - they act like CSS where it applies to everything underneath, but is overridden by anything specified lower down. Since specifying parents is only possible during node construction/cloning I believe it's impossible to construct a scene that's not a tree. 

It started to get complex enough that I took @donmccurdy's advice and switched our editor (and examples) over to Vite. It's a pretty nice build system and we don't have to worry about import maps anymore. It also makes mixing JS and TS very easy, so I took the liberty of switching our worker over to TS, which was very helpful in development. I'll probably transition the rest of the editor to TS soon. 

I was hoping you could test this version of ManifoldCAD.org before it's merged here: https://raw.githack.com/elalish/manifold/editorScene/bindings/wasm/examples/dist/index.html, but it looks like there's still some path problems.


.github/workflows/manifold.yml
```diff
@@ -127,19 +127,17 @@ jobs:
...
       uses: actions/upload-artifact@v3
       with:
         name: wasm
-        path: |
-          build/bindings/wasm/manifold.js
-          build/bindings/wasm/manifold.wasm
-          build/bindings/wasm/manifold.d.ts
-          build/bindings/wasm/manifold-global-types.d.ts
-          build/bindings/wasm/manifold-encapsulated-types.d.ts
+        path: bindings/wasm/examples/dist/
```
elalish, line 140: There's now more to deploy, but it's simpler.

bindings/wasm/examples/editor.js
```diff
@@ -12,12 +12,14 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-import {examples} from './examples.js';
-const exampleFunctions = examples.functionBodies;
+import ManifoldWorker from './worker?worker';
```
elalish, line 15: This is an interesting new Vite syntax.

bindings/wasm/examples/gltf-io.js
```diff
@@ -144,6 +155,28 @@ export function writeMesh(doc, manifoldMesh, attributes, materials) {
   const buffer = doc.getRoot().listBuffers()[0];
   const manifoldExtension = doc.createExtension(EXTManifold);
 
+  const attributeUnion = [];
+  for (const matAttributes of attributes) {
+    matAttributes.forEach((attribute, i) => {
```
elalish, line 160: The input `attributes` is now a `string[][]` instead of `string[]`, since different materials can have different attributes.

bindings/wasm/examples/package.json
```diff
@@ -0,0 +1,30 @@
+{
+  "name": "vite-starter",
```
elalish, line 2: Should change this.

bindings/wasm/examples/public/examples.js
```diff
@@ -38,6 +39,11 @@ export const examples = {
       // See the script drop-down above ("Intro") for usage examples. The
       // gl-matrix package from npm is automatically imported for convenience -
       // its API is available in the top-level glMatrix object.
+
+      // Use GLTFNode for disjoint manifolds rather than compose(), as this will
+      // keep them better organized in the GLB. This will also allow you to
+      // specify material properties, and even vertex colors via
+      // setProperties(). See Tetrahedron Puzzle example.
```
elalish, line 46: Here's the interesting part!

bindings/wasm/examples/public/examples.js
```diff
@@ -57,17 +63,43 @@ export const examples = {
...
+      // overridden by any materials defined lower down. Our default material:
+      // {
+      //   roughness = 0.2,
+      //   metallic = 1,
+      //   baseColorFactor = [1, 1, 0],
+      //   alpha = 1,
+      //   unlit = false,
+      //   name = ''
+      // }
+      const node = new GLTFNode();
```
elalish, line 101: This is a good example of what's possible with this new API, plus it's pretty.

bindings/wasm/examples/public/examples.js
```diff
@@ -409,10 +459,30 @@ export const examples = {
...
+        for (let i = 0; i < m; ++i) {
+          for (let j = i; j < m; ++j) {
+            for (let k = j; k < m; ++k) {
+              const node = new GLTFNode();
+              node.manifold = gyroidModule;
+              node.translation =
+                  [(k + i - j) * size, (k - i) * size, (-j) * size];
+              node.material = {
+                baseColorFactor:
+                    [(k + i - j + 1) / m, (k - i + 1) / m, (j + 1) / m]
```
elalish, line 476: This shows how you can reuse the same manifold many times, even with different materials, but just a single one gets packed in the GLB and referenced appropriately.

bindings/wasm/examples/worker.test.js
```diff
@@ -0,0 +1,152 @@
...
+        const docIn = await io.read(objectURL);
+        const nodes = docIn.getRoot().listNodes();
+        for (const node of nodes) {
+          const mesh = node.getMesh();
+          if (!mesh) {
+            continue;
+          }
+          const attributes = [];
+          const materials = [];
+          const manifoldMesh = readMesh(mesh, attributes, materials);
```
elalish, line 72: Our JS testing is now giving us dramatically more coverage - we're using the worker directly instead of duplicating its functionality, which means we're also round-tripping every test through write-GLB and read-GLB now.

src/manifold/include/manifold.h
```diff
@@ -207,7 +207,7 @@ class Manifold {
   Manifold Mirror(glm::vec3) const;
   Manifold Warp(std::function<void(glm::vec3&)>) const;
   Manifold SetProperties(
-      int, std::function<void(glm::vec3, const float*, float*)>) const;
+      int, std::function<void(float*, glm::vec3, const float*)>) const;
```
elalish, line 210: I switched the order here because I forgot to follow my own style of putting output arguments before input arguments.

geoffder:
Ah nice, no more of that exporting mess in `test.js`, took me a moment to figure out why the constructors/functions I was adding weren't showing up for use in `examples.js` to be honest. I suppose I'll try to rebase what I have so far onto this branch the next time I work on the bindings and see where I'm at. Still gradually figuring out the JS stuff as I work on the new bindings, but I think I'll be at the point of adding examples for testing and working out the kinks soon.

geoffder:
@elalish I've rebased the WIP bindings onto this branch, but now I'm not sure what the invocation is for running the tests (should I be using the worker one, I noticed that `test/test.js` actually still exists, but running `npm test` from that directory or `npm test/test` doesn't work). It's probably obvious, but I've got about zero familiarity with js.

elalish:
Yeah, sorry I need to update the README. Basically everything is run from the `examples` directory now, since that package.json has all the deps and such. Just `npm install`, `npm run build`, `npm test`.

Issue #328, opened 2023-02-18
ManifoldCAD doesn't work on Firefox before 111


pca006132:
https://bugzilla.mozilla.org/show_bug.cgi?id=1247687

The worker `type: 'module'` is only recently implemented (111 is not yet released), not sure if we want this on the website as some users may still use older versions of firefox in the near future.

elalish:
Wow, that bug was opened 7 years ago. Man, when Safari implements something first, you know you've really gotten behind.

I haven't really been worrying too much about cross-browser testing, but it's not a bad idea. Still, I'm not sure how ugly the work around for this will be. I'm tempted to just not worry about it - maybe put up a message if their browser is an unsupported version. What do you think?

pca006132:
Yeah I don't usually worry about cross-browser testing either (I used to even test IE, but that was years ago), but the problem is that my main driver is firefox :P

Why do we have to use modules though? I think if they allowed this bug to stay opened for so long, this should be something that can be worked around easily?

elalish:
Yeah, if it's easy to work around, let's do that. But after that ES6 module update, I really like modules a lot better - the code is nice and clean.

elalish:
Huh, I just tried this with Firefox 111 and it's still failing with `import declarations may only appear at top level of a module` in `worker.js`. Maybe they didn't fix it completely?

elalish:
Excellent, I just tried it and it does indeed work!

pca006132:
I guess we can consider this as won't fix.

PR #440, opened 2023-05-25
Update JS bindings with CrossSection + reorganize

### LINKED ISSUE #360: Add JS bindings for CrossSection
There's also a few other APIs missing, like `Split` and `Trim`.

geoffder:
Fixes #360

- add CrossSection class
- move static method (constructors) under their respective classes
- add Manifold.{split, splitByPlane}
- enable usage of `instanceof` on `Manifold` (and now `CrossSection`) objects
- simplify registry indirection in `worker.ts`, enabling method de-structuring (opening static function names out of class module name-spaces)

Build changes:
- break C++ helpers out from `bindings.cpp` to `helpers.cpp`
- ensure that `ts` files are copied during builds even when nothing is recompiled

Note that in addition to the new bindings, the most of the existing toplevel functions have been moved under their respective classes as static methods (breaking change!). I haven't added any new examples yet, but the existing ones are passing and do make use of `CrossSection` now (some transparently, and some directly), but I think enough of this is together that review is worthwhile.

### DIFF
--- a/bindings/wasm/CMakeLists.txt
+++ b/bindings/wasm/CMakeLists.txt
@@ -16,14 +16,7 @@ project(wasm)
 
 add_executable(manifoldjs bindings.cpp)
 
-set(JS_DEPS
-        ${CMAKE_CURRENT_SOURCE_DIR}/bindings.js
-        ${CMAKE_CURRENT_SOURCE_DIR}/manifold.d.ts
-        ${CMAKE_CURRENT_SOURCE_DIR}/manifold-encapsulated-types.d.ts)
-# FIXME: can we rewrite the custom command to run after these are modified when building
-# without recompiling bindings.cpp?
-# make sure that we recompile the wasm when bindings.js is being modified
-set_source_files_properties(bindings.cpp PROPERTIES OBJECT_DEPENDS ${JS_DEPS})
+set_source_files_properties(bindings.cpp PROPERTIES OBJECT_DEPENDS ${CMAKE_CURRENT_SOURCE_DIR}/bindings.js)
 target_link_libraries(manifoldjs manifold sdf cross_section polygon)
 target_compile_options(manifoldjs PRIVATE ${MANIFOLD_FLAGS})
 target_link_options(manifoldjs PUBLIC --pre-js ${CMAKE_CURRENT_SOURCE_DIR}/bindings.js --bind -sALLOW_TABLE_GROWTH=1
@@ -34,15 +27,20 @@ set_target_properties(manifoldjs PROPERTIES OUTPUT_NAME "manifold")
 
 file(MAKE_DIRECTORY ${CMAKE_CURRENT_SOURCE_DIR}/examples/built)
 
-add_custom_command(
-        TARGET manifoldjs POST_BUILD
-        COMMAND ${CMAKE_COMMAND} -E copy
-                ${CMAKE_CURRENT_BINARY_DIR}/manifold.*
-                ${CMAKE_CURRENT_SOURCE_DIR}/manifold*.d.ts
-                ${CMAKE_CURRENT_SOURCE_DIR}/examples/built/)
+# ensure that interface files are copied over when modified
+add_custom_target(js_deps ALL
+        DEPENDS ${CMAKE_CURRENT_BINARY_DIR}/manifold.*
+               ${CMAKE_CURRENT_SOURCE_DIR}/manifold*.d.ts)
 
 add_custom_command(
-        TARGET manifoldjs POST_BUILD
+        TARGET js_deps POST_BUILD
         COMMAND ${CMAKE_COMMAND} -E copy
-                ${CMAKE_CURRENT_SOURCE_DIR}/manifold*.d.ts
-                ${CMAKE_CURRENT_SOURCE_DIR}/examples/public/)
+               ${CMAKE_CURRENT_BINARY_DIR}/manifold.*
+               ${CMAKE_CURRENT_SOURCE_DIR}/manifold*.d.ts
+               ${CMAKE_CURRENT_SOURCE_DIR}/examples/built/)
+
+add_custom_command(
+        TARGET js_deps POST_BUILD
+        COMMAND ${CMAKE_COMMAND} -E copy
+               ${CMAKE_CURRENT_SOURCE_DIR}/manifold*.d.ts
+               ${CMAKE_CURRENT_SOURCE_DIR}/examples/public/)
--- a/bindings/wasm/CMakeLists.txt
+++ b/bindings/wasm/CMakeLists.txt
@@ -16,13 +16,14 @@ project(wasm)
 
 add_executable(manifoldjs bindings.cpp)
 
+set(JS_DEPS
+        ${CMAKE_CURRENT_SOURCE_DIR}/bindings.js
+        ${CMAKE_CURRENT_SOURCE_DIR}/manifold.d.ts
+        ${CMAKE_CURRENT_SOURCE_DIR}/manifold-encapsulated-types.d.ts)
+# FIXME: can we rewrite the custom command to run after these are modified when building
+# without recompiling bindings.cpp?
 # make sure that we recompile the wasm when bindings.js is being modified
-set_source_files_properties(bindings.cpp
-  OBJECT_DEPENDS ${CMAKE_CURRENT_SOURCE_DIR}/bindings.js
-  # FIXME: can we rewrite the custom command to run after these are modified when building
-  # without recompiling bindings.cpp?
-  OBJECT_DEPENDS ${CMAKE_CURRENT_SOURCE_DIR}/manifold.d.ts
-  OBJECT_DEPENDS ${CMAKE_CURRENT_SOURCE_DIR}/manifold-encapsulated-types.d.ts)
+set_source_files_properties(bindings.cpp PROPERTIES OBJECT_DEPENDS ${JS_DEPS})
 target_link_libraries(manifoldjs manifold sdf cross_section polygon)
 target_compile_options(manifoldjs PRIVATE ${MANIFOLD_FLAGS})
 target_link_options(manifoldjs PUBLIC --pre-js ${CMAKE_CURRENT_SOURCE_DIR}/bindings.js --bind -sALLOW_TABLE_GROWTH=1
--- a/bindings/wasm/CMakeLists.txt
+++ b/bindings/wasm/CMakeLists.txt
@@ -17,8 +17,12 @@ project(wasm)
 add_executable(manifoldjs bindings.cpp)
 
 # make sure that we recompile the wasm when bindings.js is being modified
-set_source_files_properties(bindings.cpp OBJECT_DEPENDS
-  ${CMAKE_CURRENT_SOURCE_DIR}/bindings.js)
+set_source_files_properties(bindings.cpp
+  OBJECT_DEPENDS ${CMAKE_CURRENT_SOURCE_DIR}/bindings.js
+  # FIXME: can we rewrite the custom command to run after these are modified when building
+  # without recompiling bindings.cpp?
+  OBJECT_DEPENDS ${CMAKE_CURRENT_SOURCE_DIR}/manifold.d.ts
+  OBJECT_DEPENDS ${CMAKE_CURRENT_SOURCE_DIR}/manifold-encapsulated-types.d.ts)
 target_link_libraries(manifoldjs manifold sdf cross_section polygon)
 target_compile_options(manifoldjs PRIVATE ${MANIFOLD_FLAGS})
 target_link_options(manifoldjs PUBLIC --pre-js ${CMAKE_CURRENT_SOURCE_DIR}/bindings.js --bind -sALLOW_TABLE_GROWTH=1
--- a/bindings/wasm/bindings.cpp
+++ b/bindings/wasm/bindings.cpp
@@ -104,8 +104,7 @@ EMSCRIPTEN_BINDINGS(whatever) {
   register_vector<glm::vec4>("Vector_vec4");
 
   class_<CrossSection>("CrossSection")
-      .constructor<std::vector<std::vector<glm::vec2>>,
-                   CrossSection::FillRule>()
+      .constructor(&cross_js::OfPolygons)
       .function("_add", &cross_js::Union)
       .function("_subtract", &cross_js::Difference)
       .function("_intersect", &cross_js::Intersection)
@@ -122,7 +121,7 @@ EMSCRIPTEN_BINDINGS(whatever) {
       .function("numContour", &CrossSection::NumContour)
       .function("_Bounds", &CrossSection::Bounds)
       .function("simplify", &CrossSection::Simplify)
-      .function("_Offset", &CrossSection::Offset)
+      .function("_Offset", &cross_js::Offset)
       .function("_RectClip", &CrossSection::RectClip)
       .function("_ToPolygons", &CrossSection::ToPolygons);
 
--- a/bindings/wasm/bindings.cpp
+++ b/bindings/wasm/bindings.cpp
@@ -14,178 +14,17 @@
 
 #include <emscripten/bind.h>
 #include <emscripten/val.h>
-
-#include <vector>
-
-using namespace emscripten;
-
 #include <manifold.h>
 #include <polygon.h>
 #include <sdf.h>
 
+#include <vector>
+
+#include "cross_section.h"
+#include "helpers.cpp"
+
+using namespace emscripten;
 using namespace manifold;
-
-Manifold Union(const Manifold& a, const Manifold& b) { return a + b; }
-
-Manifold Difference(const Manifold& a, const Manifold& b) { return a - b; }
-
-Manifold Intersection(const Manifold& a, const Manifold& b) { return a ^ b; }
-
-Manifold UnionN(const std::vector<Manifold>& manifolds) {
-  return Manifold::BatchBoolean(manifolds, OpType::Add);
-}
-
-Manifold DifferenceN(const std::vector<Manifold>& manifolds) {
-  return Manifold::BatchBoolean(manifolds, OpType::Subtract);
-}
-
-Manifold IntersectionN(const std::vector<Manifold>& manifolds) {
-  return Manifold::BatchBoolean(manifolds, OpType::Intersect);
-}
-
-std::vector<SimplePolygon> ToPolygon(
-    std::vector<std::vector<glm::vec2>>& polygons) {
-  std::vector<SimplePolygon> simplePolygons(polygons.size());
-  for (int i = 0; i < polygons.size(); i++) {
-    std::vector<glm::vec2> vertices(polygons[i].size());
-    for (int j = 0; j < polygons[i].size(); j++) {
-      vertices[j] = polygons[i][j];
-    }
-    simplePolygons[i] = {vertices};
-  }
-  return simplePolygons;
-}
-
-val MeshGL2JS(const MeshGL& mesh) {
-  val meshJS = val::object();
-
-  meshJS.set("numProp", mesh.numProp);
-  meshJS.set("triVerts",
-             val(typed_memory_view(mesh.triVerts.size(), mesh.triVerts.data()))
-                 .call<val>("slice"));
-  meshJS.set("vertProperties",
-             val(typed_memory_view(mesh.vertProperties.size(),
-                                   mesh.vertProperties.data()))
-                 .call<val>("slice"));
-  meshJS.set("mergeFromVert", val(typed_memory_view(mesh.mergeFromVert.size(),
-                                                    mesh.mergeFromVert.data()))
-                                  .call<val>("slice"));
-  meshJS.set("mergeToVert", val(typed_memory_view(mesh.mergeToVert.size(),
-                                                  mesh.mergeToVert.data()))
-                                .call<val>("slice"));
-  meshJS.set("runIndex",
-             val(typed_memory_view(mesh.runIndex.size(), mesh.runIndex.data()))
-                 .call<val>("slice"));
-  meshJS.set("runOriginalID", val(typed_memory_view(mesh.runOriginalID.size(),
-                                                    mesh.runOriginalID.data()))
-                                  .call<val>("slice"));
-  meshJS.set("faceID",
-             val(typed_memory_view(mesh.faceID.size(), mesh.faceID.data()))
-                 .call<val>("slice"));
-  meshJS.set("halfedgeTangent",
-             val(typed_memory_view(mesh.halfedgeTangent.size(),
-                                   mesh.halfedgeTangent.data()))
-                 .call<val>("slice"));
-  meshJS.set("runTransform", val(typed_memory_view(mesh.runTransform.size(),
-                                                   mesh.runTransform.data()))
-                                 .call<val>("slice"));
-
-  return meshJS;
-}
-
-MeshGL MeshJS2GL(const val& mesh) {
-  MeshGL out;
-  out.numProp = mesh["numProp"].as<int>();
-  out.triVerts = convertJSArrayToNumberVector<uint32_t>(mesh["triVerts"]);
-  out.vertProperties =
-      convertJSArrayToNumberVector<float>(mesh["vertProperties"]);
-  if (mesh["mergeFromVert"] != val::undefined()) {
-    out.mergeFromVert =
-        convertJSArrayToNumberVector<uint32_t>(mesh["mergeFromVert"]);
-  }
-  if (mesh["mergeToVert"] != val::undefined()) {
-    out.mergeToVert =
-        convertJSArrayToNumberVector<uint32_t>(mesh["mergeToVert"]);
-  }
-  if (mesh["runIndex"] != val::undefined()) {
-    out.runIndex = convertJSArrayToNumberVector<uint32_t>(mesh["runIndex"]);
-  }
-  if (mesh["runOriginalID"] != val::undefined()) {
-    out.runOriginalID =
-        convertJSArrayToNumberVector<uint32_t>(mesh["runOriginalID"]);
-  }
-  if (mesh["faceID"] != val::undefined()) {
-    out.faceID = convertJSArrayToNumberVector<uint32_t>(mesh["faceID"]);
-  }
-  if (mesh["halfedgeTangent"] != val::undefined()) {
-    out.halfedgeTangent =
-        convertJSArrayToNumberVector<float>(mesh["halfedgeTangent"]);
-  }
-  if (mesh["runTransform"] != val::undefined()) {
-    out.runTransform =
-        convertJSArrayToNumberVector<float>(mesh["runTransform"]);
-  }
-  return out;
-}
-
-val GetMeshJS(const Manifold& manifold, const glm::ivec3& normalIdx) {
-  MeshGL mesh = manifold.GetMeshGL(normalIdx);
-  return MeshGL2JS(mesh);
-}
-
-val Merge(const val& mesh) {
-  val out = val::object();
-  MeshGL meshGL = MeshJS2GL(mesh);
-  bool changed = meshGL.Merge();
-  out.set("changed", changed);
-  out.set("mesh", changed ? MeshGL2JS(meshGL) : mesh);
-  return out;
-}
-
-Manifold FromMeshJS(const val& mesh) { return Manifold(MeshJS2GL(mesh)); }
-
-Manifold Smooth(const val& mesh,
-                const std::vector<Smoothness>& sharpenedEdges = {}) {
-  return Manifold::Smooth(MeshJS2GL(mesh), sharpenedEdges);
-}
-
-Manifold Extrude(std::vector<std::vector<glm::vec2>>& polygons, float height,
-                 int nDivisions, float twistDegrees, glm::vec2 scaleTop) {
-  return Manifold::Extrude(ToPolygon(polygons), height, nDivisions,
-                           twistDegrees, scaleTop);
-}
-
-Manifold Revolve(std::vector<std::vector<glm::vec2>>& polygons,
-                 int circularSegments) {
-  return Manifold::Revolve(ToPolygon(polygons), circularSegments);
-}
-
-Manifold Transform(Manifold& manifold, const val& mat) {
-  std::vector<float> array = convertJSArrayToNumberVector<float>(mat);
-  glm::mat4x3 matrix;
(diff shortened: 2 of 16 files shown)


geoffder:
Ah looks like the previously [mentioned](https://github.com/elalish/manifold/issues/360#issuecomment-1563435673) addition to `debug` in `worker.ts` must have been dropped while I was resolving conflicts, woops. For reference, I think it might be helpful to allow `show` and `only` to work on `CrossSection` as well, similar to how 2d shapes show up in OpenSCAD's preview mode.

```javascript
function debug(shape: Manifold | CrossSection, map: Map<number, Mesh>) {
  const manifold = (shape instanceof module.CrossSection) 
                   ? shape.extrude(1).translate([0, 0, -0.5]) 
                   : shape;
  const result = manifold.asOriginal();
  map.set(result.originalID(), result.getMesh());
  return result;
};
```

Also, after looking at it again just now, if the cleanup step is required to make sure that Manifolds get freed, is `debug` currently (even without this addition) leaking the result of `asOriginal()` (in need of a   `memoryRegistry.push(result)`)?

elalish:
Good question - I *think* that gets cleaned up by the modified `asOriginal` function like the rest, but let me know if I'm missing something. I like the cross-section debug idea - how about extrude 1/100 of the max dimension to make it scale invariant?

bindings/wasm/examples/worker.ts
```diff
@@ -12,15 +12,15 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-import {Document, Material, Node, WebIO} from '@gltf-transform/core';
-import {KHRMaterialsUnlit, KHRONOS_EXTENSIONS} from '@gltf-transform/extensions';
+import { Document, Material, Node, WebIO } from '@gltf-transform/core';
```
elalish, line 15: Can we figure out what's going on with our clang formats? It's impossible to read these diffs if they're full of white space corrections. Are you using VSCode? I have a `.clang-format` file in our root, so ideally that should be standardizing this. I should really figure out how to add a format check to our CI...

bindings/wasm/examples/worker.ts
```diff
@@ -30,6 +30,7 @@ interface WorkerStatic extends ManifoldStatic {
   cleanup(): void;
 }
 
+//@ts-ignore
```
elalish, line 33: Were you getting an error here?

bindings/wasm/manifold.d.ts
```diff
@@ -37,10 +38,10 @@ export interface ManifoldStatic {
...
   reserveIDs: typeof T.reserveIDs;
+  CrossSection: typeof T.CrossSection;
   Manifold: typeof T.Manifold;
   Mesh: typeof T.Mesh;
   setup: () => void;
 }
 
-export default function Module(config?: {locateFile: () => string}):
-    Promise<ManifoldStatic>;
+export default function Module(config: { locateFile: () => string }): Promise<ManifoldStatic>;
```
elalish, line 47: Intensional?

bindings/wasm/helpers.cpp
```diff
@@ -0,0 +1,203 @@
...
+#include <polygon.h>
+#include <sdf.h>
+
+#include <vector>
+
+using namespace emscripten;
+using namespace manifold;
+
+namespace js {
+val MeshGL2JS(const MeshGL& mesh) {
```
elalish, line 13: Good idea splitting this file out. 👍

elalish (changes requested):
Looking pretty good, but let's clean up the format so I can do a better pass. The examples especially will break pretty badly, since their white space gets processed.

bindings/wasm/examples/worker.ts
geoffder: The CI errored out yea (see the first commit of the PR)

bindings/wasm/examples/worker.ts
geoffder: Yea sorry about that, I'm using the `ts-lsp` in emacs. The clang-lsp has always lined up with your formatting setup without any tweaking for me, but the js formatting has other ideas on my end.

bindings/wasm/manifold.d.ts
geoffder: That'd be autoformatting

elalish: No, you lost a `?` I added in my last PR.

geoffder:
> Good question - I _think_ that gets cleaned up by the modified `asOriginal` function like the rest, but let me know if I'm missing something. 

This doesn't go through the modified version does it? It comes before the indirection is set up.maybe if it was moved down the file.

> I like the cross-section debug idea - how about extrude 1/100 of the max dimension to make it scale invariant?

Sure, that doesn't sound bad.

bindings/wasm/manifold.d.ts
geoffder: Oh then it's my fault when I was resolving merge conflicts.

bindings/wasm/examples/worker.ts
elalish: No, that's from the missing `?` below.

geoffder: Hmm well the CI passed once that comment was added

elalish:
> This doesn't go through the modified version does it? It comes before the indirection is set up.maybe if it was moved down the file.

Oh, good catch! Let's do that.

bindings/wasm/examples/public/examples.js
```diff
@@ -502,7 +502,3 @@ for (const [func, code] of Object.entries(examples.functions)) {
       func.replace(/([a-z])([A-Z])/g, '$1 $2');  // Add spaces between words
   examples.functionBodies.set(name, body);
 };
-
-if (typeof self !== 'undefined') {
-  self.examples = examples;
```
elalish, line 507: Intentional? Have you tested ManifoldCAD.org locally?

bindings/wasm/bindings.cpp
```diff
@@ -284,19 +168,21 @@ EMSCRIPTEN_BINDINGS(whatever) {
   function("_Cube", &Manifold::Cube);
   function("_Cylinder", &Manifold::Cylinder);
   function("_Sphere", &Manifold::Sphere);
-  function("tetrahedron", &Manifold::Tetrahedron);
-  function("_Smooth", &Smooth);
-  function("_Extrude", &Extrude);
+  function("_Tetrahedron", &Manifold::Tetrahedron);
```
elalish, line 171: intentional?

bindings/wasm/bindings.cpp
```diff
@@ -284,19 +168,21 @@ EMSCRIPTEN_BINDINGS(whatever) {
...
+  function("_Smooth", &js::Smooth);
+  function("_Extrude", &Manifold::Extrude);
```
elalish, line 173: 👍

bindings/wasm/examples/vite.config.js
geoffder, line 1: Ah missed a re-reformat.

bindings/wasm/bindings.js
```diff
@@ -59,6 +83,113 @@ Module.setup = function() {
...
+    return this._Translate(vararg2vec2(vec));
+  };
+
+  Module.CrossSection.prototype.rotate = function(vec) {
+    return this._Rotate(...vec);
+  };
+
+  Module.CrossSection.prototype.scale = function(vec) {
+    if (typeof vec == 'number') {
+      return this._Scale({x: vec, y: vec});
```
elalish, line 111: Doesn't `vararg2vec2` already take care of this case?

bindings/wasm/examples/public/examples.js
```diff
@@ -502,7 +502,3 @@ for (const [func, code] of Object.entries(examples.functions)) {
       func.replace(/([a-z])([A-Z])/g, '$1 $2');  // Add spaces between words
   examples.functionBodies.set(name, body);
 };
-
-if (typeof self !== 'undefined') {
```
geoffder, line 506: Was this something you added in your recent commit that I missed again? Will add back.

bindings/wasm/examples/worker.ts
```diff
@@ -49,6 +132,7 @@ const SHOW = {
   alpha: 0.25,
   roughness: 1,
   metallic: 0
+
```
geoffder, line 135: Woops, flagged for deletion

bindings/wasm/bindings.js
```diff
@@ -291,74 +487,85 @@ Module.setup = function() {
     return manifold;
   };
 
-  Module.Manifold.prototype = Object.create(ManifoldCtor.prototype);
+  Module.Manifold.ofMesh = function(mesh) {
+    return new Module.Manifold(mesh);
```
elalish, line 491: Why `ofMesh` instead of just the plain constructor?

bindings/wasm/bindings.js
```diff
@@ -291,74 +487,85 @@ Module.setup = function() {
...
+  };
 
-  Module.cube = function(...args) {
+  Module.Manifold.tetrahedron = function() {
+    return Module._Tetrahedron();
```
elalish, line 495: Necessary?

bindings/wasm/bindings.js
```diff
@@ -375,18 +582,15 @@ Module.setup = function() {
...
-    };
-  }
+  Module.Manifold.prototype = Object.create(ManifoldCtor.prototype);
+
+  // Top-level functions
 
-  Module.union = batchbool('union');
-  Module.difference = batchbool('difference');
-  Module.intersection = batchbool('intersection');
+  Module.triangulate = function(polygons, precision = -1) {
```
elalish, line 589: Should this be on the `CrossSection` prototype instead?

bindings/wasm/examples/public/examples.js
```diff
@@ -109,22 +109,25 @@ export const examples = {
...
-          -edgeLength / 2, -edgeLength / 2, 0
-        ]);
+        const edge1 =
+            Manifold.union(corner, edge).rotate([-90, 0, 0]).translate([
+              -edgeLength / 2, -edgeLength / 2, 0
+            ]);
 
-        const edge2 = union(
-            union(edge1, edge1.rotate([0, 0, 180])),
+        const edge2 = Manifold.union(
```
elalish, line 121: Maybe we should destructure at the top: `const {union, cylinder} = Manifold;` so we don't have to add so much verbosity?

elalish (changes requested):
Looking great, thanks! Would you mind pulling master so we can double-check the formatting?

bindings/wasm/helpers.cpp
```diff
@@ -0,0 +1,203 @@
...
+
+std::vector<Manifold> Split(Manifold& a, Manifold& b) {
+  auto [r1, r2] = a.Split(b);
+  return {r1, r2};
+}
+
+std::vector<Manifold> SplitByPlane(Manifold& m, glm::vec3 normal,
+                                   float originOffset) {
+  auto [a, b] = m.SplitByPlane(normal, originOffset);
+  return {a, b};
```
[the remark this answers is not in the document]
geoffder: As far as my searches could tell, emscripten does not take care of std::pair automatically, so this just transforms into a vector which can easily be mapped into a js array (as with decompose).

bindings/wasm/bindings.js
geoffder: The triangulation is mostly used for building meshes though right? Triangulating a polygon doesn't do much good if your goal is using it as a CrossSection.

bindings/wasm/bindings.js
geoffder: I feel like `Manifold.ofMesh(m)` is clearer than `new Manifold(m)` (same with the CrossSection constructor alias I added, `ofPolygons`). Also, it stands out as an available option if it is there in the list of functions imo (with a hint of what you need to give it in the name). The regular constructor still exists too.

bindings/wasm/examples/public/examples.js
geoffder: No I haven't yet since I need to install a browser other than Firefox. It wasn't intentional though, as mentioned in the review I added to this block.

bindings/wasm/bindings.js
elalish: I just mean the input is a CrossSection, not the output. I was imagining something like `const triangles = polygon.triangulate();` instead of `const triangles = Module.triangulate(polygon);`.

bindings/wasm/bindings.js
elalish: Alright, that sounds reasonable.

bindings/wasm/examples/public/examples.js
elalish: Firefox should work too, with a small tweak: https://github.com/elalish/manifold/issues/328#issuecomment-1473847102

bindings/wasm/examples/public/examples.js
elalish: yeah, thanks.

bindings/wasm/bindings.js
geoffder: Ah right. The issue I can think of is that the points Cana change from the users original polygon on making a cross section, so they'll have to make it, put it through triangulate, then get the points from cross section so that they match up with what triangulate worked on. Depends on how low level we want to keep the API for triangulate (and its use case is generally pretty low level, making bespoke meshes).

elalish: Okay, good point. Let's come back to this another time then.

bindings/wasm/examples/public/examples.js
geoffder: That did work, thanks! Maybe we should mention that in a README somewhere.

bindings/wasm/bindings.cpp
geoffder: Yes, the `class_function` method of embind was not working for me for setting up the static methods (compiles, but then the function is not actually available in `examples.js` (function undefined). Maybe because of the way that the object is being replaced in `binding.js`?

bindings/wasm/bindings.js
geoffder: Yes, the class_function method of embind was not working for me for setting up the static methods (compiles, but then the function is not actually available in examples.js (function undefined). Maybe because of the way that the object is being replaced in binding.js? I imagine there is some reorganization that would not require this, but the way we have it setup right now it doesn't work.

bindings/wasm/bindings.cpp
geoffder: See https://github.com/elalish/manifold/pull/440#discussion_r1207452143

bindings/wasm/examples/public/examples.js
geoffder: Within the example body (where they will appear to the user when viewing the example), or above? If it was at the top, that pollute the whole namespace for the editor I guess. The shape constructors have no collision between Manifold and CrossSection if there were opened out, but the booleans would. Not sure what the best course is.

bindings/wasm/bindings.js
geoffder: Ah perhaps, I just copied what `Manifold.prototype.scale` is doing.

geoffder:
Gyroid example just refusing to not timeout it seems.

bindings/wasm/bindings.js
elalish: Interesting, thanks.

bindings/wasm/examples/public/examples.js
elalish: In the body where they appear to the user. Basically the same pattern I use with `const {vec3} = glMatrix;`.

bindings/wasm/examples/vite.config.js
```diff
@@ -3,7 +3,7 @@ import {resolve} from 'path'
 import {defineConfig} from 'vite'
 
 export default defineConfig({
-  test: {testTimeout: 15000},
+  test: {timeout: 10000},
```
elalish, line 6: It's timing out because of a rebase issue - this is the wrong key, so we're using the default of 5000.

bindings/wasm/bindings.js
elalish: Never a bad time to clean up :)

elalish (approved):
Looks like this is ready once the tests pass. Thanks! If you click the merge button, please check when deployment completes and then test ManifoldCAD.org. Since we don't have any staging, I'd like to catch problems early.

bindings/wasm/examples/public/examples.js
geoffder: Ok I'll hunt through for places where it looks like it could use a local opening like that.

bindings/wasm/examples/vite.config.js
geoffder: Ahhh ok, I thought I picked the right one on that by obviously I was going a bit crosseyed going through the autoformat infected diffs (my own fault of course).

geoffder:
> Looks like this is ready once the tests pass. Thanks! If you click the merge button, please check when deployment completes and then test ManifoldCAD.org. Since we don't have any staging, I'd like to catch problems early.

Seems like most things are resolved, but I haven't touched the interface files yet which are giving inaccurate/incomplete completion information with all of the static function reorganization and the CrossSection addition unaccounted for.

bindings/wasm/CMakeLists.txt
```diff
@@ -17,8 +17,12 @@ project(wasm)
 add_executable(manifoldjs bindings.cpp)
 
 # make sure that we recompile the wasm when bindings.js is being modified
-set_source_files_properties(bindings.cpp OBJECT_DEPENDS
-  ${CMAKE_CURRENT_SOURCE_DIR}/bindings.js)
+set_source_files_properties(bindings.cpp
+  OBJECT_DEPENDS ${CMAKE_CURRENT_SOURCE_DIR}/bindings.js
+  # FIXME: can we rewrite the custom command to run after these are modified when building
+  # without recompiling bindings.cpp?
```
elalish, line 23: Probably, but my CMake foo is weak. Feel free to play around!

geoffder:
Ah looks like I should have tested the scale cleanup before committing, might have been what broke here.

bindings/wasm/CMakeLists.txt
geoffder: I believe that I've got something working for the interface copying that does not require rebuilding the bindings needlessly. :+1:

elalish (changes requested):

bindings/wasm/examples/worker.ts
```diff
@@ -184,19 +184,29 @@ module.setMaterial = (manifold: Manifold, material: GLTFMaterial): Manifold => {
...
 
-function debug(manifold: Manifold, map: Map<number, Mesh>) {
-  const result = manifold.asOriginal();
+function debug(shape: Manifold|CrossSection, map: Map<number, Mesh>) {
+  let manifold;
+  if (shape instanceof module.CrossSection) {
+    const box = shape.bounds();
+    const x = box.max[0] - box.min[0];
+    const y = box.max[1] - box.min[1];
+    const h = Math.max(x, y) / 100;
```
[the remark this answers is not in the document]
geoffder: The box returned in JS is just a couple fields though (not a Box with all the methods).

bindings/wasm/bindings.js
```diff
@@ -87,12 +87,12 @@ Module.setup = function() {
...
   const CrossSectionCtor = Module.CrossSection;
 
-  function cross(polygons, fillrule = 'Positive') {
+  function cross(polygons, fillRule = 'Positive') {
     if (polygons instanceof CrossSectionCtor) {
       return polygons;
     } else {
       const polygonsVec = polygons2vec(polygons);
-      const cs = new CrossSectionCtor(polygonsVec, fillrule = fillrule);
+      const cs = new CrossSectionCtor(polygonsVec, fillRule = fillRule);
```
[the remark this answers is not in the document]
geoffder: I did it when I thought that keywords existed (rather than just ordering like in cpp) out of python habit (in ocaml there is name punning, so wouldn't be doing the repetition there). I'll go back and drop it.

bindings/wasm/examples/worker.ts
```diff
@@ -184,19 +184,29 @@ module.setMaterial = (manifold: Manifold, material: GLTFMaterial): Manifold => {
...
+    manifold = shape.extrude(h).translate([0, 0, -h / 2]);
+  } else {
+    manifold = shape;
+  }
+  let result = manifold.asOriginal();
```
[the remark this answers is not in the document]
geoffder: I've played with it a bit, it isn't as nice as with the manifolds for obvious reasons as you mention, but it does let you quickly red/ghost them and show them alongside manifolds if you stick the manifold made by `show` in a node for example.

bindings/wasm/helpers.cpp
```diff
@@ -104,6 +106,15 @@ Manifold Smooth(const val& mesh,
 }  // namespace js
 
 namespace cross_js {
+CrossSection OfPolygons(std::vector<std::vector<glm::vec2>> polygons,
+                        std::string fill_rule) {
```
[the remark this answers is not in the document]
geoffder: Ah ok, I'll just pass an int to the helper function from js instead.

bindings/wasm/manifold-encapsulated-types.d.ts
```diff
@@ -318,6 +632,15 @@ export class Manifold {
...
       Manifold;
 
+  /**
+   * Constructs a new manifold from a list of other manifolds. This is a purely
+   * topological operation, so care should be taken to avoid creating
+   * overlapping results. It is the inverse operation of Decompose().
+   *
+   * @param manifolds A list of Manifolds to lazy-union together.
+   */
+  static compose(manifolds: Manifold[]): Manifold;
```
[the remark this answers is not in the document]
geoffder: We could, I was just organizing them by purpose. I'm not sure what the way that this interface will normally be viewed.

bindings/wasm/examples/worker.ts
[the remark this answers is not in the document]
elalish: Ah, fair. Should probably add those someday...

geoffder:
As an aside, I have come across a funny bug, the following fails in the editor with `RuntimeError: index out of bounds`
```javascript
const box = CrossSection.square([80, 80], true).offset(10, "Round").extrude(100, 0, 0, [1, 1], false);
```
However, changing the final argument (center) to `true` (which simply translates the manifold down) works.

bindings/wasm/manifold-encapsulated-types.d.ts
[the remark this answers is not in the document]
elalish: Probably mostly by other maintainers. Just try to make it obvious where someone would put the next method.

elalish:
> As an aside, I have come across a funny bug, the following fails in the editor with `RuntimeError: index out of bounds`
> 
> ```js
> const box = CrossSection.square([80, 80], true).offset(10, "Round").extrude(100, 0, 0, [1, 1], false);
> ```
> 
> However, changing the final argument (center) to `true` (which simply translates the manifold down) works.

Ugh, WASM errors are so obtuse. Can you repro it in a C++ test? I just fixed some like that for Warp, which turned out to be because I wasn't checking inputs strictly enough.

bindings/wasm/examples/worker.ts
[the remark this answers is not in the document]
elalish: I dunno, I think this could get pretty confusing for users. Let's hold off until we've found a better way.

geoffder:
I did try doing `GetMesh` on the non-centred version in cpp, and it exported fine without crashing :grimacing:.

```cpp
TEST(CrossSection, RoundedBox) {
  auto sq = CrossSection::Square({80, 80}, true)
                .Offset(10, CrossSection::JoinType::Round);
  auto uncentred = Manifold::Extrude(sq, 100, 0, 0, {1, 1});
  auto ball = Manifold::Sphere(60, 100);
  auto mesh = (uncentred - ball).GetMesh();
#ifdef MANIFOLD_EXPORT
  if (options.exportModels)
    ExportMesh("cross_section_rounded_box.glb", mesh, {});
#endif
}
```
It *did* however take almost 37s to run though.

elalish:
Weird, it seems way too simple to take that long. How many facets for "round"? Oh well, we should probably move this discussion to a separate issue anyway.

geoffder:
> Weird, it seems way too simple to take that long. How many facets for "round"? Oh well, we should probably move this discussion to a separate issue anyway.

Ok. It is a pretty weird problem given than a mere translation makes the difference. This use of offset is bringing it out, but where the error is and how insidious I don't know.

pca006132:
I finally have a break now. I just have a look at it, there are 281010 edges in the `uncentred` mesh, so most of the time is spent on `RecursiveSwapEdge` and causing the slowness.

pca006132:
Related: http://www.angusj.com/clipper2/Docs/Units/Clipper.Offset/Classes/ClipperOffset/Properties/ArcTolerance.htm
Perhaps we should have this in CrossSection, similar to what we have with Manifold.

geoffder:
From how smooth the roundover looks I'm not too surprised. 

Calculating a default ourselves based on the delta, rather than leaving it up to the Clipper2 default or the user you mean?

pca006132:
Yeah, we can allow the user to set the `minCircularAngle`/`circularSegments` and calculate it ourselves, or ask the user to specify a `minCircularEdgeLength`. We should probably have the `minCircularAngle` set to 10 or something, so the default performance is good enough.

geoffder:
Caught what I hope to be the last bugs while testing out `CrossSection` in the editor earlier, think it is looking pretty good now.

I remembered a question I'd forgotten around the business in [getManifoldDTS](https://github.com/elalish/manifold/blob/507b8448401ddb899be5cc6ecaaf78392b4da9e1/bindings/wasm/examples/editor.js#L231) in `editor.js`. I've copied the corresponding updated block over what was there before, but I'm wondering what the purpose is, since naively, it seems incomplete. It gets `encapsulated`, but then doesn't seem to be aliased as `T` to work with the subsequent `typeof` declarations, and the top-level types aren't exported as they are in `manifold.d.ts` either. Though, again, I've come into working on this with no familiarity to the js/ts ecosystem, so I could be missing something obvious.

If this stuff is actually (still?) necessary, it seems pretty fragile given that it requires manually copying over the interface into the string block (though it isn't something that happens *too* often I suppose).

elalish:
I believe #348 is where the `getManifoldDTS` came from - linked issues there may give a little background. It's a total hack to work around Monaco not liking `.d.ts` files that import others. I don't have a chance to review right now - it's entirely possible it is now out of date or there's a better way.

geoffder:
Ok, well I think I'll leave it for another PR then if any problems arise. The types are working for completion and linting in my use of the editor locally at least.

bindings/wasm/examples/public/examples.js
```diff
@@ -259,12 +263,8 @@ export const examples = {
...
                                        n * q * majorRadius / threadRadius;
 
-        const circle = [];
-        const dPhi = 2 * 3.14159 / n;
-        const offset = 2;
-        for (let i = 0; i < n; ++i) {
-          circle.push([Math.cos(dPhi * i) + offset, Math.sin(dPhi * i)]);
-        }
+        const offset = 2.
+        const circle = CrossSection.circle(1., n).translate([offset, 0.]);
```
elalish, line 267: No need for `.` in JS - everything is a double - even integers!

bindings/wasm/examples/public/examples.js
```diff
@@ -325,16 +325,18 @@ export const examples = {
...
-        result = difference(result, hole);
-        result = difference(result, hole.rotate([90, 0, 0]));
-        result = difference(result, hole.rotate([0, 90, 0]));
+        const hole = Manifold.compose(holes);
 
+        result = Manifold.difference(
+            result,
+            hole,
+            hole.rotate([90, 0, 0]),
+            hole.rotate([0, 90, 0]),
```
elalish, line 338: 👍

bindings/wasm/examples/public/examples.js
```diff
@@ -20,8 +20,11 @@ export const examples = {
       // see the static API - these functions can also be used bare. Use
       // console.log() to print output (lower-right). This editor defines Z as
       // up and units of mm.
-
-      const box = cube([100, 100, 100], true);
+      const {sphere} = Manifold;
+      const {square} = CrossSection;
+      const box = square([70, 70], true)
+                      .offset(15, 'Round')
+                      .extrude(100, 0, 0, [1, 1], true);
```
elalish, line 27: I appreciate the idea, but I'm intentionally keeping this first landing-page example super simple. Let's revert this for now. We should add a cool example that showcases more of the CrossSection API though at some point.

bindings/wasm/helpers.cpp
```diff
@@ -0,0 +1,222 @@
...
+  void (*f)(glm::vec2&) = reinterpret_cast<void (*)(glm::vec2&)>(funcPtr);
+  return cross_section.Warp(f);
+}
+
+CrossSection Offset(CrossSection& cross_section, double delta, int join_type,
+                    double miter_limit, double arc_tolerance) {
+  auto jt = join_type == 0   ? CrossSection::JoinType::Square
+            : join_type == 1 ? CrossSection::JoinType::Round
+                             : CrossSection::JoinType::Miter;
+  return cross_section.Offset(delta, jt, miter_limit, arc_tolerance);
```
elalish, line 160: Let's do this in a separate PR, but I agree with @pca006132 that instead of `arc_tolerance` we should use our `setMinCircularAngle` etc. functions to define this.

elalish (changes requested):
Just some minor things and then I think we can merge. Let's try to avoid any more functional changes to this PR so we can close the review cycle. We'll need another one before release anyway to address the offset API.

bindings/wasm/examples/public/examples.js
geoffder: Ah ocaml muscle memory acting up again.

bindings/wasm/examples/worker.ts
```diff
@@ -39,6 +39,87 @@ glMatrix.glMatrix.setMatrixArrayType(Array);
...
+
+const memoryRegistry = new Array<Manifold|CrossSection>();
+
+function addMembers(
+    className: string, methodNames: Array<string>, areStatic: boolean) {
+  //@ts-ignore
+  const cls = module[className];
+  const obj = areStatic ? cls : cls.prototype;
+  for (const name of methodNames) {
+    if (name != 'cylinder') {
```
elalish, line 84: I must have missed this in code review - is this intentional? Just want to make sure I can safely remove it now.

hrgdavor:
love the addition of slice :)

PR #441, opened 2023-05-26
Add CI check for formatting more languages

elalish:
Let's see if this works. FYI @geoffder.

### DIFF
--- a/.github/workflows/check_format.yml
+++ b/.github/workflows/check_format.yml
@@ -16,6 +16,6 @@ jobs:
       with:
         source: '.'
         exclude: '*/third_party'
-        extensions: 'h,cpp,js,ts,py,json,html,css'
+        extensions: 'h,cpp,js,ts,html'
         clangFormatVersion: 12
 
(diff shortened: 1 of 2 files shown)


geoffder:
This will mean the CI complains, but without configs for the particular formatters we won't be coordinated to keep the checks happy will we? I'm not sure what it is that is being run for my formatting by the lsp, I think it's prettier for js. So I think we'd need a prettier config.

elalish:
I'm using clang-format for everything, with `.clang-format` controlling it. Can you set up lsp that way?

geoffder:
Oh, I wasn't aware clang-format did non C/C++ stuff. I'll look into swapping that in.

geoffder:
I've got clang-format running now, but it looks like I'm going to have to wrestle with it a bit. It is splitting `=>` into `= >`, which needless to say, is not so good.

geoffder:
Alright, well I never figured out why running clang-format through the usual formatter setup, but I went around it! Diffs are way down on the JS PR, so it seems like it is working as advertised.

PR #442, opened 2023-05-27
Fix editor warp crash

elalish:
I was finding that warp would sometimes crash (saying something really unhelpful like out of memory) and then the editor was unusable until after a reload. I fixed this three ways:

1) When the worker throws an error, we now recreate it, like when we cancel. This keeps the editor usable.
2) I added our non-finite vertex check to `Warp` - turns out I was just generating a NaN somewhere.
3) I surfaced this error through to throw in the JS bindings, so you actually get some feedback on what went wrong.

### DIFF
--- a/bindings/wasm/bindings.js
+++ b/bindings/wasm/bindings.js
@@ -72,6 +72,11 @@ Module.setup = function() {
     }, 'vi');
     const out = this._Warp(wasmFuncPtr);
     removeFunction(wasmFuncPtr);
+
+    const status = out.status();
+    if (status.value !== 0) {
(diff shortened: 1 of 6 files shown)


geoffder:
Was going to say I might not get to this and the rest of the other PR right away since I've got my bros wedding tomorrow, but you beat me to merging anyway! 😅

elalish:
Have fun! I'm out this weekend too, so wanted to merge this before I went to bed - more of an FYI in case I generated any merge conflicts for you.

geoffder:
Roger, I'll plan to pull in master when I'm back at it for the formatting checks and these fixes.

PR #444, opened 2023-06-01
Use quality globals instead of arc tolerance for offset

geoffder:
Use `circularSegments` and the `Quality` globals rather than `arc_tolerance` to control offset roundovers. 

I'll add commits for the doc strings and JS bindings soon.


elalish (approved):
Looks great, thank you!

PR #446, opened 2023-06-01
Backward compat

elalish:
We made a breaking change in #440 to pull the static methods up into `module.Manifold`, which I think is the right call, but this breaks people's existing scripts in ManifoldCAD.org. I decided to add back in the bare manifold constructors so that those scripts will still run, though I did not add them back into the `.d.ts`. This means they get a red error telling them to update while their script still runs - this feels like a reasonable compromise for now.

### DIFF
--- a/bindings/wasm/examples/public/examples.js
+++ b/bindings/wasm/examples/public/examples.js
@@ -16,10 +16,11 @@ export const examples = {
   functions: {
     Intro: function() {
       // Write code in JavaScript or TypeScript and this editor will show the
-      // API docs. Type e.g. "box." to see the Manifold API. Type "module." to
-      // see the static API - these functions can also be used bare. Use
-      // console.log() to print output (lower-right). This editor defines Z as
-      // up and units of mm.
+      // API docs. Type e.g. "box." to see the Manifold API. Type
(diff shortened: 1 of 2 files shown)


geoffder:
Seems like an easy way to quasi-deprecate those functions being at the top-level.

Is the goal to eventually drop these out of the top-level again? If so do you think a notice with a timeline to removal would be necessary, or just give people some time with the annoying red squigglies?

pca006132:
Interesting, why is there redefinition error? Can you show the error message and the exact location throwing the error? (too lazy to turn on my computer)

elalish:
It's because we now want to use the pattern e.g. `const {cube} = Manifold`, but if I've already defined `cube = Manifold.cube` inside of the worker, then JS won't allow you to redefine `cube`. I was going to try to add a QuickFix in Monaco to help deal with it, but their documentation is very sparse, so I think it's not worthwhile. I guess we'll just hope Monaco's auto-completion will help guide people to the solution.

geoffder:
I think people will be able to figure it out. Pretty quick fix to add one line at the top to cover all of the used constructors as well.

PR #447, opened 2023-06-08
Fix RemoveIfFolded() in the decimator

elalish:
I'm finally testing #429, but it doesn't repro for me locally (the test passes). I'm enabling it here (and moving it) to see if the CI catches the problem.


elalish:
Hmm, @Loosetooth the only failures I see are with a small degenerate polygon complaining about being not geometrically valid. We're robust to that by default and only make it pickier for our tests. So basically, I can't reproduce your problem, either locally or on our CI. Do you want to take another look?

elalish:
@Loosetooth Never-mind, your test case is good! I re-read your thread and tried changing the scale and found a value that infinite-loops. Then I went through and cut down the test case a bunch. It's easy to see where it fails when you run `./manifold_test --gtest_filter=Boolean.UnionError -v` - the last thing it reports is `found 1902 short edges to collapse`, then hangs. With the address sanitizer enabled it says `runtime error: addition of unsigned offset to 0x000116304800 overflowed to 0x0001163047d0` at `READ of size 4 at 0x0001163047d8 thread T0` so I think I'm getting in the neighborhood of the problem now.

src/utilities/include/vec_dh.h
```diff
@@ -226,9 +227,19 @@ class ManagedVec {
 
   const T &back() const { return *(ptr_ + size_ - 1); }
 
-  T &operator[](size_t i) { return *(ptr_ + i); }
+  T &operator[](size_t i) {
+    // ASSERT(i < size_, logicErr,
```
elalish, line 231: This makes it quicker to find seg-fault locations but in `MANIFOLD_DEBUG` mode it makes the tests take twice as long, so I'm leaving it commented so we can easily use it just if we need to look for such things. The address sanitizer is another good way.

src/manifold/src/edge_op.cpp
```diff
@@ -325,8 +325,23 @@ void Manifold::Impl::RemoveIfFolded(int edge) {
...
+      if (halfedge_[tri0edge[2]].pairedHalfedge == tri1edge[1]) {
+        vertPos_[halfedge_[tri1edge[1]].startVert] = glm::vec3(NAN);
+      }
+    }
+    PairUp(halfedge_[tri0edge[1]].pairedHalfedge,
+           halfedge_[tri1edge[2]].pairedHalfedge);
+    PairUp(halfedge_[tri0edge[2]].pairedHalfedge,
+           halfedge_[tri1edge[1]].pairedHalfedge);
     for (int i : {0, 1, 2}) {
-      vertPos_[halfedge_[tri0edge[i]].startVert] = glm::vec3(NAN);
```
elalish, line 329: The bug was that I was assuming when two triangles had folded up, they were already a disconnected two-triangle mesh that could be deleted. This is very often true, but sometimes it's an ear, still attached to a larger mesh. In this case we can only delete one vert instead of three and we need to pair up the edges we cut.

PR #448, opened 2023-06-08
Change curvature API

elalish:
I decided to update the curvature API, since I never much liked returning a list that one can only hope stays correlated to output vertices. Now that we have a properties API, I've updated this to simply fill that in as desired. I plan to follow this pattern to add a `CalculateNormals` API, which I swear I discussed with someone here, but I can't find the thread now. This is a breaking change, but I don't think we have many users of curvature yet. 

@geoffder I removed the c bindings related to the old API - the new one should be a lot simpler. I figure you can add it in when you have a chance?


bindings/wasm/bindings.js
```diff
@@ -229,7 +229,7 @@ Module.setup = function() {
   };
 
   Module.Manifold.prototype.setProperties = function(numProp, func) {
-    const oldNumProp = this.numProp;
+    const oldNumProp = this.numProp();
```
elalish, line 232: This was a bug - reading existing properties didn't work before.

geoffder:
Sounds goos, I'll plan to add the new bindings after these changes are in.

PR #449, opened 2023-06-09
Workflow builds

cartesian-theatrics:


elalish:
A preview?

geoffder:
Not sure why manifold would be moved into a "cpp" directory, with Java alongside it rather than putting java alongside the other bindings.

cartesian-theatrics:
Sorry, I meant to open this against my fork!

> Not sure why manifold would be moved into a "cpp" directory, with Java alongside it rather than putting java alongside the other bindings.

The cpp code here is mostly intended to facilitate the java bindings by providing efficient mapping to c++ structures. Plus there a little extra functionality to support more openscad functionality. I wanted to keep all the changes under one directory so upkeep would be easier.

You know you have a great library when everyone and their mother wants to bind to it lol.

hrgdavor:
>You know you have a great library when everyone and their mother wants to bind to it lol.

That is so true. I am personally impressed how fast this library became stable and useful.

PR #454, opened 2023-06-15
Workaround assimp exporting empty 3mfs

geoffder:
This is something that I actually noticed a long time ago, but forgot to mention or return to myself, but 3mf export via `MeshIO` doesn't actually work (a file containing no meshes is written).

There is actually a [long standing issue](https://github.com/assimp/assimp/issues/3816) for this in the assimp tracker, but fortunately it has an easy workaround. This PR implements that workaround by nesting the node holding the mesh into a child of the root node if the extension is `.3mf`.


elalish (approved):
Excellent! I think I ran into this at some point too and never bothered debugging it. Thanks for the simple work-around.

PR #455, opened 2023-06-17
Partial Revolve

cartesian-theatrics:
This adds an optional arity for Revolve that supports partial revolutions. It changes the implementation slightly by first taking the x>=0 slice if the bounding box contains negative x values. The cases of vertices lying on the revolve axis is handled in the same loop, arguably simplifying the logic slightly and making it easier to support partial evolves.  I believe the tessellation of full revolutions is equivalent to the previous version.

It might be worth noting it's not strictly required to implement special handling of coincident vertices. The logic is far simpler in this case and the core implementation seems to be able handle it no problem. However, it does break Genus. I think it's better to do the correct construction, but might be something to consider. The complexity overhead is quite high unfortunately... Perhaps it can be simplified.

### DIFF
--- a/src/manifold/src/constructors.cpp
+++ b/src/manifold/src/constructors.cpp
@@ -319,10 +319,26 @@ Manifold Manifold::Extrude(const CrossSection& crossSection, float height,
  * calculated by the static Defaults.
  */
 Manifold Manifold::Revolve(const CrossSection& crossSection,
-                           int circularSegments) {
-  auto polygons = crossSection.ToPolygons();
+                           int circularSegments,
+                           float revolveDegrees) {
+
+  Polygons polygons = crossSection.ToPolygons();
+
   if (polygons.size() == 0) {
     return Invalid();
+  }
+
+  Rect bounds = crossSection.Bounds();
+
+  // Take the x>=0 slice.
+  if (bounds.min.x < 0) {
+    glm::vec2 min = bounds.min;
+    glm::vec2 max = bounds.max;
+    CrossSection posBoundingBox = CrossSection({{0.0, min.y},{max.x, min.y},
+                                                {max.x,max.y},{0.0,max.y}});
+
+    // Can't use RectClip as it has many failure cases.
+    polygons = (crossSection ^ posBoundingBox).ToPolygons();
   }
 
   float radius = 0.0f;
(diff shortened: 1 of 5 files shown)


test/manifold_test.cpp
```diff
@@ -215,6 +216,15 @@ TEST(Manifold, Revolve2) {
   EXPECT_NEAR(prop.surfaceArea, 96.0f * glm::pi<float>(), 1.0f);
 }
 
+TEST(Manifold, PartialRevolve) {
+  Polygons polys = SquareHole(2.0f);
+  Manifold donutHole = Manifold::Revolve(polys, 48, 180);
+  EXPECT_EQ(donutHole.Genus(), 1);
```
elalish, line 222: Would you mind pasting an image of this object? I'm having trouble visualizing what axis it's rotating around.

src/manifold/src/constructors.cpp
```diff
@@ -317,95 +317,127 @@ Manifold Manifold::Extrude(const CrossSection& crossSection, float height,
...
-              triVerts.push_back(
-                  {startVert - 1, startVert + slice, startVert + lastSlice});
-            }
+
+        if (isFullRevolution || slice > 0) {
+          int lastSlice = (slice == 0 ? nDivisions : slice) - 1;
+          if (currPolyVertex.x > 0.0) {
+            triVerts.push_back({startPosIndex + slice,
+                startPosIndex + lastSlice,
+                // "Reuse" vertex of first slice if it lies on the revolve axis
```
elalish, line 413: Is this the part you were saying messes up the genus? Can you give an example?

src/manifold/src/constructors.cpp
```diff
@@ -317,95 +317,127 @@ Manifold Manifold::Extrude(const CrossSection& crossSection, float height,
...
+  Rect bounds = crossSection.Bounds();
+
+  // Take the x>=0 slice.
+  if (bounds.min.x < 0) {
+    glm::vec2 min = bounds.min;
+    glm::vec2 max = bounds.max;
+    CrossSection posBoundingBox = CrossSection({{0.0, min.y},{max.x, min.y},
+                                                {max.x,max.y},{0.0,max.y}});
+
+    // Can't use RectClip as it has many failure cases.
```
elalish, line 341: That seems odd - what failure cases have you seen?

src/manifold/src/constructors.cpp
```diff
@@ -317,95 +317,127 @@ Manifold Manifold::Extrude(const CrossSection& crossSection, float height,
...
   auto pImpl_ = std::make_shared<Impl>();
   auto& vertPos = pImpl_->vertPos_;
   VecDH<glm::ivec3> triVertsDH;
   auto& triVerts = triVertsDH;
-  float dPhi = 360.0f / nDivisions;
+
+  std::vector<int> startPoses;
+  std::vector<int> endPoses;
+
+  float dPhi = revolveDegrees / nDivisions;
```
elalish, line 368: I've been trying to update our style to use `const` as much as possible; would you mind applying that across your PR?

elalish (commented):
Thanks, this looks promising. Any trouble signing the CLA?

src/manifold/src/constructors.cpp
cartesian-theatrics: Kinda. Genus "breaks" when you don't have any special handling of vertices lying on the Y axis. When x=0, `{pos.x * manifold::cosd(phi), pos.x * manifold::sind(phi), pos.y}` yields coincident vertices and degenerate triangles. For example, when a square is offset from the Y axis Revolve will yield 8 triangles and 4 vertices per slice, whereas if one edge is on the Y axis, Revolve will yield 4 triangles and two vertices per slice. In my initial implementation I didn't have any special handling of such cases but surprisingly it still mostly worked. But Genus was off  I guess because Manifold considered the revolved square to be an infinitesimal toroid.

This is the original loop that "works":

```cpp
    for (const auto& poly : polygons) {
        for (std::size_t polyVert = 0; polyVert < poly.size(); ++polyVert) {

            int startVert = vertPos.size();

            if (!isFullRevolution)
                startPoses.push_back(startVert);

            // first and last slice are distinguished if not a full revolution.
            int nSlices = isFullRevolution ? nDivisions : nDivisions + 1;
            int lastStart = startVert + (polyVert == 0 ? nSlices * (poly.size() - 1) : -nSlices);

            for (int slice = 0; slice < nSlices; ++slice) {

                float phi = slice * dPhi;
                glm::vec2 pos = poly[polyVert];
                glm::vec3 p = {pos.x * manifold::cosd(phi), pos.x * manifold::sind(phi), pos.y};
                vertPos.push_back(p);

                int lastSlice = (slice == 0 ? nDivisions : slice) - 1;
                if (isFullRevolution || slice > 0) {
                    triVerts.push_back({startVert + slice, startVert + lastSlice,
                            lastStart + lastSlice});
                    triVerts.push_back(
                        {lastStart + lastSlice, lastStart + slice, startVert + slice});
                }

            }
            if (!isFullRevolution)
                endPoses.push_back(vertPos.size() -1);
        }
    }
```

src/manifold/src/constructors.cpp
cartesian-theatrics: I wish I would have kept failing cases. I will try to recreate them.

src/manifold/src/constructors.cpp
cartesian-theatrics: Yep, good call.

src/manifold/src/constructors.cpp
elalish: Right, so the problem is without special handling of the center verts you end up with an infinitesimal hole through it like a bead. Definitely important to avoid that, since I don't think our decimator can fix it. If your test case doesn't already check that, it's probably a good idea to add a test for that specifically.

test/manifold_test.cpp
cartesian-theatrics: Yes, it looks like this. I'd definitely want to do more thorough testing before merging (even if we don't want to include them all).
PartialRevolve

elalish: Thanks; I'd recommend pushing it either in or out a little, since right now it's just rounding error determining a major topological change in the center.

elalish:
@cartesian-theatrics Is anything blocking you on this PR? I would like to merge it.

cartesian-theatrics:
My apologies, got distracted with other priorities. I will work this today.

cartesian-theatrics:
@elalish Ugh, looks like I authored commits with a configured git username/email that I no longer control. Should I amend the commits or create a new branch?

elalish:
Thanks! Do whatever's easiest with the CLA - if you want to make a different branch with a new PR, that's fine. Let us know when you're ready for review.

src/manifold/src/constructors.cpp
cartesian-theatrics: Even in the plain Revolve test case this doesn't seem to work.

```c++
  if (bounds.min.x < 0) {
    glm::vec2 min = bounds.min;
    glm::vec2 max = bounds.max;
    Rect boundingBox = Rect({0.0, min.y}, {max.x, max.y});
    // Can't use RectClip as it has many failure cases.
    polygons = crossSection.RectClip(boundingBox).ToPolygons();
  }

```

Result is not manifold-2, doesn't export. Am I constructing the Rect wrong?

cartesian-theatrics:
@elalish other than the unresolved RectClip issue (see my comment), I think this is ready for review.

src/manifold/src/constructors.cpp
elalish: I'm not seeing a manifoldness failure, but I'm definitely reproducing a serious problem with `RectClip`: When I clip a polygon with a hole in half, I'm only getting the outer contour; intersection gives the correct result. @geoffder would you mind taking a look?

This is fine for this PR, thanks!

elalish (changes requested):
This is looking great, just a few minor things. Thanks for cleaning up my code!

geoffder:
This sounded familiar so I went to the Clipper2 issue tracker and it may be related to a known limitation mentioned in January https://github.com/AngusJohnson/Clipper2/issues/386.

elalish:
Thanks, @geoffder, we can follow up on those Clipper2 threads. The behavior I saw was worse than the examples he showed, but probably the same issue. If we can't get a proper fill-rule preserving RectClip, I think I'd prefer to remove it from our API, due to this kind of confusion.

elalish (approved):
Thanks!

geoffder:
> Thanks, @geoffder, we can follow up on those Clipper2 threads. The behavior I saw was worse than the examples he showed, but probably the same issue. If we can't get a proper fill-rule preserving RectClip, I think I'd prefer to remove it from our API, due to this kind of confusion.

Hmm I just tested clipping a ring (circle with opposite winding circular hole) and got the same result with intersection and rect clip in Clipper2 (via my `ocaml-clipper2` bindings), using their svg export. But, there is definitely funny business when it comes to trying to use the output in manifold.

```cpp
  auto rect = Rect({0, 0}, {5, 5});
  auto ring = CrossSection::Circle(4) - CrossSection::Circle(2);
  auto clipped_ring = ring.RectClip(rect);
  auto clipped_ring_ex = Manifold::Extrude(ring.RectClip(rect), 5); // empty result
  auto inter_ring_ex = Manifold::Extrude(ring ^ rect.AsCrossSection(), 5);  // works after AsCrossSection fix
  EXPECT_EQ(clipped_ring.NumContour(), 1); // fails, with 2
```

Going by Angus' comments in the linked discussions, this is because `RectClip` simply tracks each contour and clips them independently, while leaving the winding order alone. I would have thought that going through a union would be fine if the winding was left intact though (if you add an empty `CrossSection` to the `clipped_ring` the result is odd). Since the usual clipping algorithms aren't applied, the usual guarantees that we have for our CrossSections are out the window I think. So, maybe leaving `RectClip` out for now would be for the best.
 
While testing this, I've found a bug (my fault) with the winding of the path made in `AsCrossSection` that results in an empty shape after the Clipper2 union, so I'll make a PR to fix that along with dropping `RectClip` from our API later.

PR #457, opened 2023-06-17
Added a python bindings example where a CrossSection is missing a polygon

johnnybigert:


bindings/python/examples/missing_polygon.py
```diff
@@ -0,0 +1,39 @@
...
+                (-174.50, 58.40),
+                (-174.50, 58.60),
+                (-174.70, 58.60),
+                (-174.70, 58.40),
+            ]]
+        )
+
+    diff = test_cross - cross_section
+    if diff.area() != 0:
+        print(f"Cross-section is missing a part of the polygon! "
```
elalish, line 31: Why do you expect this area to be zero?

bindings/python/examples/missing_polygon.py
```diff
@@ -0,0 +1,39 @@
...
+            ]]
+        )
+    test_cross = \
+        CrossSection(
+            [[
+                (-174.70, 58.40),
+                (-174.50, 58.40),
+                (-174.50, 58.60),
+                (-174.70, 58.60),
+                (-174.70, 58.40),
```
elalish, line 25: Perhaps the duplicate point is throwing it off, though we ought to be robust to that.

bindings/python/examples/missing_polygon.py
```diff
@@ -0,0 +1,39 @@
...
+    cross_section = \
+        CrossSection(
+            [[
+                (-176.5416564941429, 62.267156291299955), (-177.11900294911175, 63.267144940124126),
+                (-177.11900294910944, 62.26716470718384), (-176.5416564941406, 62.26715569064245),
+                (-176.54165614735393, 61.26716470718384), (-175.9643096923828, 61.26715985212647),
+                (-175.96430934559382, 60.26716470718384), (-175.38696566496216, 60.26717407055541),
+                (-175.386962543836, 59.26716470718384), (-174.80960083007815, 59.26714979522481),
+                (-174.8096004832914, 58.26716470718384), (-174.23225470415798, 58.26715427684856),
+                (-175.01954376826416, 59.63078409350792), (-176.5416564941429, 62.267156291299955),
```
elalish, line 15: Can we further simplify this test case? I'm curious if it reproduces with simple integer coordinates.

bindings/python/examples/missing_polygon.py
pca006132: They are not duplicate? Should just be a rectangle.

geoffder:
For polygons in Manifold, you aren't supposed to duplicate the start and end points. There is always segment between the last and first points of a path. In Clipper2, which we use for CrossSection, closed-ness is indicated by a parameter, not the closing point.

bindings/python/examples/missing_polygon.py
johnnybigert: I expect there to be overlap, so I want it to be greater than zero.

johnnybigert:
> For polygons in Manifold, you aren't supposed to duplicate the start and end points. There is always segment between the last and first points of a path. In Clipper2, which we use for CrossSection, closed-ness is indicated by a parameter, not the closing point.

Ok, removed the duplicated vertex at the end of the test polygon.

Note that the misbehaving polygon didn't have any duplicated points. (I tried both with and without duplicating the first vertex, with the same result.)

johnnybigert:
Any thoughts on why this very simple example is producing the wrong result in Python (but the correct result in C++)?

I bet it's something simple that is missing/wrong in the bindings. Like not taking decomposition into account in to account in `to_polygons`? Or some off-by-one problem in the transfer of the polygon points?

@geoffder @pca006132

pca006132:
This doesn't look like an off-by-one error. You can check the following patch:
```diff
diff --git a/src/cross_section/src/cross_section.cpp b/src/cross_section/src/cross_section.cpp
index e2b7eed..a4a6b08 100644
--- a/src/cross_section/src/cross_section.cpp
+++ b/src/cross_section/src/cross_section.cpp
@@ -194,6 +194,7 @@ CrossSection::CrossSection(const SimplePolygon& contour, FillRule fillrule) {
  * contours.
  */
 CrossSection::CrossSection(const Polygons& contours, FillRule fillrule) {
+  Dump(contours[0]);
   auto ps = C2::PathsD();
   ps.reserve(contours.size());
   for (auto ctr : contours) {
```

I wonder if the coordinates are sensitive to rounding? There may be different rounding behavior in Python and C++ (especially when we need to perform conversion from double to float for example).

bindings/python/examples/missing_polygon.py
elalish: That's not what this if condition is saying. Also, even if there is overlap between these polygons, I don't believe they have equal area, so I'm not sure where the `> 0` expectation is coming from.

johnnybigert: We do a diff between the test square and the polygon under test to check for overlap (there should be a full overlap). If the difference is small (I changed it to eps now), the polygon is overlapping the test rectangle. But we get zero overlap in my test case.

johnnybigert:
> This doesn't look like an off-by-one error. You can check the following patch:
> 
> ```diff
> diff --git a/src/cross_section/src/cross_section.cpp b/src/cross_section/src/cross_section.cpp
> index e2b7eed..a4a6b08 100644
> --- a/src/cross_section/src/cross_section.cpp
> +++ b/src/cross_section/src/cross_section.cpp
> @@ -194,6 +194,7 @@ CrossSection::CrossSection(const SimplePolygon& contour, FillRule fillrule) {
>   * contours.
>   */
>  CrossSection::CrossSection(const Polygons& contours, FillRule fillrule) {
> +  Dump(contours[0]);
>    auto ps = C2::PathsD();
>    ps.reserve(contours.size());
>    for (auto ctr : contours) {
> ```
> 
> I wonder if the coordinates are sensitive to rounding? There may be different rounding behavior in Python and C++ (especially when we need to perform conversion from double to float for example).

Great thinking!

I printed all the values and they are not at really the same as the input (some of them differs in the sixth decimal):
```
i=0 j=0, -176.54165649414062 62.267154693603516   # differs a lot from 62.267156291299955
i=0 j=1, -177.11900329589844 63.267143249511719
i=0 j=2, -177.11900329589844 62.267166137695312
i=0 j=3, -176.54165649414062 62.267154693603516
i=0 j=4, -176.54165649414062 61.267166137695312
i=0 j=5, -175.96430969238281 61.267158508300781
i=0 j=6, -175.96430969238281 60.267166137695312
i=0 j=7, -175.386962890625 60.267173767089844
i=0 j=8, -175.386962890625 59.267166137695312
i=0 j=9, -174.80960083007812 59.26715087890625
i=0 j=10, -174.80960083007812 58.267166137695312
i=0 j=11, -174.23225402832031 58.267154693603516
i=0 j=12, -175.01954650878906 59.630783081054688
```
The funny thing is, if I feed these numbers back into the C++ unit test, it still passes! Very puzzling.

geoffder:
I still don't think that this is actually a great example or test case, as it involves a shape that is likely self-intersecting at a corner, leading clipper to drop the last part of the shape during clipping operations (not a bug). Also note that the maximum decimal precision on Clipper2 is 8 places (which he have it set to as a constant for now), so if you are relying on barely not having the corner fall on the line by virtue of the values beyond place 8, it could be the source of this problem.

Echoing what @elalish said before, perhaps an easier to read example with round (or close to) numbers would be more appropriate.

elalish:
@geoffder Agreed. However, would you mind detailing this?

> involves a shape that is likely self-intersecting at a corner, leading clipper to drop the last part of the shape during clipping operations

I would generally expect a small overlap to cause only a correspondingly small change in area. Do you know what Clipper's expected behavior is here?

geoffder:
Now that I'm thinking on it, I can't remember if there are particular proscriptions on overlapping corners in Clipper, my thoughts about those are also polluted by them being illegal in a bunch of operations on paths etc in BOSL2/ocadml.

johnnybigert:
Ah ok, I thought the CrossSection class is an "arrangement in the 2d plane" (since the docs for the constructor says `A boolean union operation (with Positive filling rule by default) is performed to ensure the resulting CrossSection is free of self-intersections.`). But is the requirement that the input is a set of valid/simple polygons? In my use of CrossSection, I would have hoped the union of a (slightly) self-intersecting polygon would preserve the majority of the area.

I've tried all fill rules and they all gave the same result.

geoffder:
Interestingly, it comes out as you expect if I use Clipper2 through my OCaml bindings.

```ocaml
module C =
  Clipper.MakeD' (OCADml.V2) (OCADml.Path2) (OCADml.Poly2)
    ((val Clipper.config ~fill_rule:`Positive ~precision:8 ()))

let () =
  let open OCADml in
  let a =
    Path2.of_tups
      [ -176.5416564941429, 62.267156291299955
      ; -177.11900294911175, 63.267144940124126
      ; -177.11900294910944, 62.26716470718384
      ; -176.5416564941406, 62.26715569064245
      ; -176.54165614735393, 61.26716470718384
      ; -175.9643096923828, 61.26715985212647
      ; -175.96430934559382, 60.26716470718384
      ; -175.38696566496216, 60.26717407055541
      ; -175.386962543836, 59.26716470718384
      ; -174.80960083007815, 59.26714979522481
      ; -174.8096004832914, 58.26716470718384
      ; -174.23225470415798, 58.26715427684856
      ; -175.01954376826416, 59.63078409350792
      ]
  and b =
    Path2.of_tups [ -174.70, 58.40; -174.50, 58.40; -174.50, 58.60; -174.70, 58.60 ]
  in
  C.Svg.(
    write
      ~max_height:1000
      ~max_width:1000
      "/home/geoff/sawtooth.png"
      [ subject ~fill_rule:`Positive C.(sub (path a) (path b)) ] )
  |> Result.get_ok
```
sawtooth

bindings/python/examples/missing_polygon.py
elalish: The overlap is the intersection, `^`, not the difference.

elalish:
I simplified this test case down to just a 5-point polygon that loses its lower half, but retains it if a point is moved a smidge to avoid the self-intersection:
```
SimplePolygon polygon = {
      {-7.119003, 13.2671432},  //
      {-7.119003, 12.2671661},  //
      {-4.809601, 9.2671509},   //
      {-4.809601, 8.2671661},   //
      {-4.232254, 8.2671547},   //
  };

  auto cross = CrossSection(polygon);
  EXPECT_NEAR(cross.Area(), 1.44, 0.01);
```

Either Clipper2 has a serious bug or we're using it wrong.

geoffder:
I've just double checked, and I'm on the same commit in my Clipper2 bindings as we are here, so there isn't a difference there. I'll try reproducing your simplified case in those, but since the other one worked, I feel like something is going awry in the number type conversions. I was looking at CrossSection yesterday, and we really aren't doing so much logic in our layer that could go wrong in construction/booleans.

elalish:
Okay, so I've simplified this even more to avoid anything nearly-colinear. It's failing on a simple self-intersecting polygon:
```
SimplePolygon polygon = {
      {-7, 13},   //
      {-7, 12},   //
      {-5, 9},    //
      {-5, 8.1},  //
      {-4.8, 8},  //
  };
```

Reading the Clipper2 [docs](http://www.angusj.com/clipper2/Docs/Overview.htm), it looks like Union can operate on open polylines and give strange results. Stranger still, Paths can apparently be open or closed and it's only known which by context (??). I have the feeling we may be passing in open paths accidentally. Do you want to investigate @geoffder?

geoffder:
No, the top-level helper functions that we are using from Clipper2 only operate on closed subject paths, you have to use the `Clipper2` class directly to be able to imperatively add an open subject path to be clipped.

elalish:
Well, then this looks like a Clipper2 bug. Think you can make a minimal C++ repro for them and file a bug?

geoffder:
I went and updated my Clipper2 C and OCaml bindings to the latest Clipper2 commit and your simplified overlapping edge case works fine now. I'll update us to the latest as well and this issue should go away.

geoffder:
@elalish For your last one, what do you mean by failing? On the updated Clipper2 I get the following:
nonzero_vs_positive
Which is what I would expect.

geoffder:
RE: the loss of precision moving through float, I think it would be appropriate to drop the precision we specify to Clipper2 in CrossSection down to 6 (from the maximum of 8). Those last two places in the double paths aren't really meaningful after upconversion from single precision.

elalish:
Does Clipper use fixed or absolute precision? Internally our tolerance is 1e-5, but that gets scaled by the max size of the bounding box. Probably not a bad idea to include `precision_` in `CrossSection` following the pattern in `Manifold`.

elalish:
I'm going to close this in favor of #464, since it's a simpler. The issue is fixed in #463, thanks to an upstream fix in Clipper2 - remember to use `--recurse-submodules` when you pull, or better yet, `git config --global submodule.recurse true`.

geoffder:
> Does Clipper use fixed or absolute precision? Internally our tolerance is 1e-5, but that gets scaled by the max size of the bounding box. Probably not a bad idea to include `precision_` in `CrossSection` following the pattern in `Manifold`.

For Clipper2 double paths, precision is taken as a parameter on each function, and is used for scaling the doubles into and back from int64. 

Correct me if I'm wrong, but I'm not sure that we'd gain much from dynamically setting this, as `precision_ = 6` in `cross_section.cpp` allows the maximum precision attainable by floats, and doesn't impact the max value headroom (since the maximum is not high enough to leave less than places available for decimal portion).

elalish:
Ah, okay I see what you mean. We could use this as a way to auto-remove sub-precision geometry, but that's probably more trouble than it's worth. We can rely on Manifold's decimator for that.

johnnybigert:
Thanks for looking into it! I'll pull recursively and try again!

PR #461, opened 2023-06-21
Clean up `CrossSection` constructor in pymanifold

geoffder:
Removes `vertices` intermediary and `{vertices}` assignment in the `CrossSection` constructor of pymanifold. I'm not sure why `{vertices}` was compiling obviously, and I'm not sure it'd ever be a problem, but it seemed odd to me to be wrapping the vector in braces as though the `SimplePolygon` was a vector of contours, rather than a single one. Is this clearer, or am I just not used to the cpp way of doing things?


pca006132:
`{vertices}` is interpreted as a copy constructor, and can also be interpreted as initializer list with 1 element. Welcome to C++ :P

pca006132 (approved):

geoffder:
Ahh ok, figured it must be something like that since it was still compiling, thanks for clarification!

PR #463, opened 2023-06-22
Update Clipper2 to latest commit

geoffder:
Clipping self-intersecting/collinear paths received some attention on the commits since the one we were on, which seems to have squashed the bugs observed in #457 and its surrounding discussion. 

There are no changes to our exposed API brought on by this update, with only a minor internal touchup on the binding of `RectClip`.


elalish (approved):
SG, since they have't made a release lately. I general I'd like to stay on releases rather than particular commits of submodules. Let's make a note to update when their next release comes out.

PR #465, opened 2023-06-22
Drop CrossSection precision_ (decimal places) from 8 to 6

geoffder:
Since `glm::vec2` uses single precision floats, the decimal places past the sixth once converted to doubles for Clipper2 aren't really meaningful anyway.

### DIFF
--- a/src/cross_section/src/cross_section.cpp
+++ b/src/cross_section/src/cross_section.cpp
@@ -17,7 +17,7 @@
 using namespace manifold;
 
 namespace {
-const int precision_ = 8;
+const int precision_ = 6;
 
 C2::ClipType cliptype_of_op(OpType op) {
   C2::ClipType ct = C2::ClipType::Union;


src/cross_section/src/cross_section.cpp
```diff
@@ -17,7 +17,7 @@
 using namespace manifold;
 
 namespace {
-const int precision_ = 8;
+const int precision_ = 6;
```
elalish, line 20: What does this precision mean, precisely?

geoffder: Copying my [reply](https://github.com/elalish/manifold/pull/457#issuecomment-1603406206) in #457.

> Does Clipper use fixed or absolute precision? Internally our tolerance is 1e-5, but that gets scaled by the max size of the bounding box. Probably not a bad idea to include precision_ in CrossSection following the pattern in Manifold.


For Clipper2 double paths, precision is taken as a parameter on each function, and is used for scaling the doubles into and back from int64.

Correct me if I'm wrong, but I'm not sure that we'd gain much from dynamically setting this, as precision_ = 6 in cross_section.cpp allows the maximum precision attainable by floats, and doesn't impact the max value headroom (since the maximum is not high enough to leave less than places available for decimal portion).

elalish: If this is an absolute precision, then why change it from 8? floats on the order of 10,000 will only have a precision of 1 or 2, right? So do we even care?

geoffder: Actually, that's not correct I guess. I was thinking of floats vs doubles RE max value, a precision of six still allows 12 or 13 digits for the integer portion of the float to be represented in `int64`. So, dynamically decreasing precision if the integer portion is larger than that would make some sense. Is it worth doing that do you think, or should we just leave it at 6 for now?

elalish: I'm actually less concerned about the top end than the bottom end. If your floats are on the order of 1e-6, then all your precision is gone, no matter what `precision` value you set. I think the only real way to take care of that is by scaling to normalize the bounding box. I wouldn't call it a top priority, but I don't think we want to change from 8 to 6.

geoffder: > If this is an absolute precision, then why change it from 8? floats on the order of 10,000 will only have a precision of 1 or 2, right? So do we even care?

From some (rigorous scientific) testing, the precision does drop from 7 (rounded) down as the integer portion grows.

```cpp
  double d1 = 1.12345678987654321;
  float f1 = d1;
  double d10 = d1 + 10;
  float f10 = d10;
  double d100 = d1 + 100;
  float f100 = d100;
  printf("d1 = %.8f; f1 = %.8f\n", d1, f1);
  printf("d10 = %.8f; f10 = %.8f\n", d10, f10);
  printf("d100 = %.8f; f100 = %.8f\n", d100, f100);
```
```
d1 = 1.12345679; f1 = 1.12345684
d10 = 11.12345679; f10 = 11.12345695
d100 = 101.12345679; f100 = 101.12345886
```

So as my other comment suggests, we only care insofar as we want to enable larger float values (lower precision means they wont cap out a the int64 maximum as early when being scaled from double on the Clipper2 side).

elalish: Sure, but if you set `(1.23e-8, 4.56e-8)` in Clipper2 with `precision=6`, you'll just get zeros back, right?

geoffder: Yes, that's true. From our side (via `glm::vec2`) we'd only be giving Clipper2 `(1e-8, 5e-8)` though, and when we pull the paths back out into float, we are limited to float precision. With 8 precision, we end up limited on the other side (e.g. `1.23e11` hits max `int64` on conversion at `precision_ = 8`, vs `e13` at `6`), with the extra decimal places being meaningless at that magnitude anyway.

This is just making me want to change my mind and make this PR to go with dynamic precision based on the integer part of bounding rect size now.

geoffder:
After thinking about it more through the discussion here, I decided I may as well just go ahead with dynamic precision so we aren't sacrificing values large or small. See #466

PR #466, opened 2023-06-23
Dynamically calculate decimal precision in CrossSection

geoffder:
Following discussion in #465, I decided to just bite the bullet and implement dynamic precision scaling for `CrossSection` so that we can properly enjoy the range of values that single precision floats allow us when moving over the `Clipper2` boundary.

### DIFF
--- a/bindings/wasm/bindings.js
+++ b/bindings/wasm/bindings.js
@@ -182,8 +182,9 @@ Module.setup = function() {
     return (center ? man.translate([0., 0., -height / 2.]) : man);
   };
 
-  Module.CrossSection.prototype.revolve = function(circularSegments = 0) {
-    return Module._Revolve(this, circularSegments);
+  Module.CrossSection.prototype.revolve = function(
+      circularSegments = 0, revolveDegrees = 360.0) {
+    return Module._Revolve(this, circularSegments, revolveDegrees);
   };
 
   Module.CrossSection.prototype.add = function(other) {
@@ -555,11 +556,12 @@ Module.setup = function() {
     return cs.extrude(height, nDivisions, twistDegrees, scaleTop, center);
   };
 
-  Module.Manifold.revolve = function(polygons, circularSegments = 0) {
+  Module.Manifold.revolve = function(
+      polygons, circularSegments = 0, revolveDegrees = 360.0) {
     const cs = (polygons instanceof CrossSectionCtor) ?
         polygons :
         Module.CrossSection(polygons, 'Positive');
-    return cs.revolve(circularSegments);
+    return cs.revolve(circularSegments, revolveDegrees);
   };
 
   Module.Manifold.reserveIDs = function(n) {
--- a/bindings/wasm/manifold-encapsulated-types.d.ts
+++ b/bindings/wasm/manifold-encapsulated-types.d.ts
@@ -91,7 +91,7 @@ export class CrossSection {
    * @param circularSegments Number of segments along its diameter. Default is
    * calculated by the static Defaults.
    */
-  revolve(circularSegments?: number): Manifold;
+  revolve(circularSegments?: number, revolveDegrees?: number): Manifold;
 
   // Transformations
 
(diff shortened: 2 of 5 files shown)


src/cross_section/src/cross_section.cpp
```diff
@@ -132,6 +132,13 @@ void decompose_hole(const C2::PolyTreeD* outline,
   }
 }
 
+int precision(C2::PathsD paths) {
+  auto rect = C2::GetBounds(paths);
+  float size_x = glm::max(fabs(rect.left), fabs(rect.right));
+  float size_y = glm::max(fabs(rect.bottom), fabs(rect.top));
+  int magnitude = floor(log10(glm::max(size_x, size_y)));
+  return std::max(0, max_precision - magnitude);
```
elalish, line 140: I think this will only fix float magnitudes between 1 and 1e8, anything larger or smaller will still loose precision. Floats can go from 1e38 to 1e-38. Given how Clipper2 set up their API, I think we have to physically scale our points to normalize them, then rescale the result. The lazy transformations should allow this to be reasonably efficient for repeated operations, but I'm honestly not that concerned about the speed of our 2D subsystem, compared to the 3D ops.

elalish (changes requested):

src/cross_section/src/cross_section.cpp
geoffder: I probably need to read more, but it seemed like when I was messing around earlier that the floats were imprecise on the integer part of large values anyway (when scaling the floats up into int64 according to their precision, some of the lower digits were floating).

geoffder: Ah right, I think I get what you meant, scaling down numbers to be representable as int64 then scaling them back up.

```cpp
float maxf = std::numeric_limits<float>::max();
int64_t maxf_i = maxf; 
printf("max float = %.f\n", maxf);
printf("max float to int64 = %li\n", maxf_i);
```
```
max float = 340282346638528859811704183484516925440
max float to int64 = 9223372036854775807
```

Wasn't thinking about it so much from the other direction.

geoffder: If we do lazy scaling for this, I think it will have to be separate from the stored affine matrix, as the matrix needs to be applied before any Clipper2 operations. We would just want to scale back up when getting the values out / extruding. I might as well switch us over to using the int64 API rather to spare the scaling conversions on the Clipper2 side if we go that route.

geoffder: Actually, my eyes glazed over it earlier without absorbing it, but Clipper does actually allow negative precision down to `-8`. 🤦🏻‍♂️

geoffder: Did a bit more reading, and learned that 32-bit floats store integers accurately up to 16777217 (2^24 + 1) (RE: the integer part floating at large numbers). 

Anyway, do you think that dynamic precision between -8 and 8, adding (+1) to the magnitude to make `0.0123` give a precision of 8, and checking if `size = 0` to avoid a runtime error, would make this a good enough improvement for now? We'd get a lot more of the float range than we are now at least. Or, would you rather I go straight to using int64 and scaling on our side?

elalish: I dunno, 8 out of 38 feels pretty small. I think if we're fixing this, we should fix it right, so that we can tell people floats just work and they don't have to think about how big their units are. These confusions are exactly why I dislike absolute precision.

geoffder:
I've [discussed](https://github.com/AngusJohnson/Clipper2/discussions/564) the precision limits over at Clipper2 with Angus, and he's going to make it configurable in his next update. I prepared and offered a PR but he said he was going to just do it for all three language ports himself, mine was only for CPP.

So once that is in, I'll update the Clipper2 commit once again, and adjust this PR according to the Clipper2 accurate coordinate range brought up in the discussion over there.

elalish:
Okay, so this PR will wait for the clipper update, SG. What kind of configuration is he adding? Someone still needs to check the bounding box to determine the right precision range - is that going to happen on our side or Clipper's?

geoffder:
On our side, he is just making the allowable range of `precision` values to be configured, so there isn't an artificial limit like there is now. The bounding box check will be like I have it now in the PR, adjusted to make sure that not only do we not clip our values against `int64` max, but also stay in the accurate range for Clipper2's algorithms (choosing suitably large negative precision values).

elalish:
Okay, sounds great, thanks!

geoffder:
Since I accidentally used my master to do this and I wanted to fix up my fork, I'll just re-do this when the Clipper2 changes needed are in upstream.

PR #470, opened 2023-06-25
Update HTML examples for WASM API changes

elalish:
I just noticed we forgot to update our example pages (e.g. https://manifoldcad.org/model-viewer.html) when we changed the WASM API. Would be nice to add some simple CI testing to these pages, but I can't think of an easy way.


geoffder:
Ah I don't think I've ever seen that page! Getting it into CI would help avoid same in future for sure, I'm afraid I know zero about web especially in this context though.

PR #473, opened 2023-06-27
fix nixos js build

### LINKED ISSUE #424: Nix flake update
We should clean up the flake, build the project appropriately, and perhaps upstream to nixpkgs.
Should also update emscripten once https://github.com/NixOS/nixpkgs/pull/229718 is landed.

pca006132:
This fixes the broken wasm build on nixos. Closes #424.

@elalish I am not sure if js_deps should depend on `${CMAKE_CURRENT_BINARY_DIR}/manifold.*`, as this causes error when I run the build.
```
*** No rule to make target 'bindings/wasm/manifold.*', needed by 'bindings/wasm/CMakeFiles/js_deps'.  Stop.
```

### DIFF
--- a/bindings/wasm/manifold-encapsulated-types.d.ts
+++ b/bindings/wasm/manifold-encapsulated-types.d.ts
@@ -113,16 +113,12 @@ export class CrossSection {
   translate(v: Vec2): CrossSection;
 
   /**
-   * Applies an Euler angle rotation to the cross-section, first about the X
-   * axis, then Y, then Z, in degrees. We use degrees so that we can minimize
-   * rounding error, and eliminate it completely for any multiples of 90
-   * degrees. Additionally, more efficient code paths are used to update the
-   * cross-section when the transforms only rotate by multiples of 90 degrees.
-   * This operation can be chained. Transforms are combined and applied lazily.
+   * Applies a (Z-axis) rotation to the CrossSection, in degrees. This operation
+   * can be chained. Transforms are combined and applied lazily.
    *
-   * @param v [X, Y, Z] rotation in degrees.
+   * @param degrees degrees about the Z-axis to rotate.
    */
-  rotate(v: Vec2): CrossSection;
+  rotate(v: number): CrossSection;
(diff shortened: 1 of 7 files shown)


pca006132:
Ah this still doesn't work. Probably need some more time to figure it out.

pca006132:
Instead of depending on compiled outputs in `${CMAKE_CURRENT_BINARY_DIR}/manifold.*`, we should depend on the target `manifoldjs` instead.

pca006132:
Also, I'm wondering if we should modify the build process to copy files in examples to the build directory, instead of copying the built output to src directory. That way we don't need any special gitignore rules to ignore directories in `bindings`.

bindings/wasm/manifold-encapsulated-types.d.ts
```diff
@@ -113,16 +113,12 @@ export class CrossSection {
...
-   * axis, then Y, then Z, in degrees. We use degrees so that we can minimize
-   * rounding error, and eliminate it completely for any multiples of 90
-   * degrees. Additionally, more efficient code paths are used to update the
-   * cross-section when the transforms only rotate by multiples of 90 degrees.
-   * This operation can be chained. Transforms are combined and applied lazily.
+   * Applies a (Z-axis) rotation to the CrossSection, in degrees. This operation
+   * can be chained. Transforms are combined and applied lazily.
    *
-   * @param v [X, Y, Z] rotation in degrees.
+   * @param degrees degrees about the Z-axis to rotate.
```
elalish, line 119: Thanks for the updates!

elalish (approved):
>Also, I'm wondering if we should modify the build process to copy files in examples to the build directory, instead of copying the built output to src directory. That way we don't need any special gitignore rules to ignore directories in bindings.

I think that would mess up how Vite works for JS development. But if I misunderstand, maybe open a new PR and we can discuss?

pca006132:
I am not familiar with Vite, I guess we can just leave it as it for now.

PR #478, opened 2023-07-05
Remove RectClip API + Rect::AsCrossSection fix

geoffder:
Following discussion on https://github.com/elalish/manifold/pull/455, this removes `RectClip` from our API, as it does not provide the same guarantees that Clipper2's regular boolean clipping operations give us (that make sure that `CrossSection`s can be extruded successfully).

Also, this fixes `AsCrossSection` which was returning empty `CrossSection`s due to incorrect winding followed by a `Positive` fill rule union operation. Now it is working as advertised, while skipping over the needless clipping by implementing it as a `CrossSection` constructor (building a `PathsD` directly).

### DIFF
--- a/bindings/c/cross.cpp
+++ b/bindings/c/cross.cpp
@@ -106,13 +106,6 @@ ManifoldCrossSection *manifold_cross_section_intersection(
   return to_c(new (mem) CrossSection(cs));
 }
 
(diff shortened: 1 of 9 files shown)


elalish (approved):
Perfect, thanks!

PR #482, opened 2023-07-07
Nomad

stephomi:


stephomi:
I updated the glm dependency but it failed.
I'm on Mac so I couldn't replicate on my machine so I pass, maybe a cuda-only (I think it was an simple glm warning)

elalish:
Can you give a little more context for this PR? Is this related to our CMake discussion? What problem is this solving (not trying to push back, just honestly curious)?

stephomi:
Yep, it's related but after checking more carefully I don't think there is need for the PR, you can probably close it.

---
I integrated manifold with `add_subdirectory(lib/manifold)` but my project failed to compile since any `glm/glm.hpp` include was taking the one from manifold (which is 1 year old), because manifold is linked with `PUBLIC`.

The issue was that I was adding my own `glm` with `target_include_directories(nomad SYSTEM PUBLIC lib/glm)`.
I use `SYSTEM` to easily shut down spammy warning for header-only library.
But it also does other thing, like giving it a lower priority or something, my bad!
Removing it the system fixes my issue.

elalish:
Thanks for the explanation!

Issue #483, opened 2023-07-07
Optimization (std::map, etc)


stephomi:
`std::map` should only be used if ordering needs to be preserved.

Case study: subtract a 150k verts mesh on a 400k verts one.
The entire getMeshGL() took 770ms, by replacing 2 minor thing I got it down to <600ms.

By testing this section (175ms): https://github.com/elalish/manifold/blob/96c57c0591a9081d1e3f26d31a57c603f46e11e8/src/manifold/src/manifold.cpp#L287

```diff
diff --git a/src/manifold/src/manifold.cpp b/src/manifold/src/manifold.cpp
index 48efe44..51356ee 100644
--- a/src/manifold/src/manifold.cpp
+++ b/src/manifold/src/manifold.cpp
@@ -21,6 +21,10 @@
 #include "impl.h"
 #include "par.h"
 
+#include "../parallel-hashmap/parallel_hashmap/phmap.h"
+#include <glm/gtx/hash.hpp>
+#include <unordered_map>
+
 namespace {
 using namespace manifold;
 using namespace thrust::placeholders;
@@ -284,7 +288,8 @@ MeshGL Manifold::GetMeshGL(glm::ivec3 normalIdx) const {
 
   // Duplicate verts with different props
   std::vector<int> vert2idx(impl.NumVert(), -1);
-  std::map<std::pair<int, int>, int> vertPropPair;
+  // std::unordered_map<glm::ivec2, int> vertPropPair;
+  phmap::flat_hash_map<glm::ivec2, int> vertPropPair;
   for (int run = 0; run < out.runOriginalID.size(); ++run) {
     for (int tri = out.runIndex[run] / 3; tri < out.runIndex[run + 1] / 3;
          ++tri) {
@@ -294,14 +299,10 @@ MeshGL Manifold::GetMeshGL(glm::ivec3 normalIdx) const {
         const int prop = triProp[i];
         const int vert = out.triVerts[3 * tri + i];
 
-        const auto it = vertPropPair.find({vert, prop});
-        if (it != vertPropPair.end()) {
-          out.triVerts[3 * tri + i] = it->second;
-          continue;
-        }
         const int idx = out.vertProperties.size() / out.numProp;
-        vertPropPair.insert({{vert, prop}, idx});
-        out.triVerts[3 * tri + i] = idx;
+        const auto it = vertPropPair.try_emplace({vert, prop},idx);
+        out.triVerts[3 * tri + i] = it.first->second;
+        if (!it.second) continue;
[...]

PR #480, opened 2023-07-06
Fix operations with different numbers of properties

elalish:
Fixes #479 

There were two related bugs here: one in the Boolean and one in Compose. Compose was not handling properties at all (oops!), while the Boolean was assuming both inputs had the same number of properties. I added several tests inspired by the linked issue, but made them quite a bit more comprehensive.

PR #484, opened 2023-07-08
Replace map with unordered_map

stephomi:
Context: https://github.com/elalish/manifold/issues/483
Case study: subtract a 150k verts mesh on a 1.5M verts one (with UVs).

The edge map is one of the bottleneck when doing boolean on high poly.

Total time of `GetMeshGL()`:

|       -              |      Time     |
| -------------------- | ------------- |
| std::map (current)   | 2555 ms       |
| std::unordered_map   | 1985 ms       |
| phmap::flat_hash_map | 1450 ms       |
| (without UV)         | 1000 ms       |

phmap is header-only and from https://github.com/greg7mdp/parallel-hashmap.
For now I didn't include it, but the performance is not negligeable.
Should I do it? (git submodule? just copying the include directory?)

### DIFF
--- a/src/manifold/src/boolean_result.cpp
+++ b/src/manifold/src/boolean_result.cpp
@@ -481,7 +481,10 @@ void CreateProperties(Manifold::Impl &outR, const VecDH<TriRef> &refPQ,
                           inP.halfedge_.cptrD(), inQ.halfedge_.cptrD(),
                           outR.halfedge_.cptrD(), outR.precision_}));
 
-  std::map<std::tuple<int, int, int, int>, int> propIdx;
+  using Entry = std::pair<glm::ivec3, int>;
+  int idMissProp = outR.NumVert();
+  std::vector<std::vector<Entry>> propIdx(outR.NumVert() + 1);
+  outR.meshRelation_.properties.reserve(outR.NumVert() * numProp);
   int idx = 0;
 
   for (int tri = 0; tri < numTri; ++tri) {
@@ -499,16 +502,16 @@ void CreateProperties(Manifold::Impl &outR, const VecDH<TriRef> &refPQ,
 
     for (const int i : {0, 1, 2}) {
       const int vert = outR.halfedge_[3 * tri + i].startVert;
-      const glm::vec3 uvw = bary[3 * tri + i];
+      const glm::vec3 &uvw = bary[3 * tri + i];
 
-      auto key = std::make_tuple(PQ, -1, -1, -1);
+      glm::ivec4 key(PQ, idMissProp, -1, -1);
       if (oldNumProp > 0) {
-        std::get<1>(key) = vert;
+        key[1] = vert;
         int edge = -1;
         for (const int j : {0, 1, 2}) {
           if (uvw[j] == 1) {
             // On a retained vert, the propVert must also match
-            std::get<2>(key) = triProp[j];
+            key[2] = triProp[j];
             edge = -1;
             break;
           }
@@ -518,18 +521,23 @@ void CreateProperties(Manifold::Impl &outR, const VecDH<TriRef> &refPQ,
           // On an edge, both propVerts must match
           const int p0 = triProp[Next3(edge)];
           const int p1 = triProp[Prev3(edge)];
-          std::get<2>(key) = glm::min(p0, p1);
-          std::get<3>(key) = glm::max(p0, p1);
+          key[2] = glm::min(p0, p1);
+          key[3] = glm::max(p0, p1);
         }
       }
 
-      const auto it = propIdx.find(key);
-      if (it != propIdx.end()) {
-        outR.meshRelation_.triProperties[tri][i] = it->second;
-        continue;
+      auto &bin = propIdx[key.y];
+      bool bFound = false;
+      for (int k = 0; k < bin.size(); ++k) {
+        if (bin[k].first == glm::ivec3(key.x, key.z, key.w)) {
+          bFound = true;
+          outR.meshRelation_.triProperties[tri][i] = bin[k].second;
+          break;
+        }
       }
-      outR.meshRelation_.triProperties[tri][i] = idx;
-      propIdx.insert({key, idx++});
+      if (bFound) continue;
+      bin.push_back(std::make_pair(glm::ivec3(key.x, key.z, key.w), idx));
+      outR.meshRelation_.triProperties[tri][i] = idx++;
 
       for (int p = 0; p < numProp; ++p) {
         if (p < oldNumProp) {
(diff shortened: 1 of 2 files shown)


elalish:
Thanks for the perf data! What device are you testing on and which backend are you using? Would you mind adding a comparison with the hash table we already use? https://github.com/elalish/manifold/issues/483#issuecomment-1626768739

First, because I'd like to benchmark its performance anyway, and second, because it'd be nice to use the same hash table everywhere if we find a favorite.

stephomi:
I made another [branch](https://github.com/stephomi/manifold/tree/test_map) to test.
If you want to test, you can simply check it out, everything is included.
(to switch between method, there is a `#define` at the top of `hashtable.h`).

Device is MacBookAir M1, backend is TBB but it doesn't matter since the two sections of the code I changed are both serial.
It's the same test (1.5M and 150k subtract).

| -                              | manifold.cpp  | boolean_result.cpp |
| ------------------------------ |-------------- | ------------------ |
| std::map (master)              | 670ms         | 782ms              |
| std::unordered_map             | 227ms         | 437ms              |
| phmap::flat_hash_map           | 120ms         | 153ms              |
| hashtable.h                    | 165ms         | 200ms              |
| hashtable.h + identity         | 50ms          | -                  |
| vector<anker::svector<>>       | 50ms          | 74ms               |
| (without UV)                   | 0ms           | 0ms                |

For now a simple `vector<vector>>` linear search gave the best result (hashing mesh edges).
Maybe it's possible to improve the table result by tweaking the hash, but the vector solution is likely the best one (performance and memory as well)

---
Not sure if I understand how hashtable.h is working but it's not a drop-in replacement to map/unordered_map.
I templated it (very quick-n-dirty way) to handle arbirtrary key, not just u64.
Also the hashtable is not dynamic and you need to provide a correct size at first.
Not sure what's the expected value but when I gave the exact number of entries, it crashed.
So I'm allocating twice the number of entries.

elalish:
Indeed, given the high likelihood of degree-6 verts, I think your vector of vectors approach makes a lot of sense. I'm reading about the small vector lib, but I don't quite understand why it's an improvement over a std::vector? If std::vector gives about the same perf, let's just use that (I figure the compilers have lots of experience optimizing std::vector).

pca006132:
I think those small vector libraries usually flatten the vector and perform some small size optimizations, so to reduce the number of pointer chasing when accessing an element (at the expense of more expensive move etc.). I thought I modified map with unordered_map a few months ago but it seems that this is reintroduced for other parts of the code, have to say that the naming of std::map is unfortunate...

It seems to me that using vector of vectors make sense in this case. I just got back home and is now dealing with jet lag, will look at it in more details tmr.

stephomi:
> I figure the compilers have lots of experience optimizing std::vector

It's not a compiler thing but purely a malloc thing.
In my test I got ~1.5M elements, so it's at least 1.5M malloc.
It's ok if the malloc is smart enough to give region close together.

There's a slight gain with svo, but I'm fine with std::vector<std::vector though!
I mean getMeshGL went from `2555ms` -> `1173ms` just because of the extra prop (UV)!
Without the extra prop, it was `1050ms` so I won't nitpick about ~20ms.

| -                              | manifold.cpp  | boolean_result.cpp |
| ------------------------------ |-------------- | ------------------ |
| std::vector<anker::svector<>>  | 50ms          | 74ms               |
| std::vector<std::vector<>>     | 67ms          | 80ms               |
| (without UV)                   | 0ms           | 0ms                |

Fyi `boolean_result.cpp`, here are the average size on my test (only 5 malloc)
```
size == 0 (2900)
size == 1 (1464930)
size == 2 (4932)
size == 3 (1734)
size == 4 (5)
size >= 5 (0)
```

For `manifold.cpp` I can reduce it down to 40ms (using only `std::vector<int>`) because we visit only one of `prop` neighbor.
Maybe double check that this assumption is correct.

I updated the PR, with, if you want svector, just tell me and I'll update it.

hrgdavor:
This looks significant. How moch perf gains are we looking here?

src/manifold/src/boolean_result.cpp
```diff
@@ -499,16 +501,15 @@ void CreateProperties(Manifold::Impl &outR, const VecDH<TriRef> &refPQ,
 
     for (const int i : {0, 1, 2}) {
       const int vert = outR.halfedge_[3 * tri + i].startVert;
-      const glm::vec3 uvw = bary[3 * tri + i];
+      const glm::vec3 &uvw = bary[3 * tri + i];
 
-      auto key = std::make_tuple(PQ, -1, -1, -1);
+      glm::ivec4 key(PQ, vert, -1, -1);
       if (oldNumProp > 0) {
-        std::get<1>(key) = vert;
```
elalish, line 506: These lines are something I just changed in #480 that I think we want to keep. The fact the tests still pass implies I need to add a test. This is to keep extra properties from being created when a mesh without properties is attached a mesh with properties.

elalish (changes requested):
Excellent, thank you for a significant speed up when handling properties! And also for not introducing an extra dependency. One fix and I think it's ready.

src/manifold/src/boolean_result.cpp
stephomi: Ok but maybe I'll need to revert to hashmap for `manifold.cpp` as it breaks most assumptions.
I'll check.

pca006132: One way would be to use plain array for small size hashmap and hashmap for larger size.

pca006132:
> This looks significant. How moch perf gains are we looking here?

Seems pretty large when your mesh uses UV mapping.

src/manifold/src/boolean_result.cpp
stephomi: It's fine, I can still use the `std::vector<std::vector` approach.
I made the previous assumption that one prop could belong to only 1 logical vertex but it's not the case (although gets flattened in the end).

src/manifold/src/boolean_result.cpp
```diff
@@ -481,7 +481,10 @@ void CreateProperties(Manifold::Impl &outR, const VecDH<TriRef> &refPQ,
                           inP.halfedge_.cptrD(), inQ.halfedge_.cptrD(),
                           outR.halfedge_.cptrD(), outR.precision_}));
 
-  std::map<std::tuple<int, int, int, int>, int> propIdx;
+  using Entry = std::pair<glm::ivec3, int>;
+  int idMissProp = outR.NumVert();
+  std::vector<std::vector<Entry>> propIdx(outR.NumVert() + 1);
+  outR.meshRelation_.properties.reserve(outR.NumVert() * numProp * 1.1);
```
elalish, line 487: What is `* 1.1` for?

elalish (approved):
Looks great, thanks!

src/manifold/src/boolean_result.cpp
stephomi: I've removed it.

PR #486, opened 2023-07-10
Add 2d hull operation

geoffder:
This adds a 2d convex hull operation to CrossSection (ported from OCADml, which I ported based on BOSL2). I'll do similar for Manifold (3d convex hull) next.

### DIFF
--- a/src/cross_section/src/cross_section.cpp
+++ b/src/cross_section/src/cross_section.cpp
@@ -132,30 +132,26 @@ void decompose_hole(const C2::PolyTreeD* outline,
   }
 }
 
-bool v2_lesser(glm::vec2 a, glm::vec2 b) {
+bool V2Lesser(glm::vec2 a, glm::vec2 b) {
   if (a.x == b.x) return a.y < b.y;
   return a.x < b.x;
 }
 
 double cross(glm::vec2 a, glm::vec2 b) { return a.x * b.y - a.y * b.x; }
 
-double distance(glm::vec2 a, glm::vec2 b) {
-  auto d = a - b;
-  return std::sqrt(d.x * d.x + d.y * d.y);
-}
-
-bool is_cw(glm::vec2 a, glm::vec2 b, glm::vec2 c) {
+bool IsCw(glm::vec2 a, glm::vec2 b, glm::vec2 c) {
   double lhs = cross(a - c, b - c);
-  double rhs = 1e-9 * distance(a, c) * distance(b, c);
-  return lhs < -rhs;
+  double rhs = 1e-9 * glm::distance(a, c) * glm::distance(b, c);
+  // return lhs < -rhs;
+  return lhs <= rhs;
 }
 
-void backtrack(const SimplePolygon& pts, const int idx, std::vector<int>& keep,
-               const int hold) {
+void HullBacktrack(const SimplePolygon& pts, const int idx,
+                   std::vector<int>& keep, const int hold) {
   const int stop = keep.size() - hold;
   int i = 0;
-  while (i < stop && !is_cw(pts[idx], pts[keep[keep.size() - 1]],
-                            pts[keep[keep.size() - 2]])) {
+  while (i < stop && !IsCw(pts[idx], pts[keep[keep.size() - 1]],
+                           pts[keep[keep.size() - 2]])) {
     keep.pop_back();
     i++;
   }
@@ -163,19 +159,19 @@ void backtrack(const SimplePolygon& pts, const int idx, std::vector<int>& keep,
 
 // Based on method described here:
 // https://www.hackerearth.com/practice/math/geometry/line-sweep-technique/tutorial/
-C2::PathD hull(SimplePolygon& pts) {
+C2::PathD HullImpl(SimplePolygon& pts) {
   int len = pts.size();
   if (len < 2) return C2::PathD();
-  std::sort(pts.begin(), pts.end(), v2_lesser);
+  std::sort(pts.begin(), pts.end(), V2Lesser);
   auto keep = std::vector<int>{0, 1};
   for (int i = 2; i < len; i++) {
-    backtrack(pts, i, keep, 1);
+    HullBacktrack(pts, i, keep, 1);
     keep.push_back(i);
   }
   int n_lower = keep.size();
   for (int i = 0; i < len - 1; i++) {
     int idx = len - 2 - i;
-    backtrack(pts, idx, keep, n_lower);
+    HullBacktrack(pts, idx, keep, n_lower);
     if (idx > 0) keep.push_back(idx);
   }
   auto path = C2::PathD(keep.size());
@@ -626,7 +622,7 @@ CrossSection CrossSection::Hull(
       }
     }
   }
-  return CrossSection(C2::PathsD{hull(pts)});
+  return CrossSection(C2::PathsD{HullImpl(pts)});
 }
 
 CrossSection CrossSection::Hull() const {
@@ -634,8 +630,8 @@ CrossSection CrossSection::Hull() const {
 }
 
 CrossSection CrossSection::Hull(const SimplePolygon poly) {
-  auto pts = poly;  // hull will sort inplace, so we copy here
-  return CrossSection(C2::PathsD{hull(pts)});
+  auto pts = poly;  // HullImpl will sort inplace, so we copy here
+  return CrossSection(C2::PathsD{HullImpl(pts)});
 }
 
 CrossSection CrossSection::Hull(const Polygons polys) {
--- a/src/cross_section/src/cross_section.cpp
+++ b/src/cross_section/src/cross_section.cpp
@@ -168,10 +168,10 @@ C2::PathD HullImpl(SimplePolygon& pts) {
     HullBacktrack(pts, i, keep, 1);
     keep.push_back(i);
   }
-  int n_lower = keep.size();
+  int nLower = keep.size();
   for (int i = 0; i < len - 1; i++) {
     int idx = len - 2 - i;
-    HullBacktrack(pts, idx, keep, n_lower);
+    HullBacktrack(pts, idx, keep, nLower);
     if (idx > 0) keep.push_back(idx);
   }
   auto path = C2::PathD(keep.size());
--- a/src/cross_section/src/cross_section.cpp
+++ b/src/cross_section/src/cross_section.cpp
@@ -161,7 +161,7 @@ void HullBacktrack(const SimplePolygon& pts, const int idx,
 // https://www.hackerearth.com/practice/math/geometry/line-sweep-technique/tutorial/
 C2::PathD HullImpl(SimplePolygon& pts) {
   int len = pts.size();
-  if (len < 2) return C2::PathD();
+  if (len < 3) return C2::PathD();  // not enough points to create a polygon
   std::sort(pts.begin(), pts.end(), V2Lesser);
   auto keep = std::vector<int>{0, 1};
   for (int i = 2; i < len; i++) {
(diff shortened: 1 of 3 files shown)


src/cross_section/src/cross_section.cpp
```diff
@@ -132,6 +132,58 @@ void decompose_hole(const C2::PolyTreeD* outline,
...
+double cross(glm::vec2 a, glm::vec2 b) { return a.x * b.y - a.y * b.x; }
+
+double distance(glm::vec2 a, glm::vec2 b) {
+  auto d = a - b;
+  return std::sqrt(d.x * d.x + d.y * d.y);
+}
+
+bool is_cw(glm::vec2 a, glm::vec2 b, glm::vec2 c) {
+  double lhs = cross(a - c, b - c);
+  double rhs = 1e-9 * distance(a, c) * distance(b, c);
```
elalish, line 149: Can you skip the `sqrt`s by squaring both sides?

test/cross_section_test.cpp
```diff
@@ -167,3 +169,24 @@ TEST(CrossSection, FillRule) {
...
 }
+
+TEST(CrossSection, Hull) {
+  auto circ = CrossSection::Circle(10);
+  auto circs = std::vector<CrossSection>{circ, circ.Translate({0, 30}),
+                                         circ.Translate({30, 0})};
+  auto circ_tri = CrossSection::Hull(circs);
+  auto centres = SimplePolygon{{0, 0}, {0, 30}, {30, 0}, {15, 5}};
+  auto tri = CrossSection::Hull(centres);
+  auto circ_tri_ex = Manifold::Extrude(circ_tri, 10);
```
elalish, line 180: If this isn't part of the test, you can move it into the export define.

test/cross_section_test.cpp
```diff
@@ -167,3 +169,24 @@ TEST(CrossSection, FillRule) {
...
+
+#ifdef MANIFOLD_EXPORT
+  if (options.exportModels)
+    ExportMesh("cross_section_hull_circ_tri.glb", circ_tri_ex.GetMesh(), {});
+#endif
+
+  auto circ_area = circ.Area();
+  EXPECT_FLOAT_EQ(circ_area, (circ - circ.Scale({0.8, 0.8})).Hull().Area());
+  EXPECT_FLOAT_EQ(
+      circ_area * 2.5,
```
elalish, line 190: Where does the 2.5 come from?

src/cross_section/src/cross_section.cpp
```diff
@@ -132,6 +132,58 @@ void decompose_hole(const C2::PolyTreeD* outline,
...
+  return lhs < -rhs;
```
elalish, line 150: If you have a bunch of points that are colinear within precision with their neighbors, but slowly curve convex or concave, which one will get incorrect results? What's the worst case?

elalish (changes requested):
Good idea, thanks!

src/cross_section/src/cross_section.cpp
```diff
@@ -132,6 +132,58 @@ void decompose_hole(const C2::PolyTreeD* outline,
...
+bool v2_lesser(glm::vec2 a, glm::vec2 b) {
+  if (a.x == b.x) return a.y < b.y;
+  return a.x < b.x;
+}
+
+double cross(glm::vec2 a, glm::vec2 b) { return a.x * b.y - a.y * b.x; }
+
+double distance(glm::vec2 a, glm::vec2 b) {
+  auto d = a - b;
+  return std::sqrt(d.x * d.x + d.y * d.y);
```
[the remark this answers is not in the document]
geoffder: Ah it does look like they have distance actually, but cross is for `vec3`, so I think I'll keep mine to spare constructing the extra vecs.

src/cross_section/src/cross_section.cpp
```diff
@@ -560,6 +612,42 @@ CrossSection CrossSection::Offset(double delta, JoinType jointype,
   return CrossSection(ps);
 }
 
+CrossSection CrossSection::Hull(
+    const std::vector<CrossSection>& crossSections) {
```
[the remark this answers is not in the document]
geoffder: We do have `Decompose` for `CrossSection`, and `Compose` is just an alias for a batch union (to make sure that the output is still a valid `CrossSection`). Allowing a set of `CrossSections` to be hulled like here is nice since then a needless clipping operation is avoided.

src/cross_section/src/cross_section.cpp
```diff
@@ -560,6 +612,42 @@ CrossSection CrossSection::Offset(double delta, JoinType jointype,
...
+  }
+  return CrossSection(C2::PathsD{hull(pts)});
+}
+
+CrossSection CrossSection::Hull() const {
+  return Hull(std::vector<CrossSection>{*this});
+}
+
+CrossSection CrossSection::Hull(const SimplePolygon poly) {
+  auto pts = poly;  // hull will sort inplace, so we copy here
```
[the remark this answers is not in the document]
geoffder: Ah right, I'll drop the line then :+1:. Wasn't thinking of that.

src/cross_section/src/cross_section.cpp
```diff
@@ -132,6 +132,58 @@ void decompose_hole(const C2::PolyTreeD* outline,
...
+}
+
+double cross(glm::vec2 a, glm::vec2 b) { return a.x * b.y - a.y * b.x; }
+
+double distance(glm::vec2 a, glm::vec2 b) {
+  auto d = a - b;
+  return std::sqrt(d.x * d.x + d.y * d.y);
+}
+
+bool is_cw(glm::vec2 a, glm::vec2 b, glm::vec2 c) {
```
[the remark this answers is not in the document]
geoffder: Ok, I'll try to fix my snake habit.

src/cross_section/src/cross_section.cpp
```diff
@@ -132,6 +132,58 @@ void decompose_hole(const C2::PolyTreeD* outline,
...
+  auto keep = std::vector<int>{0, 1};
+  for (int i = 2; i < len; i++) {
+    backtrack(pts, i, keep, 1);
+    keep.push_back(i);
+  }
+  int n_lower = keep.size();
+  for (int i = 0; i < len - 1; i++) {
+    int idx = len - 2 - i;
+    backtrack(pts, idx, keep, n_lower);
+    if (idx > 0) keep.push_back(idx);
```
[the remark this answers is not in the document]
geoffder: The 3d hull algorithm I'm using in OCADml does have some similarity to this yea. Next up is porting it over like I did this 2d one.

src/cross_section/src/cross_section.cpp
geoffder: Ahh good idea.

test/cross_section_test.cpp
geoffder: The hull triangle cuts a quarter out of the circle on the origin, and an eight out of the other two.

src/cross_section/src/cross_section.cpp
geoffder: I'm not totally sure, but this is the more strict version that doesn't include collinear points in the solution, `lhs <= rhs` is supposed to keep them (going by the note in my docs, which was probably based off of BOSL2 since that is what I followed when implementing it in OCaml). The more lax conditional might be more robust to the case that you describe.

src/cross_section/src/cross_section.cpp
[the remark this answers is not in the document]
geoffder: Oh right, the reason why I did it is that I actually get an error because the parameter is const and I'm passing it to the internal `hull` which is not `const` (because it needs to sort).

src/cross_section/src/cross_section.cpp
geoffder: I tried and didn't get it to work, not sure if I'm missing something. I've switch to `glm::distance` anyway, which works and lets me drop the helper.

src/cross_section/src/cross_section.cpp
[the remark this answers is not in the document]
elalish: No need to call it const when it's not taken by reference :).

src/cross_section/src/cross_section.cpp
elalish: Maybe because you lose the sign of the cross? Could try `cross * abs(cross)`.

test/cross_section_test.cpp
elalish: ah, as in 3 - 0.5 = 2.5 - I get it now, thanks.

src/cross_section/src/cross_section.cpp
elalish: Anyway, might make for a good test. We had the bug earlier on a high-res sphere where my decimator epic-failed. A high-res circle might be good here, though using doubles will make it pretty hard - maybe just a secant.

src/cross_section/src/cross_section.cpp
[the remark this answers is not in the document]
elalish: I guess, but they can already make a vector of polygons pretty easily. I'd like to keep our API small where possible.

src/cross_section/src/cross_section.cpp
geoffder: Right, of course. Tests are passing with this:
```cpp
double SquaredDistance(glm::vec2 a, glm::vec2 b) {
  auto d = a - b;
  return d.x * d.x + d.y * d.y;
}

bool IsCw(glm::vec2 a, glm::vec2 b, glm::vec2 c) {
  double lhs = Cross(a - c, b - c);
  double rhs = 1e-18 * SquaredDistance(a, c) * SquaredDistance(b, c);
  return lhs * std::abs(lhs) <= rhs;
}
```

src/cross_section/src/cross_section.cpp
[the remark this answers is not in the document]
elalish: Nice job on `Decompose` by the way - I'd totally forgotten you already put that in. 💯

src/cross_section/src/cross_section.cpp
geoffder: Got it to fail (a little) for `circularSegments=1e6`.
```
/home/geoff/git/manifold/test/cross_section_test.cpp:189: Failure
  circ_area
    Which is: 314.15927
  (circ - circ.Scale({0.8, 0.8})).Hull().Area()
    Which is: 314.15906
```

src/cross_section/src/cross_section.cpp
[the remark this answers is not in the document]
geoffder: > I guess, but they can already make a vector of polygons pretty easily. I'd like to keep our API small where possible.

I think this ones worth having. A lot of basic use case for 2d hull is making rounding thing to extrude, and being able to do that directly with some Circles is pretty nice.

src/cross_section/src/cross_section.cpp
```diff
@@ -132,50 +132,50 @@ void decompose_hole(const C2::PolyTreeD* outline,
...
 }
 
-double cross(glm::vec2 a, glm::vec2 b) { return a.x * b.y - a.y * b.x; }
+double Cross(glm::vec2 a, glm::vec2 b) { return a.x * b.y - a.y * b.x; }
 
-double distance(glm::vec2 a, glm::vec2 b) {
+double SquaredDistance(glm::vec2 a, glm::vec2 b) {
   auto d = a - b;
-  return std::sqrt(d.x * d.x + d.y * d.y);
+  return d.x * d.x + d.y * d.y;
```
elalish, line 144: glm::dot(d, d) works too.

elalish (approved):
Thanks!

src/cross_section/src/cross_section.cpp
geoffder: 🤦‍♂️

Issue #361, opened 2023-03-10
Add python bindings for the rest of our APIs


elalish:
I think we've fallen behind a bit...

pca006132:
Still require some follow up

elalish:
After #445, how are we feeling about the completeness of the python bindings? Looks like there are a few that could use docs, and there's nothing for MeshGL yet. @pca006132 @geoffder @johnnybigert, thoughts? I think we could do a release pretty much any time, and I'd like to push our python bindings to pypi as part of that.

pca006132:
Yeah I don't think this should block our release. The python binding is usable even without the MeshGL stuff yet. I don't have much time to work on this for now.

geoffder:
I'll try to get the curvature stuff in for the C bindings soon at least RE release inclusion.

I haven't really looked at Python bindings outside of the recent PR, but the CrossSection stuff in particular was pretty doc-less for sure. Are the magic methods that people would generally expect for nice python APIs already in there (I haven't checked or used those bindings).

axel-angel:
Hello there! I really enjoy the library (and I'm writing a python wrapper for it). I noticed Manifold.mirror is missing (only CrossSection.mirror was defined). It's pretty easy to add and it works:

```cpp
      .def("mirror", // TODO add doc
           [](Manifold &self, Float3 ax) {
             return self.Mirror(glm::vec3(std::get<0>(ax), std::get<1>(ax), std::get<2>(ax)));
           })
```

Unless it's there and I missed it?
Thanks.

elalish:
Sounds good, want to send a PR? Also, I hope you can share on #340 - I'd love to see what you're building.

axel-angel:
Hello @elalish! Great, glad to see your enthusiasm. I can try to make such PR by Sunday, let's see. I also noticed Mesh(…) constructor force the last two arguments even tho they are optional in C++. I may find more cases like this, I'll try to collect them together.

PR #488, opened 2023-07-13
updated python doc strings

elalish:
Fixes some more of #361, but I'd still like to get a `MeshGL` binding in before we publish to pypi.


pca006132:
It seems that we should also change the name of `pymanifold.cpp` to `manifold3d.cpp`, or there will be weird import errors. Otherwise this looks good to me.

elalish:
Good call, thanks!

axel-angel:
aaa breaking change :sweat_smile: 

> It seems that we should also change the name of pymanifold.cpp to manifold3d.cpp, or there will be weird import errors. Otherwise this looks good to me.

it's probably obvious but can you explain?

pca006132:
@axel-angel This is because we want to use the same name across NPM and PyPi later, so we want to change the module name. Changing the module name also require changing the filename due to how python works.

elalish:
Exactly, I'm trying to get our breaking changes out of the way before doing our first pypi release.

PR #489, opened 2023-07-13
3d Convex Hull

geoffder:
This adds a 3d convex hull operation to Manifold. Implementation is a port of what I have in OCADml, which is a port of BOSL2. I've made some effort to get the performance reasonable (for this algorithm anyway, there may be smarter ones out there). From my rough measurement, the vast majority of the time is spent on the `DistanceToPlane` operation, so I'm not sure what can be wrung out.

I've set it up to try and preserve some properties info, but I'm not too familiar with that API, so I'm hoping that what I've missed can get filled in with review.


src/manifold/src/manifold.cpp
```diff
@@ -71,6 +73,67 @@ Manifold Halfspace(Box bBox, glm::vec3 normal, float originOffset) {
   float zDeg = glm::degrees(glm::atan(normal.y, normal.x));
   return cutter.Rotate(0.0f, yDeg, zDeg);
 }
+
+double DistanceToVector(const glm::vec3& pt, const glm::vec3& v) {
+  return glm::length(pt - v * glm::dot(pt, v));
+}
+
+inline double DistanceToPlane(const glm::vec3& pt, const glm::vec4& plane) {
```
pca006132, line 81: do we need double?

src/manifold/src/manifold.cpp
```diff
@@ -767,4 +830,104 @@ Manifold Manifold::TrimByPlane(glm::vec3 normal, float originOffset) const {
...
+  mesh.numProp = numProp;
+  mesh.vertProperties = vertProps;
+  mesh.triVerts.reserve((triangles.size() - dropped.size()) * 3);
+  for (int i = 0; i < triangles.size(); i++) {
+    if (!kept[i]) continue;
+    mesh.triVerts.push_back(triangles[i].x);
+    mesh.triVerts.push_back(triangles[i].y);
+    mesh.triVerts.push_back(triangles[i].z);
+  }
+  return Manifold(mesh);
```
pca006132, line 907: I think we should probably not use MeshGL in the internal, as conversion between Manifold and MeshGL is pretty costly.

src/manifold/src/manifold.cpp
```diff
@@ -767,4 +830,104 @@ Manifold Manifold::TrimByPlane(glm::vec3 normal, float originOffset) const {
...
+  const auto plane = PlaneOfPts(pts[trip.x], pts[trip.y], pts[trip.z]);
+  const auto [d, dIsAboveTrip] = NonCoplanar(pts, plane, precision);
+  if (d < 0) return Invalid();  // All points coplanar.
+  const int a = trip.x;
+  const int b = dIsAboveTrip ? trip.z : trip.y;
+  const int c = dIsAboveTrip ? trip.y : trip.z;
+  std::vector<glm::ivec3> triangles;
+  std::vector<glm::vec4> planes;
+  std::unordered_set<glm::ivec2, IVec2Hash> halfEdges{};
+  std::vector<int> dropped;
```
pca006132, line 854: We can avoid using dropped and kept vector by swapping dropped triangle/plane with the last element in the vector and then shrink the vector.

src/manifold/src/manifold.cpp
```diff
@@ -71,6 +73,67 @@ Manifold Halfspace(Box bBox, glm::vec3 normal, float originOffset) {
...
+  if (n == 0.) return glm::vec4();
+  return glm::vec4(crx, glm::dot(crx, a)) / n;
+}
+
+// Return a trio of non-collinear indices into pts. If none are found, indices
+// will be -1 (and hulling cannot be performed).
+glm::ivec3 NonCollinearTriple(const std::vector<glm::vec3>& pts,
+                              const double precision) {
+  const int len = pts.size();
+  int furthest = 1;
```
pca006132, line 98: Just wondering, do we need to find the furthest point or any non-collinear triple is fine? If we don't need the furthest point, we can probably just try each point repeatedly until some non-collinear triple is found, which should be faster than iterating through all the points and calculating their distances.

src/manifold/src/manifold.cpp
```diff
@@ -71,6 +73,67 @@ Manifold Halfspace(Box bBox, glm::vec3 normal, float originOffset) {
...
+
+inline double DistanceToPlane(const glm::vec3& pt, const glm::vec4& plane) {
+  return glm::dot({plane.x, plane.y, plane.z}, pt) - plane.w;
+}
+
+glm::vec4 PlaneOfPts(const glm::vec3& a, const glm::vec3& b,
+                     const glm::vec3& c) {
+  auto crx = glm::cross(c - a, b - a);
+  auto n = glm::length(crx);
+  if (n == 0.) return glm::vec4();
```
elalish, line 89: I believe it's faster to `glm::normalize` and check `isfinite`. Also, for complex types, `auto` is fine, but I prefer to be explicit with simple ones, and also to put `const` wherever possible.

src/manifold/src/manifold.cpp
```diff
@@ -71,6 +73,67 @@ Manifold Halfspace(Box bBox, glm::vec3 normal, float originOffset) {
...
+  return glm::vec4(crx, glm::dot(crx, a)) / n;
+}
+
+// Return a trio of non-collinear indices into pts. If none are found, indices
+// will be -1 (and hulling cannot be performed).
+glm::ivec3 NonCollinearTriple(const std::vector<glm::vec3>& pts,
+                              const double precision) {
+  const int len = pts.size();
+  int furthest = 1;
+  float dist = glm::distance(pts[0], pts[1]);
```
elalish, line 99: Also, if all you care about is the max/min, always used squared distance (dot-product) since it's much faster than a sqrt.

src/manifold/src/manifold.cpp
```diff
@@ -71,6 +73,67 @@ Manifold Halfspace(Box bBox, glm::vec3 normal, float originOffset) {
...
+  if (third < 0) return glm::ivec3(-1);
+  return glm::ivec3(0, furthest, third);
+}
+
+std::pair<int, bool> NonCoplanar(const std::vector<glm::vec3>& pts,
+                                 const glm::vec4& plane,
+                                 const double precision) {
+  for (int i = 0; i < pts.size(); i++) {
+    auto dist = DistanceToPlane(pts[i], plane);
+    if (std::abs(dist) > precision) return {i, dist > precision};
```
elalish, line 127: Likewise, prefer checking `d^2 > precision^2`

src/manifold/src/manifold.cpp
```diff
@@ -767,4 +830,104 @@ Manifold Manifold::TrimByPlane(glm::vec3 normal, float originOffset) const {
 }
 
 ExecutionParams& ManifoldParams() { return params; }
+
+Manifold Manifold::Hull(const std::vector<float>& vertProps, const int numProp,
+                        const double precision) {
+  const int len = vertProps.size() / numProp;
+  // FIXME: Should there be a hull specific error?
+  if (len < 4) return Invalid();
```
elalish, line 838: Why `std::vector<float>` instead of `std::vector<glm::vec3>`? `numProp` seems odd since this can only ever be used for 3D data, right?

test/manifold_test.cpp
```diff
@@ -711,4 +711,28 @@ TEST(Manifold, PinchedVert) {
   EXPECT_FALSE(touch.IsEmpty());
   EXPECT_EQ(touch.Status(), Manifold::Error::NoError);
   EXPECT_EQ(touch.Genus(), 0);
-}
\ No newline at end of file
+}
+
+TEST(Manifold, Hull) {
+#ifdef MANIFOLD_EXPORT
```
elalish, line 717: This should only wrap `ExportMesh` or else the test often won't run.

test/manifold_test.cpp
```diff
@@ -711,4 +711,28 @@ TEST(Manifold, PinchedVert) {
...
+  auto cube = Manifold::Hull(cubePts);
+
+  if (options.exportModels) {
+    ExportMesh("tictac_hull.glb", tictac.GetMesh(), {});
+  }
+
+  EXPECT_FLOAT_EQ(hollow.Hull().GetProperties().volume,
+                  sphere.GetProperties().volume);
+  EXPECT_EQ(spheres.NumVert() / 2 + 36, tictac.NumVert());
+  EXPECT_FLOAT_EQ(cube.GetProperties().volume, 1);
```
elalish, line 736: It looks like there are three independent tests here - lets make them separate `TEST`s.

test/manifold_test.cpp
```diff
@@ -711,4 +711,28 @@ TEST(Manifold, PinchedVert) {
   EXPECT_FALSE(touch.IsEmpty());
   EXPECT_EQ(touch.Status(), Manifold::Error::NoError);
   EXPECT_EQ(touch.Genus(), 0);
-}
\ No newline at end of file
+}
+
+TEST(Manifold, Hull) {
```
elalish, line 716: I'd like to see another test with two high-res spheres - one before the precision limit, testing that the number of verts is the same after hulling as before, and another past the limit where it's less (but the volume should still be close).

src/manifold/src/manifold.cpp
```diff
@@ -767,4 +830,104 @@ Manifold Manifold::TrimByPlane(glm::vec3 normal, float originOffset) const {
...
+  std::vector<glm::vec4> planes;
+  std::unordered_set<glm::ivec2, IVec2Hash> halfEdges{};
+  std::vector<int> dropped;
+  std::vector<bool> kept;
+  const auto AddTri = [&pts, &triangles, &planes, &dropped, &kept](int a, int b,
+                                                                   int c) {
+    auto tri = glm::ivec3(c, b, a);
+    auto plane = PlaneOfPts(pts[a], pts[b], pts[c]);
+    if (dropped.size() > 0) {
+      const int idx = dropped[dropped.size() - 1];
```
elalish, line 861: how about `dropped.back()`?

src/manifold/src/manifold.cpp
```diff
@@ -767,4 +830,104 @@ Manifold Manifold::TrimByPlane(glm::vec3 normal, float originOffset) const {
...
+  // FIXME: Should there be a hull specific error?
+  if (len < 4) return Invalid();
+  std::vector<glm::vec3> pts(len);
+  for (int i = 0; i < len; i++) {
+    const int j = i * numProp;
+    pts[i] = {vertProps[j], vertProps[j + 1], vertProps[j + 2]};
+  }
+  const auto trip = NonCollinearTriple(pts, precision);
+  const auto plane = PlaneOfPts(pts[trip.x], pts[trip.y], pts[trip.z]);
+  const auto [d, dIsAboveTrip] = NonCoplanar(pts, plane, precision);
```
elalish, line 846: Let's split this into two functions.

src/manifold/src/manifold.cpp
```diff
@@ -767,4 +830,104 @@ Manifold Manifold::TrimByPlane(glm::vec3 normal, float originOffset) const {
 }
 
 ExecutionParams& ManifoldParams() { return params; }
+
+Manifold Manifold::Hull(const std::vector<float>& vertProps, const int numProp,
+                        const double precision) {
+  const int len = vertProps.size() / numProp;
```
elalish, line 836: Let's try to be more specific with our naming - maybe `numVert`?

src/manifold/src/manifold.cpp
```diff
@@ -767,4 +830,104 @@ Manifold Manifold::TrimByPlane(glm::vec3 normal, float originOffset) const {
...
+      planes.push_back(plane);
+      kept.push_back(true);
+    }
+  };
+  AddTri(a, b, c);
+  AddTri(d, b, a);
+  AddTri(c, d, a);
+  AddTri(b, d, c);
+  for (int i = 0; i < len; i++) {
+    if (i == a || i == b || i == c || i == d) continue;  // skip starting points
```
elalish, line 877: Maybe just swap these to the front of the vector and start at `i=4`?

src/manifold/src/manifold.cpp
```diff
@@ -767,4 +830,104 @@ Manifold Manifold::TrimByPlane(glm::vec3 normal, float originOffset) const {
...
+        // mark triangle/plane as dropped (slot to be reused)
+        dropped.push_back(j);
+        kept[j] = false;
+      }
+    }
+    // form new triangles with the outer perimeter (horizon) of the set of
+    // conflicting triangles and the point at idx
+    for (auto e : halfEdges) {
+      if (halfEdges.erase({e.y, e.x}) == 0) AddTri(e.x, e.y, i);
+    }
```
elalish, line 894: Is this O(n^2)? Or greater? I believe a hull can be O(n log n). How does this compare to others like qhull for performance?

elalish (changes requested):
I guess before we go any further we should figure out the performance of this algorithm. If we're not close to what other libraries achieve, we should probably take a dependency, because our `n` will tend to be quite large.

src/manifold/src/manifold.cpp
geoffder: Of course, starting at 4 makes sense. That's a holdover from the way I was folding in the OCaml version.

src/manifold/src/manifold.cpp
geoffder: The algorithm as a whole may be greater. After I got this one working I started reading more and have some things to try.

Basically, the thing that takes all the time here is checking if each point is outside of the existing triangles in the first loop. There are some optimizations that cut down on the amount of triangles that need to be checked by throwing out points based on a rough low triangle approximations. I'll see if I can find more detail on those modern ones and adapt this or rewrite accordingly.

src/manifold/src/manifold.cpp
geoffder: The triangles to be dropped aren't necessarily the last one on the vector though.

src/manifold/src/manifold.cpp
pca006132: Yeah I guess picking *good* triangles will be important for the algorithm. However, picking the furthest point may result in a sliver triangles as well.

geoffder: I should say, it's more about the plane that the triangle forms, so if it's a sliver it doesn't really matter. So long as it cuts across the shape I think. In any case, this is very much not where the performance needs to be improved.

src/manifold/src/manifold.cpp
geoffder: There can be many dropped triangles before new ones are added though. New drops would then swap the old drops back into the vector.

geoffder:
RE: libraries [this one](https://github.com/andreacasalino/Fast-Quick-hull)  is the top result, and seems like it'd be easy to vendor and use if we want to just go the dependency route. It's an implementation of [QuickHull](https://github.com/andreacasalino/Fast-Quick-hull/blob/master/doc/Fast_QHull.pdf)  which I believe is better than what I have here. There are a couple papers that say they have something faster than QHull but I just looked at summaries/abstracts last night.

geoffder:
There is also this one with code available to try: https://github.com/EricOuellet2/ConvexHull

test/manifold_test.cpp
geoffder: Ah I just totally forgot about those ifdefs. Fixed, since the few test lines are about the only things that are going to survive from these initial commits (replace with a vendored algorithm).

src/manifold/src/manifold.cpp
elalish: Yeah, but n^2 is a non-starter compared to nlogn, that could easily be a factor of a million for us. I'm pretty sure the efficient algorithms for this are all sweep planes. Before we review this any further, let's compare with something we can take a dependency on (which doesn't include the one you linked, because it has no license), probably https://github.com/qhull/qhull

elalish: Hmm, those further links look more promising 👍

geoffder: I don't disagree, that's why I've been looking around. I did notice the lack of license when I came back to look. I didn't really think about how bad the one I ported was until I was trying bigger inputs, since my main use of it was doing simple hulls over points to join things together, not trying to hull over multiple shapes (would use OpenSCAD for that since I was generating SCAD files).

geoffder:
I'll go ahead try out the header quickhulls implementations that I linked above on then. Not necessary to start with I don't think, but might be interesting to try out a strategy like described here in this [video](https://www.youtube.com/watch?v=iFNUm5rFHaE) and [article](https://www.eecs.ucf.edu/~jjl/pubs/ISVC2020.pdf) in the future. I'm curious how this and another similar paper I've seen along the way actually play out with the kind of inputs we'd have in Manifold, since many of the benchmarks that are shown off in these things are clouds rather than separate sphere meshes (that are already hollow) for example.

src/manifold/src/manifold.cpp
pca006132: Yeah true, I was thinking about parallelization and whether it is possible to do things using atomics. I guess this need some actual benchmark to see which way is faster.

geoffder:
So far out of the non-qhull options, `akuuka` performs the best, as in it doesn't blow up and OOM my 32GM completely at `circularSegments = 1000` (does at `10000`) in the basic tictac hull test, *but* it is capping out the resolution below even that level (the hull output has lower resolution triangles than the input).

image
image

I guess I'm going to actually have to try `qhull` next.

geoffder:
After trying the above mentioned libs it seems like Akuukka's QuickHull implementation is the way to go, with very lines of code and `CMakeLists.txt` required to get it working to boot.

geoffder:
The time to hull I mentioned before was actually an over-estimate since it included the time to get the huge mesh and export it :sweat_smile:, here is a rough idea of scaling over vertex count.
```cpp
TEST(Manifold, TictacBench) {
  const float tictacRad = 100;
  const float tictacHeight = 500;
  const float tictacMid = tictacHeight - 2 * tictacRad;
  const std::vector<int> tictacSegs{36,   360,  720,  1000, 2000,
                                    3000, 4000, 5000, 6000};
  for (int seg : tictacSegs) {
    auto sphere = Manifold::Sphere(tictacRad, seg);
    auto spheres = sphere + sphere.Translate({0, 0, tictacMid});
    auto t0 = std::chrono::system_clock::now();
    auto tictac = spheres.Hull();
    double ms = std::chrono::duration_cast<std::chrono::milliseconds>(
                    std::chrono::system_clock::now() - t0)
                    .count();
    std::printf("segs = %i; verts = %i; time = %f\n", seg, spheres.NumVert(),
                ms);
  }
}
```
```
segs = 36; verts = 652; time = 9.000000
segs = 360; verts = 64804; time = 154.000000
segs = 720; verts = 259204; time = 696.000000
segs = 1000; verts = 500004; time = 1453.000000
segs = 2000; verts = 2000004; time = 6134.000000
segs = 3000; verts = 4500004; time = 14119.000000
segs = 4000; verts = 8000004; time = 26975.000000
segs = 5000; verts = 12500004; time = 48744.000000
segs = 6000; verts = 18000004; time = 77166.000000
```
akuukka_quickhull_bench

src/manifold/src/manifold.cpp
```diff
@@ -775,4 +778,61 @@ Manifold Manifold::TrimByPlane(glm::vec3 normal, float originOffset) const {
 }
 
 ExecutionParams& ManifoldParams() { return params; }
-}  // namespace manifold
\ No newline at end of file
+
+Manifold Manifold::Hull(const std::vector<glm::vec3>& pts,
+                        const std::vector<float>& props, const int numProp) {
+  const int numVert = pts.size();
+  // FIXME: Should there be a hull specific error?
```
elalish, line 785: Sure, I suppose we can add an error status for this.

src/manifold/src/manifold.cpp
```diff
@@ -775,4 +778,61 @@ Manifold Manifold::TrimByPlane(glm::vec3 normal, float originOffset) const {
...
+  // bools: correct triangle winding, and use original indices
+  auto hull = qh.getConvexHull(vertices, false, true);
+  const auto& triangles = hull.getIndexBuffer();
+  const int numTris = triangles.size() / 3;
+
+  Mesh mesh;
+  Impl::MeshRelationD relation;
+  relation.numProp = numProp;
+  relation.properties = props;
+  mesh.vertPos = pts;
```
elalish, line 803: Oh right, since the Manifold constructor gets rid of unreferenced verts and props now. 👍

src/manifold/src/manifold.cpp
```diff
@@ -775,4 +778,61 @@ Manifold Manifold::TrimByPlane(glm::vec3 normal, float originOffset) const {
...
+  mesh.triVerts.reserve(numTris);
+  for (int i = 0; i < numTris; i++) {
+    const int j = i * 3;
+    mesh.triVerts.push_back({triangles[j], triangles[j + 1], triangles[j + 2]});
+  }
+  return Manifold(mesh);
+}
+
+Manifold Manifold::Hull() const {
+  auto mesh = GetMeshGL();
```
elalish, line 814: So this is to get 1:1 properties to positions? Of course that means we'll get duplicate positions with different properties, so it's going to be basically random which properties end up in the output. Also, it'll be hard to make sense of them without an `originalID`, since some of those properties might mean different things. Actually it's worse that than: you could end up with a face interpolating UV coordinates into vertex colors or something, since there's no guarantee the verts are from the same input object. 

I wonder if we should just remove properties from the Hull entirely and do just geometry for this reason. I can't think of a way to make it sensical otherwise.

src/manifold/src/manifold.cpp
```diff
@@ -775,4 +778,61 @@ Manifold Manifold::TrimByPlane(glm::vec3 normal, float originOffset) const {
...
+              mesh.vertProperties[j + 2]};
+    for (int k = 0; k < extraProps; k++) {
+      props.push_back(mesh.vertProperties[j + 3 + k]);
+    }
+  }
+  return Hull(pts, props, extraProps);
+}
+
+Manifold Manifold::Hull(const std::vector<Manifold>& manifolds) {
+  return BatchBoolean(manifolds, OpType::Add).Hull();
```
elalish, line 832: I think a `Compose` would be a quite a bit faster here and I don't think you care about intersections.

src/third_party/CMakeLists.txt
```diff
@@ -6,5 +6,9 @@ set(CLIPPER2_EXAMPLES OFF)
 set(CLIPPER2_TESTS OFF)
 set(CLIPPER2_USINGZ "OFF" CACHE STRING "Preempt cache default of USINGZ (we only use 2d)")
 add_definitions(-D_HAS_EXCEPTIONS=0) # disable exceptions for STL
-
 add_subdirectory(clipper2/CPP)
+
+file(GLOB_RECURSE QH_SOURCE_FILES CONFIGURE_DEPENDS quickhull/*.cpp)
```
elalish, line 11: I seem to recall it's better practice to list out the files specifically than to use GLOB.

elalish (changes requested):
I like this library you found, thanks!

src/third_party/CMakeLists.txt
geoffder: Sure, I was just mimicing what was done in the manifold CMakeLists.

src/manifold/src/manifold.cpp
geoffder: Right, I didn't think to use Compose, that makes more sense.

src/manifold/src/manifold.cpp
geoffder: I wasn't really sure myself since I don't know much about theesh properties business. I just did it this way to try and preserve properties as you said, hoping that you or pca could point out what more I should be preserving (or not). I did consider triangles ending up with weirdly disparate properties as a possibility, but I don't know/understand what is done with properties during booleans either.

src/manifold/src/manifold.cpp
geoffder: Partly as a reminder to myself, I've forgotten to check so far what this QuickHull returns if all points are coplanar. However it's found though, will have to have that result in an error Manifold as well.

geoffder: So, `getConvexHull` returns 4 triangles when a coplanar rectangle of vertices is given to it, but after passing through the `Manifold(mesh)` constructor, the result is an empty `Manifold` (`NoError`). Should we just return an empty Manifold in both cases, have an error for one and empty for the other, or an error for both?

If error(s):
- `AllVerticesCoplanar` (or simply `NotManifold`)
- `TooFewVertices` (which also technically is coplanar, and not manifold)

elalish: Yeah, let's prefer no error if there isn't a huge reason to have one.

src/manifold/src/manifold.cpp
elalish: The nice thing about Booleans is we can guarantee a bunch of things: every result triangle is a subset of exactly one input triangle, so it will only every interpolate properties from that triangle. Our decimator is also aware, so it won't merge triangles that don't come from the same original triangle. 

Each triangle in a hull is either an input triangle, or a new triangle. I suppose we could try to maintain properties (and faceIDs) for input triangles, but we'd have no choice but to drop them for new triangles. Personally I think it sounds like more trouble than it's worth.

geoffder: Ok, sounds good :+1:. I've stripped out all of the properties stuff and just get the vertices via `GetMesh`.

src/manifold/src/manifold.cpp
```diff
@@ -12,9 +12,12 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+#include <QuickHull.hpp>
 #include <algorithm>
+#include <cstdio>
```
elalish, line 17: necessary?

elalish (approved):
Thanks!

src/manifold/src/manifold.cpp
geoffder: Nope, just missed cleaning it up after doing some debugging.

geoffder:
I'll follow up soon with some bindings updates (2d and 3d hull).

geoffder:
I'll follow up soon with some bindings updates (2d and 3d hull).

PR #492, opened 2023-07-15
Update python mesh

### LINKED ISSUE #361: Add python bindings for the rest of our APIs

elalish:
Fixes #361 - at least I think so; please remind me if I've forgotten anything.

The main thing I'm adding here is `MeshGL` support, which is to say updating the Python Mesh class to work as MeshGL instead, akin to our JS binding. This PR is still WIP, as I need to update our `trimesh` integration, but we should now be able to export vertex colors and such right through to glTF with that soon.

### DIFF
--- a/bindings/python/examples/run_all.py
+++ b/bindings/python/examples/run_all.py
@@ -33,7 +33,7 @@ if __name__ == "__main__":
         module = importlib.import_module(f)
         t0 = time()
         model = module.run()
-        mesh = model.to_meshgl()
+        mesh = model.to_mesh()
         if export_models:
             vertices = np.reshape(mesh.vert_properties, (-1, mesh.num_prop))
             if mesh.num_prop > 3:
--- a/bindings/python/examples/run_all.py
+++ b/bindings/python/examples/run_all.py
@@ -37,7 +37,7 @@ if __name__ == "__main__":
         if export_models:
             vertices = np.reshape(mesh.vert_properties, (-1, mesh.num_prop))
             if mesh.num_prop > 3:
-                vertices = np.hsplit(vertices, 3)
+                vertices = np.hsplit(vertices, 3)[0]
             meshOut = trimesh.Trimesh(
                 vertices=vertices, faces=np.reshape(mesh.tri_verts, (-1, 3)))
             trimesh.exchange.export.export_mesh(meshOut, f'{f}.glb', 'glb')
--- a/bindings/python/examples/run_all.py
+++ b/bindings/python/examples/run_all.py
@@ -35,7 +35,7 @@ if __name__ == "__main__":
         model = module.run()
         mesh = model.to_mesh()
         if export_models:
-            if mesh.num_prop > 3:
+            if mesh.vert_properties.shape[1] > 3:
                 vertices = np.hsplit(mesh.vert_properties, 3)[0]
             else:
                 vertices = mesh.vert_properties
--- a/bindings/python/examples/run_all.py
+++ b/bindings/python/examples/run_all.py
@@ -35,11 +35,11 @@ if __name__ == "__main__":
         model = module.run()
         mesh = model.to_mesh()
         if export_models:
-            vertices = np.reshape(mesh.vert_properties, (-1, mesh.num_prop))
             if mesh.num_prop > 3:
-                vertices = np.hsplit(vertices, 3)[0]
-            meshOut = trimesh.Trimesh(
-                vertices=vertices, faces=np.reshape(mesh.tri_verts, (-1, 3)))
+                vertices = np.hsplit(mesh.vert_properties, 3)[0]
+            else:
+                vertices = mesh.vert_properties
+            meshOut = trimesh.Trimesh(vertices=vertices, faces=mesh.tri_verts)
             trimesh.exchange.export.export_mesh(meshOut, f'{f}.glb', 'glb')
             print(f'Exported model to {f}.glb')
         t1 = time()
--- a/bindings/python/examples/run_all.py
+++ b/bindings/python/examples/run_all.py
@@ -18,6 +18,7 @@ import pathlib
 import sys
 import importlib
 import trimesh
+import numpy as np
 from time import time
 
 if __name__ == "__main__":
@@ -32,10 +33,13 @@ if __name__ == "__main__":
         module = importlib.import_module(f)
         t0 = time()
         model = module.run()
-        mesh = model.to_mesh()
+        mesh = model.to_meshgl()
         if export_models:
+            vertices = np.reshape(mesh.vert_properties, (-1, mesh.num_prop))
+            if mesh.num_prop > 3:
+                vertices = np.hsplit(vertices, 3)
             meshOut = trimesh.Trimesh(
-                vertices=mesh.vert_pos, faces=mesh.tri_verts)
+                vertices=vertices, faces=np.reshape(mesh.tri_verts, (-1, 3)))
             trimesh.exchange.export.export_mesh(meshOut, f'{f}.glb', 'glb')
             print(f'Exported model to {f}.glb')
         t1 = time()
--- a/bindings/python/manifold3d.cpp
+++ b/bindings/python/manifold3d.cpp
@@ -29,6 +29,10 @@ typedef std::tuple<float, float, float> Float3;
 
 constexpr auto pyArrayFlags = py::array::c_style | py::array::forcecast;
 
+PYBIND11_MAKE_OPAQUE(std::vector<int>);
+PYBIND11_MAKE_OPAQUE(std::vector<uint32_t>);
+PYBIND11_MAKE_OPAQUE(std::vector<float>);
+
 template <typename T>
 std::vector<T> toVector(const py::array_t<T> &arr) {
   return std::vector<T>(arr.data(), arr.data() + arr.size());
@@ -493,7 +497,7 @@ PYBIND11_MODULE(manifold3d, m) {
                   "marking sets of triangles that can be looked up after "
                   "further operations. Assign to MeshGL.runOriginalID vector");
 
-  py::class_<MeshGL>(m, "MeshGL")
+  py::class_<MeshGL>(m, "Mesh")
       .def(
           // note that reshape requires mutable array_t, but this will not
           // affect the original array passed into the function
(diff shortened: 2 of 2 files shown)


pca006132:
Will have a look at this. I think we should also add at least one python example using MeshGL.

pca006132:
1. The original Mesh type is now deleted, but it is still used for things like to_mesh which are used in older python code by us. I think we should keep it.
2. For immutable access to fields of MeshGL, I think there are ways to make it into some python object without copying the entire array.

geoffder:
I think the only way to hide things in python is to put them in a closure. But there is no way to hide that closure if it is stored as a field of an object, so if it returns a reference to the object it wraps, users will still be able to get it. A wrapper class on the CPP side that stores the MeshGL as private then gives

geoffder:
Sorry I was writing a comment then decided it was wrong and tried to delete, and somehome closed the while PR (on phone). My bad!

elalish:
I'm going to update our examples to this, because I'd prefer to only use MeshGL in the bindings. I may even deprecate Mesh in the C++, as I don't love supporting both. If you know a more efficient way to access the vectors, that would be great. I'm no python expert. I just want the API to be the right shape so I can feel good publishing it.

pca006132:
It seems that we don't really need to copy vectors of primitive types into a python array, pybind11 does that automatically for us already.

I also updated the python example to use the `MeshGL` API, and converted `to_mesh` into `to_meshgl`

pca006132:
For the other python changes, e.g. hull, should I open another PR or should I add them here? should be relatively small change.

geoffder:
I've got a bindings PR in the works that I included python in. Just haven't finished the JS bits yet.

pca006132:
@axel-angel this should be our final breaking change. You can see the updated `run_all.py` for how to use the updated `MeshGL` that replaces `Mesh`.

pca006132:
@elalish maybe we can merge this? As I modified this, it doesn't make sense for me to review it again, and I cannot choose you to be the reviewer.

elalish:
Thank you so much, @pca006132! I had intended to get back to this earlier, but I got laid up with a sinus infection instead. Thanks for figuring this out. I'll review soon.

pca006132:
Btw I am thinking if we should return a flattened array, i.e. the same as our C++ interface, or a reshaped array, as numpy can do reshaping without actually copying the memory.

bindings/python/manifold3d.cpp
```diff
@@ -238,7 +238,33 @@ PYBIND11_MODULE(manifold3d, m) {
           "\n"
           ":param n: The number of pieces to split every edge into. Must be > "
           "1.")
-      .def("to_mesh", &Manifold::GetMesh)
+      .def(
+          "to_meshgl",
```
elalish, line 242: I was thinking of leaving this as just "mesh", like we do in the JS bindings, since there's only the one kind here.

bindings/python/manifold3d.cpp
```diff
@@ -459,111 +485,117 @@ PYBIND11_MODULE(manifold3d, m) {
...
+          py::arg("vert_properties"), py::arg("tri_verts"),
+          py::arg("merge_from_vert") = py::none(),
+          py::arg("merge_to_vert") = py::none(),
+          py::arg("run_index") = py::none(),
+          py::arg("run_original_id") = py::none(),
+          py::arg("run_transform") = py::none(),
+          py::arg("face_id") = py::none(),
+          py::arg("halfedge_tangent") = py::none(), py::arg("precision") = 0)
+      .def_readonly("vert_properties", &MeshGL::vertProperties)
+      .def_readonly("tri_verts", &MeshGL::triVerts)
```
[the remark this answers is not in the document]
pca006132: I'm wondering what the names should be. I guess I will leave it to you to decide on this :)

bindings/python/examples/run_all.py
```diff
@@ -32,10 +33,13 @@
...
         model = module.run()
-        mesh = model.to_mesh()
+        mesh = model.to_meshgl()
         if export_models:
+            vertices = np.reshape(mesh.vert_properties, (-1, mesh.num_prop))
+            if mesh.num_prop > 3:
+                vertices = np.hsplit(vertices, 3)
             meshOut = trimesh.Trimesh(
-                vertices=mesh.vert_pos, faces=mesh.tri_verts)
+                vertices=vertices, faces=np.reshape(mesh.tri_verts, (-1, 3)))
```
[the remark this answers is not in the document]
pca006132: I think it would be better to finish everything in this PR, to test the new APIs are working correctly. (in case we understood python incorrectly...)

bindings/python/manifold3d.cpp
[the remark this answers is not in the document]
elalish: These names sound reasonable to me. Anything you want to change?

bindings/python/examples/run_all.py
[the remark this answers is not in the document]
elalish: Yeah, fair enough. Do you want to take a stab at it? Otherwise I can work on it tomorrow.

bindings/python/manifold3d.cpp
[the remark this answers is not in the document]
pca006132: Oh I was replying to the incorrect comment. I was thinking about something like vert_pos, after splitting the properties.

bindings/python/examples/run_all.py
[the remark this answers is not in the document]
pca006132: yeah you can work on it

bindings/python/manifold3d.cpp
```diff
@@ -459,111 +486,90 @@ PYBIND11_MODULE(manifold3d, m) {
...
-            for (int i = 0; i < vertPos_view.shape(0); i++)
-              for (const int j : {0, 1, 2})
-                vertPos_vec[i][j] = vertPos_view(i, j);
-            for (int i = 0; i < triVerts_view.shape(0); i++)
-              for (const int j : {0, 1, 2})
-                triVerts_vec[i][j] = triVerts_view(i, j);
+          // note that reshape requires mutable array_t, but this will not
+          // affect the original array passed into the function
+          py::init([](py::array_t<float> &vertProp, py::array_t<int> &triVerts,
+                      const std::optional<std::vector<uint32_t>> &mergeFromVert,
```
elalish, line 494: Why are some of these `std::vector` and some `py::array_t`? If I understand https://pybind11.readthedocs.io/en/stable/advanced/cast/stl.html and https://pybind11.readthedocs.io/en/stable/advanced/pycpp/numpy.html correctly, the first makes a copy and the second does not?

bindings/python/manifold3d.cpp
```diff
@@ -459,111 +486,90 @@ PYBIND11_MODULE(manifold3d, m) {
...
+                      const std::optional<std::vector<uint32_t>> &mergeToVert,
+                      const std::optional<std::vector<uint32_t>> &runIndex,
+                      const std::optional<std::vector<uint32_t>> &runOriginalID,
+                      std::optional<py::array_t<float>> &runTransform,
```
elalish, line 498: Do we need `py::array_t<double, py::array::c_style | py::array::forcecast>`? https://pybind11.readthedocs.io/en/stable/advanced/pycpp/numpy.html#arrays

bindings/python/manifold3d.cpp
```diff
@@ -459,111 +486,90 @@ PYBIND11_MODULE(manifold3d, m) {
...
+          }),
+          py::arg("vert_properties"), py::arg("tri_verts"),
+          py::arg("merge_from_vert") = py::none(),
+          py::arg("merge_to_vert") = py::none(),
+          py::arg("run_index") = py::none(),
+          py::arg("run_original_id") = py::none(),
+          py::arg("run_transform") = py::none(),
+          py::arg("face_id") = py::none(),
+          py::arg("halfedge_tangent") = py::none(), py::arg("precision") = 0)
+      .def_readonly("vert_properties", &MeshGL::vertProperties)
```
elalish, line 563: So, I think this is using the basic STL interop by doing a copy, right? Should we instead do like on the input side and construct a `py::array` from the pointer and apply the correct shape?

bindings/python/manifold3d.cpp
pca006132: Yes. Btw I just noticed that I forgot to do move for the vectors.

bindings/python/manifold3d.cpp
pca006132: yeah I think we should do this, I forgot that numpy arrays can have various styles

bindings/python/manifold3d.cpp
pca006132: yes

pca006132:
Note that some parameters are not constant because reshape is not a constant function.
Alternatively we can also make every parameter of the MeshGL constructor not constant.

bindings/python/manifold3d.cpp
```diff
@@ -459,111 +493,117 @@ PYBIND11_MODULE(manifold3d, m) {
...
+                               return py::array(self.halfedgeTangent.size(),
+                                                self.halfedgeTangent.data())
+                                   .reshape(std::array<int, 3>{-1, 3, 4});
+                             })
+      .def_readonly("merge_from_vert", &MeshGL::mergeFromVert)
+      .def_readonly("merge_to_vert", &MeshGL::mergeToVert)
+      .def_readonly("run_index", &MeshGL::runIndex)
+      .def_readonly("run_original_id", &MeshGL::runOriginalID)
+      .def_readonly("face_id", &MeshGL::faceID)
+      .def_readonly("num_prop", &MeshGL::numProp);
```
elalish, line 606: I don't think we need this anymore, since it's part of the shape of `vertProperties` now.

bindings/python/manifold3d.cpp
```diff
@@ -459,111 +493,117 @@ PYBIND11_MODULE(manifold3d, m) {
...
+                                                self.runTransform.data())
+                                   .reshape(std::array<int, 3>{-1, 4, 3});
+                             })
+      .def_property_readonly("halfedge_tangent",
+                             [](const MeshGL &self) {
+                               return py::array(self.halfedgeTangent.size(),
+                                                self.halfedgeTangent.data())
+                                   .reshape(std::array<int, 3>{-1, 3, 4});
+                             })
+      .def_readonly("merge_from_vert", &MeshGL::mergeFromVert)
```
elalish, line 601: Do these automatically take the pointer, or are they copying the `std::vector` instead?

bindings/python/manifold3d.cpp
```diff
@@ -459,111 +493,117 @@ PYBIND11_MODULE(manifold3d, m) {
                   "marking sets of triangles that can be looked up after "
                   "further operations. Assign to MeshGL.runOriginalID vector");
 
-  py::class_<Mesh>(m, "Mesh")
+  py::class_<MeshGL>(m, "MeshGL")
```
elalish, line 496: Let's go with `Mesh`, like our JS bindings.

bindings/python/manifold3d.cpp
pca006132: I found that I forgot to add opaque types. https://pybind11.readthedocs.io/en/stable/advanced/cast/stl.html#making-opaque-types

Now, it should auotmatically take the pointer.

elalish (commented):
Looks good to me - want to go ahead and merge it?

Issue #460, opened 2023-06-19
Crash while building


emil-peters:
While building on my computer I get the following error:

```
[ 22%] Building CXX object src/manifold/CMakeFiles/manifold.dir/src/boolean_result.cpp.o
cd /app/manifold/build/src/manifold && /usr/bin/c++ -Dmanifold_EXPORTS -I/app/manifold/src/manifold/include -I/app/manifold/src/utilities/include -I/app/manifold/src/third_party/thrust/dependencies/libcudacxx/include -I/app/manifold/src/third_party/thrust -I/app/manifold/src/third_party/glm/glm/.. -I/app/manifold/src/cross_section/include -I/app/manifold/src/third_party/clipper2/CPP/Clipper2Lib/include -I/app/manifold/src/collider/include -I/app/manifold/src/polygon/include -I/app/manifold/src/third_party/graphlite/include -O3 -DNDEBUG -fPIC -fPIC -Werror -Wall -Wno-sign-compare -Wno-unused -O3 -DTHRUST_DEVICE_SYSTEM=THRUST_DEVICE_SYSTEM_CPP -std=gnu++17 -MD -MT src/manifold/CMakeFiles/manifold.dir/src/boolean_result.cpp.o -MF CMakeFiles/manifold.dir/src/boolean_result.cpp.o.d -o CMakeFiles/manifold.dir/src/boolean_result.cpp.o -c /app/manifold/src/manifold/src/boolean_result.cpp
In file included from /app/manifold/src/utilities/include/sparse.h:22,
                 from /app/manifold/src/collider/include/collider.h:17,
                 from /app/manifold/src/manifold/src/impl.h:18,
                 from /app/manifold/src/manifold/src/boolean3.h:16,
                 from /app/manifold/src/manifold/src/boolean_result.cpp:18:
In static member function ‘static void manifold::ManagedVec<T>::mallocManaged(T**, size_t) [with T = int]’,
[...]

PR #494, opened 2023-07-15
Optimize sort_by_key (SortFaces/SortVerts) 

stephomi:
Remap boxes separately

MacOS M1/TBB: 1485ms -> 1385ms

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -16,7 +16,7 @@
             , doCheck ? true
             , build-tools ? [ ]
             , ...
-            }: pkgs.stdenv.mkDerivation {
+            }: pkgs.llvmPackages_15.stdenv.mkDerivation {
               inherit doCheck;
               pname =
                 if cuda-support then
--- a/src/manifold/src/sort.cpp
+++ b/src/manifold/src/sort.cpp
@@ -312,8 +312,9 @@ void Manifold::Impl::SortVerts() {
 
   VecDH<int> vertNew2Old(numVert);
   sequence(policy, vertNew2Old.begin(), vertNew2Old.end());
+
   sort_by_key(policy, vertMorton.begin(), vertMorton.end(),
-              zip(vertPos_.begin(), vertNew2Old.begin()));
+              vertNew2Old.begin());
 
   ReindexVerts(vertNew2Old, numVert);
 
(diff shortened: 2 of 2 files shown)


pca006132:
interesting, I guess these recent optimizations means that sort_by_key and sort using zip iterators are very slow? I originally thought that the performance should be fine.
will have a look more carefully later today

stephomi:
Nah sort_by_key is fine (actually I could have reverted to sort_by_key in the previous PR, it would give similar performance).

The only thing that matters is the size of the object to move (both keys or values), here by sorting boxes we are moving big objects for nothing.
If you plan on sync-sorting multiple array it’s best to to perform the sort on the minimal amount of memory and re-index the rest once.

pca006132:
For the build error, I think we hit the #460. The safer way I can now think of is to use clang. Try to add the following commit:
```diff
diff --git a/flake.nix b/flake.nix
index 04ffc90..4c37ab3 100644
--- a/flake.nix
+++ b/flake.nix
@@ -16,7 +16,7 @@
             , doCheck ? true
             , build-tools ? [ ]
             , ...
-            }: pkgs.stdenv.mkDerivation {
+            }: pkgs.llvmPackages_15.stdenv.mkDerivation {
               inherit doCheck;
               pname =
                 if cuda-support then
```

> Not sure if I'm breaking the whole CUDA thing though, if I don't get the pointer first, it just takes forever (unless I capture by reference [&] instead of value [=])

it seems that this is no longer relevant?

> MacOS M1/TBB: 1485ms -> 1385ms

Wondering what benchmark are you using? Wonder if you can share them to the unit tests? This can help us track performance later.

pca006132 (approved):

stephomi:
> it seems that this is no longer relevant?

Yep, removed the comment to avoid confusion

> Wondering what benchmark are you using? Wonder if you can share them to the unit tests? This can help us track performance later.

A simple sphere/cylinder subtraction.

Capture d’écran 2023-07-16 à 20 02 37

I'm always benching the part that I'm editing, since there's a bit of variance by looking at the whole thing.
For the PR I'm giving timing difference for the whole operation to give an idea of the gain (minor here, but the idea is also to give insight on what might be improved)

The part in question:
|             | Master        | PR           |
|-------------| ------------- | ------------ |
| SortVerts 1 | 22 ms         | 17 ms        |
| SortVerts 2 | 4 ms          | 3  ms        |
| SortVerts 3 | 20 ms         | 14 ms        |
| SortFaces 1 | 100 ms        | 60 ms        |
| SortFaces 2 | 15 ms         | 10 ms        |
| SortFaces 3 | 97 ms         | 56 ms        |

PR #497, opened 2023-07-19
Add hull bindings (c, wasm/js, and python)

geoffder:
This updates the bindings to cover the addition of convex hulling to CrossSection and Manifold, including the static versions to spare unnecessary booleans (which may actually get in the way of hulling over arbitrary points, since one would need to pack them into valid shapes before hulling with the instance methods).

### DIFF
--- a/bindings/wasm/bindings.js
+++ b/bindings/wasm/bindings.js
@@ -472,6 +472,47 @@ Module.setup = function() {
   Module.CrossSection.difference = crossSectionBatchbool('DifferenceN');
   Module.CrossSection.intersection = crossSectionBatchbool('IntersectionN');
 
+  Module.CrossSection.hull = function(...args) {
+    if (args.length == 1) args = args[0];
+    let pts = new Module.Vector_vec2();
+    for (const cs of args) {
+      if (cs instanceof CrossSectionCtor) {
+        const polyVecs = cs._ToPolygons();
+        for (let i = 0; i < polyVecs.size(); i++) {
+          const vec = polyVecs.get(i);
+          for (let j = 0; j < vec.size(); j++) {
+            pts.push_back(vec.get(j));
+          }
+          vec.delete();
+        }
+        disposePolygons(polyVecs);
+      } else if (
+          cs instanceof Array && cs.length == 2 && typeof cs[0] == 'number') {
+        pts.push_back({x: cs[0], y: cs[1]});
+      } else if (cs.x) {
+        pts.push_back(cs);
+      } else {
+        const polys = cs instanceof Array &&
+                ((cs[0].length == 2 && typeof cs[0][0] == 'number') ||
+                 cs[0].x) ?
+            [cs] :
+            cs;
+        for (const poly of polys) {
+          for (const p of poly) {
+            if (p instanceof Array) {
+              pts.push_back({x: p[0], y: p[1]});
+            } else {
+              pts.push_back(p);
+            }
+          }
+        }
+      }
+    }
+    const result = Module._crossSectionHullPoints(pts);
+    pts.delete();
+    return result;
+  };
+
   Module.CrossSection.prototype = Object.create(CrossSectionCtor.prototype);
 
   // Because the constructor and prototype are being replaced, instanceof will
--- a/bindings/wasm/bindings.js
+++ b/bindings/wasm/bindings.js
@@ -472,40 +472,29 @@ Module.setup = function() {
   Module.CrossSection.difference = crossSectionBatchbool('DifferenceN');
   Module.CrossSection.intersection = crossSectionBatchbool('IntersectionN');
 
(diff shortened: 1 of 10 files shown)


bindings/wasm/bindings.js
```diff
@@ -596,6 +626,33 @@ Module.setup = function() {
...
+    for (const m of args) {
+      if (m instanceof ManifoldCtor) {
+        Module._manifoldCollectVertices(pts, m);
+      } else if (
+          m instanceof Array && m.length == 3 && typeof m[0] == 'number') {
+        pts.push_back({x: m[0], y: m[1], z: m[2]});
+      } else if (m.x) {
+        pts.push_back(m);
+      } else {
+        pushVec3(pts, m);
```
elalish, line 648: I feel like we have some helper functions for things like this already - any chance we can make use of those? It also seems a little weird that we can have an array with a manifold and a few extra points, but maybe that'll be useful to someone...

bindings/wasm/manifold-encapsulated-types.d.ts
```diff
@@ -111,6 +111,8 @@ export class CrossSection {
    * @param v The vector to add to every vertex.
    */
   translate(v: Vec2): CrossSection;
+  translate(x: number): CrossSection;
```
elalish, line 114: Can we really take just an x-value? Feels weird...

bindings/wasm/manifold-encapsulated-types.d.ts
```diff
@@ -507,6 +516,9 @@ export class Manifold {
    * @param v The vector to add to every vertex.
    */
   translate(v: Vec3): Manifold;
+  translate(x: number): Manifold;
+  translate(x: number, y: number): Manifold;
+  translate(x: number, y: number, z: number): Manifold;
```
elalish, line 521: If this is really what we want, it can be condensed to `translate(x: number, y = 0, z = 0): Manifold;`

bindings/wasm/manifold-encapsulated-types.d.ts
elalish: Oh, I guess this is how OpenSCAD does it, right?

src/cross_section/src/cross_section.cpp
```diff
@@ -629,14 +635,31 @@ CrossSection CrossSection::Hull(
...
 CrossSection CrossSection::Hull(SimplePolygon pts) {
   return CrossSection(C2::PathsD{HullImpl(pts)});
 }
 
+/**
+ * Compute the convex hull of a set of points/polygons. If the given points are
+ * fewer than 3, an empty CrossSection will be returned.
+ *
+ * @param pts A vector of vectors of 2-dimensional points over which to compute
+ * a convex hull.
```
elalish, line 661: 👍

elalish (approved):
Thanks!

bindings/wasm/bindings.js
geoffder: They aren't quite the same, polygons2vec is for vectors of vectors.

bindings/wasm/manifold-encapsulated-types.d.ts
geoffder: The functions already let you give a variable number of arguments, this just adds an overload to the typescript interface so you don't get res squigglies for legal use.

bindings/wasm/manifold-encapsulated-types.d.ts
geoffder: Ah right didn't think of that. I'll condense it. As mentioned on the other comment, since we are allowed to give these inputs by the implementation, we might as well have it represented in the typescript interface.

bindings/wasm/bindings.js
geoffder: I don't think it's too likely that people would mix manifolds with points, but it was simple enough to just allow the mixing while supporting both lists of points and lists of manifolds separately. 

It might be conventient to hull a single point with a cube to quickly get a roof though for example.

bindings/wasm/manifold-encapsulated-types.d.ts
geoffder: Condensed to `translate(x: number, y?: number, z?:number)` which does the trick of not warning on these inputs (which are valid for the implementation).

bindings/wasm/manifold-encapsulated-types.d.ts
geoffder: > Oh, I guess this is how OpenSCAD does it, right?

Yeap that's right. We pretty well mirror that, since the `vararg2vec` helpers used are also flexible to array inputs not including all dimensions. Typescript will complain though, since that still isn't covered in the signature, e.g.: 
```js
translate(v: [number] | Vec2 | Vec3): Manifold
```

elalish (approved):

Issue #495, opened 2023-07-17
Infinite RecursiveEdgeSwap


stephomi:
I'm testing manifold on a simple mesh that already self-intersects (yes it's bad blablaha) and it crashes in `SimplifyTopology`.

<img width="723" alt="Capture d’écran 2023-07-17 à 18 00 27" src="https://github.com/elalish/manifold/assets/4750281/12ea0896-4ce1-47f0-a1de-b16543917bcf">

It recurse indefinitely in `RecursiveEdgeSwap`.
To avoid the crash I added a check that we don't call RecursiveEdgeSwap on the same edge but I have no idea if it makes sense. 

```diff
diff --git a/src/manifold/src/edge_op.cpp b/src/manifold/src/edge_op.cpp
index c0afbfd..60b6a33 100644
--- a/src/manifold/src/edge_op.cpp
+++ b/src/manifold/src/edge_op.cpp
@@ -570,10 +570,10 @@ void Manifold::Impl::RecursiveEdgeSwap(const int edge) {
     if (glm::dot(e23, e23) < precision_ * precision_) {
       CollapseEdge(tri0edge[2]);
     } else {
-      RecursiveEdgeSwap(tri0edge[0]);
-      RecursiveEdgeSwap(tri0edge[1]);
-      RecursiveEdgeSwap(tri1edge[0]);
-      RecursiveEdgeSwap(tri1edge[1]);
+      if (edge != tri0edge[0]) RecursiveEdgeSwap(tri0edge[0]);
+      if (edge != tri0edge[1]) RecursiveEdgeSwap(tri0edge[1]);
+      if (edge != tri1edge[0]) RecursiveEdgeSwap(tri1edge[0]);
+      if (edge != tri1edge[1]) RecursiveEdgeSwap(tri1edge[1]);
     }
     return;
   } else if (CCW(v[0], v[3], v[2], precision_) <= 0 ||
@@ -582,8 +582,12 @@ void Manifold::Impl::RecursiveEdgeSwap(const int edge) {
   }
   // Normal path
   SwapEdge();
-  RecursiveEdgeSwap(halfedge_[tri0edge[1]].pairedHalfedge);
-  RecursiveEdgeSwap(halfedge_[tri1edge[0]].pairedHalfedge);
+  if (edge != halfedge_[tri0edge[1]].pairedHalfedge) {
+    RecursiveEdgeSwap(halfedge_[tri0edge[1]].pairedHalfedge);
+  }
[...]

PR #501, opened 2023-07-21
Preventing crash with self intersection

stephomi:
Issue: https://github.com/elalish/manifold/issues/495

The code snippet in the issue makes `Samples.Sponge4` fails:
```
Expected: (sponge.NumDegenerateTris()) <= (40000), actual: 42386 vs 40000
```
If I change the logic a bit the test passes and it doesn't crash on my example, but

1. No idea if the fix is robust, as it's hard to replicate the self-intersection crash
2. Using `ImportMesh` to test the crash scenario but not sure relative path is the way to go for the test


pca006132:
I think the relative path should be `../../samples/models/self_intersectA.glb`, as the directory we execute the test is in `build/test`.

stephomi:
It will  fail with the README build instruction for example(`test/manifold_test`, so executed from `build/`)
Ideally it should be relative to the executable folder (or from where it's executed). 
For now I'll simply make relative to __FILE__, should for all cases.

stephomi:
> To make the test pass, you can add
> 
> ```c++
> manifold::PolygonParams().processOverlaps = true;
> ```
> 
> before your test case, and set it back to false afterwards.

Thanks!
For some reason, even without it the test was running on my machine™

src/manifold/src/edge_op.cpp
```diff
@@ -222,9 +222,10 @@ void Manifold::Impl::SimplifyTopology() {
                      precision_};
     for_each_n(policy, countAt(0), nbEdges,
                [=] __host__ __device__(int i) { bflagsPtr[i] = se(i); });
+    std::vector<uint32_t> visited(halfedge_.size(), -1);
```
elalish, line 225: Setting a uint to -1 feels a little awkward...

src/manifold/src/edge_op.cpp
```diff
@@ -483,12 +484,15 @@ void Manifold::Impl::CollapseEdge(const int edge) {
...
-void Manifold::Impl::RecursiveEdgeSwap(const int edge) {
+void Manifold::Impl::RecursiveEdgeSwap(const int edge,
+                                       std::vector<uint32_t>& visited) {
   VecDH<TriRef>& triRef = meshRelation_.triRef;
 
   if (edge < 0) return;
   const int pair = halfedge_[edge].pairedHalfedge;
   if (pair < 0) return;
+  if (visited[edge] == pair) return;
+  visited[edge] = pair;
```
elalish, line 495: Would it be safer to mark these as booleans and set both the `[edge]` and `[pair]` to visited?

test/samples_test.cpp
```diff
@@ -255,3 +255,16 @@ TEST(Samples, Sponge4) {
 #endif
 }
 #endif
+
+#ifdef MANIFOLD_EXPORT
+TEST(Samples, SelfIntersect) {
+  manifold::PolygonParams().processOverlaps = true;
+  std::string file = __FILE__;
+  std::string dir = file.substr(0, file.rfind('/'));
+  Manifold m1 = ImportMesh(dir + "/../samples/models/self_intersectA.glb");
```
elalish, line 264: Is there a chance these are simple enough they can be generated by e.g. an Extrude and Warp? Still, there is value in being able to add test models - this may be nice as a helper function. I think I'd prefer to put them in a dedicated models directory in `test` rather than `samples` though.

elalish (changes requested):
Thanks!

test/samples_test.cpp
stephomi: Probably not, it’s a bit sensitive and I couldn’t replicate when I tried with other self intersected tubes,
Maybe if you managed to output the exact same sweep but probably lot of trial & error, and I’m lazy.

src/manifold/src/edge_op.cpp
stephomi: It will likely output more degenerate tris, but I’ll try.
Otherwise maybe tagging `visited[pair] = edge`.
Ultimately, slight changes in the logic gives different result for the ‘Sponge4’ test

elalish: Yeah, I'm not super worried about the variations in Sponge4 - really I need to come back to it and figure out why it has so many in the first place - I had managed to get rid of most of them at one time. 

Curious what you find here - I need to look into this logic more deeply at some point. It's been awhile.

test/samples_test.cpp
elalish: Okay, no prob - but let's at least change the directory.

src/manifold/src/edge_op.cpp
stephomi: Found a way to get the same number of degenerate Tris as master, while avoiding recursion.
Instead of booleans I use tag that I reset when there is a `CollapseEdge`.

---
Alternative version: (in case the current solution skips valid case)
Using a stack and hash pair of edges instead of checking both edges separately (I don't think it's necessary)

```diff
diff --git a/src/manifold/src/edge_op.cpp b/src/manifold/src/edge_op.cpp
index 26fd983..e1c2e73 100644
--- a/src/manifold/src/edge_op.cpp
+++ b/src/manifold/src/edge_op.cpp
@@ -222,12 +222,11 @@ void Manifold::Impl::SimplifyTopology() {
                      precision_};
     for_each_n(policy, countAt(0), nbEdges,
                [=] __host__ __device__(int i) { bflagsPtr[i] = se(i); });
-    std::vector<int> visited(halfedge_.size(), -1);
-    int tag = 0;
+    std::vector<glm::ivec2> visited(halfedge_.size(), glm::ivec2(-1));
     for (int i = 0; i < nbEdges; ++i) {
       if (bflags[i]) {
-        tag++;
-        RecursiveEdgeSwap(i, visited, tag);
+        visited.resize(0);
+        RecursiveEdgeSwap(i, visited);
         numFlagged++;
       }
     }
@@ -487,7 +486,7 @@ void Manifold::Impl::CollapseEdge(const int edge) {
 }
 
 void Manifold::Impl::RecursiveEdgeSwap(const int edge,
-                                       std::vector<int>& visited, int& tag) {
+                                       std::vector<glm::ivec2>& visited) {
   VecDH<TriRef>& triRef = meshRelation_.triRef;
 
   if (edge < 0) return;
@@ -495,7 +494,9 @@ void Manifold::Impl::RecursiveEdgeSwap(const int edge,
   if (pair < 0) return;
 
   // avoid infinite recursion
-  if (visited[edge] == tag && visited[pair] == tag) return;
+  for (int i = 0; i < visited.size(); ++i) {
+    if (visited[i] == glm::ivec2(edge, pair)) return;
+  }
 
   const glm::ivec3 tri0edge = TriOf(edge);
   const glm::ivec3 tri1edge = TriOf(pair);
@@ -575,15 +576,14 @@ void Manifold::Impl::RecursiveEdgeSwap(const int edge,
     SwapEdge();
     const glm::vec2 e23 = v[3] - v[2];
     if (glm::dot(e23, e23) < precision_ * precision_) {
-      tag++;
+      visited.resize(0);
       CollapseEdge(tri0edge[2]);
     } else {
-      visited[edge] = tag;
-      visited[pair] = tag;
-      RecursiveEdgeSwap(tri0edge[0], visited, tag);
-      RecursiveEdgeSwap(tri0edge[1], visited, tag);
-      RecursiveEdgeSwap(tri1edge[0], visited, tag);
-      RecursiveEdgeSwap(tri1edge[1], visited, tag);
+      visited.emplace_back(glm::ivec2(edge, pair));
+      RecursiveEdgeSwap(tri0edge[0], visited);
+      RecursiveEdgeSwap(tri0edge[1], visited);
+      RecursiveEdgeSwap(tri1edge[0], visited);
+      RecursiveEdgeSwap(tri1edge[1], visited);
     }
     return;
   } else if (CCW(v[0], v[3], v[2], precision_) <= 0 ||
@@ -592,10 +592,9 @@ void Manifold::Impl::RecursiveEdgeSwap(const int edge,
   }
   // Normal path
   SwapEdge();
-  visited[edge] = tag;
-  visited[pair] = tag;
-  RecursiveEdgeSwap(halfedge_[tri0edge[1]].pairedHalfedge, visited, tag);
-  RecursiveEdgeSwap(halfedge_[tri1edge[0]].pairedHalfedge, visited, tag);
+  visited.emplace_back(glm::ivec2(edge, pair));
+  RecursiveEdgeSwap(halfedge_[tri0edge[1]].pairedHalfedge, visited);
+  RecursiveEdgeSwap(halfedge_[tri1edge[0]].pairedHalfedge, visited);
 }
 
 void Manifold::Impl::SplitPinchedVerts() {
diff --git a/src/manifold/src/impl.h b/src/manifold/src/impl.h
index dc4f57b..cb890df 100644
--- a/src/manifold/src/impl.h
+++ b/src/manifold/src/impl.h
@@ -121,7 +121,7 @@ struct Manifold::Impl {
   void SimplifyTopology();
   void DedupeEdge(int edge);
   void CollapseEdge(int edge);
-  void RecursiveEdgeSwap(int edge, std::vector<int>& visited, int& tag);
+  void RecursiveEdgeSwap(int edge, std::vector<glm::ivec2>& visited);
   void RemoveIfFolded(int edge);
   void PairUp(int edge0, int edge1);
   void UpdateVert(int vert, int startEdge, int endEdge);
```

elalish (approved):
Nice solution, thanks!

src/manifold/src/edge_op.cpp
elalish: Interesting, I tend to prefer to simpler version you have now, though I don't quite grok the difference. If you want, it's probably best to put this as a follow-on PR and add a test that shows the difference.

PR #503, opened 2023-07-22
some more optimizations

pca006132:
This is mainly optimization for tbb, with as much as -30% running time for some extreme example.

1. Use parallel stl algorithms when thrust doesn't use the parallel version when `thrust::tbb::par` is specified (https://github.com/NVIDIA/thrust/issues/1977)
2. Use parallel sort from stl, which seems to be slightly faster than the implementation in thrust.
3. Optimize triangulator using monotonic memory pool (list nodes are not freed until the triangulator is destroyed), and change `std::stack` which uses deque backend to vector.

## Performance differences with tbb

Surprisingly, the performance improvement is quite significant for such simple changes.

`perfTest` (max: **-8%**):
```
nTri = 512, time = 0.00169123 sec
nTri = 2048, time = 0.00497043 sec
nTri = 8192, time = 0.01013 sec
nTri = 32768, time = 0.0302614 sec
nTri = 131072, time = 0.0534231 sec
nTri = 524288, time = 0.15851 sec
nTri = 2097152, time = 0.691089 sec
nTri = 8388608, time = 2.94239 sec

-------------------------------------

nTri = 512, time = 0.00170492 sec
nTri = 2048, time = 0.00422943 sec
nTri = 8192, time = 0.00834682 sec
nTri = 32768, time = 0.0227696 sec
nTri = 131072, time = 0.0404606 sec
nTri = 524288, time = 0.139214 sec
nTri = 2097152, time = 0.619758 sec
nTri = 8388608, time = 2.69757 sec
```

Selected long-running tests (max: **-30%**)
```
Boolean.Close (852 ms -> 614ms)
Boolean.Sweep (245 ms -> 209ms)
Samples.Bracelet (330 ms -> 231ms)
Samples.GyroidModule (309 ms -> 248ms)
Samples.Sponge4 (7234 ms -> 6854ms)
```

And I did not observe notable performance differences for wasm build (-60ms in total), so it should not cause regression.


pca006132:
oh the rebase somehow changed the committer, weird

elalish:
> oh the rebase somehow changed the committer, weird

In general, I would recommend avoiding the use of rebase entirely. Merge is safer and since we do a squash and merge, there's no pollution of our commit history anyway. It also avoids the need for force-push and makes it easier to read the sequential diffs on Github.

elalish (changes requested):
Let's see if we can pull master to simplify this diff, since it seems to have a bunch of previous PRs in it. Then I'll review, but this sounds great!

pca006132:
> Let's see if we can pull master to simplify this diff, since it seems to have a bunch of previous PRs in it. Then I'll review, but this sounds great!

Should be simplified now.

pca006132:
I give up with macos fix, maybe @elalish you can see what preprocessor macro does macos provide

src/polygon/src/polygon.cpp
```diff
@@ -15,8 +15,16 @@
 #include "polygon.h"
 
 #include <algorithm>
+#if MANIFOLD_PAR == 'T'
+#include <execution>
+#endif
 #include <list>
 #include <map>
+#if _APPLE_
```
elalish, line 23: Looks like the problem here is it needs to be `__APPLE__` instead of `_APPLE_`.

src/polygon/src/polygon.cpp
```diff
@@ -15,8 +15,16 @@
...
+#include <experimental/memory_resource>
```
elalish, line 24: However, it looks like apple doesn't even has complete `pmr` support in experimental: https://discourse.llvm.org/t/std-pmr-maturity/62200

elalish (changes requested):
I don't know much about `pmr`, but it looks like it's not ready on Apple's compiler. Can we fall back without making the code too crazy?

elalish (approved):
Thanks!

PR #504, opened 2023-07-26
adding python set_properties

elalish:
Adding our colored Menger Sponge example to verify that mesh properties are working properly. I'll also update our trimesh usage to write a colored GLB.

### DIFF
--- a/bindings/python/examples/sponge.py
+++ b/bindings/python/examples/sponge.py
@@ -1,4 +1,5 @@
 from manifold3d import Manifold
+import numpy as np
 
 
 def fractal(holes, hole, w, position, depth, maxDepth):
@@ -7,7 +8,7 @@ def fractal(holes, hole, w, position, depth, maxDepth):
         [position[0], position[1], 0.0]))
     if depth == maxDepth:
         return
-    offsets = [
+    offsets = np.array([
         [-w, -w],
         [-w, 0.0],
         [-w, w],
@@ -16,7 +17,7 @@ def fractal(holes, hole, w, position, depth, maxDepth):
         [w, 0.0],
         [w, -w],
         [0.0, -w],
-    ]
+    ])
     for offset in offsets:
         fractal(holes, hole, w, position + offset, depth + 1, maxDepth)
 
@@ -28,7 +29,7 @@ def posColors(pos, _):
 def run(n=1):
     result = Manifold.cube([1, 1, 1], True)
     holes = []
-    fractal(holes, result, 1.0, [0.0, 0.0], 1, n)
+    fractal(holes, result, 1.0, np.array([0.0, 0.0]), 1, n)
 
     hole = Manifold.compose(holes)
 
--- a/bindings/python/examples/sponge.py
+++ b/bindings/python/examples/sponge.py
@@ -3,7 +3,8 @@ from manifold3d import Manifold
 
 def fractal(holes, hole, w, position, depth, maxDepth):
     w /= 3
-    holes.append(hole.scale([w, w, 1.0]).translate([position[0], position[1], 0.0]))
+    holes.append(hole.scale([w, w, 1.0]).translate(
+        [position[0], position[1], 0.0]))
     if depth == maxDepth:
         return
     offsets = [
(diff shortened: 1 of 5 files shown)


bindings/python/manifold3d.cpp
```diff
@@ -254,6 +254,31 @@ PYBIND11_MODULE(manifold3d, m) {
...
+          [](Manifold &self, int newNumProp,
+             const std::function<void(py::array_t<float> &, Float3,
+                                      const py::array_t<float> &)> &f) {
+            const int oldNumProp = self.NumProp();
+            return self.SetProperties(newNumProp, [newNumProp, oldNumProp, &f](
+                                                      float *newProps,
+                                                      glm::vec3 v,
+                                                      const float *oldProps) {
+              f(py::array(newNumProp, newProps), std::make_tuple(v.x, v.y, v.z),
+                py::array(oldNumProp, oldProps));
```
elalish, line 268: I'm getting a compile error here that I don't understand. Am I missing something?

pca006132:
I guess I understand the problem here, but this is not a simple issue to solve. I think the best way would be to change the python API for set_properties (and warp as well) due to fundamental issue with lifetime management and efficiency concerns.

The problem here is that py::array will and must perform copying when we give it a pointer, so the performance is slow due to many allocations and we need to copy the changes back.
Buffer protocol can allow us to give the raw buffer to python directly, but lifetime management would be a pain: we need to avoid deallocating the original vector until the buffer is released, and this requires reference counting.

elalish:
Ugh, that's too bad - I had thought the point of giving `py::array` a pointer was to avoid copying. I'm surprised that functional binding is so much worse in Python than WASM. I don't really want to expose more internals in our public API just to get around Python's weirdness. What you're saying sounds a lot like `GetMeshGL`, rebuild `vertProperties`, then make a new Manifold. Maybe we should just tell people to do that and skip `set_properties` entirely?

pca006132:
No it is not much worse comparing with WASM, we are also doing copying in our js binding or we will have similar issues regarding lifetime. We can do something similar, but I think the performance will likely be bad with Python.

Yeah, I think skipping `set_properties` is reasonable.

pca006132:
It should be pretty simple, just don't think about avoid copying and you are done, basically just the logic in our js code, but need to convert into python code. https://github.com/elalish/manifold/blob/59760b3369b6faac5a46f1d30ad9d3bbda16b1f1/bindings/wasm/bindings.js#L220-L245

I can only do it this weekend, as I am in the middle of doing something else.

pca006132:
Finished, and I also updated our trimesh export example to export vertex colors.

bindings/python/examples/sponge.py
```diff
@@ -0,0 +1,40 @@
...
+        [w, 0.0],
+        [w, -w],
+        [0.0, -w],
+    ])
+    for offset in offsets:
+        fractal(holes, hole, w, position + offset, depth + 1, maxDepth)
+
+
+def posColors(pos, _):
+    return [1 - p / 2 for p in pos] + [1.0]
```
elalish, line 26: Love the simplicity!

bindings/python/examples/sponge.py
```diff
@@ -0,0 +1,40 @@
...
+    holes = []
+    fractal(holes, result, 1.0, np.array([0.0, 0.0]), 1, n)
+
+    hole = Manifold.compose(holes)
+
+    result -= hole
+    result -= hole.rotate([90, 0, 0])
+    result -= hole.rotate([0, 90, 0])
+
+    return result.trim_by_plane([1, 1, 1], 0).set_properties(4, posColors).scale(100)
```
elalish, line 40: How is the new number of properties 4? It seems like it would be 3, and anything else would trigger your invalid shape error below? Did you check that the resulting GLB looks like our logo?

elalish (commented):
Thanks for fixing this for me!

bindings/python/examples/sponge.py
pca006132: trimesh expects rgba for vert_color (and max is 255, but I did the conversion in the export part)

elalish: Oh, my python is just weak - I totally missed that ` + [1.0]` was pushing an additional item into the array. Thanks!

Issue #502, opened 2023-07-21
Difference of Cube and Complex SVG Results in Incomplete Render (openscad/openscad#4679)


pca006132:
Downstream issue: https://github.com/openscad/openscad/issues/4679

Reproduce (requires #492):
```python
import trimesh
from manifold3d import Manifold, MeshGL

def export(model, filename, extension='glb'):
    mesh = model.to_mesh()
    meshOut = trimesh.Trimesh(
        vertices=mesh.vert_properties, faces=mesh.tri_verts)
    trimesh.exchange.export.export_mesh(meshOut, f'{filename}', extension)

woodgrain = trimesh.load('woodgrain.stl')
woodgrain = Manifold.from_mesh(MeshGL(woodgrain.vertices, woodgrain.faces))
model = Manifold.cube([145, 100, 1], True) - woodgrain

export(model, 'test.glb')
```

with the `woodgrain.stl` in [woodgrain.zip](https://github.com/openscad/openscad/files/12113817/woodgrain.zip).

Current Output:
![image](https://github.com/elalish/manifold/assets/12198657/b686de1b-967f-4722-a28a-8282c2b56e7a)

![image](https://github.com/elalish/manifold/assets/12198657/c4d967c9-c902-4df9-98dc-59a2f7ec3aa8)

Note that the output from union is correct:
![image](https://github.com/elalish/manifold/assets/12198657/00bb0015-b647-4d73-9992-44612c80f5be)

pca006132:
It seems that this is due to out-of-bound access, should be easy to debug. Will work on this later today.

pca006132:
OK the out-of-bound access is due to https://github.com/NVIDIA/thrust/issues/1648.
The main issue here is that the result surface is considered not geometrically valid. I verified that the input mesh is indeed a 2 manifold (using the `Is2Manifold` method), and it seems that the input mesh does not contain any self intersection, but probably many small line segments messing things up.

pca006132:
![image](https://github.com/elalish/manifold/assets/12198657/dff70753-b2c4-43ce-8f56-3dccd3560fcf)

In particular, these triangles are likely causing problem...

elalish:
So, triangulator failed is not itself a bug. We're robust to that unless process overlaps is false, which is only for testing/debugging. Where is the out of bounds access occurring? Is that actually related to the overlap?

elalish:
Could this be #501?

pca006132:
[...]

PR #507, opened 2023-07-29
ugly hack to work around #502

pca006132:
https://github.com/elalish/manifold/issues/502#issuecomment-1656556780

This is an ugly hack to workaround issue #502, which seems to affect quite a lot of OpenSCAD users when they extrude from SVG and perform boolean operations.

The idea is that #502 is caused by skipping vertices incorrectly, so we try to triangulate the polygon without considering precision first, and triangulate it the second time if the first attempt failed. This makes some of the geometry checks necessary, so I modified some of the debug related code to make things compile and working without `MANIFOLD_DEBUG=ON`. Note that the performance with `MANIFOLD_DEBUG=OFF` is still very fast because we only enabled *some* checks in `polygon.cpp`, instead of all other checks that `MANIFOLD_DEBUG=ON` will enable.

What I am uncertain about:
1. Should we also do this when `params.overlapChecks = false`?
2. Should we merge this before we get a real fix for #502? Or should we wait a bit?


pca006132:
image
image

I tried building openscad with this patch, and it did work.

elalish:
Okay, I am more and more confused as to what this bug actually is. Is it a crash, or non-manifold output, or geometrically incorrect output? Are the images above correct or incorrect output?

elalish:
Okay, rereading the original openscad bug, I think I've answered my own question: no crash, all output is manifold, but the geometry is off because of a bad triangulation. Let me see if I can fix the underlying problem before we put in a hack. Thanks for all the effort narrowing this down!

PR #508, opened 2023-07-29
tighten CCW

### LINKED ISSUE #502: Difference of Cube and Complex SVG Results in Incomplete Render (openscad/openscad#4679)

elalish:
Partially solves #502 

Alternative to #507. I found a problem with `CCW`, where it was not properly conservative because the base could be up to twice as long as calculated if the `v0` is near the midpoint between `v1` and `v2`.

### DIFF
--- a/src/utilities/include/public.h
+++ b/src/utilities/include/public.h
@@ -106,7 +106,7 @@ inline HOST_DEVICE int CCW(glm::vec2 p0, glm::vec2 p1, glm::vec2 p2,
   glm::vec2 v2 = p2 - p0;
   float area = v1.x * v2.y - v1.y * v2.x;
   float base2 = glm::max(glm::dot(v1, v1), glm::dot(v2, v2));
-  if (area * area <= base2 * tol * tol)
+  if (area * area * 4 <= base2 * tol * tol)
     return 0;
   else
     return area > 0 ? 1 : -1;
--- a/test/polygon_test.cpp
+++ b/test/polygon_test.cpp
@@ -301,6 +301,23 @@ TEST(Polygon, ColinearY) {
   TestPoly(polys, 16);
 }
 
+TEST(Polygon, NearlyColinearY) {
+  Polygons polys;
+  polys.push_back({
+      {3.5, 0},       //
+      {3.0, 0.5},     //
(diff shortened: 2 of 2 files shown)


pca006132:
Wow it is amazing that the cause of issue is so simple!

However, I found that although this fixes the simplified repro in #502, there is still triangulation error for the original mesh in https://github.com/openscad/openscad/issues/4679, which you can reproduce by extracting the [`woodgrain.stl`](https://github.com/openscad/openscad/files/12113817/woodgrain.zip) to `test/models`, and then run

```c++
#ifdef MANIFOLD_EXPORT
TEST(Samples, SvgBreak) {
  Manifold m1 = ImportMesh("../../test/models/woodgrain.stl");
  Manifold output = Manifold::Cube({145, 100, 1}, true) - m1;
  ExportMesh("SvgBreak.glb", output.GetMesh(), {});
}
#endif
```

elalish:
Thanks, yeah I just noticed that your other simple polygon test is still failing - it's actually an unrelated issue. As part of debug I just realized `intermediateChecks` is not actually running during our tests like I thought it was. I am fundamentally misunderstanding something about how our `ExecutionParameters` are getting compiled.

elalish:
Okay, I'm going to go ahead and merge this one and deal with the other triangulation issues on a separate branch.

PR #516, opened 2023-08-02
more triangulator check

pca006132:
This enables triangle overlap check when `MANIFOLD_DEBUG` is enabled. And I also added a tool called `minimizeTestcase` for minimizing triangulator testcases. ~The added checks should not slow the triangulator down too much when `MANIFOLD_DEBUG=on`. On my computer with TBB enabled, `Samples.Sponge4` runs in 6838ms comparing with > 7000ms previously.~

`minimizeTestcase` does the following:
1. `Usage: ./minimizeTestcase <input file>` where input file is a path to a text file containing the error output from the triangulator when `MANIFOLD_DEBUG` is on. For example:
```
Triangulation failed! Precision = 0.00144999998
Error in file: /home/pca006132/code/manifold/src/polygon/src/polygon.cpp (47): 'condition' is false: Not Geometrically Valid! None of the skipped verts is valid.
polys.push_back({
    {-15.694437, 33.8253479},  //
    {13.2702093, -3.28426218},  //
    {53.4751701, -32.0106316},  //
    {61.4174232, 44.4299698},  //
});
polys.push_back({
    {17.61759, 25.2365837},  //
    {17.6270828, 25.2370052},  //
    {17.1236248, 26.0772629},  //
    {20.4602833, -7.37109232},  //
});
polys.push_back({
    {44.9942932, 39.2275734},  //
    {45.120945, 29.0997658},  //
    {44.896656, 25.2351685},  //
});
polys.push_back({
    {61.4174271, 33.8253479},  //
    {90.3820724, -3.28426218},  //
    {130.587036, -32.0106316},  //
    {138.529282, 44.4299698},  //
});
polys.push_back({
    {94.729454, 25.2365837},  //
    {94.738945, 25.2370052},  //
    {94.2354889, 26.0772629},  //
    {97.5721436, -7.37109232},  //
});
polys.push_back({
    {122.106155, 39.2275734},  //
    {122.232803, 29.0997658},  //
    {122.008514, 25.2351685},  //
});
```
2. The tool will read the *first* `Polygons` in the input file, and try to remove vertices and simple polygons one by one while keeping the resulting polygon *correct* (does not introduce overlapping edges, and not inverting winding directions), while keeping the same error output.
3. The tool will output the minimal instance that it can get, along with the erroneous triangulation that you can visualize with matplotlib. If a triangle is not in CCW orientation, it will point it out by writing `^ not CCW` in the line below. (should be relatively efficient, I can minimize a 50k vertices test case into 10 vertices in 32s)

~For now, the tool will not check if the input polygon is valid, but doing this should not be hard.~ The tool will check for self-intersection but not check if the polygon winding directions are correct. (not sure what is the best way for checking that when we can have subtle self-intersections that are epsilon-valid)

I also added two test cases that I got from the woodgrain stl file.


extras/CMakeLists.txt
```diff
@@ -24,6 +24,13 @@ target_link_libraries(largeSceneTest manifold)
 target_compile_options(largeSceneTest PRIVATE ${MANIFOLD_FLAGS})
 target_compile_features(largeSceneTest PUBLIC cxx_std_17)
 
+if(MANIFOLD_DEBUG)
```
elalish, line 27: Since this is an extra, should we just go ahead and compile it always?

extras/minimize_testcase.cpp
```diff
@@ -0,0 +1,251 @@
...
+std::vector<int> getChildren(Polygons &polys, int i) {
+  std::vector<int> results;
+  for (int j = 0; j < polys.size(); j++) {
+    if (i == j) continue;
+    glm::vec2 point = polys[j][0];
+    auto k1 = [&](size_t k) {
+      return k == (polys[i].size() - 1) ? 0 : (k + 1);
+    };
+    int count = std::count_if(
+#if MANIFOLD_PAR == 'T' && !(__APPLE__)
```
elalish, line 96: Can we use `__has_include` instead?

extras/minimize_testcase.cpp
```diff
@@ -0,0 +1,251 @@
...
+    int points = 0;
+    for (auto &poly : polys) points += poly.size();
+    std::cout << "trying to simplify " << points << " points" << std::endl;
+
+    removedSomething = false;
+    // try to remove simple polygons
+    for (int i = 0; i < polys.size(); i++) {
+      std::vector<int> children = getChildren(polys, i);
+      // if there are children, we can't remove it or we will mess up with
+      // winding direction...
```
elalish, line 123: I'll have to study your algorithm sometime when I'm less tired, but I trust you. This will be a great convenience function even if it misses an edge case or two (not that I've found any).

src/polygon/src/polygon.cpp
```diff
@@ -83,6 +85,7 @@ std::vector<PolyEdge> Polygons2Edges(const PolygonsIdx &polys) {
 std::vector<PolyEdge> Triangles2Edges(
     const std::vector<glm::ivec3> &triangles) {
   std::vector<PolyEdge> halfedges;
+  halfedges.reserve(triangles.size() * 3);
```
elalish, line 88: 👍

src/polygon/src/polygon.cpp
```diff
@@ -155,6 +158,46 @@ void CheckGeometry(const std::vector<glm::ivec3> &triangles,
...
+        edges[3 * i + 1] =
+            std::make_pair(vertPos[triangles[i].y], vertPos[triangles[i].z]);
+        edges[3 * i + 2] =
+            std::make_pair(vertPos[triangles[i].z], vertPos[triangles[i].x]);
+      });
+  std::mutex mutex;
+  std::vector<std::pair<size_t, size_t>> overlappingPairs;
+  auto p = [&](size_t i) {
+    return std::all_of(countAt(i + 1), countAt(edges.size()), [&](size_t j) {
+      if (intersect(edges[i].first, edges[i].second, edges[j].first,
```
elalish, line 179: Related to my comment on your last PR - I don't *think* you need to check intersection of edges. I believe a manifold triangulation has every triangle CCW if and only if no triangles overlap/intersect. That's why I never had this intersection check before. Have you found a counter example?

test/polygon_test.cpp
```diff
@@ -3086,3 +3086,46 @@ TEST(Polygon, BigSponge) {
...
+  polys.push_back({
+      {27.7341537, -49.9853325},  //
+      {27.4095287, -49.6340637},  //
+      {27.9608383, -49.758522},   //
+  });
+  polys.push_back({
+      {-1.56910598, -49.98423},    //
+      {-1.58251643, -49.9838409},  //
+      {-1.19169295, -48.0386696},  //
+      {-1.02179492, -49.9855652},  //
```
elalish, line 3106: 👍

extras/minimize_testcase.cpp
elalish: I'd like to avoid `__APPLE__` and any other platform-specific macros if possible.

extras/CMakeLists.txt
pca006132: no because it requires `MANIFOLD_DEBUG=on` so `geometryErr` is defined.

src/polygon/src/polygon.cpp
pca006132: We did have a counter example previously: https://github.com/elalish/manifold/issues/502#issuecomment-1653968014

Not sure if this can still happen after #508, but I think it doesn't hurt to check? The check is not too slow (comparing with the original `checkGeometry`.)

extras/minimize_testcase.cpp
pca006132: I can add more comments to explain it as well.

src/polygon/src/polygon.cpp
```diff
@@ -155,6 +158,46 @@ void CheckGeometry(const std::vector<glm::ivec3> &triangles,
...
+                    edges[j].second, precision)) {
+        mutex.lock();
+        overlappingPairs.push_back(std::make_pair(i, j));
+        mutex.unlock();
+        return true;
```
pca006132, line 184: I found that I actually flipped the true/false. The actual check takes much longer even with some optimization, maybe I should hide it after another parameter. Will fix.

pca006132: When this is fixed, Sponge4 takes ~15s to run, which is pretty slow...

src/polygon/src/polygon.cpp
pca006132: An interesting thing I found is that, some triangulation is only epsilon-valid but not strictly valid. I guess this is a consequence of the input mesh/polygon being only epsilon-valid?

pca006132:
Here is an optimization that runs triangulation in parallel, so test cases like `Sample.Sponge4` can run faster (15s -> 7s)

```diff
diff --git a/src/manifold/src/face_op.cpp b/src/manifold/src/face_op.cpp
index 0437ad3..680ff0a 100644
--- a/src/manifold/src/face_op.cpp
+++ b/src/manifold/src/face_op.cpp
@@ -12,6 +12,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+#include <future>
 #include <map>
 
 #include "impl.h"
@@ -39,6 +40,10 @@ void Manifold::Impl::Face2Tri(const VecDH<int>& faceEdge,
   triNormal.reserve(halfedge_.size());
   triRef.reserve(halfedge_.size());
 
+  std::vector<
+      std::pair<int, std::future<std::pair<int, std::vector<glm::ivec3>>>>>
+      slowTasks;
+
   for (int face = 0; face < faceEdge.size() - 1; ++face) {
     const int firstEdge = faceEdge[face];
     const int lastEdge = faceEdge[face + 1];
@@ -116,11 +121,24 @@ void Manifold::Impl::Face2Tri(const VecDH<int>& faceEdge,
         triNormal.push_back(normal);
         triRef.push_back(halfedgeRef[firstEdge]);
       }
+    } else if (numEdge > 512) {
+      slowTasks.push_back(std::make_pair(
+          numEdge, std::async(
+                       std::launch::async,
+                       [&](int face) {
+                         const glm::vec3 normal = faceNormal_[face];
+                         const glm::mat3x2 projection =
+                             GetAxisAlignedProjection(normal);
+                         const PolygonsIdx polys =
+                             Face2Polygons(face, projection, faceEdge);
+                         std::vector<glm::ivec3> newTris =
+                             TriangulateIdx(polys, precision_);
+                         return std::make_pair(face, std::move(newTris));
+                       },
+                       face)));
     } else {  // General triangulation
       const glm::mat3x2 projection = GetAxisAlignedProjection(normal);
-
       const PolygonsIdx polys = Face2Polygons(face, projection, faceEdge);
-
       std::vector<glm::ivec3> newTris = TriangulateIdx(polys, precision_);
 
       for (auto tri : newTris) {
@@ -130,6 +148,22 @@ void Manifold::Impl::Face2Tri(const VecDH<int>& faceEdge,
       }
     }
   }
+
+  std::sort(slowTasks.begin(), slowTasks.end(),
+            [](const auto& p1, const auto& p2) { return p1.first < p2.first; });
+  for (auto& task : slowTasks) {
+    task.second.wait();
+    const auto pair = task.second.get();
+    const int face = pair.first;
+    const int firstEdge = faceEdge[face];
+    const glm::vec3 normal = faceNormal_[face];
+    for (auto tri : pair.second) {
+      triVerts.push_back(tri);
+      triNormal.push_back(normal);
+      triRef.push_back(halfedgeRef[firstEdge]);
+    }
+  }
+
   faceNormal_ = std::move(triNormal);
   CreateHalfedges(triVerts);
 }
@@ -143,7 +177,7 @@ PolygonsIdx Manifold::Impl::Face2Polygons(int face, glm::mat3x2 projection,
   const int firstEdge = faceEdge[face];
   const int lastEdge = faceEdge[face + 1];
 
-  std::map<int, int> vert_edge;
+  std::unordered_map<int, int> vert_edge;
   for (int edge = firstEdge; edge < lastEdge; ++edge) {
     const bool inserted =
         vert_edge.emplace(std::make_pair(halfedge_[edge].startVert, edge))
```

The main problem is that I am not sure if we can reorder faces like this.

src/polygon/src/polygon.cpp
elalish: How much difference does removing the intersection checks make now? Those look like they would be pretty slow...

elalish:
For winding directions, I think it's probably adequate to check if the overall area is positive. It might miss an inverted hole or something, but that should become pretty obvious from the reduced example. The right way is to find each polygon's winding number by testing a non-degenerate point, then testing the sign of the area against that, but that's likely more trouble than it's worth.

src/polygon/src/polygon.cpp
pca006132: Indeed, one of them is CW. I guess I can remove this check.

pca006132:
> I think it's okay to make the faces in a different order - they're independent calculations and the triangles get resorted again anyway in `Finish()`. If it's making things faster I think it's fine.

It increases the number of degenerate triangles in `Samples.Sponge4` by about 1000, from ~39500 to ~40500.

> For winding directions, I think it's probably adequate to check if the overall area is positive. It might miss an inverted hole or something, but that should become pretty obvious from the reduced example. The right way is to find each polygon's winding number by testing a non-degenerate point, then testing the sign of the area against that, but that's likely more trouble than it's worth.

Yeah sounds quite complicated, I think the user can visualize the polygon and look at winding direction by themselves.

elalish:
The extra 1000 degenerates sounds like it's in the noise to me - no surprise there's some order-dependence there. I need to come back and see if I can get that down again sometime. It should be possible to remove all of them, honestly.

extras/minimize_testcase.cpp
```diff
@@ -0,0 +1,344 @@
...
+      if (!children.empty()) continue;
+      SimplePolygon poly = std::move(polys[i]);
+      polys.erase(polys.begin() + i);
+      try {
+        auto result = Triangulate(polys, precision);
+        polys.insert(polys.begin() + i, std::move(poly));
+      } catch (geometryErr &e) {
+        if (msg.size() > 0 && msg.compare(e.what()) != 0) {
```
elalish, line 140: What's happening here? The first time we get an error, it proceeds, but if we get it twice then we try something different? Do we need to reset `msg` at some point?

src/manifold/src/face_op.cpp
```diff
@@ -130,6 +148,22 @@ void Manifold::Impl::Face2Tri(const VecDH<int>& faceEdge,
       }
     }
   }
+
+  std::sort(slowTasks.begin(), slowTasks.end(),
+            [](const auto& p1, const auto& p2) { return p1.first < p2.first; });
+  for (auto& task : slowTasks) {
+    task.second.wait();
+    const auto pair = task.second.get();
```
elalish, line 156: `auto` is great for complex types, but it might be nice to be explicit here for readability.

extras/minimize_testcase.cpp
```diff
@@ -0,0 +1,344 @@
...
+  }
+  Edge(const Polygons &polys, size_t i, size_t j) : i(i), j(j) {
+    size_t j1 = j == polys[i].size() - 1 ? 0 : j + 1;
+    east = polys[i][j];
+    west = polys[i][j1];
+    if (west.x > east.x) std::swap(east, west);
+  }
+};
+
+int isValid(const Polygons &polys, float precision = -1) {
```
elalish, line 210: Do we need this function at all anymore? Or if the goal is to print where the problems are, maybe just list the CW triangles?

pca006132:
ah I have no idea why the wasm build failed. it seems that the code runs fine, but the js part is now wrong?

src/utilities/include/utils.h
```diff
@@ -220,5 +215,52 @@ class ConcurrentSharedPtr {
   std::shared_ptr<std::recursive_mutex> mutex =
       std::make_shared<std::recursive_mutex>();
 };
+
+inline float cross(glm::vec2 p, glm::vec2 q) { return p.x * q.y - p.y * q.x; }
+
+// return true if p intersects with q
+// note that we don't care about collinear, intersection in the ends etc.
+// https://stackoverflow.com/questions/563198/how-do-you-detect-where-two-line-segments-intersect
+inline bool intersect(glm::vec2 p0, glm::vec2 p1, glm::vec2 q0, glm::vec2 q1,
```
elalish, line 224: Is this function still needed here?

elalish (changes requested):

extras/minimize_testcase.cpp
pca006132: The goal of this is to check if the input polygon is valid, so we can know if the geometrically invalid error is really correct.

src/utilities/include/utils.h
pca006132: Yeah it is not needed here. Will fix.

extras/minimize_testcase.cpp
pca006132: We just want to reduce the testcase so we get the same error, this is checking if the error messages are the same.

pca006132:
I moved the face reordering out because it breaks the mengerSponge js example. I will open a separate issue for it.

src/polygon/src/polygon.cpp
```diff
@@ -143,7 +144,7 @@ void CheckTopology(const std::vector<glm::ivec3> &triangles,
 
 void CheckGeometry(const std::vector<glm::ivec3> &triangles,
                    const PolygonsIdx &polys, float precision) {
-  std::map<int, glm::vec2> vertPos;
+  std::unordered_map<int, glm::vec2> vertPos;
```
pca006132, line 147: ok, this seemingly innocent change is causing bug...

elalish: Interesting, thanks! I have a plan on the triangulator, but I need to find a couple days to put it into action.

elalish (approved):

PR #517, opened 2023-08-04
3mf node export

elalish:
This makes our 3MF export consistent with our GLB export when `GLTFNode` is used, creating the same scene hierarchy with mesh reuse. No materials yet, but we do include object names and a proper header. 

This is blocked on https://github.com/hrgdavor/jscadui/pull/37, which will need an npm release for us.

### DIFF
--- a/bindings/wasm/examples/package.json
+++ b/bindings/wasm/examples/package.json
@@ -16,7 +16,7 @@
     "@gltf-transform/core": "^3.2.1",
     "@gltf-transform/extensions": "^3.2.1",
     "@gltf-transform/functions": "^3.2.1",
-    "@jscadui/3mf-export": "^0.4.0",
+    "@jscadui/3mf-export": "^0.4.2",
     "fflate": "^0.8.0",
     "gl-matrix": "^3.4.3",
     "simple-dropzone": "0.8.3",
--- a/bindings/wasm/examples/package.json
+++ b/bindings/wasm/examples/package.json
@@ -16,7 +16,7 @@
     "@gltf-transform/core": "^3.2.1",
     "@gltf-transform/extensions": "^3.2.1",
     "@gltf-transform/functions": "^3.2.1",
-    "@jscadui/3mf-export": "^0.3.0",
+    "@jscadui/3mf-export": "^0.4.0",
     "fflate": "^0.8.0",
     "gl-matrix": "^3.4.3",
     "simple-dropzone": "0.8.3",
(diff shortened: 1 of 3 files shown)


elalish:
FYI @hrgdavor

hrgdavor:
just released `@jscadui/3mf-export/0.4.0` with changes incorporated.

I would like to point out that the export supports embedding a thumbnail, it would be cool if manifold would take advantage of that when exporting. There is example of it in the repo called testGen.js

bindings/wasm/examples/package.json
```diff
@@ -16,7 +16,7 @@
     "@gltf-transform/core": "^3.2.1",
     "@gltf-transform/extensions": "^3.2.1",
     "@gltf-transform/functions": "^3.2.1",
-    "@jscadui/3mf-export": "^0.3.0",
+    "@jscadui/3mf-export": "^0.4.0",
```
hrgdavor, line 19: needs to be 0.4.1

hrgdavor (changes requested):
version change needed

elalish:
> there was a small bug introduced in `to3dmodel` ... parameter `...header` was changed to `header` which is not correct ... header is supposed to pickup all other keys from the function options param, and fwd it.

Hmm, I actually changed that intentionally in order to make the header parameters work (and tested it). Since `header` is passed as a complete object to `pushHeader`, if you use `...` you don't actually end up with a variable called `header` at all (you only get it's sub-members individually).

hrgdavor:
What does your calling the function look like ? This dotted variant works in the test script within the package

hrgdavor:
@elalish I checked the header parameter and the types you proposed. I agree with separating the header object instead of mixing the keys in the options.

v0.4.2 is published that is aligned with your call to `to3dmodel` that I found more practical. I also separated out `to3dmodelSimple` variant of the function for cases where I have just set of meshes without a hiararchy (that is usual for jscad).
I transfered the proposed interfaces to jsdoc format with minor tweaks.

elalish:
@hrgdavor Awesome, thank you! I'll get back to this shortly, but I need to go to a wedding first.

elalish:
@pca006132 Did you say something about a problem with the windows TBB build?

pca006132:
yes, maybe we can disable it for now until I find a way to fix it.

elalish:
It's already marked "not required", so I guess it's fine for the moment. Ideally we should scope down this CI list a bit (if we go full TBB) and make all the remaining ones required.

PR #519, opened 2023-08-04
fixed for_each_host and for_each_n_host

pca006132:
this is a really silly mistake... luckily this is only used in csg_tree and we rarely have over 4096 nodes to operate over.

### DIFF
--- a/src/utilities/include/par.h
+++ b/src/utilities/include/par.h
@@ -134,9 +134,9 @@ inline ExecutionPolicy autoPolicy(int size) {
         thrust::NAME(thrust::MANIFOLD_PAR_NS::par, args...); \
         break;                                               \
       case ExecutionPolicy::Seq:                             \
+        thrust::NAME(thrust::cpp::par, args...);             \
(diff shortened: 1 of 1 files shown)


src/utilities/include/par.h
```diff
@@ -134,9 +134,9 @@ inline ExecutionPolicy autoPolicy(int size) {
         thrust::NAME(thrust::MANIFOLD_PAR_NS::par, args...); \
         break;                                               \
       case ExecutionPolicy::Seq:                             \
+        thrust::NAME(thrust::cpp::par, args...);             \
```
elalish, line 137: Is the same error in the function above? Or is there some reason it works differently there?

pca006132: the function above will return the result, so it cna only fall through for `ExecutionPolicy::Seq`. Also, the reason I need to write it in this weird form instead of putting the last return in the switch case is due to compiler warning, the compiler somehow thinks that policy can fall through and never return anything.

Issue #362, opened 2023-03-10
Publish python library to pypi


elalish:
This is blocked by #361

pca006132:
I am trying to build python wheels in https://github.com/pca006132/manifold/tree/python-wheel, but now is stuck with windows build... There are two problems:

1. Somehow cmake cannot find the tbb library installed via pip (tbb-devel) or nuget. I am now disabling tbb for windows, but hopefully we can get this sorted out later.
2. CMake does not build the `ALL_TARGET` automatically, I am now trying to copy args from our normal CI commands.

elalish:
Thanks for looking into this! I just found what might be a good [template](https://github.com/scikit-build/scikit-build-sample-projects/blob/master/projects/hello-cmake-package/README.md) to follow. I also found an interesting wheels CI [example](https://github.com/pybind/cmake_example/blob/master/.github/workflows/wheels.yml). TBB can wait, no problem. I'll defer to your superior Python and build setup skills.

pca006132:
Oh and there are other issues as well:

1. manylinux2014 uses gcc 10, which ships with pstl 10, and will cause issues when using newer tbb versions (>= 2021), the one provided by the system is too old (2014), and cmake/pkg-config cannot find the tbb installed from the intel provided script.
2. somehow our extrude example will cause python 3.7 importer error (syntax error), I have no idea why, so I am limiting it to Python >= 3.8.
3. auditwheel which is used for statically linking the libraries (libtbb.so for example) somehow is unable to find tbb installed from pip, and can only install from the system.

I am familiar with building for linux, but not so much for MacOS and Windows... I wonder if there are others that are familiar with Python packaging that can help.

elalish:
Do you want to open a PR with your WIP? I think this is pretty much the last thing to take care of for our release (aside from triangulator bugs I'm still working on).

pca006132:
OK now the remaining thing is to push the wheels to PyPI.

pca006132:
I think we should push to TestPyPi first. I think automated release should not be hard, we already have the wheels anyway.

Eric-Vin:
Super excited to try this out! Is there a rough timeline of when the wheels will be pushed to PyPi?

pca006132:
I think we will do that after the triangulator PR is merged.

pca006132:
@elalish maybe you can setup a python package first?

pca006132:
[...]

PR #521, opened 2023-08-04
build Python wheel

pca006132:
Attempt at #362.

Now the problematic thing is windows...


pca006132:
@elalish maybe you can have a look at this later? both windows and macos have trouble finding the installed package. You can use the setup.py to build and inspect the generated wheel file.

elalish:
Could it be [this](https://stackoverflow.com/a/75284960)?

pca006132:
> Could it be [this](https://stackoverflow.com/a/75284960)?

I don't think so, we only have a dynamic library file in there and we should have no python files.

elalish:
I don't know, seems like package installation didn't go well. Did it work on Linux? Did you ever look into scikit-build: https://github.com/elalish/manifold/issues/362#issuecomment-1651021130? Seems like it's designed specifically to streamline our type of building process. Wild guess says if it was easy, they wouldn't have bothered making a package for it...

pca006132:
no, I haven't tried scikit-build, but I tried another cmake extension for setuptools https://github.com/diegoferigo/cmake-build-extension without success.

elalish:
Did you see [this](https://cibuildwheel.readthedocs.io/en/stable/faq/#macos-modulenotfounderror), regarding the mac issue?

pca006132:
> Did you see [this](https://cibuildwheel.readthedocs.io/en/stable/faq/#macos-modulenotfounderror), regarding the mac issue?

yeah I saw that, but we are not calling cibuildwheel from a python script so this should not be an issue, we are running the test from within cibuildwheel (so the virtualenv should be the same).

pca006132:
I guess I should tri scikit-build later.

pca006132:
ok it seems scikit build works
I think we should merge #533 first

pca006132:
@elalish https://cibuildwheel.readthedocs.io/en/stable/deliver-to-pypi/

we probably want to run the build wheel actions on tag/release only as well, as they are too time consuming.

pca006132:
the cibuildwheel errors here are weird... will have a look at them later

Eric-Vin:
Thanks so much for your work on this, it would be great to use this tool from Python! The `distutils` error on 3.12 is probably because of this: https://peps.python.org/pep-0632/ . I've managed to work around it by specifying the following numpy versions:
```
  'numpy; python_version<"3.12.0rc1"',
  'numpy>=1.26.0b1; python_version>="3.12.0.rc1"',
```
in the following PR for a different library: https://github.com/BerkeleyAutomation/python-fcl/pull/70

pca006132:
> 'numpy; python_version<"3.12.0rc1"',

Thanks a lot!

pca006132:
I found that in order to make python source distribution work, we have to put everything in the top level directory.

pca006132:
oh and we can switch to build backend as well, which verifies the sdist for us

pca006132:
ah I have no idea why the macos build keeps running for 2 hours. I guess it is stuck somewhere

pca006132:
I am not entirely sure if my current way of using MANIFEST.in file to specify source file is idiomatic or not. It seems that it is mainly for data files, but it works.

src/cross_section/src/cross_section.cpp
```diff
@@ -291,7 +291,7 @@ std::shared_ptr<const PathImpl> CrossSection::GetPaths() const {
   if (transform_ == glm::mat3x2(1.0f)) {
     return paths_;
   }
-  paths_ = shared_paths(transform(paths_->paths_, transform_));
+  paths_ = shared_paths(::transform(paths_->paths_, transform_));
```
elalish, line 294: What does `::` do?

elalish (approved):

.github/workflows/build_wheels.yml
```diff
@@ -0,0 +1,23 @@
...
+    steps:
+      - run: git config --global submodule.fetchJobs 8
+      - uses: actions/checkout@v3
+        with:
+          submodules: recursive
+      - name: Build wheels
+        uses: pypa/cibuildwheel@v2.15.0
+      - uses: actions/upload-artifact@v3
+        with:
+          path: ./wheelhouse/*.whl
```
elalish, line 21: So, we upload the artifact to the wheelhouse, but what do with it? Should we add a release action? Should we try TestPyPI first?

elalish:
Excellent work! I can't wait to see this published as a proper Python package!

src/cross_section/src/cross_section.cpp
pca006132: It search for the transform function defined in the anonymous namespace. By default it should search for the one in the current namespace, i.e. the transform function defined in par.
I found this error when I was experimenting with precompiled headers (minor compile time improvement)

.github/workflows/build_wheels.yml
pca006132: https://github.com/pypa/cibuildwheel/blob/e8bb208374117ef8e54e28f8fada054fbe701cc8/examples/github-deploy.yml#L44-L63

test/boolean_test.cpp
```diff
@@ -636,9 +636,9 @@ TEST(Boolean, BooleanVolumes) {
   // m7 = m1 + m2 + m3
   auto m1 = Manifold::Cube({1, 1, 1});
   auto m2 = Manifold::Cube({2, 1, 1}).Transform(
-      glm::translate(glm::mat4(1.0f), glm::vec3(1.0f, 0, 0)));
+      glm::mat4x3(glm::translate(glm::mat4(1.0f), glm::vec3(1.0f, 0, 0))));
   auto m4 = Manifold::Cube({4, 1, 1}).Transform(
-      glm::translate(glm::mat4(1.0f), glm::vec3(3.0f, 0, 0)));
+      glm::mat4x3(glm::translate(glm::mat4(1.0f), glm::vec3(3.0f, 0, 0))));
```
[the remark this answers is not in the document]
pca006132: yeah, I found that we actually set `GLM_FORCE_EXPLICIT_CTOR` in `public.h`, and compile with precompiled header will cause errors. I think it is something related to include order which causes it to pass compilation in normal builds, but error when I tried using precompiled header.

pca006132:
https://pyo3.rs/v0.14.5/building_and_distribution#py_limited_apiabi3 has a good summary:

> By default, Python extension modules can only be used with the same Python version they were compiled against. For example, an extension module built for Python 3.5 can't be imported in Python 3.8. [PEP 384](https://www.python.org/dev/peps/pep-0384/) introduced the idea of the limited Python API, which would have a stable ABI enabling extension modules built with it to be used against multiple Python versions. This is also known as abi3.
> The advantage of building extension modules using the limited Python API is that package vendors only need to build and distribute a single copy (for each OS / architecture), and users can install it on all Python versions from the [minimum version](https://pyo3.rs/v0.14.5/building_and_distribution#minimum-python-version-for-abi3) and up.

So basically the idea is to have forward compatibility. Old versions can be used in higher python versions without recompiling. Potentially when python keeps evolving we still can build a limited set of python wheels. (because we need to build one for every supported python version, multiplied by the supported OS)

But the problem is that while nanobind can target abi3, it can only target abi3-py312, which is not even released yet...
So it is more of a future problem. For now, I think the main benefit is just slightly faster build time. The old pybind11 build will take quite some time to link, but nanobind can complete in 1s or something.

pca006132:
another downside of using nanobind is it only supports python 3.8+ (although 3.7 already reached EOL)

elalish:
I have no way to judge how much work it would be to switch to nanobind, but if you think it's worthwhile, I'm fine with it. I'm not *too* concerned about the long builds, but short is nice. It's also funny that it says "*By default* Python extension modules can only be used with the same Python version they were compiled against." - Does that mean they can in fact work against multiple versions as is? Seems hard to believe a minor version bump would make every module break.

pca006132:
> It's also funny that it says "By default Python extension modules can only be used with the same Python version they were compiled against." - Does that mean they can in fact work against multiple versions as is? Seems hard to believe a minor version bump would make every module break.

It depends on the ABI. By default there is no guarantee, so yeah a version bump can make every module break. I think there are also performance concerns making libraries want to use *unstable* APIs. For Python 3.12, it seems that they introduced some very efficient stable APIs so minibind can target them. (just my guess, I don't know enough about Python internals for this)

pca006132:
For pybind11 vs nanobind, I will just open another PR and see what others think. I think nanobind will be nicer in the long run but I am not sure if the subtle differences in the API will break existing code that depends on our binding.

Issue #518, opened 2023-08-04
face reordering makes js mengerSponge test fail


pca006132:
Reordering code:
```c++
diff --git a/src/manifold/src/face_op.cpp b/src/manifold/src/face_op.cpp
index 0437ad3..680ff0a 100644
--- a/src/manifold/src/face_op.cpp
+++ b/src/manifold/src/face_op.cpp
@@ -12,6 +12,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+#include <future>
 #include <map>
 
 #include "impl.h"
@@ -39,6 +40,10 @@ void Manifold::Impl::Face2Tri(const VecDH<int>& faceEdge,
   triNormal.reserve(halfedge_.size());
   triRef.reserve(halfedge_.size());
 
+  std::vector<
+      std::pair<int, std::future<std::pair<int, std::vector<glm::ivec3>>>>>
+      slowTasks;
+
   for (int face = 0; face < faceEdge.size() - 1; ++face) {
     const int firstEdge = faceEdge[face];
     const int lastEdge = faceEdge[face + 1];
@@ -116,11 +121,24 @@ void Manifold::Impl::Face2Tri(const VecDH<int>& faceEdge,
         triNormal.push_back(normal);
         triRef.push_back(halfedgeRef[firstEdge]);
       }
+    } else if (numEdge > 512) {
+      slowTasks.push_back(std::make_pair(
+          numEdge, std::async(
+                       std::launch::async,
+                       [&](int face) {
+                         const glm::vec3 normal = faceNormal_[face];
+                         const glm::mat3x2 projection =
+                             GetAxisAlignedProjection(normal);
+                         const PolygonsIdx polys =
+                             Face2Polygons(face, projection, faceEdge);
+                         std::vector<glm::ivec3> newTris =
+                             TriangulateIdx(polys, precision_);
+                         return std::make_pair(face, std::move(newTris));
+                       },
+                       face)));
     } else {  // General triangulation
       const glm::mat3x2 projection = GetAxisAlignedProjection(normal);
-
       const PolygonsIdx polys = Face2Polygons(face, projection, faceEdge);
-
       std::vector<glm::ivec3> newTris = TriangulateIdx(polys, precision_);
 
       for (auto tri : newTris) {
@@ -130,6 +148,22 @@ void Manifold::Impl::Face2Tri(const VecDH<int>& faceEdge,
       }
     }
   }
+
[...]

PR #522, opened 2023-08-04
parallel some more

pca006132:
Use tbb for some optimizations directly:
1. Parallelize complex face triangulation. The approach with `std::async` will create a thread for every invocation, which is too much overhead, so tbb `thread_arena` is needed.
2. Parallelize `AddNewEdgeVerts`. This requires `concurrent_map` as we will be obtaining elements concurrently.

Combining these two optimizations, we can cut down the running time for `Samples.Sponge4` from ~4s to ~3s. When `MANIFOLD_DEBUG=on`, the running time for `Samples.Sponge4` is reduced from ~7s to ~3.7s. The main bottleneck for now should be simplification which takes about 1s.


pca006132:
it seems to me that the CI somehow stalled, but I have no idea why (I don't think the way I use mutex can cause deadlock?)

pca006132:
Maybe enqueue can cause a deadlock if itself is the sole worker thread... I guess I should use task_group instead.

pca006132:
It seems that face reordering causes `SimplifyTopology` failure (I triggered both segfault and infinite loop). May be related to #518.

pca006132:
thinking about it, it should be possible to parallelize this without reordering. will try it tmr

pca006132:
Admittedly I am getting a bit crazy now, I was seeing so many optimization opportunities and can hardly sleep without implementing them.

Samples.Sponge4 now runs in 2600ms with tbb (it took 4000ms previously) and MANIFOLD_DEBUG=off, the entire test suite completes in 4.5s, and the perfTest with tbb:
```
nTri = 512, time = 0.00160684 sec
nTri = 2048, time = 0.00310806 sec
nTri = 8192, time = 0.00775844 sec
nTri = 32768, time = 0.0200173 sec
nTri = 131072, time = 0.029914 sec
nTri = 524288, time = 0.0974839 sec
nTri = 2097152, time = 0.434753 sec
nTri = 8388608, time = 1.94247 sec
```

pca006132:
the segfault is weird, it doesn't seem to me that the collider update is doing anything that will cause segfault.

pca006132:
@elalish is it required that the collider output order agrees with the query input order? e.g. say the query is [a, b] and a collides with [1, 2], b collides with [3, 4], can we output [3, 4, 1, 2]?

pca006132:
I found the issue: CsgOpNode children is somehow empty in some runs, but I cannot find why it can be empty (I tried adding checks everywhere and can only see it to be empty when we call GetChildren). Perhaps there is some issue with the way I use thread local. Anyway, I am removing that collider optimization which seems to be causing the issue (but I have no idea why...).

pca006132:
@elalish this should be ready now

src/manifold/src/boolean_result.cpp
```diff
@@ -145,6 +165,81 @@ void AddNewEdgeVerts(
...
+              // each EdgePos need to be locked independently,
+              // as pushing elements into a vector may reallocate and
+              // invalidate the pointer
+              mutexes[hash].lock();
+              for (int j = 0; j < glm::abs(inclusion); ++j) {
+                edgePosP.push_back(edgePos);
+                ++edgePos.vert;
+              }
+              mutexes[hash].unlock();
+            }
```
elalish, line 199: For readability, it might be nice to turn these into actual functions, especially if it allows us to show that they only affect a subset of the local variables here.

src/manifold/src/boolean_result.cpp
```diff
@@ -145,6 +165,81 @@ void AddNewEdgeVerts(
...
+              EdgePos edgePos = {vert, 0.0f, inclusion < 0};
+              EdgePos edgePosRev = edgePos;
+              edgePosRev.isStart = !edgePos.isStart;
+              mutexes[hash].lock();
+              for (int j = 0; j < glm::abs(inclusion); ++j) {
+                edgePosLeft.push_back(forward ? edgePosRev : edgePos);
+                ++edgePos.vert;
+                ++edgePosRev.vert;
+              }
+              mutexes[hash].unlock();
```
elalish, line 235: There looks to be some duplicated code between these blocks - again functions would be nice so we don't have to keep changes in sync down the road.

src/manifold/src/edge_op.cpp
```diff
@@ -576,14 +587,14 @@ void Manifold::Impl::RecursiveEdgeSwap(const int edge,
...
+      CollapseEdge(tri0edge[2], edges);
+      edges.resize(0);
     } else {
       visited[edge] = tag;
       visited[pair] = tag;
-      RecursiveEdgeSwap(tri0edge[0], visited, tag);
-      RecursiveEdgeSwap(tri0edge[1], visited, tag);
-      RecursiveEdgeSwap(tri1edge[0], visited, tag);
-      RecursiveEdgeSwap(tri1edge[1], visited, tag);
+      // note: the order should be inverted
```
elalish, line 595: This is a touch ambiguous - do you mean it is currently correct, or should be changed?

src/manifold/src/face_op.cpp
```diff
@@ -35,10 +40,130 @@ void Manifold::Impl::Face2Tri(const VecDH<int>& faceEdge,
...
+      [&](int face) {
+        triCount[face] = faceEdge[face + 1] - faceEdge[face] - 2;
+        ASSERT(triCount[face] >= 1, topologyErr,
+               "face has less than three edges.");
+        if (triCount[face] > 2)
+          group.run([&, face] {
+            const glm::vec3 normal = faceNormal_[face];
+            const glm::mat3x2 projection = GetAxisAlignedProjection(normal);
+            const PolygonsIdx polys = Face2Polygons(face, projection, faceEdge);
+            std::vector<glm::ivec3> newTris = TriangulateIdx(polys, precision_);
```
elalish, line 62: Is this now calling `Triangulate` on even 3 and 4 sided polygons? We were separating those out before to reduce overhead since the vast majority of the polygons are those and don't require the full triangulator.

src/manifold/src/face_op.cpp
```diff
@@ -35,10 +40,130 @@ void Manifold::Impl::Face2Tri(const VecDH<int>& faceEdge,
...
+            std::swap(tri[1], tri[2]);
+            std::swap(ends[1], ends[2]);
+          }
+          ASSERT(ends[0] == tri[1] && ends[1] == tri[2] && ends[2] == tri[0],
+                 topologyErr, "These 3 edges do not form a triangle!");
+
+          triVerts[triCount[face]] = tri;
+          triNormal[triCount[face]] = normal;
+          triRef[triCount[face]] = halfedgeRef[firstEdge];
+        } else if (numEdge == 4) {  // Pair of triangles
```
elalish, line 103: Can we reuse the old code from below for this with functions? It looks pretty duplicate.

elalish (changes requested):
This looks great! Remind me, which of our platforms is TBB not available on? If it's only WASM and that's in progress, then we should put in a TODO to remove the old code when that's ready.

src/manifold/src/edge_op.cpp
pca006132: it is currently correct

src/manifold/src/face_op.cpp
pca006132: No, we are skipping those with less than 5 sides.

src/manifold/src/face_op.cpp
pca006132: yes we can do that

pca006132:
Only not available on WASM. Yes, we can mark the old code as about to be removed.
I actually tried to compile tbb from source for WASM, it seems that it can work, but the performance is not great. I can look into it later.

elalish (approved):
Thanks! If the WASM compiles with TBB and isn't worse than without it, I'd say we can go ahead and remove the alternate code paths. Take a look at the effect on binary size too, just in case.

pca006132:
I will check that. My main concern is that the PR is still under review, and tbb with wasm is not much tested, so I am afraid that it may not be stable.

pca006132:
weird, it seems that this somehow causes windows build to fail (when tbb is enabled)

PR #523, opened 2023-08-05
third_party dependencies

pca006132:
Reincarnation of #402, this uses system gtest if possible and automatically download gtest from github otherwise (e.g. for windows). It will also download tbb and compile from source if the user specified `MANIFOLD_PAR=TBB` and CMake is unable to find the installation.

I also added optional dependency to [Tracy profiler](https://github.com/wolfpld/tracy), which is a handy graphical profiler.


src/manifold/src/manifold.cpp
```diff
@@ -260,7 +260,10 @@ MeshGL Manifold::GetMeshGL(glm::ivec3 normalIdx) const {
       out.triVerts[3 * tri + i] = impl.halfedge_[3 * oldTri + i].startVert;
 
     if (meshID != lastID) {
-      addRun(out, runNormalTransform, tri, meshIDtransform.at(meshID));
+      Impl::Relation rel;
+      auto it = meshIDtransform.find(meshID);
+      if (it != meshIDtransform.end()) rel = it->second;
```
elalish, line 265: Is there a case where this is expected, or should this be an assert?

src/utilities/CMakeLists.txt
```diff
@@ -29,11 +29,32 @@ if(MANIFOLD_PAR STREQUAL "OMP")
...
 elseif(MANIFOLD_PAR STREQUAL "TBB")
-    find_package(PkgConfig REQUIRED)
-    pkg_check_modules(TBB REQUIRED tbb)
-    target_include_directories(${PROJECT_NAME} PUBLIC ${TBB_INCLUDE_DIRS})
+    find_package(PkgConfig)
+    if (PKG_CONFIG_FOUND)
+        pkg_check_modules(TBB tbb)
+    endif()
+    if(NOT TBB_FOUND)
+        message(STATUS "tbb not found, downloading from source")
```
elalish, line 37: 💯

src/utilities/include/public.h
```diff
@@ -19,17 +19,14 @@
...
 #include <memory>
+#include <sstream>
 #include <unordered_map>
 #include <vector>
 
-#ifdef MANIFOLD_DEBUG
-#include <iomanip>
-#include <iostream>
-#include <sstream>
-#endif
```
elalish, line 31: Why this change? Part of my purpose here was to keep the WASM build size down since including things like `std::string` are notoriously large.

test/CMakeLists.txt
```diff
@@ -24,9 +22,45 @@ if(MANIFOLD_CBIND)
...
 endif()
 
+set(gtest_force_shared_crt ON CACHE BOOL "" FORCE)
+# Prevent installation of GTest with your project
+set(INSTALL_GTEST OFF CACHE BOOL "" FORCE)
+set(INSTALL_GMOCK OFF CACHE BOOL "" FORCE)
+
+find_package(GTest 1.12)
+if((NOT GTest_FOUND) OR NOT(GTest_VERSION GREATER_EQUAL 1.12))
+  message(STATUS "GTest not found, downloading from source")
```
elalish, line 32: @kintel - what do you think of this approach? Would this make consuming our library easier for the likes of OpenSCAD?

elalish (commented):
Awesome, this looks great! Can you give any example of what you can do with Tracy? I'm not familiar.

src/manifold/src/manifold.cpp
pca006132: I am not entirely sure, it seems that the our python test cases would somehow trigger this.

src/utilities/include/public.h
pca006132: Oh I forgot to revert this change, I was trying to debug the ubuntu failure when using system provided gtest.

pca006132:
I am not familiar with Tracy either :) We should be able to add more client markups to make the profiler easier to use.
For now, we can see the CPU utilization and roughly what functions are running.

image
image

pca006132:
Sadly it seems that some features are not available on OSX.
image

src/manifold/src/manifold.cpp
elalish: That seems worth investigating. I think it implies we have marked some faces as coming from an ID for which there is no transform. Those things should be 1:1. We definitely don't have much testing of this in the Python bindings, so we could have a bug there.

pca006132: indeed, I think I saw this happening previously as well

pca006132: maybe we can make an issue for this and merge this first?

elalish: Sounds good.

elalish (approved):
Thanks!

test/CMakeLists.txt
kintel: Looks reasonable. It does make the build system depend on having access to github.com, which is a known issue from some countries, as well as for people with limited connectivity, but that's a separate challenge. For OpenSCAD we'll likely try to hit the find_package call, so this should work well for us!

pca006132: Yes it depends on access to github by default, but if for some reason that is not possible and `find_package` still cannot find the package, you can specify the directory by setting `-DFETCHCONTENT_SOURCE_DIR_GOOGLETEST`. I tried using this for the nix package (it blocks internet connection when building) and it works.

kintel: Awesom! Not familiar with FetchContent..

elalish: Nice - would be good to add that trick to the README somewhere.

PR #525, opened 2023-08-10
performance tuning and disabling deprecated backends

### LINKED ISSUE #524: deprecate OMP and CUDA backend
need to update our README and CMakefile, we can probably leave the code cleanup a bit later

pca006132:
There are various things in this PR:

1. Disabled OMP and CUDA backend. The CUDA related code is still there because it will take a large cleanup to remove them. (e.g. those attributes).
2. Added hooks for allocation/deallocation, so Tracy can be used for looking at memory allocation data (number of allocation, size of allocations, and when they are done). This is only limited to `VecDH` for now. I can remove this commit if we don't really need this.
3. Performance optimization by using a single vector for `SparseIndices` instead of two, and use faster binary search adopted from https://github.com/scandum/binary_search/blob/master/binary_search.c#L264. I will post benchmark data later today.
4. Use thread sanitizer to debug data races. I ignored `src/third_party/thrust/thrust/system/tbb/detail/reduce.inl` using [ignorelist](https://clang.llvm.org/docs/SanitizerSpecialCaseList.html#format) as it is known that thrust has data race for parallel reduce. I fixed some of the simpler ones by using atomics, but there are still two cases: `UpdateProperties` and `CoplanarEdge`. Note that by c++ standard, data race between threads without using atomics or memory fence is undefined behavior.

Closes #524


pca006132:
and for some reason windows tbb build is now working, nice!

src/utilities/include/sparse.h
```diff
@@ -110,59 +106,41 @@ class SparseIndices {
...
 #ifdef MANIFOLD_DEBUG
   void Dump() const {
-    const auto& p = Get(0);
-    const auto& q = Get(1);
     std::cout << "SparseIndices = " << std::endl;
+    const int* p = ptr();
     for (int i = 0; i < size(); ++i) {
-      std::cout << i << ", p = " << p[i] << ", q = " << q[i] << std::endl;
+      std::cout << i << ", p = " << p[2 * i + pOffset]
+                << ", q = " << p[2 * i + 1 - pOffset] << std::endl;
```
elalish, line 136: I've seen these accessors used a lot - should we just define `int Get(int i, bool useQ)` to clean it up?

elalish: Or even `int Get<useQ>(int i)`

src/utilities/include/vec_dh.h
```diff
@@ -17,6 +17,13 @@
 #include <cuda.h>
 #endif
 
+#if TRACY_ENABLE
+#include <tracy/Tracy.hpp>
+#else
+#define TracyAllocS(ptr, size, n) (void)0
+#define TracyFreeS(ptr, n) (void)0
+#endif
```
elalish, line 25: The whole point of VecDH (at least originally) was to handle the fact that CUDA needs separate host and device vectors that need to be synced. With CUDA now removed, is there a chance we can remove VecDH entirely? Doesn't have to be in this PR, but it would be great if we can simplify here.

elalish (changes requested):
This looks amazing! I'll update our list of required tests.

elalish:
Nice job on the benchmarks! We should really add our spheres perf benchmark to our standard tests so we remember to run them. I consider that to be more indicative of common real-world usage and scaling than these degenerate-focused tests.

src/utilities/include/sparse.h
pca006132: I just realized as we disabled cuda, we can just make these functions take a reference to the original SparseIndices and use functions on the SparseIndices instead of doing all those pointer arithmetic. Will fix.

src/manifold/src/impl.cpp
```diff
@@ -185,7 +186,8 @@ struct CoplanarEdge {
   const int numProp;
   const float precision;
 
-  __host__ __device__ void operator()(
+  // FIXME: race condition
```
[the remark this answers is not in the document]
pca006132: In theory it can be worse, as this is UB. In practice, it may or may not cause non-determinism depending on whether or not the assigned value is different.

src/utilities/include/vec_dh.h
pca006132: Yes this is what I am going to do, but I plan to do this in another PR as that will be a lot of changes.

pca006132:
> Nice job on the benchmarks! We should really add our spheres perf benchmark to our standard tests so we remember to run them. I consider that to be more indicative of common real-world usage and scaling than these degenerate-focused tests.

Agreed. I actually checked the spheres benchmark as well, but I forgot to include it in the plots.

src/manifold/src/boolean3.cpp
```diff
@@ -473,34 +499,30 @@ std::tuple<VecDH<int>, VecDH<glm::vec3>> Intersect12(
...
+                       inP.halfedge_.cptrD(), inQ.halfedge_.cptrD(),
+                       inP.vertPos_.cptrD(), forward, p1q2.ptr()}));
 
   p1q2.KeepFinite(v12, x12);
 
   return std::make_tuple(x12, v12);
 };
 
-VecDH<int> Winding03(const Manifold::Impl &inP, SparseIndices &p0q2,
+VecDH<int> Winding03(const Manifold::Impl &inP, VecDH<int> &indices,
```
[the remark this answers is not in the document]
pca006132: No, it can only be either p0 or q0, depending on `reverse`.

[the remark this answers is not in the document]
pca006132: I think they are just the list of vertices.

src/utilities/include/vec_dh.h
pca006132: Thinking about it, I'm thinking if we may want to have vulkan acceleration later. Vulkan would require something similar or perhaps a bit more complicated (as we may not have unified memory, depending on the target hardware).

pca006132: also, as our `VecDH` class has specialized parallel fill and uninitialized memory routines, changing to `std::vector<T>` can result in more than 20% performance regression for some test cases.

pca006132:
image

image

I also did some optimization for csg_tree today, I think the result is noticeable. However, a weird thing is that after changing SparseIndices API the tests in general run a bit slower, I suspect that this is due to pointer arithmetic not being optimized enough by GCC (or not inlined enough, as we originally use pointers directly but we now use methods).

src/utilities/include/vec_dh.h
elalish: I have no problem with leaving the class if it's actually doing something useful, but if it can be simplified to one internal storage instead of two, that would be great. We can always redesign it later if needed for Vulkan, but I don't want to plan early for that, especially considering that TBB ended up being just as fast as CUDA. That makes me think Vulkan may be more trouble than it's worth.

src/manifold/src/csg_tree.cpp
```diff
@@ -432,14 +436,59 @@ std::shared_ptr<CsgLeafNode> CsgOpNode::ToLeafNode() const {
...
-                  std::shared_ptr<const Manifold::Impl> b) {
-    // invert the order because we want a min heap
-    return a->NumVert() > b->NumVert();
+  auto cmpFn = [&getImplPtr](const SharedImpl &a, const SharedImpl &b) {
+    return getImplPtr(a)->NumVert() < getImplPtr(b)->NumVert();
   };
 
+  // common cases
+  if (results.size() == 0) return std::make_shared<Manifold::Impl>();
+  if (results.size() == 1)
```
elalish, line 456: 💯

src/utilities/CMakeLists.txt
```diff
@@ -31,8 +30,14 @@ if (TRACY_ENABLE)
...
   )
   FetchContent_MakeAvailable(tracy)
-  target_link_libraries(${PROJECT_NAME} PUBLIC TracyClient)
-  target_compile_options(${PROJECT_NAME} PUBLIC -DTRACY_NO_CODE_TRANSFER=1)
+  target_link_libraries(${PROJECT_NAME} INTERFACE TracyClient)
+  target_compile_options(${PROJECT_NAME} INTERFACE -DTRACY_NO_CODE_TRANSFER=1)
+endif()
+
+if(EMSCRIPTEN)
+    set(TBB_GIT_TAG 8db4efacad3b8aa6eea62281a9e8444e5dc8f16a)
```
elalish, line 38: What's this?

src/utilities/include/hashtable.h
```diff
@@ -26,7 +26,7 @@ typedef Uint64 (*hash_fun_t)(Uint64);
 constexpr Uint64 kOpen = std::numeric_limits<Uint64>::max();
 
 template <typename T>
-__host__ __device__ T AtomicCAS(T& target, T compare, T val) {
+T AtomicCAS(T& target, T compare, T val) {
```
elalish, line 29: Looks like some more `__CUDA_ARCH__` below here we can remove.

src/utilities/include/par.h
```diff
@@ -60,47 +54,13 @@ inline ExecutionPolicy autoPolicy(int size) {
   if (size <= (1 << 12)) {
     return ExecutionPolicy::Seq;
   }
-  if (size <= (1 << 16) || !CudaEnabled()) {
-    return ExecutionPolicy::Par;
-  }
-  return ExecutionPolicy::ParUnseq;
+  return ExecutionPolicy::Par;
```
elalish, line 57: Look like another good `constexpr` candidate above.

src/utilities/include/par.h
```diff
@@ -60,47 +54,13 @@ inline ExecutionPolicy autoPolicy(int size) {
...
 }
 
-#ifdef MANIFOLD_USE_CUDA
-#define THRUST_DYNAMIC_BACKEND_VOID(NAME)                    \
```
elalish, line 72: 💯

src/utilities/include/par.h
```diff
@@ -138,13 +95,12 @@ inline ExecutionPolicy autoPolicy(int size) {
     }                                                        \
   }
 
-#if MANIFOLD_PAR == 'T' && !(__APPLE__)
+#if MANIFOLD_PAR == 'T' && __has_include(<pstl/glue_execution_defs.h>)
```
elalish, line 98: 👍

src/utilities/include/sparse.h
```diff
@@ -132,15 +147,19 @@ class SparseIndices {
...
+                << std::endl;
     }
     std::cout << std::endl;
   }
 #endif
 
  private:
   VecDH<int64_t> data;
+  inline int* ptr() { return reinterpret_cast<int32_t*>(data.ptrD()); }
+  inline const int* ptr() const {
```
elalish, line 160: Nice job making these private! Love the readability improvements here.

src/utilities/include/vec_dh.h
pca006132: Yeah I can simplify that in another PR. It was really surprising when I see that switching to std::vector causes a large slowdown.

src/utilities/include/vec_dh.h
```diff
@@ -255,42 +232,19 @@ class ManagedVec {
     if (bytes >= (1ull << 63)) {
       throw std::bad_alloc();
     }
-#ifdef MANIFOLD_USE_CUDA
-    if (CudaEnabled())
-      cudaMallocManaged(reinterpret_cast<void **>(ptr), bytes);
```
elalish, line 253: Oh, that's right - I forgot you already refactored VecDH for managed memory. Looks like you've taken care of everything I'd hoped for!

elalish (approved):
This is excellent! A few nits, but I think this is ready when you are.

src/utilities/include/vec_dh.h
pca006132: Regarding Vulkan, I think the only case where GPU can be faster is SDF.

src/utilities/CMakeLists.txt
pca006132: Oh, this is the TBB on wasm thing that I accidentally commited, will remove them.

src/utilities/include/vec_dh.h
elalish: Good to know. Maybe someday SDF will be used enough that that'll be worth looking into, but for now I'm thinking TBB is just fine.

pca006132:
and I also went ahead to implement the vec_dh refactoring.

src/manifold/src/impl.cpp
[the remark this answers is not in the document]
pca006132: I will just put this into a new issue

pca006132:
It seems that some thrust functions will trigger GCC warnings, not sure if they are false positives or not. The code should be basically identical to the old code, and so far I haven't got address sanitizer errors. I am not too worried about them for now.

pca006132:
@elalish do you want to further review this or I will just merge it?

elalish (approved):
Looks great, thanks!

PR #528, opened 2023-08-14
Triangulator: shift on merge

### LINKED ISSUE #502: Difference of Cube and Complex SVG Results in Incomplete Render (openscad/openscad#4679)

elalish:
Fixes #502 

Lots of subtlety in this one, as usual with my triangulator. I probably need to rewrite it yet again to simplify the logic, but I haven't found a better mental model for it yet. 

I added a bunch of polygon tests from the bugs that came in and other things that failed along the way and we're now down to just one disabled test. I also verified https://github.com/elalish/manifold/issues/502#issuecomment-1675840390 is now working, so I figure the test coverage I have here now covers that case.


test/polygon_test.cpp
```diff
@@ -3107,25 +3107,3 @@ TEST(Polygon, DISABLED_Woodgrain1) {
   });
   TestPoly(polys, 12, 0.00145);
 }
-
-TEST(Polygon, DISABLED_Woodgrain2) {
```
elalish, line 3111: I removed this one because upon inspection it's not actually geometrically valid - one of the inside contours in CCW instead of CW.

src/polygon/src/polygon.cpp
```diff
@@ -536,8 +542,13 @@ class Monotones {
...
           return End;
-        } else if (westPair != activePairs_.end() &&
-                   std::next(westPair) == eastPair) {
+        } else if (!eastPair->westCertain || !westPair->eastCertain ||
+                   (westPair != activePairs_.end() &&
+                    std::next(westPair) == eastPair)) {
+          if (!eastPair->westCertain)
+            activePairs_.splice(std::next(westPair), activePairs_, eastPair);
+          if (!westPair->eastCertain)
+            activePairs_.splice(eastPair, activePairs_, westPair);
```
elalish, line 551: This is the key change, reordering uncertain pairs based on how they get attached to each other.

src/polygon/src/polygon.cpp
```diff
@@ -381,9 +381,12 @@ class Monotones {
   struct EdgePair {
     VertItr vWest, vEast, vMerge;
     PairItr nextPair;
-    bool westCertain, eastCertain, startCertain;
+    bool westCertain, eastCertain;
 
     int WestOf(VertItr vert, float precision) const {
+      if (vEast->pos.x + precision < vert->pos.x &&
+          vEast->right->pos.x + precision < vert->pos.x)
```
elalish, line 388: This makes the check robust to short segments.

src/polygon/src/polygon.cpp
```diff
@@ -747,8 +758,13 @@ class Monotones {
     }
     if (isHole) return true;
 
-    activePairs_.splice(activePairs_.end(), activePairs_, inputPair);
-    inputPair->eastCertain = true;
+    if (activePairs_.size() > 1) {
+      activePairs_.splice(activePairs_.end(), activePairs_, inputPair);
+      inputPair->eastCertain =
+          std::prev(inputPair)->WestOf(vert, precision_) > 0;
```
elalish, line 764: This makes the uncertainty more accurate. `Polygon.Small` still fails for the duplicated case due to an edge being incorrectly marked as certain, but I haven't figured out how to deal with that yet, and it's a pretty nasty test case.

pca006132:
Nice! I compiled openscad to test this and indeed this fixes the simple repro and the thin gap thing, but woodgrain is still failing. It seems that the triangulator input in that case is somehow geometrically invalid, as indicated by the minimized test case woodgrain2 (the minimization process should preserve geometry validity).

pca006132 (approved):

elalish:
Okay, I added your updated test and fixed it. Makes sense - it was basically the symmetric case to one of the others I fixed, so I probably should have guessed.

pca006132:
Everything now passes!

PR #531, opened 2023-08-15
refactor

pca006132:
This does 5 things:

1. Added bound check to vectors by default. I verified that the overhead is not large so we should probably enable this even when debug is off. Downstream projects can catch the exception and do whatever they like instead of segfault. (we don't have much shared states so it should be fine)
2. Refactored VecDH to remove nearly all pointer usages, added VecView for the cases that requires pointer offsetting. I also renamed VecDH to Vec as we are no longer having differences between host and device.
3. Moved sdf into a source file, and only expose the minimal public API to the user. We are no longer using CUDA so this should be fine
4. Refactored `cross_section.h` so we no longer expose Clipper2 dependency to our users.
5. Added CMake install target. Users can now use cmake install and get the following result:
```
result
├── include
│   └── manifold
│       ├── cross_section.h
│       ├── manifoldc.h
│       ├── manifold.h
│       ├── public.h
│       ├── sdf.h
│       └── types.h
└── lib
    ├── libcollider.so
    ├── libcross_section.so
    ├── libmanifoldc.so
    ├── libmanifold.so
    ├── libsdf.so
    └── manifold3d.cpython-39-x86_64-linux-gnu.so
```

Originally I only wanted to remove pointer usages, but as I was modifying things I suddenly want to do some refactor as the changes is already pretty large.

This change breaks the sdf API, as the original way of exposing sdf in header will make all our internal header files public.

Users of our library can now use compiled binaries and our header files. The only third party dependency they need is glm.


elalish:
This looks amazing! I very much like the changes you've listed - I'll start reviewing now. I think we can probably leave it as one, but I'll let you know. Sounds like our next release will be v3.0 instead of v2.2, but I think that's fine.

src/sdf/src/sdf.cpp
```diff
@@ -0,0 +1,357 @@
+// Copyright 2022 The Manifold Authors.
```
elalish, line 1: 2023

src/sdf/src/sdf.cpp
```diff
@@ -0,0 +1,357 @@
...
+ * @param sdf The signed-distance functor, containing this function signature:
+ * `float operator()(glm::vec3 point)`, which returns the
+ * signed distance of a given point in R^3. Positive values are inside,
+ * negative outside.
+ * @param bounds An axis-aligned box that defines the extent of the grid.
+ * @param edgeLength Approximate maximum edge length of the triangles in the
+ * final result. This affects grid spacing, and hence has a strong effect on
+ * performance.
+ * @param level You can inset your Mesh by using a positive value, or outset
+ * it with a negative value.
```
elalish, line 294: Looks like we forgot to add the `canParallel` param. Can probably just move the comment up here as a doc.

src/utilities/include/vec_dh.h
```diff
@@ -40,244 +125,195 @@ namespace manifold {
...
-  VecDc(const VecDH<T> &vec) : ptr_(vec.ptrD()), size_(vec.size()) {}
-
-  __host__ __device__ const T &operator[](int i) const { return ptr_[i]; }
-  __host__ __device__ int size() const { return size_; }
-
- private:
-  T const *const ptr_;
-  const int size_;
-};
+  VecDHView<const T> get_cview(int offset = 0, int length = -1) const {
```
elalish: Since we use these a lot, let's go for a short name. How about just `view/cview`?

src/utilities/include/vec.h
```diff
@@ -0,0 +1,318 @@
...
+      if (this->ptr_ != nullptr) {
+        TracyFreeS(ptr_, 3);
+        free(this->ptr_);
+      }
+    }
+    this->ptr_ = newBuffer;
+    capacity_ = this->size_;
+  }
+
+  VecView<T> get_view(int offset = 0, int length = -1) {
```
elalish, line 283: How much is get_view used with non-default arguments? Curious if that will all go away with the next refactor? Likewise, what is different about `Vec` and `std::vector` at this point? I recall you said you found a perf difference, but do you know why?

src/utilities/include/vec.h
```diff
@@ -0,0 +1,318 @@
...
+    ptr_ = other.ptr_;
+    size_ = other.size_;
+    return *this;
+  }
+
+  // allows conversion to a const VecView
+  operator VecView<const T>() const { return {ptr_, size_}; }
+
+  inline const T &operator[](int i) const {
+    if (i < 0 || i >= size_) throw std::out_of_range("Vec out of range");
```
elalish, line 62: Does this really not make a perf difference? I'm surprised, but that would be great.

src/cross_section/include/cross_section.h
```diff
@@ -169,10 +166,10 @@ class CrossSection {
   ///@}
 
  private:
-  mutable std::shared_ptr<const C2::PathsD> paths_;
+  mutable std::shared_ptr<const PathImpl> paths_;
```
elalish, line 169: FYI @geoffder

bindings/c/box.cpp
```diff
@@ -1,7 +1,6 @@
-#include <conv.h>
-#include <manifold.h>
-#include <public.h>
-
+#include "conv.h"
+#include "manifold.h"
+#include "public.h"
```
elalish, line 3: FYI @geoffder

src/manifold/CMakeLists.txt
```diff
@@ -26,3 +26,15 @@ target_link_libraries(${PROJECT_NAME}
...
+
+install(
+  TARGETS ${PROJECT_NAME}
+  LIBRARY DESTINATION ${CMAKE_INSTALL_FULL_LIBDIR}
+  COMPONENT runtime
+)
+install(
+  DIRECTORY include/
+  DESTINATION ${CMAKE_INSTALL_FULL_INCLUDEDIR}/manifold
+  COMPONENT devel
```
elalish, line 38: I love it! Are we testing the install somehow in our CI?

elalish (approved):
This is fantastic! Just some minor comments.

elalish:
@googlebot I signed it!

src/utilities/include/vec_dh.h
pca006132: sounds good. actually we can overload implicit conversion for the majority of cases, I wonder if you like that or not

src/utilities/include/vec.h
pca006132: it is quite rare, in boolean3 and boolean_result where we need to separate a pq vector into two parts.

differences from std::vector
1. we use parallel fill for large vectors
2. initialization can have uninitialized memory (this is pretty unsafe in general, but we can do this because we will immediately write to them in parallel)
3. we have bound checks (std::vector only perform bound checks if you use at instead of [])

src/utilities/include/vec.h
pca006132: yeah, there is a very tiny performance difference (50ms/3000ms)

this kind of checks are fast because the compiler can often optimize them away when inlining, e.g. in a loop and the compiler can prove that the bound must hold
even when the compiler cannot optimize away this, it is generally in the cold path and does not contribute much to the latency.

in fact, rust does bound check for safe code, and rust can write fast programs without using unsafe.

src/manifold/CMakeLists.txt
pca006132: the nix build uses install, but we are not checking the file contents

src/sdf/src/sdf.cpp
pca006132: I think we are missing copyright header in the cbind files, I will add them later as well

src/utilities/include/vec_dh.h
elalish: Overloads are fine, I just like the const to be clear for readability, but that is already specified in the functor, which is clearer anyway.

src/utilities/include/vec.h
elalish: Huh, I thought `std::vector` did uninitialized memory too. 1 and 3 sound like reason enough though.

elalish:
Nice bug find - this all looks good to me; feel free to merge.

>This change breaks the sdf API, as the original way of exposing sdf in header will make all our internal header files public.

I don't quite understand what you mean - looking at our samples and tests it doesn't appear there was any breaking change to the SDF API.

pca006132:
The policy parameter is replaced with a boolean

elalish:
Ah, thanks, I see it now. That's pretty minor, I'm pretty comfortable glossing over that as far as a breaking change. Probably still good with 2.2.

PR #532, opened 2023-08-16
fix memory leak for vec

pca006132:
it turns out I placed this in the wrong position, causing memory leak after resizing to 0.

### DIFF
--- a/src/utilities/include/vec.h
+++ b/src/utilities/include/vec.h
@@ -273,10 +273,10 @@ class Vec : public VecView<T> {
       TracyAllocS(newBuffer, size_ * sizeof(T), 3);
       uninitialized_copy(autoPolicy(this->size_), this->ptr_,
                          this->ptr_ + this->size_, newBuffer);
-      if (this->ptr_ != nullptr) {
-        TracyFreeS(ptr_, 3);
-        free(this->ptr_);
-      }
+    }
+    if (this->ptr_ != nullptr) {
+      TracyFreeS(ptr_, 3);
+      free(this->ptr_);
     }
     this->ptr_ = newBuffer;
     capacity_ = this->size_;


pca006132:
I tried running all our tests with address sanitizer in single threaded mode and there is no leak now. However, when TBB is enabled, it reports many leaks in tbb allocator and scheduler, potentially related to thrust as well.
Running thread sanitizer suggests that thrust has severe data race, but ignoring thrust reduce and two know race conditions in our code reports no additional data race.

pca006132:
also removed the last few remaining things from CUDA that I forgot to remove previously

elalish (approved):
Good find! I know in GPUs the order of execution is non-deterministic, hence basically everything is a data race. I wonder if that design mentality explains thrust's behavior? In general it should still be reliable modulo rounding error, but I'm not sure if there is some other difference between CPU and GPU architectures that would cause something worse.

pca006132:
No, non-deterministic execution does not imply data race (otherwise the concept it pretty useless).
Data race are usually write-write (last write wins) and read-write (the read value depends on when the write is being performed), with the requirement that there is no ordering between the two events (not in the same thread, no locks barriers, and no atomic updates).

This kind of race conditions are undefined because the compiler wants to happily remove memory access that it thinks the user will not observe. Making every memory operation atomic by default, even just with relaxed order, will break this assumption, because the compiler does not know if there is a thread elsewhere reading this value, making the compiler unable to optimize anything. In our case the race condition will probably not cause anything wild, but it is still a good idea to fix it.

Changing the updates to atomic update makes the behavior well defined, but the behavior can still be non-deterministic. Usually atomics are used to avoid getting into inconsistent states (we use the ordering imposed by the memory model to rule out some execution results), but in this case we just need to make it well defined.

Btw, the reason I did not fix this is because I want to rule out potential non-deterministic behavior in this part as well. The outcome should be the same regardless of the threads ordering.

PR #537, opened 2023-08-25
nanobind

pca006132:
Use nanobind instead of pybind11, which should support python ABI3 (with Python 3.12).

API changes: Python list is in general not accepted for int/float array arguments, users will need to use numpy (or tensorflow etc.) arrays, except for Vec3.

@Loosetooth what do you think about this? will this break your existing code?


Loosetooth:
My python is a bit rusty, haven't used it in years, so I can't immediately comment on the changes.
Also, I haven't used the python bindings at all, only the javascript / wasm bindings. Perhaps you wanted to tag someone else?

pca006132:
Ahh sorry, I should tag @axel-angel

bindings/python/manifold3d.cpp
```diff
@@ -77,25 +113,26 @@ PYBIND11_MODULE(manifold3d, m) {
...
-      .def(py::self ^ py::self, "Boolean intersection.")
+  nb::class_<Manifold>(m, "Manifold")
+      .def(nb::init<>())
+      .def(
+          "__init__",
+          [](Manifold *self, std::vector<Manifold> &manifolds) {
+            if (manifolds.size() >= 1) {
+              // for some reason using Manifold() as the initial object
+              // will cause failure for python specifically
+              // unable to reproduce with c++ directly
```
elalish, line 124: I wonder if this was a pybind11 bug? Does it still repro?

bindings/python/manifold3d.cpp
```diff
@@ -116,19 +153,18 @@ PYBIND11_MODULE(manifold3d, m) {
           "Compute the convex hull enveloping a set of 3d points.")
       .def(
           "transform",
-          [](Manifold &self, py::array_t<float> &mat) {
-            auto mat_view = mat.unchecked<2>();
```
elalish, line 120: Nice that we can remove this - I take it nanobind doesn't do any checking by default?

bindings/python/manifold3d.cpp
```diff
@@ -140,74 +176,45 @@ PYBIND11_MODULE(manifold3d, m) {
...
           "\n\n"
           ":param x: X axis translation. (default 0.0).\n"
           ":param y: Y axis translation. (default 0.0).\n"
           ":param z: Z axis translation. (default 0.0).")
-      .def(
-          "translate",
-          [](Manifold &self, py::array_t<float> &t) {
-            auto t_view = t.unchecked<1>();
-            if (t.ndim() != 1 || t.shape(0) != 3)
-              throw std::runtime_error("Invalid vector shape");
```
elalish, line 155: What happens now if the wrong shape array is input?

bindings/python/manifold3d.cpp
```diff
@@ -257,24 +264,35 @@ PYBIND11_MODULE(manifold3d, m) {
...
+              } else {
+                throw std::runtime_error(
+                    "Callback in set_properties should return an array");
+              }
             });
           },
-          py::arg("new_num_prop"), py::arg("f"),
-          "Create a new copy of this manifold with updated vertex properties "
+          nb::arg("new_num_prop"), nb::arg("f"),
+          "Create a new conb of this manifold with updated vertex properties "
```
elalish, line 295: find/replace?

elalish (approved):
This looks great, especially since nanobind is reducing how much binding code we need to maintain. Considering we haven't made an official Python release yet, I'm not particularly concerned about potential breaking changes here, as long as it's easy for people to use.

bindings/python/manifold3d.cpp
pca006132: oh I did not check this

bindings/python/manifold3d.cpp
pca006132: it does check, the problem is that they only allows the use of ndarrays, e.g. numpy/tensorflow arrays, instead of doing explicit cast of python lists.

bindings/python/manifold3d.cpp
pca006132: it will say that there is no appropriate function to call and print the type signature. (Vec3 requires a sequence with length 3, which is checked in the type caster I wrote above)

bindings/python/manifold3d.cpp
elalish: That seems pretty reasonable to me.

bindings/python/manifold3d.cpp
pca006132: oh forgot to fix this, yes

elalish:
This looks like a good direction; should we go ahead and merge this?

pca006132:
yeah I think we can use this instead of #521, but let me limit the python build wheel CI to tag/release first.

PR #544, opened 2023-08-30
fix ndarray creation (#543)

### LINKED ISSUE #543: Python Interface Vertex and Face Types
It seems that the vertices and faces of Mesh objects in the Python interface are PyCapsule objects, not numpy arrays. Is there a way to easily convert these?

As an example, running the following script:
```
in_mesh = trimesh.creation.box(extents=[2,0.5,0.2])
manifold = Manifold.from_mesh(Mesh(vert_properties=np.asarray(a.vertices), tri_verts=np.asarray(a.faces)))
out_mesh = manifold.to_mesh()

print(type(out_mesh.vert_properties))
print(type(out_mesh.tri_verts))
```

Outputs the following:
```
<class 'PyCapsule'>
<class 'PyCapsule'>
```

pca006132:
Fixes #543

It seems that ndarray defaults to exporting a PyCapsule. Although we can convert it to numpy array in the python code, it seems easier to convert them directly in our c++ code.

This also adds mesh export to our CI.

### DIFF
--- a/.github/workflows/manifold.yml
+++ b/.github/workflows/manifold.yml
@@ -41,7 +41,7 @@ jobs:
         cd build/test
         ./manifold_test
         cd ../../
-        python3 bindings/python/examples/run_all.py
+        python3 bindings/python/examples/run_all.py -e
     - name: Coverage Report
       # only do code coverage for default sequential backend, it seems that TBB
       # backend will cause failure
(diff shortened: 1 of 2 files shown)


elalish (approved):
Excellent, thanks!

PR #548, opened 2023-09-01
misc

pca006132:
1. Use `std::unique` instead of `thrust::unique`, as it is known that `thrust::unique` can cause out of bound access.
2. Use standard sort instead of sort_by_key. This is an effort to reduce the thrust dependencies.
3. Use stl algorithms instead of thrust algorithms when we use single threaded backend. This is for wasm binding.

### DIFF
--- a/src/utilities/include/par.h
+++ b/src/utilities/include/par.h
@@ -23,18 +23,15 @@
 #include <thrust/sort.h>
 #include <thrust/system/cpp/execution_policy.h>
 #include <thrust/uninitialized_copy.h>
-#include <thrust/unique.h>
-#if MANIFOLD_PAR == 'O'
-#include <thrust/system/omp/execution_policy.h>
-#define MANIFOLD_PAR_NS omp
-#elif MANIFOLD_PAR == 'T'
+
+#include <algorithm>
+#include <numeric>
+#if MANIFOLD_PAR == 'T'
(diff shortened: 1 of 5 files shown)


src/utilities/include/par.h
```diff
@@ -87,9 +84,9 @@ inline constexpr ExecutionPolicy autoPolicy(int size) {
     return thrust::NAME(thrust::cpp::par, args...);                 \
   }
 
-#if MANIFOLD_PAR == 'T' && TBB_INTERFACE_VERSION >= 10000 && \
-    __has_include(<pstl/glue_execution_defs.h>)
-// sometimes stl variant is faster
+#if MANIFOLD_PAR != 'T' || \
+    (TBB_INTERFACE_VERSION >= 10000 && __has_include(<pstl/glue_execution_defs.h>))
+#if MANIFOLD_PAR == 'T'
```
elalish, line 89: What does this mean with two `#if` statements in a row?

elalish (approved):
Thanks! I assume performance is roughly the same?

src/utilities/include/par.h
pca006132: I was changing this so that we will use stl as well when we only use single thread

pca006132:
yeah there is no measurable difference to me

PR #550, opened 2023-09-05
added deterministic option

### LINKED ISSUE #545: Volume calculation is not deterministic
In particular, this leads to intermittent test failure.
This should be related to data race issue.

https://github.com/elalish/manifold/actions/runs/6020517957/job/16331969929

pca006132:
This disables some racy optimizations when user wants deterministic output. We should now consider non-deterministic output when deterministic option is enabled a bug.

The area and volume calculation now uses Kahan summation, which is numerically stable.

Also, it turns out parallel compose will somehow cause non-determinism. Considering we already do parallel for inside compose, I just disabled doing multiple compose in parallel and rewrote it in a simpler way.

Fixes #545.

### DIFF
--- a/src/manifold/src/boolean_result.cpp
+++ b/src/manifold/src/boolean_result.cpp
@@ -207,8 +207,8 @@ void AddNewEdgeVerts(
   // parallelize operations, requires concurrent_map so we can only enable this
   // with tbb
   if (!ManifoldParams().deterministic && p1q2.size() > kParallelThreshold) {
-    // ideally we should have 1 mutex per key, but kParallelThreshold is enough to avoid
-    // contention for most of the cases
+    // ideally we should have 1 mutex per key, but kParallelThreshold is enough
+    // to avoid contention for most of the cases
     std::array<std::mutex, kParallelThreshold> mutexes;
     static tbb::affinity_partitioner ap;
     auto processFun = std::bind(
--- a/src/manifold/src/boolean_result.cpp
+++ b/src/manifold/src/boolean_result.cpp
@@ -204,7 +204,7 @@ void AddNewEdgeVerts(
 #if MANIFOLD_PAR == 'T' && __has_include(<tbb/tbb.h>)
   // parallelize operations, requires concurrent_map so we can only enable this
   // with tbb
-  if (p1q2.size() > 128) {
+  if (!ManifoldParams().deterministic && p1q2.size() > 128) {
     // ideally we should have 1 mutex per key, but 128 is enough to avoid
     // contention for most of the cases
     std::array<std::mutex, 128> mutexes;
@@ -240,8 +240,8 @@ std::vector<Halfedge> PairUp(std::vector<EdgePos> &edgePos) {
                                [](EdgePos x) { return x.isStart; });
   ASSERT(middle - edgePos.begin() == nEdges, topologyErr, "Non-manifold edge!");
   auto cmp = [](EdgePos a, EdgePos b) { return a.edgePos < b.edgePos; };
-  std::sort(edgePos.begin(), middle, cmp);
-  std::sort(middle, edgePos.end(), cmp);
+  std::stable_sort(edgePos.begin(), middle, cmp);
+  std::stable_sort(middle, edgePos.end(), cmp);
   std::vector<Halfedge> edges;
   for (int i = 0; i < nEdges; ++i)
     edges.push_back({edgePos[i].vert, edgePos[i + nEdges].vert, -1, -1});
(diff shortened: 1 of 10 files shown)


src/manifold/src/boolean_result.cpp
```diff
@@ -204,7 +204,7 @@ void AddNewEdgeVerts(
 #if MANIFOLD_PAR == 'T' && __has_include(<tbb/tbb.h>)
   // parallelize operations, requires concurrent_map so we can only enable this
   // with tbb
-  if (p1q2.size() > 128) {
+  if (!ManifoldParams().deterministic && p1q2.size() > 128) {
```
elalish, line 207: Let's make 128 a named `constexpr`.

src/utilities/include/public.h
```diff
@@ -292,17 +292,17 @@ struct Box {
   /**
    * Does this box overlap the one given (including equality)?
    */
-  bool DoesOverlap(const Box& box) const {
-    return min.x <= box.max.x && min.y <= box.max.y && min.z <= box.max.z &&
-           max.x >= box.min.x && max.y >= box.min.y && max.z >= box.min.z;
+  inline bool DoesOverlap(const Box& box) const {
+    return (min.x <= box.max.x) & (min.y <= box.max.y) & (min.z <= box.max.z) &
+           (max.x >= box.min.x) & (max.y >= box.min.y) & (max.z >= box.min.z);
```
elalish, line 297: Do you not want to short-circuit the comparisons?

test/boolean_test.cpp
```diff
@@ -898,6 +898,7 @@ TEST(Boolean, Sweep) {
   }
 
   Manifold shape = Manifold::BatchBoolean(result, OpType::Add);
+  EXPECT_EQ(shape.NumTri(), 4400);
```
elalish, line 901: This is checking 3, rather than 2. I'm willing to try it, but I'm guessing this will need updates with future changes. If I'm wrong, so much the better!

src/polygon/src/polygon.cpp
```diff
@@ -838,9 +838,10 @@ class Monotones {
     }
 #if MANIFOLD_PAR == 'T' && TBB_INTERFACE_VERSION >= 10000 && \
     __has_include(<pstl/glue_execution_defs.h>)
-    std::sort(std::execution::par_unseq, starts.begin(), starts.end(), cmp);
+    std::stable_sort(std::execution::par_unseq, starts.begin(), starts.end(),
+                     cmp);
 #else
-    std::sort(starts.begin(), starts.end(), cmp);
+    std::stable_sort(starts.begin(), starts.end(), cmp);
```
elalish, line 844: Do these stable sorts make much difference?

src/utilities/include/public.h
pca006132: ah I was modifying this for something else, forgot to remove it
I was trying to see if the short-circuiting is causing branch misprediction and stuff, but it seems that there is no significant performance implication

src/polygon/src/polygon.cpp
pca006132: probably not, this is just in case. the performance of stable sort is not a lot different than sort from my testing

(FYI: rust uses stable sort by default, and the c++ sort is named unstable sort in rust and we will only use it when profile shows it is beneficial)

test/boolean_test.cpp
pca006132: Indeed, this is quite sensitive to the behavior of the decimator and many other stuff.
I guess I should probably put the deterministic tests elsewhere, and just run the samples 10 times or something to test if they have different results

src/manifold/src/properties.cpp
```diff
@@ -313,11 +313,26 @@ int Manifold::Impl::NumDegenerateTris() const {
 
 Properties Manifold::Impl::GetProperties() const {
   if (IsEmpty()) return {0, 0};
-  auto areaVolume = transform_reduce<thrust::pair<float, float>>(
-      autoPolicy(NumTri()), countAt(0), countAt(NumTri()),
-      FaceAreaVolume({halfedge_, vertPos_, precision_}),
-      thrust::make_pair(0.0f, 0.0f), SumPair());
-  return {areaVolume.first, areaVolume.second};
+  // Kahan summation
```
elalish, line 316: This is cool - I hadn't heard of this before.

elalish (approved):
Thanks!

src/manifold/src/properties.cpp
pca006132: Same here, I just know this after searching for numerically stable summation algorithm.

PR #558, opened 2023-09-12
[FEAT] Add Status() to the Python Bindings

zalo:
This will help consumers of the Python bindings diagnose why their input meshes are not translating into manifolds properly.

### DIFF
--- a/bindings/python/manifold3d.cpp
+++ b/bindings/python/manifold3d.cpp
@@ -689,30 +689,20 @@ NB_MODULE(manifold3d, m) {
       .def_ro("face_id", &MeshGL::faceID);
 
   nb::enum_<Manifold::Error>(m, "Error")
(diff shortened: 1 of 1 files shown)


elalish (approved):
Good call, thanks!

pca006132:
Note: I recommend turning on MANIFOLD_DEBUG as well, which will have slight performance impact, but it can catch some geometry/topology errors.
And I think we will throw an exception when MANIFOLD_DEBUG is enabled and did not enable suppressErrors.

elalish:
Hmm, I'd rather not turn on `MANIFOLD_DEBUG` in production builds. If once I've got the triangulator refactor done we're still having trouble with errors, let's discuss a design for exceptions generally. Some of those debug checks are not entirely trustworthy...

Issue #555, opened 2023-09-10
Emscripten build


stephomi:
Can we add an option to skip the wasm bindings when targeting EMSCRIPTEN? Or an option to skip the nixos thing.

For my wasm integration I used to do something like:

```cmake
add_compile_definitions(_LIBCUDACXX_HAS_THREAD_API_EXTERNAL)
set(EMSCRIPTEN 0)
add_subdirectory(lib/manifold/)
set(EMSCRIPTEN 1)
```

But now the build fails because of nixos thing:
https://github.com/elalish/manifold/blob/5e103931f1191eba8eb25d89b1039aa58a03e0db/CMakeLists.txt#L63-L65
If EMSCRIPTEN is 1, then the wasm bindings fail (don't really want to investigate why, as I don't use it)

Other tiny question: Is `D_LIBCUDACXX_HAS_THREAD_API_CUDA` doing something for [wasm](https://github.com/elalish/manifold/blob/5e103931f1191eba8eb25d89b1039aa58a03e0db/CMakeLists.txt#L46C74-L46C106)?

pca006132:
Sure, we can add a `MANIFOLD_JSBIND` option. Also it is a bit interesting that the implicit include directory will cause compilation error, I was not expecting it. Maybe you can make a PR to make the wasm bindings optional? (we want to enable MANIFOLD_JSBIND for emscripten by default, but can be overridden by the user.)

IIRC `D_LIBCUDACXX_HAS_THREAD_API_CUDA` is required for solving some compilation error with thrust, and is needed for wasm.

Also, you can try to use the latest tbb as it supports emscripten now. See https://github.com/elalish/manifold/discussions/520#discussioncomment-6693526 for the patch. We did not enable it right now because the emscripten worker file does not support modules yet.

elalish:
Why would you want to target Emscripten without the WASM bindings? What would your output be?

pca006132:
> (By the way, there is now a collider.a, etc for every folder except of just a single manifold.a, not sure if done on purpose, bu I now need to link 5/6 lib in Xcode instead of a single one).

No this is not on purpose. Can you show the output of `tree`? FYI this is my output:
```
├── src
│   ├── CMakeFiles
│   ├── cmake_install.cmake
│   ├── collider
│   │   ├── CMakeFiles
│   │   │   └── collider.dir
│   │   │       └── src
│   │   │           └── collider.cpp.o
│   │   ├── cmake_install.cmake
[...]

PR #559, opened 2023-09-13
dependency update

pca006132:
Address #554 and #555.

- Removed submodule for thrust and glm, use system library if possible and resort to git clone if there is no such package installed.
- Allows users to not build js buildings when emscripten is used.
- Updated documentation for our dependencies, and guide for building in offline mode without some of the packages.

### DIFF
--- a/.github/workflows/manifold.yml
+++ b/.github/workflows/manifold.yml
@@ -219,7 +219,7 @@ jobs:
     - uses: actions/checkout@v3
       with:
         submodules: recursive
-    - uses: cachix/install-nix-action@v15
+    - uses: cachix/install-nix-action@v22
     - run: nix build -L '.?submodules=1#manifold-${{matrix.variant}}'
 
   build_nix_python:
@@ -230,5 +230,5 @@ jobs:
     - uses: actions/checkout@v3
       with:
         submodules: recursive
-    - uses: cachix/install-nix-action@v15
+    - uses: cachix/install-nix-action@v22
     - run: nix build -L '.?submodules=1#manifold3d'
--- a/.github/workflows/manifold.yml
+++ b/.github/workflows/manifold.yml
@@ -148,7 +148,7 @@ jobs:
     - name: Build ${{matrix.backend}}
       shell: powershell
       run: |
-        cmake . -DCMAKE_BUILD_TYPE=Release -B build -DBUILD_SHARED_LIBS=ON -DMANIFOLD_DEBUG=ON -DPYBIND11_FINDPYTHON=OFF -DMANIFOLD_PAR=${{matrix.parallel_backend}} -A x64
+        cmake . -DCMAKE_BUILD_TYPE=Release -B build -DMANIFOLD_DEBUG=ON -DPYBIND11_FINDPYTHON=OFF -DMANIFOLD_PAR=${{matrix.parallel_backend}} -A x64
         cd build
         cmake --build . --target ALL_BUILD --config Release
     - name: Test ${{matrix.parallel_backend}}
--- a/.github/workflows/manifold.yml
+++ b/.github/workflows/manifold.yml
@@ -22,7 +22,7 @@ jobs:
     - name: Install dependencies
       run: |
         sudo apt-get -y update
-        DEBIAN_FRONTEND=noninteractive sudo apt install -y libgtest-dev libassimp-dev git libtbb-dev pkg-config libpython3-dev python3 python3-distutils python3-pip lcov
+        DEBIAN_FRONTEND=noninteractive sudo apt install -y libglm-dev libgtest-dev libassimp-dev git libtbb-dev pkg-config libpython3-dev python3 python3-distutils python3-pip lcov
         pip install trimesh
     - uses: actions/checkout@v3
       with:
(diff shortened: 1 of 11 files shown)


pca006132:
hmmm, I have no idea why the nix builds fail.
will try to fix them later. I am unable to reproduce those problems on my machines even though flakes are supposed to be deterministic.

elalish:
Hmm, this sounds ominous:

```
CMake Warning at build/_deps/tbb-src/CMakeLists.txt:118 (message):
  You are building oneTBB as a static library.  This is highly discouraged
  and such configuration is not supported.  Consider building a dynamic
  library to avoid unforeseen issues.
```

pca006132:
indeed, but that can easily be fixed anyway

pca006132:
ok, the thrust cmake problem with nix is due to cmake non-deterministically applying the config files... I have no idea why this happens but I have worked around it by overwriting the output into a known correct result.
this only applies to our CI, and will not affect other packages.

.github/workflows/manifold.yml
```diff
@@ -219,7 +219,7 @@ jobs:
     - uses: actions/checkout@v3
       with:
         submodules: recursive
-    - uses: cachix/install-nix-action@v15
+    - uses: cachix/install-nix-action@v22
```
pca006132, line 222: I later found that the nix non-deterministic issue is not related to this, but updating should not cause any problem.

test/CMakeLists.txt
```diff
@@ -27,27 +27,22 @@ set(gtest_force_shared_crt ON CACHE BOOL "" FORCE)
...
+  GIT_PROGRESS   TRUE
+  FIND_PACKAGE_ARGS NAMES GTest gtest
+)
+FetchContent_MakeAvailable(googletest)
 
-if(NOT TARGET GTest::GTest)
-    add_library(GTest::GTest ALIAS gtest)
-    add_library(GTest::Main ALIAS gtest_main)
+if(NOT TARGET GTest::gtest_main)
+    add_library(GTest::gtest_main ALIAS gtest_main)
```
pca006132, line 41: `GTest::Main` may not exist, but `GTest::gtest_main` should exist for every version.

src/utilities/CMakeLists.txt
```diff
@@ -44,13 +57,15 @@ if(MANIFOLD_PAR STREQUAL "TBB")
...
+        set(BUILD_SHARED_LIBS ON)
         FetchContent_Declare(TBB
-          GIT_REPOSITORY https://github.com/oneapi-src/oneTBB.git
-          GIT_TAG        v2021.10.0
-          GIT_SHALLOW    TRUE
-          GIT_PROGRESS   TRUE
+            GIT_REPOSITORY https://github.com/oneapi-src/oneTBB.git
+            GIT_TAG        v2021.10.0
+            GIT_SHALLOW    TRUE
+            GIT_PROGRESS   TRUE
```
pca006132, line 65: originally I wanted to use `find_package` for TBB as well (in addition to pkg-config), but it seems that the `FindTBB.cmake` provided in some versions are broken, and calling `find_package(tbb)` will cause cmake error, so I just keep the old behavior.

pca006132:
It seems that `BUILD_SHARED_LIB=ON` behaves weirdly on Windows, so tbb will still be built statically.
Actually I found this to be quite weird, as tbb should be built as dynamic library by default and we did not set `BUILD_SHARED_LIB` in our library.

src/utilities/CMakeLists.txt
```diff
@@ -67,8 +80,19 @@ else()
...
+target_link_libraries(${PROJECT_NAME} INTERFACE glm::glm)
+if(DEFINED Thrust_SOURCE_DIR)
+target_include_directories(${PROJECT_NAME} INTERFACE
+    ${Thrust_SOURCE_DIR}
+    ${Thrust_SOURCE_DIR}/dependencies/libcudacxx/include
+)
+else()
+target_include_directories(${PROJECT_NAME} INTERFACE
+    ${THRUST_ROOT}
+    ${THRUST_ROOT}/dependencies/libcudacxx/include
```
elalish, line 93: I'm a little surprised CMake doesn't take care of this automatically...

elalish (approved):
Nice work, and thanks for the docs updates too!

src/utilities/CMakeLists.txt
pca006132: CMake should be able to handle it, but I found that for some reason FetchContent does not execute the thrust cmake file correctly. This is what we did previously as well when using submodules.

pca006132:
I will merge this after I ensure that the python wheel build is working.

pca006132:
Hmm... Nice job windows

```
CUSTOMBUILD : error : unable to create file libcxx/test/std/thread/thread.mutex/thread.mutex.requirements/thread.sharedtimedmutex.requirements/thread.sharedtimedmutex.class/try_lock_until_deadlock_bug.pass.cpp: Filename too long [C:\Users\runneradmin\AppData\Local\Temp\tmp6pcu43tk\build\_deps\thrust-subbuild\thrust-populate.vcxproj]
```
Never thought of filename too long will be an issue.

PR #564, opened 2023-09-29
Earclip

### LINKED ISSUE #530: Refactor triangulator
#528 helped a fair bit with reliability, but it certainly didn't solve everything (see the disabled test, as well as #529). The trick with the Boolean algorithm is I have a proof of topological robustness (manifoldness guarantee), but getting correct geometry is still dependent on the triangulator giving a geometrically valid result, and not just using its manifold fallback. I invented this triangulator, and it has no such mathematical guarantee, which is why it is the source of most of our bugs (along with the decimator). I've rewritten it several times as it got too complicated and I came to understand the problem space better - I think it's time to do this again.

There are some complexities I really don't like, as they feel like hacks and make it difficult to reason about. They mostly come down to logic that skips the sweep line, the worst offender being `isHole` which searches way ahead to determine the winding direction. This was only necessary because I didn't have a way to swap active edges out of being a hole, so I had to know it was a hole when it was first found in order to swap it in once and leave it there. However, there are plenty of other degenerate cases where a hole might need to be swapped, and it would fit more cleanly in the logic if everything can happen at the sweep line. 

I would also like to build with a requirement in mind, inspired by @pca006132: any valid polygon should have the exact same result (and follow all the same branches) at any value of epsilon, including zero. Only polygons that are epsilon-valid, but invalid at a lower epsilon should get different results. This should allow us to better separate the code based on uncertainty, and I might even be able to check it automatically by having our tests compare verbose-mode outputs. 

I think the general logic of the vertical ordering (sweep forward and sweep back) and the final monotone triangulator are good. It's the horizontal ordering (shiftEast, shiftWest) that needs refactoring, as I think the activePairs data structure is not adequate. Pairs are not the right unit, as they must be split and combined when swapping to or from a hole. 

I think instead a list of active edges that only point to one vert, but implicitly always alternate down, up, down, up, ... will be more flexible. I believe we also need to keep track of how these edges have already been found to attach to each other: they each need to point to their end of a list of start and merge verts which are always ordered in rightward vertex progression. As currently, the key piece of information to store is the Eastern edge each merge vert should be inserted next to in the sweep back pass. Now we will be able to update these after the sweep line passes them, as we may need to use y-value information to determine where degenerate holes can be inserted in this list. I believe we can also get more concrete about what `uncertain` means, having it apply only eastwards and therefore representing the relative order uncertainty *between* adjacent edges, rather than in a particular direction.

### LINKED ISSUE #535: triangulator bugs related to #528
Things from #529. Bugs:

1. 
```
Error message for triangulation:
Error in file: /home/pca006132/code/manifold/src/polygon/src/polygon.cpp (158): 'std::all_of(triangles.begin(), triangles.end(), [&vertPos, precision](const glm::ivec3 &tri) { return CCW(vertPos[tri[0]], vertPos[tri[1]], vertPos[tri[2]], precision) >= 0; })' is false: triangulation is not entirely CCW!
------------
Final polygon:
polys.push_back({
    {2.09305096, 5.86389685},  //
    {2.06142592, 5.85550117},  //
    {1.93763256, 5.4309001},  //
});
polys.push_back({
    {1.84193516, 5.90364313},  //
    {1.93763256, 5.4309001},  //
    {1.84193516, 5.90364313},  //
    {1.93763256, 5.4309001},  //
    {1.9408412, 5.90364313},  //
});
show(array([
  [2.09305096, 5.86389685],
  [2.06142592, 5.85550117],
  [1.93763256, 5.4309001],
]))
show(array([
  [1.84193516, 5.90364313],
  [1.93763256, 5.4309001],
  [1.84193516, 5.90364313],
  [1.93763256, 5.4309001],
  [1.9408412, 5.90364313],
]))
------------
Erroneous triangulation:
show(array([[1.8419352, 5.9036431], [1.9376326, 5.4309001], [1.9408412, 5.9036431]]))
show(array([[1.9408412, 5.9036431], [1.8419352, 5.9036431], [1.9376326, 5.4309001]]))
show(array([[1.8419352, 5.9036431], [1.9408412, 5.9036431], [1.9376326, 5.4309001]]))
^ not CCW
show(array([[2.0930510, 5.8638968], [2.0614259, 5.8555012], [1.9376326, 5.4309001]]))
```
2.
```
Error in file: /home/pca006132/code/manifold/src/polygon/src/polygon.cpp (103): 'std::distance(forward.begin(), end) == n_edges' is false: Half of halfedges should be forward.
polys.push_back({
    {20.1267338, 2.00670671},  //
    {20.12673, 2.00670576},  //
    {19.7702141, 1.63188827},  //
    {19.7702141, 1.63188827},  //
    {19.7702141, 1.63188827},  //
    {19.5752983, 1.81728792},  //
    {19.5623493, 1.81283963},  //
    {19.7630997, 1.54846776},  //
    {19.7630997, 1.54846776},  //
});
polys.push_back({
    {19.7702141, 1.63188827},  //
    {19.898922, 1.92845321},  //
    {19.5752983, 1.81728792},  //
});
polys.push_back({
    {19.7630997, 1.54846776},  //
    {19.7630997, 1.54846776},  //
    {19.7630997, 1.54846776},  //
});
array([
  [20.1267338, 2.00670671],
  [20.12673, 2.00670576],
  [19.7702141, 1.63188827],
  [19.7702141, 1.63188827],
  [19.7702141, 1.63188827],
  [19.5752983, 1.81728792],
  [19.5623493, 1.81283963],
  [19.7630997, 1.54846776],
  [19.7630997, 1.54846776],
])
array([
  [19.7702141, 1.63188827],
  [19.898922, 1.92845321],
  [19.5752983, 1.81728792],
])
array([
  [19.7630997, 1.54846776],
  [19.7630997, 1.54846776],
  [19.7630997, 1.54846776],
])
```

elalish:
Fixes #530 
Fixes #535 

Well, this turned into a big one. I spent a week or two trying to refactor my monotone triangulator to be more robust given the new test cases, and finally decided it was just too convoluted, mostly because it's a sweep-line approach, which doesn't handle the concept of epsilon perturbations well because there's not a great concept of certainty in a sorted list. 

Instead, I went back to the drawing board and remembered years ago when I first wrote this triangulator that I was on the fence between using monotone triangulation or ear clipping. Either way I needed to write my own, since none existed for dealing with epsilon-valid polygons. At the time I went with monotones because it is O(nlogn), while ear clipping is O(n^2). This time, I decided that the simplicity associated with ear clipping makes it worth the extra time, especially since our polygon edges n is generally a lot smaller than our triangles n. 

It's taken 2 or 3 weeks to build and robustify an ear clipping algorithm, but I must say that beats the many months worth of effort that went into the monotone triangulator. Ear clipping is also significantly less code, even with all my added precision-related code. 

@pca006132 your minimize test case function was invaluable! It cut 5,000 point polygons down to 12 point test cases - you can see I've added quite a few. 

Please help me kick the tires on this - the real goal is to get all our OpenSCAD problem models working properly.


bindings/c/include/manifoldc.h
```diff
@@ -301,8 +301,6 @@ ManifoldRect *manifold_rect_mul(void *mem, ManifoldRect *r, float x, float y);
 int manifold_rect_does_overlap_rect(ManifoldRect *a, ManifoldRect *r);
 int manifold_rect_is_empty(ManifoldRect *r);
 int manifold_rect_is_finite(ManifoldRect *r);
-ManifoldCrossSection *manifold_rect_as_cross_section(void *mem,
-                                                     ManifoldRect *r);
```
elalish, line 305: @geoffder I also did a refactor here to move Rect into public like Box, because I wanted to use it without depending on cross_section. The only real change is that I had to remove `AsCrossSection`. Doesn't seem like a big loss, since there's already a `CrossSection(Rect)` constructor. FYI in case this has wider implications.

src/polygon/src/polygon.cpp
```diff
@@ -192,900 +165,486 @@ void PrintFailure(const std::exception &e, const PolygonsIdx &polys,
...
+ * triangles in a mesh, and relatively few faces even need triangulation.
+ *
+ * The main adjustments for robustness involve clipping the sharpest ears first
+ * (a known technique to get higher triangle quality), and doing an exhaustive
+ * search to determine ear convexity exactly if the first geometric result is
+ * within precision.
  */
-class Monotones {
+
+class EarClip {
```
elalish, line 189: This is a terrible diff - this file is basically a rewrite. Considering the 1,000 lines of added tests, this PR is really a net loss of ~500 lines of code.

src/polygon/src/polygon.cpp
```diff
@@ -192,900 +165,486 @@ void PrintFailure(const std::exception &e, const PolygonsIdx &polys,
...
+      const glm::vec2 center = 0.5f * (left->pos + right->pos);
+      const float scale = 4 / glm::dot(openSide, openSide);
+      openSide = glm::normalize(openSide);
+
+      float totalCost = glm::dot(left->rightDir, rightDir) - 1 - precision;
+      if (CCW(pos, left->pos, right->pos, precision) == 0) {
+        return totalCost < -1 ? kBest : 0;
+      }
+      VertItr test = right;
+      while (test != left) {
```
elalish, line 363: This loop is what makes ear clipping O(n^2). I haven't been paying very close attention to the change in performance; hopefully it's not too bad, though I bet a few of our test cases will probably show it more dramatically than the common cases.

pca006132:
It seems that the CondensedMatter test case will cause invalid triangulation when getting disjoint polygons:
```
Triangulation failed! Precision = 0.000230642
Error in file: /home/pca006132/code/manifold/src/polygon/src/polygon.cpp (104): 'forward[i - 1].startVert != forward[i].startVert || forward[i - 1].endVert != forward[i].endVert' is false: Not a 2-manifold.
polys.push_back({
    {-4.95815706, 4.55738497},  //
    {-4.85924578, 4.65629625},  //
    {-4.85924578, 4.65629625},  //
    {-4.91438103, 4.64493704},  //
    {-4.91438103, 4.64493704},  //
    {-4.95815706, 4.65629625},  //
    {-4.95815706, 4.65629625},  //
    {-4.95815706, 4.65629625},  //
});
polys.push_back({
    {-4.95815706, 4.95815706},  //
    {-4.95815706, 4.85725117},  //
    {-4.95815706, 4.85725117},  //
    {-4.95815706, 4.85725117},  //
    {-4.95815706, 4.95815706},  //
    {-4.90248775, 4.9224453},  //
    {-4.87105799, 4.95158386},  //
    {-4.87105799, 4.95158386},  //
    {-4.85925102, 4.95815706},  //
});
polys.push_back({
    {-4.55738497, 4.95815706},  //
    {-4.65629101, 4.95815706},  //
    {-4.65105772, 4.95158339},  //
    {-4.64647388, 4.89330769},  //
    {-4.64647388, 4.89330769},  //
    {-4.65829086, 4.85725117},  //
    {-4.65829086, 4.85725117},  //
});
```

image

pca006132:
the disjoint polygon case is probably a problem in the boolean algorithm instead of triangulator? I think we are not supposed to have this polygon as input to triangulator.

pca006132:
OK found a bug:

```
Triangulation failed! Precision = 1e-05
Error in file: /home/pca006132/code/manifold/src/polygon/src/polygon.cpp (104): 'forward[i - 1].startVert != forward[i].startVert || forward[i - 1].endVert != forward[i].endVert' is false: Not a 2-manifold.
polys.push_back({
    {0.462962955, -0.5},  //
    {0.462962955, 0.129629627},  //
    {0.432784647, -0.314814836},  //
    {0.462962955, -0.316186577},  //
    {0.461591214, -0.351851851},  //
    {0.431412905, -0.351851851},  //
    {0.428669423, -0.42592594},  //
    {0.462962955, -0.427297682},  //
    {0.42592594, -0.5},  //
});
polys.push_back({
    {0.438271612, -0.413122982},  //
    {0.438728869, -0.401234567},  //
    {0.450617284, -0.401691824},  //
});
polys.push_back({
    {0.438271612, -0.302011877},  //
    {0.438728869, -0.290123463},  //
    {0.450617284, -0.29058072},  //
});
```

image

Some of the verbose outputs:
```
connected 11 to 4
vert: 0, left: 8, right: 1, cost: 0.0123457
vert: 1, left: 0, right: 2, cost: 0.0123457
vert: 2, left: 1, right: 3, cost: 0.0123457
vert: 3, left: 2, right: 4, cost: 0
vert: 4, left: 3, right: 11, cost: 0
vert: 11, left: 4, right: 9, cost: -0.376891
vert: 9, left: 11, right: 10, cost: 0
vert: 10, left: 9, right: 11, cost: 0
vert: 11, left: 10, right: 4, cost: 0
vert: 4, left: 11, right: 5, cost: 0
vert: 5, left: 4, right: 6, cost: -0.962998
vert: 6, left: 5, right: 7, cost: -0.975249
vert: 7, left: 6, right: 8, cost: 0
vert: 8, left: 7, right: 0, cost: -1.00001
vert: 8, left: 7, right: 0, cost: -1.00001
output tri: 8, 0, 7
vert: 0, left: 7, right: 1, cost: -inf
output tri: 0, 1, 7
vert: 6, left: 5, right: 7, cost: -0.975249
output tri: 6, 7, 5       <------- ??????
vert: 5, left: 4, right: 7, cost: -1.00001
output tri: 5, 7, 4
vert: 11, left: 4, right: 9, cost: -0.376891
output tri: 11, 9, 4
vert: 7, left: 4, right: 1, cost: 0
output tri: 7, 1, 4
vert: 4, left: 11, right: 1, cost: -1.00001
output tri: 4, 1, 11
vert: 11, left: 10, right: 1, cost: -1.00001
output tri: 11, 1, 10
vert: 1, left: 10, right: 2, cost: -0.0405289
output tri: 1, 2, 10
vert: 10, left: 9, right: 2, cost: -1.00001
output tri: 10, 2, 9
vert: 2, left: 9, right: 3, cost: -1.00001
output tri: 2, 3, 9
No ear found!
```

pca006132:
It seems that the cause is due to
https://github.com/elalish/manifold/blob/85e9a5957c8a727a8bc50e80bfbfd812271ffb04/src/polygon/src/polygon.cpp#L631-L632

When you add an ear, you did not check if there is any point inside the triangle (from what I can see).

Btw, I think the `CutKeyholes` function is section 4 of https://www.geometrictools.com/Documentation/TriangulationByEarClipping.pdf, but it seems that there are some differences from the algorithm in the documentation. Can you explain it a bit?

pca006132:
I am not sure about the second case, but for the CondensedMatter16 one, it seems that there is indeed non-manifold output because the CI got index out of range error, which is typical when we have non-manifold triangulation.

elalish:
Hmm, yeah I'm troubled by the Vec out of range error. I thought this was checking it was an even-manifold, just not a 2-manifold. SimplifyTopology already has code to fix even-manifold edges. If you know any way to pinpoint where that Vec out of range error is happening, or what geometry is causing it, that would be very helpful. It feels like I might just be triggering a latent bug in the decimator. It's also strange that it doesn't seem to be deterministic. I've seen it locally once or twice, but it's hard to repro.

pca006132:
For determinism, try to build it with MANIFOLF_PAR=NONE.
For where the out of range error occurs, I usually add __builtin_trap() to the out of range code path in vec.h to trigger gdb when the error occurs. TBB will consume the error and rethrow it when a task exits so this is the only way that can give you a reliable call stack.

pca006132:
btw, I think the triangulation for the second case is indeed incorrect:
image

elalish:
Perfect, thanks. I'll look into it.

elalish:
@pca006132 I added comments to document the algorithm; let me know if it's easier to follow now.

elalish:
Okay, I fixed your two bugs, @pca006132 - thanks for nice test cases! CondensedMatter seems much happier now. Oddly, I have every build failing CoplanarProp now, just getting extra verts, yet it never fails on my machine. I have no idea what's going on there. The Windows build spat out another test polygon that repros its issue for me locally, so I'll deal with that next. Not this weekend though...

pca006132:
also if you build with MANIFOLD_PAR=NONE, Sponge4 will fail with non-CCW triangulation.

minimized case:
```
Precision = 1e-05
Error in file: /home/pca006132/code/manifold/src/polygon/src/polygon.cpp (130): 'std::all_of(triangles.begin(), triangles.end(), [&vertPos, precision](const glm::ivec3 &tri) { return CCW(vertPos[tri[0]], vertPos[tri[1]], vertPos[tri[2]], precision) >= 0; })' is false: triangulation is not entirely CCW!
------------
Final polygon:
polys.push_back({
    {-0.376543224, 0.0802469105},  //
    {-0.277777791, 0.277777791},  //
    {-0.388888896, 0.277777791},  //
});
polys.push_back({
    {-0.364197552, 0.104938269},  //
    {-0.376543224, 0.117283955},  //
    {-0.364197552, 0.117283955},  //
});
polys.push_back({
    {-0.364197552, 0.141975313},  //
    {-0.376543224, 0.154320985},  //
    {-0.364197552, 0.154320985},  //
});
polys.push_back({
    {-0.339506179, 0.19135803},  //
    {-0.327160507, 0.19135803},  //
    {-0.327160507, 0.179012358},  //
});
polys.push_back({
    {-0.314814836, 0.203703716},  //
    {-0.351851851, 0.240740761},  //
    {-0.314814836, 0.240740761},  //
});
show(array([
  [-0.376543224, 0.0802469105],
  [-0.277777791, 0.277777791],
  [-0.388888896, 0.277777791],
]))
show(array([
  [-0.364197552, 0.104938269],
  [-0.376543224, 0.117283955],
  [-0.364197552, 0.117283955],
]))
show(array([
  [-0.364197552, 0.141975313],
  [-0.376543224, 0.154320985],
  [-0.364197552, 0.154320985],
]))
show(array([
  [-0.339506179, 0.19135803],
  [-0.327160507, 0.19135803],
  [-0.327160507, 0.179012358],
]))
show(array([
  [-0.314814836, 0.203703716],
  [-0.351851851, 0.240740761],
  [-0.314814836, 0.240740761],
]))
```

image

pca006132:
OK I found the issues. There are two problems: we did not order the bridges appropriately and the bridges we found may be invalid. They occur pretty frequently for Sponge4 when MANIFOLD_PAR=NONE. Here is my attempt at fixing it. All test are passed in both MANIFOLD_PAR=NONE and MANIFOLD_PAR=TBB except the Boolean.CoplanarProp where there are more verts. Diff:
```diff
diff --git a/extras/minimize_testcase.cpp b/extras/minimize_testcase.cpp
index 7ccf60d..633af52 100644
--- a/extras/minimize_testcase.cpp
+++ b/extras/minimize_testcase.cpp
@@ -306,6 +306,7 @@ int main(int argc, char **argv) {
   PolygonParams().intermediateChecks = true;
   PolygonParams().processOverlaps = false;
   PolygonParams().suppressErrors = true;
+  PolygonParams().verbose = true;
 
   if (argc < 2) {
     std::cerr << "Usage: " << argv[0] << " <input file>" << std::endl;
diff --git a/src/polygon/src/polygon.cpp b/src/polygon/src/polygon.cpp
index b8fa938..472bde4 100644
--- a/src/polygon/src/polygon.cpp
+++ b/src/polygon/src/polygon.cpp
@@ -172,6 +172,48 @@ void PrintFailure(const std::exception &e, const PolygonsIdx &polys,
 #define PRINT(msg)
 #endif
 
+inline bool intersect(glm::vec2 p0, glm::vec2 p1, glm::vec2 q0, glm::vec2 q1,
+                      float precision) {
+  glm::vec2 r = p1 - p0;
+  glm::vec2 s = q1 - q0;
+  // allow some error in the intersection point
+  float epsilon_r = 2.0 * precision / glm::length(r);
+  float epsilon_s = 2.0 * precision / glm::length(s);
+  float rxs = glm::determinant(glm::mat2(r, s));
+  // in case they are nearly collinear, ignore them...
+  // this is to avoid treating degenerate triangles as intersecting
+  //
+  // note that this does not correspond to a fixed angle,
+  // but seems to work well enough
+  if (rxs < kTolerance && rxs > -kTolerance) {
+    return false;
+  }
+  float u = glm::determinant(glm::mat2(q0 - p0, r)) / rxs;
+  float t = glm::determinant(glm::mat2(q0 - p0, s)) / rxs;
+  // we only care about intersection in the middle of both lines, excluding
+  // their ends
+  if (epsilon_r <= u && u <= (1 - epsilon_r) &&  //
+      epsilon_s <= t && t <= (1 - epsilon_s)) {
+    // there can be cases which r and s are close to collinear,
+    // so even though the lines are not intersecting,
+    // perturbation along r/s is not enough.
+    //
+    // in that case, apply perturbation perpendicular to r
+    glm::vec2 r_orth = glm::normalize(glm::vec2(-r.y, r.x)) * precision;
+    float u1 = glm::determinant(glm::mat2(q0 - p0 + r_orth, r)) / rxs;
+    float u2 = glm::determinant(glm::mat2(q0 - p0 - r_orth, r)) / rxs;
+    float t1 = glm::determinant(glm::mat2(q0 - p0 + r_orth, s)) / rxs;
+    float t2 = glm::determinant(glm::mat2(q0 - p0 - r_orth, s)) / rxs;
+    return 0 <= u1 && u1 <= 1 &&  //
+           0 <= t1 && t1 <= 1 &&  //
+           0 <= u2 && u2 <= 1 &&  //
+           0 <= t2 && t2 <= 1;
+    return true;
+  } else {
+    return false;
+  }
+}
+
 /**
  * Ear-clipping triangulator based on David Eberly's approach from Geometric
  * Tools, but adjusted to handle epsilon-valid polygons, and including a
@@ -617,6 +659,16 @@ class EarClip {
     }
   }
 
+  bool validBridge(glm::vec2 start, glm::vec2 end, VertItr polyStart) const {
+    VertItr vert = polyStart->right;
+    while (vert != polyStart) {
+      if (intersect(start, end, vert->pos, vert->right->pos, precision_))
+        return false;
+      vert = vert->right;
+    }
+    return true;
+  }
+
   // This converts the initial guess for the keyhole location into the final one
   // and returns it. It does so by finding any reflex verts inside the triangle
   // containing the guessed connection and the initial horizontal line, and
@@ -631,6 +683,8 @@ class EarClip {
     const glm::vec2 left = start->pos - guess->pos;
     const glm::vec2 right = intersection - guess->pos;
     float minD2 = glm::dot(left, left);
+    std::vector<VertItr> bests = {best};
+    std::set<int> invalidIds;
     while (vert != guess) {
       const glm::vec2 offset = vert->pos - guess->pos;
       const glm::vec2 diff = vert->pos - start->pos;
@@ -639,11 +693,48 @@ class EarClip {
           above * glm::determinant(glm::mat2(left, offset)) > -precision_ &&
           above * glm::determinant(glm::mat2(offset, right)) > -precision_ &&
           !vert->IsConvex(precision_)) {
+        if (invalidIds.find(vert->mesh_idx) != invalidIds.end() ||
+            !validBridge(start->pos, vert->pos, guess)) {
+#ifdef MANIFOLD_DEBUG
+          if (params.verbose) {
+            std::cout << "skipped " << start->mesh_idx << " to "
+                      << best->mesh_idx << " because it is an invalid bridge"
+                      << std::endl;
+          }
+#endif
+          // cache the invalid ids so we don't have to do expensive validBridge
+          // check next time
+          invalidIds.insert(vert->mesh_idx);
+          vert = vert->right;
+          continue;
+        }
         minD2 = d2;
+        if (best->mesh_idx != vert->mesh_idx) {
+          bests.clear();
+        }
         best = vert;
       }
+      if (best->mesh_idx == vert->mesh_idx) bests.push_back(vert);
       vert = vert->right;
     }
+    if (bests.size() > 1) {
+      // make sure that incoming edges are in CW order
+      glm::vec2 base = glm::normalize(start->pos - best->pos);
+      float baseAngle = std::atan2(base.x, base.y);
+      float maxAngle = -std::numeric_limits<float>::infinity();
+      for (VertItr itr : bests) {
+        glm::vec2 v = glm::normalize(itr->left->pos - best->pos);
+        float angle = std::atan2(v.x, v.y) - baseAngle;
+        // in the special case in which they are collinear
+        // we can assume that we should place after it...
+        while (angle <= 0) angle += 2 * glm::pi<float>();
+        while (angle > 2 * glm::pi<float>()) angle -= 2 * glm::pi<float>();
+        if (angle > maxAngle) {
+          maxAngle = angle;
+          best = itr;
+        }
+      }
+    }
 #ifdef MANIFOLD_DEBUG
     if (params.verbose) {
       std::cout << "connected " << start->mesh_idx << " to " << best->mesh_idx
```

And here are some tests cases:

<details>

```
Note: Google Test filter = Samples.Sponge4
[==========] Running 1 test from 1 test suite.
[----------] Global test environment set-up.
[----------] 1 test from Samples
[ RUN      ] Samples.Sponge4
-----------------------------------
Triangulation failed! Precision = 1e-05
Error in file: /home/pca006132/code/manifold/src/polygon/src/polygon.cpp (130): 'std::all_of(triangles.begin(), triangles.end(), [&vertPos, precision](const glm::ivec3 &tri) { return CCW(vertPos[tri[0]], vertPos[tri[1]], vertPos[tri[2]], precision) >= 0; })' is false: triangulation is not entirely CCW!
polys.push_back({
    {-0.209190682, 0.351851851},  //
    {-0.206447199, 0.42592594},  //
    {-0.229766816, 0.42592594},  //
    {-0.240740761, 0.351851851},  //
});
polys.push_back({
    {-0.228395075, 0.364197552},  //
    {-0.228395075, 0.372885257},  //
    {-0.219707385, 0.376543224},  //
});
polys.push_back({
    {-0.216049403, 0.401234567},  //
    {-0.224737093, 0.401234567},  //
    {-0.228395075, 0.409922272},  //
});
```

```
Precision = 1e-05
Error in file: /home/pca006132/code/manifold/src/polygon/src/polygon.cpp (130): 'std::all_of(triangles.begin(), triangles.end(), [&vertPos, precision](const glm::ivec3 &tri) { return CCW(vertPos[tri[0]], vertPos[tri[1]], vertPos[tri[2]], precision) >= 0; })' is false: triangulation is not entirely CCW!
------------
Final polygon:
polys.push_back({
    {-0.376543224, 0.0802469105},  //
    {-0.277777791, 0.277777791},  //
    {-0.388888896, 0.277777791},  //
});
polys.push_back({
    {-0.364197552, 0.104938269},  //
    {-0.376543224, 0.117283955},  //
    {-0.364197552, 0.117283955},  //
});
polys.push_back({
    {-0.364197552, 0.141975313},  //
    {-0.376543224, 0.154320985},  //
    {-0.364197552, 0.154320985},  //
});
polys.push_back({
    {-0.339506179, 0.19135803},  //
    {-0.327160507, 0.19135803},  //
    {-0.327160507, 0.179012358},  //
});
polys.push_back({
    {-0.314814836, 0.203703716},  //
    {-0.351851851, 0.240740761},  //
    {-0.314814836, 0.240740761},  //
});
```

```
Triangulation failed! Precision = 1e-05
polys.push_back({
    {0.055555556, 0.277777791},  //
    {-0.055555556, 0.277777791},  //
    {-0.055555556, 0.166666672},  //
    {0, 0.166666672},  //
    {0.0308641978, 0.228395075},  //
    {0.0308641978, 0.228395075},  //
    {0.0308641978, 0.228395075},  //
});
polys.push_back({
    {0.00617283955, 0.179012358},  //
    {0.00480109733, 0.179012358},  //
    {-0.00617283955, 0.179012358},  //
    {-0.00617283955, 0.180384099},  //
    {-0.00617283955, 0.19135803},  //
    {-0.00480109733, 0.19135803},  //
    {0.00617283955, 0.19135803},  //
    {0.00617283955, 0.189986289},  //
});
polys.push_back({
    {0.0185185187, 0.203703716},  //
    {0.0144032929, 0.203703716},  //
    {-0.0185185187, 0.203703716},  //
    {-0.0185185187, 0.20781894},  //
    {-0.0185185187, 0.240740761},  //
    {-0.0144032929, 0.240740761},  //
    {0.0185185187, 0.240740761},  //
    {0.0185185187, 0.236625537},  //
});
polys.push_back({
    {0.0432098769, 0.253086448},  //
    {0.0418381356, 0.253086448},  //
    {0.0308641978, 0.253086448},  //
    {0.0308641978, 0.254458189},  //
    {0.0308641978, 0.265432119},  //
    {0.0322359391, 0.265432119},  //
    {0.0432098769, 0.265432119},  //
    {0.0432098769, 0.264060378},  //
});
polys.push_back({
    {0.00617283955, 0.253086448},  //
    {0.00480109733, 0.253086448},  //
    {-0.00617283955, 0.253086448},  //
    {-0.00617283955, 0.254458189},  //
    {-0.00617283955, 0.265432119},  //
    {-0.00480109733, 0.265432119},  //
    {0.00617283955, 0.265432119},  //
    {0.00617283955, 0.264060378},  //
});
polys.push_back({
    {-0.0432098769, 0.179012358},  //
    {-0.0432098769, 0.180384099},  //
    {-0.0432098769, 0.19135803},  //
    {-0.0418381356, 0.19135803},  //
    {-0.0308641978, 0.19135803},  //
    {-0.0308641978, 0.189986289},  //
    {-0.0308641978, 0.179012358},  //
    {-0.0322359391, 0.179012358},  //
});
polys.push_back({
    {-0.0308641978, 0.216049403},  //
    {-0.0322359391, 0.216049403},  //
    {-0.0432098769, 0.216049403},  //
    {-0.0432098769, 0.217421144},  //
    {-0.0432098769, 0.228395075},  //
    {-0.0418381356, 0.228395075},  //
    {-0.0308641978, 0.228395075},  //
    {-0.0308641978, 0.227023333},  //
});
polys.push_back({
    {-0.0308641978, 0.253086448},  //
    {-0.0322359391, 0.253086448},  //
    {-0.0432098769, 0.253086448},  //
    {-0.0432098769, 0.254458189},  //
    {-0.0432098769, 0.265432119},  //
    {-0.0418381356, 0.265432119},  //
    {-0.0308641978, 0.265432119},  //
    {-0.0308641978, 0.264060378},  //
});
```

```
Triangulation failed! Precision = 1e-05
polys.push_back({
    {0.277777791, 0.055555556},  //
    {0.290123463, 0.080246903},  //
    {0.290123463, 0.0802469105},  //
    {0.290123463, 0.0802469105},  //
    {0.388888896, 0.277777791},  //
    {0.277777791, 0.277777791},  //
});
polys.push_back({
    {0.302469134, 0.104938269},  //
    {0.292866945, 0.104938269},  //
    {0.290123463, 0.104938269},  //
    {0.290123463, 0.114540473},  //
    {0.290123463, 0.117283955},  //
    {0.299725652, 0.117283955},  //
    {0.302469134, 0.117283955},  //
    {0.302469134, 0.107681751},  //
});
polys.push_back({
    {0.290123463, 0.141975313},  //
    {0.290123463, 0.151577502},  //
    {0.290123463, 0.154320985},  //
    {0.299725652, 0.154320985},  //
    {0.302469134, 0.154320985},  //
    {0.302469134, 0.144718796},  //
    {0.302469134, 0.141975313},  //
    {0.292866945, 0.141975313},  //
});
polys.push_back({
    {0.302469134, 0.179012358},  //
    {0.292866945, 0.179012358},  //
    {0.290123463, 0.179012358},  //
    {0.290123463, 0.188614547},  //
    {0.290123463, 0.19135803},  //
    {0.299725652, 0.19135803},  //
    {0.302469134, 0.19135803},  //
    {0.302469134, 0.181755841},  //
});
polys.push_back({
    {0.290123463, 0.216049403},  //
    {0.290123463, 0.225651592},  //
    {0.290123463, 0.228395075},  //
    {0.299725652, 0.228395075},  //
    {0.302469134, 0.228395075},  //
    {0.302469134, 0.218792886},  //
    {0.302469134, 0.216049403},  //
    {0.292866945, 0.216049403},  //
});
polys.push_back({
    {0.339506179, 0.179012358},  //
    {0.32990399, 0.179012358},  //
    {0.327160507, 0.179012358},  //
    {0.327160507, 0.188614547},  //
    {0.327160507, 0.19135803},  //
    {0.336762697, 0.19135803},  //
    {0.339506179, 0.19135803},  //
    {0.339506179, 0.181755841},  //
});
polys.push_back({
    {0.314814836, 0.203703716},  //
    {0.314814836, 0.232510313},  //
    {0.314814836, 0.240740761},  //
    {0.343621403, 0.240740761},  //
    {0.351851851, 0.240740761},  //
    {0.351851851, 0.211934164},  //
    {0.351851851, 0.203703716},  //
    {0.323045284, 0.203703716},  //
});
polys.push_back({
    {0.290123463, 0.253086448},  //
    {0.290123463, 0.262688637},  //
    {0.290123463, 0.265432119},  //
    {0.299725652, 0.265432119},  //
    {0.302469134, 0.265432119},  //
    {0.302469134, 0.25582993},  //
    {0.302469134, 0.253086448},  //
    {0.292866945, 0.253086448},  //
});
polys.push_back({
    {0.327160507, 0.253086448},  //
    {0.327160507, 0.262688637},  //
    {0.327160507, 0.265432119},  //
    {0.336762697, 0.265432119},  //
    {0.339506179, 0.265432119},  //
    {0.339506179, 0.25582993},  //
    {0.339506179, 0.253086448},  //
    {0.32990399, 0.253086448},  //
});
polys.push_back({
    {0.364197552, 0.253086448},  //
    {0.364197552, 0.262688637},  //
    {0.364197552, 0.265432119},  //
    {0.373799741, 0.265432119},  //
    {0.376543224, 0.265432119},  //
    {0.376543224, 0.25582993},  //
    {0.376543224, 0.253086448},  //
    {0.366941035, 0.253086448},  //
});
```

</details>

pca006132:
In general I found the new triangulator quite readable, at least it is not like the previous triangulator which just feels like black magic to me :).

I am not too worried on the performance. This PR did not show significant performance regression even for some pretty extreme examples (Sponge4 for example).
Perhaps it will be slow for extreme cases, e.g. complex svg file extrusion, but we can pretend this is not an issue for now until users complain about it. I think getting a robust and readable triangulator is most important.

pca006132:
Intermittent bug:

```
Note: Google Test filter = Boolean.Sweep
[==========] Running 1 test from 1 test suite.
[----------] 1 test from Boolean
[ RUN      ] Boolean.Sweep
i = 4568, n = 2312
i = 4568, n = 2312
i = 4570, n = 2312

Thread 11 "manifold_test" received signal SIGILL, Illegal instruction.
[Switching to Thread 0x7fffeb47b6c0 (LWP 427350)]
(anonymous namespace)::FaceMortonBox::operator() (inout=..., this=0x7fffeb47a550) at /home/pca006132/code/manifold/src/manifold/src/sort.cpp:95
95	   if (halfedge[3 * face].pairedHalfedge < 0) {
(gdb) list
90	   int face = thrust::get<2>(inout);
91	
92	   // Removed tris are marked by all halfedges having pairedHalfedge = -1, and
93	   // this will sort them to the end (the Morton code only uses the first 30 of
94	   // 32 bits).
95	   if (halfedge[3 * face].pairedHalfedge < 0) {
96	     mortonCode = kNoCode;
97	     return;
98	   }
99	
```

have to say this is really weird...

pca006132:
Another bug that can cause segfault:
```
diff --git a/src/polygon/src/polygon.cpp b/src/polygon/src/polygon.cpp
index 472bde4..218cbb5 100644
--- a/src/polygon/src/polygon.cpp
+++ b/src/polygon/src/polygon.cpp
@@ -819,8 +819,10 @@ class EarClip {
       ClipEar(v);
       --numTri;
 
-      ProcessEar(v->left);
-      ProcessEar(v->right);
+      if (v->left != v)
+        ProcessEar(v->left);
+      if (v->right != v)
+        ProcessEar(v->right);
       // This is a backup vert that is used if the queue is empty (geometrically
       // invalid polygon), to ensure manifoldness.
       v = v->right;
```

which is triggered by this:
```
polys.push_back({
    {9.42795753, 9.36543179},  //
    {9.41050148, 9.39931583},  //
    {9.39856625, 9.36543179},  //
});
polys.push_back({
    {9.3985672, 9.36543179},  //
    {9.39856625, 9.36543179},  //
    {9.39856625, 9.36543179},  //
    {9.39856625, 9.36543274},  //
    {9.41050148, 9.39931488},  //
    {9.41945648, 9.38193321},  //
    {9.41326141, 9.36543179},  //
});
show(array([
  [9.42795753, 9.36543179],
  [9.41050148, 9.39931583],
  [9.39856625, 9.36543179],
]))
show(array([
  [9.3985672, 9.36543179],
  [9.39856625, 9.36543179],
  [9.39856625, 9.36543179],
  [9.39856625, 9.36543274],
  [9.41050148, 9.39931488],
  [9.41945648, 9.38193321],
  [9.41326141, 9.36543179],
]))
```

pca006132:
And this is the bug I got for sweep:
```
polys.push_back({
    {0, -21.8057232},  //
    {0.142438874, -21.3274956},  //
    {0.118176684, -21.1638756},  //
    {2.24720502, -20.7749138},  //
    {0.0947285369, -20.7749138},  //
    {0.0428545363, -19.7292175},  //
});
polys.push_back({
    {0.094730109, -20.7749138},  //
    {1.17382383, -20.7749138},  //
    {0.118176684, -21.1638756},  //
});
show(array([
  [0, -21.8057232],
  [0.142438874, -21.3274956],
  [0.118176684, -21.1638756],
  [2.24720502, -20.7749138],
  [0.0947285369, -20.7749138],
  [0.0428545363, -19.7292175],
]))
show(array([
  [0.094730109, -20.7749138],
  [1.17382383, -20.7749138],
  [0.118176684, -21.1638756],
]))
```

image

which outputs non-CCW triangle, and break manifoldness (`PolygonParams().processOverlaps = true;` disabled the check for Boolean.Sweep).

pca006132:
The bug for sweep is due to a bug with SignedDist. Consider the following over-exaggerated graph:
image

where in reality (2,8) and (4, 6) are in the same position. This is the *only* epsilon-valid interpretation, because if you move 8 and 6 in the other direction the winding direction will fail. However, SignedDist did not consider winding direction and instead think that 8 is not inside the ear 4,2,7, returned -inf cost, and clip 4,2,7 leading to a bug.
Not sure what is the best way of implementing this though.

elalish:
I assume there's a bridge and several verts at 7? I think ideally these get mostly collapsed as degenerate triangles before the ear queue even starts running, so that 4,2,7 is no longer an option. 

There's quite a bit of code in your patch - can you describe geometrically what you're checking for with `validBridge` and how it differs from the current checks? I'm also curious how ever get a situation where `vert->left == vert` - that makes me think something else is wrong.

Thanks for all your work identifying and fixing issues here, and I'm glad to hear the performance is not much affected.

pca006132:
> I assume there's a bridge and several verts at 7? I think ideally these get mostly collapsed as degenerate triangles before the ear queue even starts running, so that 4,2,7 is no longer an option.

There is a bridge at 7, but there are no other vertices with the same position as 7.

> There's quite a bit of code in your patch - can you describe geometrically what you're checking for with validBridge and how it differs from the current checks?

It is just checking the bridge against all edges in the polygon to see if there is any intersection. It seems to me that the current code in `FindBridge` does not check against all edges.

> I'm also curious how we ever get a situation where vert->left == vert - that makes me think something else is wrong.

This may be possible if the other ears are non-convex? This should trigger the issue.

```
polys.push_back({
    {9.42795753, 9.36543179},  //
    {9.41050148, 9.39931583},  //
    {9.39856625, 9.36543179},  //
});
polys.push_back({
    {9.3985672, 9.36543179},  //
    {9.39856625, 9.36543179},  //
    {9.39856625, 9.36543179},  //
    {9.39856625, 9.36543274},  //
    {9.41050148, 9.39931488},  //
    {9.41945648, 9.38193321},  //
    {9.41326141, 9.36543179},  //
});
show(array([
  [9.42795753, 9.36543179],
  [9.41050148, 9.39931583],
  [9.39856625, 9.36543179],
]))
show(array([
  [9.3985672, 9.36543179],
  [9.39856625, 9.36543179],
  [9.39856625, 9.36543179],
  [9.39856625, 9.36543274],
  [9.41050148, 9.39931488],
  [9.41945648, 9.38193321],
  [9.41326141, 9.36543179],
]))
```

I provided some test cases for every patch I made, so you can verify them.

pca006132:
> I assume there's a bridge and several verts at 7? I think ideally these get mostly collapsed as degenerate triangles before the ear queue even starts running, so that 4,2,7 is no longer an option.

Also, while the particular case above may be solved by clipping degenerate triangles, in general it may not be. Consider the following graph where 3 and 0 are in the same position (drawn differently to ease presentation)
image

1-2 and 4-5 can be made larger to make them normal triangles.
Point 3 will be considered outside the ear 5-0-1 but it will be not geometrically valid.

elalish:
Hmm, I think we'll correctly find 3 is invalidating the 5-0-1 ear due to how `SignedDistance` checks the attached edge when the vert is too close to call. Anyway, thanks for the examples; I'll look into them more this evening/tomorrow.

src/polygon/src/polygon.cpp
```diff
@@ -637,10 +656,15 @@ class EarClip {
       const float d2 = glm::dot(diff, diff);
       if (d2 < minD2 && vert->pos.y * above > start->pos.y * above &&
           above * glm::determinant(glm::mat2(left, offset)) > -precision_ &&
-          above * glm::determinant(glm::mat2(offset, right)) > -precision_ &&
-          !vert->IsConvex(precision_)) {
```
elalish: This was the cause of the incorrect bridge ordering. Convexity is not the right check - instead we need to check that our start vert is on the inside of the connector vert's two edges.

src/polygon/src/polygon.cpp
```diff
@@ -265,34 +265,53 @@ class EarClip {
     // it finds a clear geometric result. In the vast majority of cases the loop
     // will never run, and when it does, it usually only needs one iteration.
     bool IsConvex(float precision) const {
-      int convexity = CCW(left->pos, pos, right->pos, precision);
-      if (convexity != 0) {
-        return convexity > 0;
-      }
-
-      // Uncertain - walk the polygon to get certainty.
+      const float p2 = precision * precision;
```
elalish, line 268: The second test case problem came from failing to clip degenerate triangles, the reason being that IsConvex had a significant bug. I ended up going back to my old IsHole function and fixing it up for this, since I'd done a better job with it before.

test/boolean_test.cpp
```diff
@@ -189,11 +187,9 @@ TEST(Boolean, CoplanarProp) {
   MeshGL cylinderGL = WithIndexColors(cylinder.GetMesh());
   cylinder = Manifold(cylinderGL);
 
-  Manifold cylinder2 = cylinder.Scale({0.5f, 0.5f, 1.0f})
-                           .Rotate(0, 0, 15)
-                           .Translate({0.25f, 0.25f, 0.0f});
+  Manifold cylinder2 = cylinder.Scale({0.8f, 0.8f, 1.0f}).Rotate(0, 0, 5);
   Manifold out = cylinder - cylinder2;
-  ExpectMeshes(out, {{42, 84, 3, 68}});
+  ExpectMeshes(out, {{52, 104, 3, 88}});
```
elalish, line 192: I finally figured out why CoplanarProp was giving inconsistent results between my Mac and our CI: the test case is dependent on how many of the top-face triangles get intersected in the Boolean op. Our new triangulator is less deterministic (more dependent on vert sorting order). I fixed the test by making it so no matter what circle triangulation is used, every edge will always get intersected, which makes the test result deterministic, even when the triangulation is not.

elalish (commented):
Okay, I think I've got all these fixed now - please kick the tires some more; does it work better with those problem OpenSCAD cases now?

src/polygon/src/polygon.cpp
```diff
@@ -727,12 +728,17 @@ class EarClip {
       }
       if (Clipped(v)) {
         start = v->right->left;
+        if (!Clipped(start)) {
+          ProcessEar(start);
+          ++numTri;
```
elalish, line 733: Here was the real source of the seg fault - I had fixed this loop in FindStarts, but failed to make a similar fix here. I also found a better way, so I updated them both. I was failing to process the start vert if it had been updated, and I was also getting an incorrect triangle count. I believe the second one was the error causing the seg fault, as numTri is what's used to terminate the ear clipping loop.

elalish:
I have a repro with the Windows failure - two coincident verts are options for the start, and if it gets the wrong one, then it thinks it's a hole and connects it to something incorrectly. I'm going to fix it by checking holes with an area calculation instead.

pca006132:
It is still producing some incorrect triangulation, but it seems that there are less incorrect ones:
```
Triangulation failed! Precision = -1
Error in file: /home/pca006132/code/manifold/src/polygon/src/polygon.cpp (102): 'forward[i - 1].startVert != forward[i].startVert || forward[i - 1].endVert != forward[i].endVert' is false: Not a 2-manifold.
polys.push_back({
    {31.5, -40.5},  //
    {31.5, 24.5602417},  //
    {29.4397583, 22.5},  //
    {28.5, 7.5},  //
    {24.5602417, -22.5},  //
    {31.5, -31.5},  //
    {23.5843372, -31.5},  //
});
polys.push_back({
    {30.5, -6.27108431},  //
    {29.5, -5.72891569},  //
    {30.2710838, -5.5},  //
});
polys.push_back({
    {30.5, 18.5},  //
    {29.5, 17.5},  //
    {29.5, 18.5},  //
});
polys.push_back({
    {30.5, 20.7289162},  //
    {29.5, 21.2710838},  //
    {30.2710838, 21.5},  //
});
```

```
Triangulation failed! Precision = 0.000708769
Error in file: /home/pca006132/code/openscad/submodules/manifold/src/polygon/src/polygon.cpp (130): 'std::all_of(triangles.begin(), triangles.end(), [&vertPos, precision](const glm::ivec3 &tri) { return CCW(vertPos[tri[0]], vertPos[tri[1]], vertPos[tri[2]], precision) >= 0; })' is false: triangulation is not entirely CCW!
polys.push_back({
    {-33.2870407, 9.25113201},  //
    {-30.2349148, 9.25113201},  //
    {-29.3892632, 16.9009457},  //
});
polys.push_back({
    {-33.2869797, 9.25124931},  //
    {-33.2869797, 9.25125027},  //
    {-33.2869797, 9.25125217},  //
    {-29.3892632, 16.9009457},  //
    {-31.7609406, 9.25125027},  //
});
show(array([
  [-33.2870407, 9.25113201],
  [-30.2349148, 9.25113201],
  [-29.3892632, 16.9009457],
]))
show(array([
  [-33.2869797, 9.25124931],
  [-33.2869797, 9.25125027],
  [-33.2869797, 9.25125217],
  [-29.3892632, 16.9009457],
  [-31.7609406, 9.25125027],
]))
```

pca006132:
btw, we can fuzz the triangulator by applying random perturbation of less than epsilon for every vertices and check if the result triangulation is valid.
this will not be a test because it will not complete, but may be a useful way to check the robustness (especially handling epsilon-valid geometries) by utilizing our existing testcases.

elalish:
Agreed, a fuzzer is a good way to find new error cases to turn into tests. It'll also be good to try one with completely random verts to verify manifoldness for geometrically-invalid input.

elalish:
@pca006132 Okay, I fixed everything I can repro on my end, but I could use some help on these "vec out of range" issues on the CI. It makes sense that they would be caused by something non-manifold coming out of the triangulator, but I feel like I have a pretty solid debug check with `CheckTopology` that should cause the triangulator to throw before anything downstream. I can't figure out what's getting through - I'm comfortable fixing geometric errors, but I'm much more concerned about something topological causing the whole thing to seg fault. That really shouldn't ever happen, even if the algorithm goes badly wrong. Any ideas? I need to sleep.

pca006132:
I managed to get some errors for openscad tests, I will try to get the stacktrace so you can fix later. Thanks!

pca006132:
Interestingly the error is in sort.cpp GetFaceBoxMorton. Maybe the halfedge data structure is somehow corrupted?

pca006132:
ah, I found that I probably forgot to enable MANIFOLD_DEBUG. If you look at the test failure on mac, it says

```
[ RUN      ] Samples.CondensedMatter16
-----------------------------------
Triangulation failed! Precision = 0.000230642
Error in file: /Users/runner/work/manifold/manifold/src/polygon/src/polygon.cpp (104): 'forward[i - 1].startVert != forward[i].startVert || forward[i - 1].endVert != forward[i].endVert' is false: Not a 2-manifold.
polys.push_back({
    {9.82835007, -0.939956188},  //
    {10.1291866, -1.12029469},  //
    {10.1291866, -0.759618163},  //
});
polys.push_back({
    {9.82835007, -0.939956307},  //
    {9.82835007, -0.939956307},  //
    {9.82835007, -0.939956307},  //
    {10.053978, -0.80470264},  //
    {9.97876835, -1.0301255},  //
    {9.82835007, -0.939956307},  //
});
show(array([
  [9.82835007, -0.939956188],
  [10.1291866, -1.12029469],
  [10.1291866, -0.759618163],
]))
show(array([
  [9.82835007, -0.939956307],
  [9.82835007, -0.939956307],
  [9.82835007, -0.939956307],
  [10.053978, -0.80470264],
  [9.97876835, -1.0301255],
  [9.82835007, -0.939956307],
]))
produced this triangulation:
49950, 49949, 49951
49954, 49950, 49951
49954, 49951, 49952
13217, 13233, 49953
13234, 13217, 49953
49953, 13233, 13234
13234, 49953, 49954
13234, 49954, 49952
49953, 13234, 49952
unknown file: Failure
C++ exception with description "Error in file: /Users/runner/work/manifold/manifold/src/polygon/src/polygon.cpp (104): 'forward[i - 1].startVert != forward[i].startVert || forward[i - 1].endVert != forward[i].endVert' is false: Not a 2-manifold." thrown in the test body.
```

elalish:
@pca006132 That last test case is an interesting one. It has an earclip error that I'll fix, but the triangulation is basically fine - geometrically invalid, but it is an even manifold - not 2-manifold, but `DedupeEdge` should take care of that in `SimplifyTopology` to make a proper 2-manifold. I'd really like to figure out how we're getting a seg fault in this situation. It certainly won't be the correct result geometrically, but our topology should be much more robust than that. Unfortunately I can't repro this locally in the mesh. Is there a chance you can dig into it to see what's going wrong with our data structure while I keep digging through triangulator bugs?

pca006132:
Sure, I can try to add checks after each step in SimplifyTopology to make sure that the result is still valid. This will probably take some time as I have other things to do this weekend.

elalish:
If it's easier, you might just be able to output the intermediate even-manifold mesh. I'm hoping that'll repro the problem within the Manifold constructor, since that also runs `SimplifyTopology`. If I can just get a reliable test case, I can debug.

pca006132:
OK it seems that the problem is the halfedges after`SimplifyTopology` still references removed vertices (those with NAN coordinates). I managed to get deterministic repro with OpenSCAD (maze.scad) but the mesh is quite large and the build is complicated. Let me know if you need help debugging this.

elalish:
> OK it seems that the problem is the halfedges after`SimplifyTopology` still references removed vertices (those with NAN coordinates). I managed to get deterministic repro with OpenSCAD (maze.scad) but the mesh is quite large and the build is complicated. Let me know if you need help debugging this.

Thanks, this looks like the heart of the most serious bug. I haven't even tried debugging manifold within the context of OpenSCAD yet. If you're up for debugging this since you have the repro, that would be amazing. Does it still repro if you write out the mesh just prior to `SimplifyTopology` (as, say, a GLB) and then try to load it as a new Manifold? If that can repro the problem, then just send me that GLB, even if it's huge.

elalish:
Also, @pca006132, I'm review the sorting changes [here](https://github.com/elalish/manifold/blob/master/src/manifold/src/edge_op.cpp#L165) and I wonder if I missed something in code review. Your `entries` appear to be halfedges, basically, but `DedupeEdge` actually adds new halfedges, but I don't see a corresponding change to `entries`. Could it be that those halfedge vectors are getting out of sync? 

As an aside, I don't actually remember when `entries` got added or exactly what its purpose is. Do you?

elalish:
Ah, don't worry about the repro - I have one now. It occurred to me that all I need to do is make the triangulator bad enough that it produces a lot of non-2-manifold output (even manifolds) and then disable the triangulation checks and just wait for a Vec out of range error. So I just disabled the ear queue so it clips verts around in a circle and sure enough, I found a few that repro reliably. I should be able to debug now.

pca006132:
I am not sure about the entries. I guess if the input is a 4-manifold, the added halfedges should be valid and do not require further checks, but otherwise it may require further check? The SortEntry thing is really old, and iirc my changes is rather minor so we did not think much about it.

src/manifold/src/edge_op.cpp
```diff
@@ -380,6 +427,10 @@ void Manifold::Impl::RemoveIfFolded(int edge) {
   }
 }
 
+// Collapses the given edge by removing startVert. May split the mesh
+// topologically if the collapse would have resulted in a 4-manifold edge. Do
+// not collapse an edge if startVert is pinched - the vert will be marked NaN,
+// but other edges may still be pointing to it.
```
elalish, line 433: This was the primary cause of the seg fault, which I fixed in several ways. This problem has existed for quite some time, but was illuminated by unusual triangulations.

pca006132:
```
Error in file: /home/pca006132/code/openscad/submodules/manifold/src/polygon/src/polygon.cpp (127): 'std::all_of(triangles.begin(), triangles.end(), [&vertPos, precision](const glm::ivec3 &tri) { return CCW(vertPos[tri[0]], vertPos[tri[1]], vertPos[tri[2]], precision) >= 0; })' is false: triangulation is not entirely CCW!
polys.push_back({
    {-6.53556156, 12.4571409},  //
    {-6.72912979, 12.5610456},  //
    {-6.7918272, 12.5019674},  //
});
polys.push_back({
    {-6.72903824, 12.5008945},  //
    {-6.72912979, 12.5008078},  //
    {-6.72931194, 12.5008106},  //
    {-6.7918272, 12.5019674},  //
    {-6.72912979, 12.5610456},  //
    {-6.72894907, 12.501153},  //
    {-6.72894812, 12.5009804},  //
});
show(array([
  [-6.53556156, 12.4571409],
  [-6.72912979, 12.5610456],
  [-6.7918272, 12.5019674],
]))
show(array([
  [-6.72903824, 12.5008945],
  [-6.72912979, 12.5008078],
  [-6.72931194, 12.5008106],
  [-6.7918272, 12.5019674],
  [-6.72912979, 12.5610456],
  [-6.72894907, 12.501153],
  [-6.72894812, 12.5009804],
]))
```

elalish:
your latest test is passing for me since those fixes - working for you too?

pca006132:
no, it gave

```
Triangulation failed! Precision = 0.0005
Error in file: /home/pca006132/code/manifold/src/polygon/src/polygon.cpp (122): 'std::all_of(triangles.begin(), triangles.end(), [&vertPos, precision](const glm::ivec3 &tri) { return CCW(vertPos[tri[0]], vertPos[tri[1]], vertPos[tri[2]], precision) >= 0; })' is false: triangulation is not entirely CCW!
polys.push_back({
    {-6.53556156, 12.4571409},  //
    {-6.72912979, 12.5610456},  //
    {-6.7918272, 12.5019674},  //
});
polys.push_back({
    {-6.72903824, 12.5008945},  //
    {-6.72912979, 12.5008078},  //
    {-6.72931194, 12.5008106},  //
    {-6.7918272, 12.5019674},  //
    {-6.72912979, 12.5610456},  //
    {-6.72894907, 12.501153},  //
    {-6.72894812, 12.5009804},  //
});
show(array([
  [-6.53556156, 12.4571409],
  [-6.72912979, 12.5610456],
  [-6.7918272, 12.5019674],
]))
show(array([
  [-6.72903824, 12.5008945],
  [-6.72912979, 12.5008078],
  [-6.72931194, 12.5008106],
  [-6.7918272, 12.5019674],
  [-6.72912979, 12.5610456],
  [-6.72894907, 12.501153],
  [-6.72894812, 12.5009804],
]))
produced this triangulation:
9, 3, 4
8, 9, 4
7, 8, 4
7, 4, 5
2, 0, 1
6, 7, 5
```

elalish:
Ah, I've got the repro now - I forgot to specify the precision. Thanks!

pca006132:
hmmm

```
Triangulation failed! Precision = 0.000178177
Error in file: /home/pca006132/code/openscad/submodules/manifold/src/polygon/src/polygon.cpp (127): 'std::all_of(triangles.begin(), triangles.end(), [&vertPos, precision](const glm::ivec3 &tri) { return CCW(vertPos[tri[0]], vertPos[tri[1]], vertPos[tri[2]], precision) >= 0; })' is false: triangulation is not entirely CCW!
polys.push_back({
    {9.76247501, -8.75089836},  //
    {9.78284931, -8.75129509},  //
    {9.78242302, -8.75128555},  //
    {9.76939392, -8.74437904},  //
    {9.76939392, -8.74437904},  //
});
polys.push_back({
    {9.76247501, -8.75089836},  //
    {9.76939392, -8.74437904},  //
    {9.77548981, -8.75115108},  //
    {9.77501488, -8.7511425},  //
});
show(array([
  [9.76247501, -8.75089836],
  [9.78284931, -8.75129509],
  [9.78242302, -8.75128555],
  [9.76939392, -8.74437904],
  [9.76939392, -8.74437904],
]))
show(array([
  [9.76247501, -8.75089836],
  [9.76939392, -8.74437904],
  [9.77548981, -8.75115108],
  [9.77501488, -8.7511425],
]))
produced this triangulation:
14523, 363974, 462691
462691, 462692, 363944
14523, 462691, 363944
462689, 462690, 462688
462689, 462688, 462687
terminate called after throwing an instance of 'manifold::geometryErr'
  what():  Error in file: /home/pca006132/code/openscad/submodules/manifold/src/polygon/src/polygon.cpp (127): 'std::all_of(triangles.begin(), triangles.end(), [&vertPos, precision](const glm::ivec3 &tri) { return CCW(vertPos[tri[0]], vertPos[tri[1]], vertPos[tri[2]], precision) >= 0; })' is false: triangulation is not entirely CCW!
```

pca006132:
I feel like we should probably integrate the openscad tests into our CI, so it is easier to test them (some examples are more comprehensive than our unit tests). Will do it later.

pca006132:
I wrote the CI file which seems to work, but it is really slow (the tests are slow, the build is fine) and the tests failed with current master (expected though). https://github.com/pca006132/manifold/blob/e692c6ce0c1a25e15b0a8feab0c4683557e98ec5/.github/workflows/openscad.yml

I think you may want to run this locally. On mac, you should do the following:
```sh
git checkout --recursive --depth 0 https://github.com/pca006132/manifold.git
# get dependencies:
./scripts/macosx-build-homebrew.sh
brew install ninja
# substitute submodules/manifold with the current branch of manifold you are working with
git submodule set-branch -b earclip submodules/manifold
git submodule update --remote --recursive submodules/manifold
mkdir build
cd build
cmake -DUSE_BUILTIN_OPENCSG=ON -DEXPERIMENTAL=ON -DMANIFOLD_DEBUG=ON  -GNinja ..
ninja

# copy test repo
git clone https://gist.github.com/dd4e50d46610daa3f6502c6ee75b8cb4.git openscad-test
cd openscad-test
OPENSCAD=../build/openscad OPENSCADPATH=./libs ./test *.scad
```

it is referencing my fork because my fork contains a patch that enables intermediate checks in manifold.

elalish:
Is it possible you meant https://github.com/pca006132/openscad.git instead of https://github.com/pca006132/manifold.git?

elalish:
And can you give me an example of what `git submodule set-branch -b earclip submodules/manifold` should look like after replacement for a local working copy? I'm not sure what I should replace `submodules/manifold` by.

pca006132:
you can cd into the subdirectory and check git log. Also, if the changes are not pushed yet (only exists in your local repo) you can remove the directory completely and symlink it. it is just a regular directory pointing to manifold repo

elalish:
okay, everything built, but on your last line I'm getting `./render: line 58: ../build/openscad: No such file or directory`, which is true - I have OpenSCAD.app though. Also, what is `git clone https://gist.github.com/dd4e50d46610daa3f6502c6ee75b8cb4.git openscad-test` doing? Do you really want that cloned into the build directory? Isn't this already a part of the openscad repo?

pca006132:
1. You need to set `OPENSCAD` to the built openscad file. I am not familiar with mac so I guess I made a mistake here.
2. No, that repo is not inside openscad. openscad repo only contains some basic tests, I am using the benchmarks for the manifold pr in openscad previously and made some changes so each test only run once. It doesn't really matter where you clone it, as long as you set the `OPENSCAD` variable correctly so it can find the binary.

elalish:
Awesome, it's working! I'm going to edit your instructions in case I need to come back to them.

elalish:
Before updating to my local manifold I ran the tests to get a baseline. A bunch passed, but then it error out here:

```
minkowski-of-minkowski-difference.scad
CGAL error: assertion violation!
Expression : itl != it->second.end()
File       : /opt/homebrew/include/CGAL/Nef_3/SNC_external_structure.h
Line       : 1078
```

Is that expected? Is there a way to skip a test?

pca006132:
yeah this is expected, there are some weird bugs (likely on openscad side) regarding this.
no there is currently no way to skip tests, one temporary hack would be to rename the file such that it doesn't end with .scad and will not be selected by the wildcard.

elalish:
I'm also seeing 
```
scalemail.scad:N=2 $fn=100
WARNING: Can't open library 'sweep.scad'. in file scalemail.scad, line 64
```
Should sweep.scad be included in your openscad-test repo? Somehow the tests still pass anyway.

elalish:
Actually, maybe it's quite a bit worse than that? box.scad has been running for 15 minutes now with no sign of stopping. How long are these tests taking for you?

pca006132:
weird, iirc I got the errors pretty fast

maybe you can check if the polygon intermediate checks are running?

pca006132:
e.g. this is what I got when running box.scad:
```
box.scad
WARNING: variable center not specified as parameter in file box.scad, line 49
WARNING: variable center not specified as parameter in file box.scad, line 49
WARNING: variable center not specified as parameter in file box.scad, line 52
WARNING: variable center not specified as parameter in file box.scad, line 52
-----------------------------------
Triangulation failed! Precision = 0.000415995
Error in file: /home/pca006132/code/openscad/submodules/manifold/src/polygon/src/polygon.cpp (127): 'std::all_of(triangles.begin(), triangles.end(), [&vertPos, precision](const glm::ivec3 &tri) { return CCW(vertPos[tri[0]], vertPos[tri[1]], vertPos[tri[2]], precision) >= 0; })' is false: triangulation is not entirely CCW!
polys.push_back({
    {8.93379593, -41.2573624},  //
    {10.5178413, -41.2440758},  //
    {8.99354267, -41.1997375},  //
});
polys.push_back({
    {9.08317471, -41.2453728},  //
    {9.08610725, -41.242527},  //
    {9.0890007, -41.2397537},  //
    {9.08943081, -41.2436409},  //
    {9.08946514, -41.2439499},  //
    {9.08924961, -41.2441826},  //
    {9.0890007, -41.2444496},  //
});
show(array([
  [8.93379593, -41.2573624],
  [10.5178413, -41.2440758],
  [8.99354267, -41.1997375],
]))
show(array([
  [9.08317471, -41.2453728],
  [9.08610725, -41.242527],
  [9.0890007, -41.2397537],
  [9.08943081, -41.2436409],
  [9.08946514, -41.2439499],
  [9.08924961, -41.2441826],
  [9.0890007, -41.2444496],
]))
produced this triangulation:
32852, 32853, 32849
32852, 32849, 32848
32852, 32848, 32850
23618, 24421, 23625
32851, 32852, 32850
32851, 32850, 32847
terminate called recursively
```

also for the scalemail.scad, I also don't have the sweep.scad. I think it is not present in the original benchmark set as well. (there are also some other files that doesn't work and I have fixed them, so this one is probably just another issue)

elalish:
This is weird - I don't get errors at all (e.g. box.scad runs to completion, it just takes several minutes). Likewise, when I add a `std::cout` line to manifold and rebuild, it still doesn't print any messages to the command line in the openscad tests. The fact that it now runs super slow makes me think my symlink did something, and when I change manifold and run ninja, it does appear to recompile. I see `intermediateChecks = true` in your code. Sort of seems like all the problems are being somehow swallowed and maybe using some slow fallback?

pca006132:
Ah! This sounds plausible. maybe we can make the assertion call abort immediately after printing useful messages.
The behavior on my computer should probably be a bug as manifold failure should fallback to the slower CGAL routine.

elalish:
I tried the abort() trick - I did get it to abort, and also to print messages, so I know I'm controlling the code it's using. However, bolts.scad is not failing, nor using a fallback as best I can tell. I have no idea why it's so slow, nor why I'm getting a different result than you. Very strange.

pca006132:
weird. maybe you can ignore this for now and fix other failures first. hopefully their causes are the same.

elalish:
Took a good 20 minutes, but I got all your openscad tests to run, @pca006132. They all appear to have passed for me. How's it looking for you? 

I still have some ideas to further improve robustness of this triangulator, but I think it's better than it was before at least.

pca006132:
It seems that the CCW check is not enabled when `processOverlaps = true` (default). I got bad triangulation for
```
$fn=350; // ignore this since it is just for testing purposes
module hole_plate(size, hole_dm, hole_margin, hole_count = [ 2, 2 ])
{
    difference()
    {
        cube(size);
        abs_margin = hole_margin + hole_dm / 2;
        x_hole_dist = (size.x - 2 * abs_margin) / (hole_count.x - 1);
        y_hole_dist = (size.y - 2 * abs_margin) / (hole_count.y - 1);
        x_values = [abs_margin:x_hole_dist:size.x - abs_margin + 0.1];
        y_values = [abs_margin:y_hole_dist:size.y - abs_margin + 0.1];
        // holes
        for (x = x_values, y = y_values)
            translate([ x, y, -1 ]) cylinder(d = hole_dm, h = size.z + 2);
    }
}
hole_plate([ 100, 50, 5 ], 6, 4);
translate([ 0, 60, 0 ]) hole_plate(size = [ 50, 50, 5 ], hole_dm = 3, hole_margin = 2);
translate([ 60, 60, 0 ]) hole_plate([ 50, 50, 5 ], hole_dm = 5, hole_margin = 5);
translate([ 110, 0, 0 ]) hole_plate([ 100, 50, 5 ], 6, 4);
translate([ 0, -60, 0 ]) hole_plate(size = [ 50, 50, 5 ], hole_dm = 3, hole_margin = 2, hole_count = [ 10, 10 ]);
```

but it only throws an error when I set `processOverlaps = false`. Setting this also makes `box.scad` throws:
```
Triangulation failed! Precision = 0.000415995
Error in file: /home/pca006132/code/openscad/submodules/manifold/src/polygon/src/polygon.cpp (127): 'std::all_of(triangles.begin(), triangles.end(), [&vertPos, precision](const glm::ivec3 &tri) { return CCW(vertPos[tri[0]], vertPos[tri[1]], vertPos[tri[2]], precision) >= 0; })' is false: triangulation is not entirely CCW!
polys.push_back({
    {-22.7753239, 8.03145695},  //
    {-22.8791199, 8.04628658},  //
    {-22.8791199, 8.0424366},  //
    {-22.8792648, 8.04237938},  //
    {-22.8794041, 8.04239368},  //
    {-22.9786491, 8.05282307},  //
    {-22.9923439, 8.06109047},  //
    {-22.995327, 8.06289101},  //
    {-23.0484314, 8.07047844},  //
    {-23.056942, 8.06713009},  //
    {-23.0641346, 8.06429958},  //
    {-23.0679455, 8.06355},  //
    {-23.0689487, 8.06335258},  //
    {-23.0691357, 8.06331635},  //
    {-23.0723953, 8.06429958},  //
    {-23.1114883, 8.0794878},  //
    {-24.4342442, 8.26848221},  //
    {-24.4342442, 8.2057848},  //
});
show(array([
  [-22.7753239, 8.03145695],
  [-22.8791199, 8.04628658],
  [-22.8791199, 8.0424366],
  [-22.8792648, 8.04237938],
  [-22.8794041, 8.04239368],
  [-22.9786491, 8.05282307],
  [-22.9923439, 8.06109047],
  [-22.995327, 8.06289101],
  [-23.0484314, 8.07047844],
  [-23.056942, 8.06713009],
  [-23.0641346, 8.06429958],
  [-23.0679455, 8.06355],
  [-23.0689487, 8.06335258],
  [-23.0691357, 8.06331635],
  [-23.0723953, 8.06429958],
  [-23.1114883, 8.0794878],
  [-24.4342442, 8.26848221],
  [-24.4342442, 8.2057848],
]))
```

elalish:
Ah, indeed, `processOverlaps` should be false for tests. Okay, with that change and an abort trap I'm now seeing problems in box.scad. I'll keep digging.

pca006132:
I thought that the ccw check is enabled regardless of the value of processOverlaps, because iirc the result may not be manifold if the triangulation is not all ccw?

elalish:
No, manifoldness should never be affected by geometrically invalid results. We did have that happen because of a bug in the decimator I just fixed, which is why that had me so concerned.

elalish:
By the way, if you happen to find a way to change that openscad test so that it will terminate a given test when the triangulator throws, but still proceed to the next test, that would be very helpful. These shell scripts are making me really appreciate our GTest framework...

pca006132:
removing the exit on error flag for bash should do the trick. will change it when I have access to my computer later today

pca006132:
i.e.

```diff
diff --git a/test b/test
index ec163b2..a6bca8a 100755
--- a/test
+++ b/test
@@ -4,7 +4,7 @@
 #
 # ./bench foo.scad
 #
-set -euo pipefail
+# set -euo pipefail
 
 RUNS=${RUNS:-1}
 export OUTPUT_DIR=${OUTPUT_DIR:-$PWD/out}
```

elalish:
> it seems that the mesh produced is fine even though it throws error when `processOverlaps = false`.

As in, when you set `processOverlaps = true` you get a good result? Yeah, that's why it's the default. We still guarantee the result is manifold, and *often* the CW triangles are small enough not to be noticeable. Or the errors may be in the interior where they're harder to see. I think we're definitely getting closer now, but that CCW check is still very helpful for testing.

pca006132:
I think we can merge this now. This is definitely more robust!

test/polygon_test.cpp
```diff
@@ -131,6 +131,474 @@ TEST(Polygon, MultiMerge) {
...
+      {-0.462962955, -0.129629627},  //
+      {-0.42592594, -0.092592597},   //
+      {-0.42592594, -0.092592597},   //
+      {-0.42592594, -0.129629627},   //
+  });
+  TestPoly(polys, 15);
+};
+
+TEST(Polygon, Eberly) {
+  Polygons polys;
```
elalish, line 528: Long time, no see @davideberly! I've extended your earclipping algorithm to handle &epsilon;-valid polygons (polygons that may self-intersect, but for which a perturbation of each vertex by less than &epsilon; exists for which the polygon does not self-intersect). It's taken a lot of work, but I'm pleased with the results. 

In the process I found a small error in your paper. It turns out Mark Spoor was wrong and your original minimize-the-angle-approach is correct. Here is a counter-example polygon for finding a mutually visible vertex.

davideberly: Thanks for letting me know. I am in the process of dealing with another issue for the ear-clipping algorithm and will resolve that soon I hope. I will restore the previous algorithm.

elalish:
@pca006132 Sorry, I did another significant refactor - I promise this is the last! I believe it's significantly more robust now and there are no disabled tests anymore, as well as several new ones added. If you wouldn't mind trying it out on whatever OpenSCAD models we had before that were failing, I'd appreciate it. Of course the reduced polygon test cases all pass, but those reductions doesn't really apply anymore now that the algorithm has changed.

pca006132:
This is nice. I don't have more openscad models to check, I think it would be fine if the models I mentioned previously works.

PR #577, opened 2023-10-14
fix osx 10.13 error

### LINKED ISSUE #576: Support for OS X 10.13?
Hello! I am not familiar with C++ but was wondering if it would be feasible to make changes for Manifold to be compatible with Mac OS X 10.13 (and possibly lower.)

I'm building with `make -k` on OS X 10.13 and the two errors below in `csg_tree.cpp` (lines 85 and 87) appear to be the only blockers to a full build:

`manifold/src/manifold/src/csg_tree.cpp:85:14: error: call to unavailable function 'get': introduced in macOS 10.14`
`     return std::get<std::shared_ptr<const Manifold::Impl>>(p).get();`
`manifold/src/manifold/src/csg_tree.cpp:87:14: error: call to unavailable function 'get': introduced in macOS 10.14`
`     return std::get<std::shared_ptr<Manifold::Impl>>(p).get();`

Naïve searching turned up a StackOverflow answer suggesting that there were [doable workarounds](https://stackoverflow.com/questions/52521388/stdvariantget-does-not-compile-with-apple-llvm-10-0) using `std::get_if` and some boilerplate error handling code, but again I'm not versed in C++ and am unsure of the real amount of work to ensure compatibility.

Thank you for your work on this library, its inclusion in OpenSCAD has made my workflow much faster and easier :)

pca006132:
Fix #576

@kahays can you verify if this works for you now?

### DIFF
--- a/bindings/python/manifold3d.cpp
+++ b/bindings/python/manifold3d.cpp
@@ -322,8 +322,7 @@ NB_MODULE(manifold3d, m) {
              std::optional<nb::ndarray<uint32_t, nb::shape<3>>> &normalIdx) {
             glm::ivec3 v(0);
             if (normalIdx.has_value()) {
-              if (normalIdx->ndim() != 1 ||
-                  normalIdx->shape(0) != 3)
+              if (normalIdx->ndim() != 1 || normalIdx->shape(0) != 3)
                 throw std::runtime_error("Invalid vector shape, expected (3)");
               auto value = *normalIdx;
               v = glm::ivec3(value(0), value(1), value(2));
--- a/bindings/python/manifold3d.cpp
+++ b/bindings/python/manifold3d.cpp
@@ -322,10 +322,10 @@ NB_MODULE(manifold3d, m) {
              std::optional<nb::ndarray<uint32_t, nb::shape<3>>> &normalIdx) {
             glm::ivec3 v(0);
             if (normalIdx.has_value()) {
-              if (normalIdx.value().ndim() != 1 ||
-                  normalIdx.value().shape(0) != 3)
+              if (normalIdx->ndim() != 1 ||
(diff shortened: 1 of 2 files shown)


kahays:
Thank you, this solves the issues with `csg_tree.cpp`!

Looks like a few more issues cropped up in `bindings/python/manifold3d.cpp`, all of the form `error: call to unavailable member function 'value': introduced in macOS 10.14`.

`bindings/python/manifold3d.cpp:325:29: error: call to unavailable member function 'value': introduced in macOS 10.14`
`              if (normalIdx.value().ndim() != 1 ||`
`bindings/python/manifold3d.cpp:326:29: error: call to unavailable member function 'value': introduced in macOS 10.14`
`                  normalIdx.value().shape(0) != 3)`
`bindings/python/manifold3d.cpp:328:38: error: call to unavailable member function 'value': introduced in macOS 10.14`
`              auto value = normalIdx.value();`
`bindings/python/manifold3d.cpp:627:49: error: call to unavailable member function 'value': introduced in macOS 10.14`
`              auto runTransform1 = runTransform.value();`
`bindings/python/manifold3d.cpp:640:55: error: call to unavailable member function 'value': introduced in macOS 10.14`
`              auto halfedgeTangent1 = halfedgeTangent.value();`

Not sure if it preserves semantics appropriately but I changed each `variable.value().method()` invocation with `(*variable).method()`, which allows it to compile.

elalish (approved):
Thanks!

pca006132:
@kahays updated now, please check if it works. Also it is annoying that we cannot specify OSX 10.13 in github action because they don't plan to support such an old version.

kahays:
Thank you, this fixes all the compilation issues! Tests pass just fine too. I really appreciate all your help getting this running on my ancient machine.

Issue #579, opened 2023-10-25
Wrong vertex normals and uvs after boolean subtraction


jirihon:
I have two manifolds `a` and `b`. Both have 8 vertex properties - 3 for coordinates, 3 for normals, 2 for uvs).

Manifold `a` is a result of previous boolean operations and you can see it in the following picture. One of its surfaces has a UV debug texture assigned. The short blue segments pointing out from the textured surface shows vertex normals for all the vertices that are involved in any triangle belonging to this surface. 

![image](https://github.com/elalish/manifold/assets/16903172/03d05cf0-29f1-4356-8d35-5a8d54db274f)

Manifold `b` is a simple cuboid larger then manifold `a` and translated down a little bit. Here is a picture of both manifold `a` and `b` in one scene.

![image](https://github.com/elalish/manifold/assets/16903172/6624d49c-5681-4238-9cb3-d2b6d17c9295)

When `b` is subtracted from `a` (manifold `a_minus_b`), something bad happens with the normals and uvs on the textured surface. See the picture below.

![image](https://github.com/elalish/manifold/assets/16903172/4b011d94-5c20-49c4-b7db-7162d014def8)

The problematic surface is composed of three triangles. The problem is that two of the triangles uses a vertex which has normal and uv that would be good for the left adjacent surface, but not for this.

The problem occurs in specific situations only. For example, if I move the manifold `b` down by a different offset, the problem disappears. It seems to me that offsets which don't have accurate numeric representation (like multiples of 1/3) triggers this more often.

In case you would have time to look into this, I attach a MeshGL dump of all the manifolds - `a`, `b` and `a_minus_b`. The problematic triangles are 40 and 46. They use vertex 50 which has inappropriate normal and uv. I would expect that these triangles will use vertex 68 instead as it is used in the only correct triangle 38 for this surface.

[wrong_normals_and_uvs_mesh_dump.txt](https://github.com/elalish/manifold/files/13167961/wrong_normals_and_uvs_mesh_dump.txt)

jirihon:
Here is a new mesh dump which shows the same problem for the current master.

[wrong_normals_and_uvs_mesh_dump_2.txt](https://github.com/elalish/manifold/files/13168335/wrong_normals_and_uvs_mesh_dump_2.txt)

The problematic triangles here are 17 and 28 which use vertex 8 with inappropriate normal and uv.

[...]

PR #581, opened 2023-10-27
Add test case for wrong normals and uvs after boolean subtraction

jirihon:
Related to #579.

I know the test case is quite large, but I don't observe the problem for simpler geometry, at least for now.

The `RelatedGL` assertion fails here. I hope I used it correctly.


elalish (commented):
Thank you, this is very helpful! Would you mind following this example [`TEST`](https://github.com/elalish/manifold/blob/master/test/samples_test.cpp#L258-L269) to load the model as a GLB instead?

elalish:
Actually, making a GLB in this case is probably more trouble than it's worth. However, upon debugging your test, I found that roughly half of your normals for A are inverted. Can you fix that and update the test? I'm also curious if that will solve the problem anyway.

jirihon:
Thanks for looking into this. The normals should be fixed now and the test is still failing.

elalish:
closed in favor of #587. Thanks again for the test case!

PR #582, opened 2023-10-30
fix ubuntu build

pca006132:
Managed to reproduce #580. It seems that with the CMake version in Ubuntu 22.04, `FetchContent_Declare` without `GIT_SHALLOW TRUE` and `GIT_PROGRESS TRUE` generates incorrect cmake file which attempts to clone the repository in directory '' and failed.

@jirihon can you check if it works for you now? I tried running it inside a docker and it seems to work fine.


jirihon:
@pca006132 It works! Thanks a lot.

PR #587, opened 2023-11-01
Fix properties

### LINKED ISSUE #579: Wrong vertex normals and uvs after boolean subtraction

elalish:
Follow-up on #581
Fixes #579 

Third time's the charm! Of course I had to write a hideous amount of code and get nowhere to finally realize the simple solution. This even allowed me to simplify our code a bit, since I removed a previous convoluted bit I had put in trying to fix related issues.


src/manifold/src/boolean_result.cpp
```diff
@@ -567,7 +567,7 @@ void CreateProperties(Manifold::Impl &outR, const Vec<TriRef> &refPQ,
           }
           if (uvw[j] == 0) edge = j;
         }
-        if (edge >= 0) {
+        if (edge >= 0) {  // This misses some edges
```
elalish, line 570: This is the root cause of the problem: when an edge is collapsed (before this step) you can end up with verts that refer (through refPQ) to a triangle that they are now outside of, since they were merged with their neighboring triangle but the references were not updated. This would be fine, except here I'm assuming I can tell when a vert is on a retained edge by checking its barycentric coordinates - but I miss these ones. So I end up assigning a key with only the `meshID` and `vert`, which isn't enough when the other side of the retained edge might have different properties.

src/manifold/src/boolean_result.cpp
```diff
@@ -772,6 +772,7 @@ Manifold::Impl Boolean3::Result(OpType op) const {
 
   Vec<TriRef> refPQ = UpdateReference(outR, inP_, inQ_, invertQ);
 
+  // This causes some verts to be outside of their refPQ triangle.
   outR.SimplifyTopology();
```
elalish, line 776: I tried just reversing the order - create the properties first, and then simplify the topology. However, in this case it's very difficult to keep from duplicating prop verts (CoplanarProp test does a good job catching this). This might not seem like a huge deal, but they tend to continue to multiply, until your final output going to the renderer is basically triangle soup, even if the geometry is manifold.

test/test_main.cpp
```diff
@@ -396,7 +396,7 @@ void RelatedGL(const Manifold& out, const std::vector<MeshGL>& originals,
             const float volumeP =
                 glm::dot(edgesP[0], glm::cross(edgesP[1], edgesP[2]));
 
-            ASSERT_LE(volumeP, area * 100 * out.Precision());
+            ASSERT_LE(volumeP, area * out.Precision());
```
elalish, line 399: Our regular tests would have caught this bug if I hadn't put in this *100, which has the look of me just getting frustrated at some point.

pca006132: interesting, I think we should probably go over all tests to see if there are suspicious ones

src/manifold/src/boolean_result.cpp
```diff
@@ -541,23 +540,22 @@ void CreateProperties(Manifold::Impl &outR, const Vec<TriRef> &refPQ,
...
+                                : PQ ? inP.meshRelation_.triProperties[ref.tri]
+                                     : inQ.meshRelation_.triProperties[ref.tri];
 
     for (const int i : {0, 1, 2}) {
       const int vert = outR.halfedge_[3 * tri + i].startVert;
       const glm::vec3 &uvw = bary[3 * tri + i];
 
       glm::ivec4 key(PQ, idMissProp, -1, -1);
       if (oldNumProp > 0) {
-        key[1] = vert;
```
elalish, line 559: This was the key fix: for anything close to a retained vert, only key on the propVert, since the separate verts are close together and will get collapsed anyway - this was causing us to duplicate propVerts unnecessarily.

src/manifold/src/boolean_result.cpp
```diff
@@ -770,11 +771,11 @@ Manifold::Impl Boolean3::Result(OpType op) const {
...
     ASSERT(outR.IsManifold(), logicErr, "triangulated mesh is not manifold!");
 
-  Vec<TriRef> refPQ = UpdateReference(outR, inP_, inQ_, invertQ);
+  CreateProperties(outR, inP_, inQ_);
 
-  outR.SimplifyTopology();
+  UpdateReference(outR, inP_, inQ_, invertQ);
 
-  CreateProperties(outR, refPQ, inP_, inQ_);
+  outR.SimplifyTopology();
```
elalish, line 778: No need for `refPQ` anymore since now everything happens in a logical order.

Issue #592, opened 2023-11-06
Python binding to Manifold.Split() does not work


rsaccon:
I am using latest release and tried out the Python bindings. All works great so far except the binding to `Manifold.Split`

the signature of the original function (looked up in documentation) is:
```
std::pair< Manifold, Manifold > Split ( const Manifold &  cutter ) const
```

the Python binding definition however looks incomplete to me:
```
.def("split", &Manifold::Split,
           "Split cuts this manifold in two using the cutter manifold. The "
           "first result is the intersection, second is the difference. This "
           "is more efficient than doing them separately.")
```

zalo:
Ah yeah, nanobind doesn't like that it returns an std::pair...

It would have to be replaced with something like:
```c++
      .def("split", [](Manifold &self, Manifold& cutter) {
            auto p = self.Split(cutter);
            return nb::make_tuple(p.first, p.second);
           },
           nb::arg("cutter"),
           "Split cuts this manifold in two using the cutter manifold. The "
           "first result is the intersection, second is the difference. This "
           "is more efficient than doing them separately."
           "\n\n"
           ":param cutter: This is the manifold to cut by.")
```

rsaccon:
I also tried to build locally on my old intel mac, finally was able to build, but my attempt to use the local built bindings (the not yet modified ones) failed, so many things can go wrong, well, I guess I have to wait for the next pipy release for this Split() binding to work ...

zalo:
If a version of this PR https://github.com/elalish/manifold/pull/593 works, you might be able to extract a working version of the Python wheel from the Github Actions build action.

rsaccon:
I just managed to get my local build working, tested the modified binding, @zalo `Split` works now as expected, thanks.

PR #593, opened 2023-11-07
[FIX] Packaged Split() output to Tuple

zalo:
This PR attempts to resolve https://github.com/elalish/manifold/issues/592 but is untested (due to Thrust related build issues on my end).

Hopefully Github Actions will vindicate me 🤞

### DIFF
--- a/bindings/python/examples/split_cube.py
+++ b/bindings/python/examples/split_cube.py
@@ -1,0 +1,22 @@
+"""
+ Copyright 2022 The Manifold Authors.
+
+ Licensed under the Apache License, Version 2.0 (the "License");
+ you may not use this file except in compliance with the License.
+ You may obtain a copy of the License at
+
+      https://www.apache.org/licenses/LICENSE-2.0
+
+ Unless required by applicable law or agreed to in writing, software
+ distributed under the License is distributed on an "AS IS" BASIS,
+ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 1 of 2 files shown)


pca006132:
Thanks! Maybe you can add a python test example that uses `Split()` as well?

zalo:
> Thanks! Maybe you can add a python test example that uses `Split()` as well?

Just added an example that should work.   On the current public Windows wheel, this:
```python
from manifold3d import Manifold
def run():
    a = Manifold.cube(1.0, 1.0, 1.0)
    b = Manifold.cube(1.0, 1.0, 1.0).rotate(45.0, 45.0, 45.0)
    return a.split(b)[0]
```

Yields:
```python
Traceback (most recent call last):
  File "c:\Repositories\manifold\bindings\python\examples\run_all.py", line 35, in <module>
    model = module.run()
  File "c:\Repositories\manifold\bindings\python\examples\split_cube.py", line 22, in run
    return a.split(b)[0]
TypeError: Unable to convert function return value to a Python type! The signature was
    split(self, arg: manifold3d.Manifold, /) -> std::pair<manifold::Manifold,manifold::Manifold>
```

Does the Github Action run `run_all.py`?

pca006132:
> Does the Github Action run `run_all.py`?

yes it does

elalish:
Excellent, thank you! We should probably put out a point release with this fix, but let's wait a couple days to see if any more bugs come in first.

PR #597, opened 2023-11-09
[FEAT] Add Batch Boolean to Python Bindings

zalo:
This PR adds the Batch Boolean function to the Python Bindings...

But I'm not sure if there is a speed improvement; is there any benefit to doing this, or are the manifolds already lazily evaluated when converted to meshes?  If they are, then this PR can be safely closed and ignored.

### DIFF
--- a/bindings/python/examples/swiss_cheese.py
+++ b/bindings/python/examples/swiss_cheese.py
@@ -1,0 +1,37 @@
+"""
+ Copyright 2022 The Manifold Authors.
+
+ Licensed under the Apache License, Version 2.0 (the "License");
+ you may not use this file except in compliance with the License.
+ You may obtain a copy of the License at
+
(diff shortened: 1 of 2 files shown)


zalo:
Speed through Mesh baking appears to be perfectly identical, so this API is not necessarily necessary...
> python3.10-manifold3d> Individual cheese: 541.5008080000234 ms
python3.10-manifold3d> Batch cheese: 542.6143089999869 ms
python3.10-manifold3d> Took 1156.7ms for swiss_cheese

Closing tomorrow unless anyone can think of a reason to keep it in...

pca006132:
Yeah I think batch boolean is exposed in C++ because it is simple to do, and exposed in JS because of reference counting issues (#220). I don't feel the need to have such API for python, as we already have lazy evaluation and optimizations to flatten the tree.

PR #598, opened 2023-11-09
[FEAT] Add LevelSet Batch Function to Python Bindings

zalo:
This PR expands on https://github.com/elalish/manifold/pull/616 by adding a batch function that allows client libraries to use their own threading to compute the level set bindings.

### DIFF
--- a/bindings/python/examples/levelset_sdf.py
+++ b/bindings/python/examples/levelset_sdf.py
@@ -24,8 +24,8 @@ def sphere_implicit(x, y, z):
     sphere2 = 0.5 - math.sqrt((x*x) + (y*y) + (z*z))
     return min(sphere1, -sphere2)
 
-def sphere_implicit_batched(coords):
-    coords = coords.copy()
+def sphere_implicit_batched(coords_readonly):
+    coords = coords_readonly.copy() # Copy read-only coords to modify in-place
     x = coords[:, 0]; y = coords[:, 1]; z = coords[:, 2]
     sphere1 = 0.5 - np.sqrt((x*x) + (y*y) + (z*z))
     x += 0.25; y += 0.25; z -= 0.25
--- a/bindings/python/examples/levelset_sdf.py
+++ b/bindings/python/examples/levelset_sdf.py
@@ -1,0 +1,41 @@
+"""
+ Copyright 2023 The Manifold Authors.
+
+ Licensed under the Apache License, Version 2.0 (the "License");
+ you may not use this file except in compliance with the License.
+ You may obtain a copy of the License at
+
+      https://www.apache.org/licenses/LICENSE-2.0
+
+ Unless required by applicable law or agreed to in writing, software
+ distributed under the License is distributed on an "AS IS" BASIS,
+ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ See the License for the specific language governing permissions and
+ limitations under the License.
+ """
+
+import math
+import numpy as np
+from manifold3d import Manifold, Mesh
+
+def sphere_implicit(x, y, z):
+    sphere1 = 0.5 - math.sqrt((x*x) + (y*y) + (z*z))
+    x += 0.25; y += 0.25; z -= 0.25
+    sphere2 = 0.5 - math.sqrt((x*x) + (y*y) + (z*z))
+    return min(sphere1, -sphere2)
+
+def sphere_implicit_batched(coords):
+    coords = coords.copy()
+    x = coords[:, 0]; y = coords[:, 1]; z = coords[:, 2]
+    sphere1 = 0.5 - np.sqrt((x*x) + (y*y) + (z*z))
+    x += 0.25; y += 0.25; z -= 0.25
+    sphere2 = 0.5 - np.sqrt((x*x) + (y*y) + (z*z))
+    return np.minimum(sphere1, -sphere2)
+
+def run():
+    levelset_mesh_slow = Mesh.levelset      (sphere_implicit,         [-0.5, -0.5, -0.5, 0.5, 0.5, 0.5], 0.05, 0.0)
+    levelset_mesh_fast = Mesh.levelset_batch(sphere_implicit_batched, [-0.5, -0.5, -0.5, 0.5, 0.5, 0.5], 0.05, 0.0, False)
+    model_slow = Manifold.from_mesh(levelset_mesh_slow)
+    model_fast = Manifold.from_mesh(levelset_mesh_fast)
+    assert abs(model_slow.get_volume() - model_fast.get_volume()) < 0.0000001
+    return model_slow
--- a/bindings/python/manifold3d.cpp
+++ b/bindings/python/manifold3d.cpp
@@ -580,15 +580,16 @@ NB_MODULE(manifold3d, m) {
           [](const std::function<float(Float3)> &f, std::vector<float> bounds,
              float edgeLength, float level = 0.0, bool canParallel = false) {
             // Same format as Manifold.bounding_box
-            Box bound = { glm::vec3(bounds[0], bounds[1], bounds[2]),
-                          glm::vec3(bounds[3], bounds[4], bounds[5]) };
+            Box bound = {glm::vec3(bounds[0], bounds[1], bounds[2]),
+                         glm::vec3(bounds[3], bounds[4], bounds[5])};
             Mesh result = LevelSet(
-                [&f](glm::vec3 &v) { 
-                    return f(std::make_tuple(v.x, v.y, v.z)); },
+                [&f](glm::vec3 &v) {
+                    return f(std::make_tuple(v.x, v.y, v.z));
+                },
                 bound, edgeLength, level, canParallel);
             return Manifold::Manifold(result);
           },
-          nb::arg("f"), nb::arg("bounds"), nb::arg("edgeLength"), 
+          nb::arg("f"), nb::arg("bounds"), nb::arg("edgeLength"),
           nb::arg("level") = 0.0, nb::arg("canParallel") = false,
           "Constructs a level-set Mesh from the input Signed-Distance Function "
           "(SDF) This uses a form of Marching Tetrahedra (akin to Marching "
--- a/bindings/python/manifold3d.cpp
+++ b/bindings/python/manifold3d.cpp
@@ -17,6 +17,7 @@
 
 #include "cross_section.h"
 #include "manifold.h"
+#include "sdf.h"
 #include "nanobind/nanobind.h"
 #include "nanobind/ndarray.h"
 #include "nanobind/operators.h"
@@ -574,6 +575,45 @@ NB_MODULE(manifold3d, m) {
           "four, as this sphere is constructed by refining an octahedron. This "
           "means there are a circle of vertices on all three of the axis "
           "planes. Default is calculated by the static Defaults.")
+      .def_static(
+          "levelset",
+          [](const std::function<float(Float3)> &f, std::vector<float> bounds,
+             float edgeLength, float level = 0.0, bool canParallel = false) {
+            // Same format as Manifold.bounding_box
+            Box bound = { glm::vec3(bounds[0], bounds[1], bounds[2]),
+                          glm::vec3(bounds[3], bounds[4], bounds[5]) };
+            Mesh result = LevelSet(
+                [&f](glm::vec3 &v) { 
+                    return f(std::make_tuple(v.x, v.y, v.z)); },
+                bound, edgeLength, level, canParallel);
+            return Manifold::Manifold(result);
+          },
+          nb::arg("f"), nb::arg("bounds"), nb::arg("edgeLength"), 
+          nb::arg("level") = 0.0, nb::arg("canParallel") = false,
+          "Constructs a level-set Mesh from the input Signed-Distance Function "
+          "(SDF) This uses a form of Marching Tetrahedra (akin to Marching "
+          "Cubes, but better for manifoldness). Instead of using a cubic grid, "
+          "it uses a body-centered cubic grid (two shifted cubic grids). This "
+          "means if your function's interior exceeds the given bounds, you "
+          "will see a kind of egg-crate shape closing off the manifold, which "
+          "is due to the underlying grid."
+          "\n\n"
+          ":param f: The signed-distance functor, containing this function "
+          "signature: `def sdf(xyz : tuple) -> float:`, which returns the "
+          "signed distance of a given point in R^3. Positive values are "
+          "inside, negative outside."
+          ":param bounds: An axis-aligned box that defines the extent of the "
+          "grid."
+          ":param edgeLength: Approximate maximum edge length of the triangles "
+          "in the final result.This affects grid spacing, and hence has a "
+          "strong effect on performance."
+          ":param level: You can inset your Mesh by using a positive value, or "
+          "outset it with a negative value."
+          ":param canParallel: Parallel policies violate will crash language "
+          "runtimes with runtime locks that expect to not be called back by "
+          "unregistered threads.This allows bindings use LevelSet despite "
+          "being compiled with MANIFOLD_PAR active."
+          ":return Manifold: This is guaranteed to be manifold.")
       .def_static("reserve_ids", Manifold::ReserveIDs, nb::arg("n"),
                   "Returns the first of n sequential new unique mesh IDs for "
                   "marking sets of triangles that can be looked up after "
--- a/bindings/python/manifold3d.cpp
+++ b/bindings/python/manifold3d.cpp
@@ -616,7 +616,9 @@ NB_MODULE(manifold3d, m) {
           ":return Manifold: This is guaranteed to be manifold.")
       .def_static(
           "levelset_batch",
-          [](const std::function<std::vector<float>(std::vector<float>)> &f,
+          [](const std::function<
+                 nb::ndarray<nb::numpy, const float, nb::c_contig>(
+                 nb::ndarray<nb::numpy, const float, nb::c_contig>)> &f,
              std::vector<float> bounds, float edgeLength, float level = 0.0,
              bool canParallel = false) {
             // Same format as Manifold.bounding_box
@@ -625,13 +627,11 @@ NB_MODULE(manifold3d, m) {
 
             Mesh result = LevelSetBatch(
                 [&f](std::vector<glm::vec3> &v) {
-                  std::vector<float> coords(v.size()*3);
-                  for (int i = 0; i < v.size(); i++) {
-                    coords[(i*3) + 0] = v[i].x;
-                    coords[(i*3) + 1] = v[i].y;
-                    coords[(i*3) + 2] = v[i].z;
-                  }
-                  return f(coords);
+                  size_t coordsShape[2] = {v.size(), 3};
+                  nb::ndarray<nb::numpy, const float, nb::c_contig> d =
+                      f(nb::ndarray<nb::numpy, const float, nb::c_contig>(
+                          v.data(), 2, coordsShape));
+                  return std::vector<float>(d.data(), d.data() + v.size());
                 },
                 bound, edgeLength, level, canParallel);
             return Manifold(result);
@@ -639,13 +639,12 @@ NB_MODULE(manifold3d, m) {
           nb::arg("f"), nb::arg("bounds"), nb::arg("edgeLength"),
           nb::arg("level") = 0.0, nb::arg("canParallel") = false,
           "Similar to LevelSet, this constructs a level-set Mesh from the input"
-          "Signed-Distance Function (SDF). This variant feeds an std:vector of points"
-          "to the function, allowing the function to compute signed distances using"
-          ":param f: The signed-distance functor, containing this function "
-          "its own threading/parallelization paradigm."
+          "Signed-Distance Function (SDF). This variant feeds a [N,3] ndarray"
+          "of points to the function, allowing the function to compute signed"
+          "distances using its own threading/parallelization paradigm."
           "\n\n"
           ":param f: The signed-distance functor, containing this function "
-          "signature: `def sdf(xyz : tuple) -> float:`, which returns the "
+          "signature: `def sdf(xyz : ndarray) -> ndarray:`, which returns the "
           "signed distance of a given point in R^3. Positive values are "
           "inside, negative outside."
           ":param bounds: An axis-aligned box that defines the extent of the "
--- a/bindings/python/manifold3d.cpp
+++ b/bindings/python/manifold3d.cpp
@@ -578,8 +578,8 @@ NB_MODULE(manifold3d, m) {
       .def_static(
           "levelset",
           [](const std::function<float(float, float, float)> &f,
-              std::vector<float> bounds, float edgeLength, float level = 0.0,
-              bool canParallel = false) {
+             std::vector<float> bounds, float edgeLength, float level = 0.0,
+             bool canParallel = false) {
             // Same format as Manifold.bounding_box
             Box bound = {glm::vec3(bounds[0], bounds[1], bounds[2]),
                          glm::vec3(bounds[3], bounds[4], bounds[5])};
--- a/bindings/python/manifold3d.cpp
+++ b/bindings/python/manifold3d.cpp
@@ -577,15 +577,13 @@ NB_MODULE(manifold3d, m) {
           "planes. Default is calculated by the static Defaults.")
       .def_static(
           "levelset",
-          [](const std::function<float(Float3)> &f, std::vector<float> bounds,
+          [](const std::function<float(float, float, float)> &f, std::vector<float> bounds,
              float edgeLength, float level = 0.0, bool canParallel = false) {
             // Same format as Manifold.bounding_box
             Box bound = {glm::vec3(bounds[0], bounds[1], bounds[2]),
                          glm::vec3(bounds[3], bounds[4], bounds[5])};
             Mesh result = LevelSet(
-                [&f](glm::vec3 &v) {
-                  return f(std::make_tuple(v.x, v.y, v.z));
-                },
+                [&f](glm::vec3 &v) {return f(v.x, v.y, v.z);},
                 bound, edgeLength, level, canParallel);
             return Manifold(result);
           },
--- a/bindings/python/manifold3d.cpp
+++ b/bindings/python/manifold3d.cpp
@@ -17,7 +17,6 @@
 
 #include "cross_section.h"
 #include "manifold.h"
-#include "sdf.h"
 #include "nanobind/nanobind.h"
 #include "nanobind/ndarray.h"
 #include "nanobind/operators.h"
@@ -25,6 +24,7 @@
 #include "nanobind/stl/optional.h"
 #include "nanobind/stl/tuple.h"
 #include "nanobind/stl/vector.h"
+#include "sdf.h"
 
 template <>
 struct nanobind::detail::type_caster<glm::vec3> {
@@ -584,7 +584,7 @@ NB_MODULE(manifold3d, m) {
                          glm::vec3(bounds[3], bounds[4], bounds[5])};
             Mesh result = LevelSet(
                 [&f](glm::vec3 &v) {
-                    return f(std::make_tuple(v.x, v.y, v.z));
+                  return f(std::make_tuple(v.x, v.y, v.z));
                 },
                 bound, edgeLength, level, canParallel);
             return Manifold::Manifold(result);
--- a/bindings/python/manifold3d.cpp
+++ b/bindings/python/manifold3d.cpp
@@ -587,7 +587,7 @@ NB_MODULE(manifold3d, m) {
                   return f(std::make_tuple(v.x, v.y, v.z));
                 },
                 bound, edgeLength, level, canParallel);
-            return Manifold::Manifold(result);
+            return Manifold(result);
           },
           nb::arg("f"), nb::arg("bounds"), nb::arg("edgeLength"),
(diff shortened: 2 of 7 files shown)


zalo:
Strange that it builds on Windows and not Linux...

pca006132:
Hmmm interesting. The way I think about implementing SDF for python is to have an expression tree interface, similar to that provided by libfive, because we cannot easily perform parallel evaluation for Python using callbacks due to GIL.

pca006132:
and the expression API will be related to some future optimization as well, for context you can have a look at the (rather long) discussion in #561.

zalo:
I’ve gone hard down that path before (even wrote a big Unity plugin for libfive: https://github.com/zalo/libfive-unity ), but my ultimate conclusion is that it’s severely limited by the expression tree.

It makes it difficult to mesh volumetric textures, or fill holes in meshes, or do implicit evaluations that use for-loops, etc.

If manifold can batch sample an [N, 3] or [3, N] list of query points, then NumPy can do array-parallel evaluation of them and get close to native speeds.

[PyMCubes] (https://github.com/pmneila/PyMCubes), for example, accepts regular grid evaluations of implicit functions, but I haven’t checked if the tetrahedral grid is a regular grid.   If it’s not, or if the algorithm needs to iteratively sample the implicit, then it makes sense to have C++ iteratively call the delegate with sample points instead.

pca006132:
We can support grid evaluation, but we probably need to modify the sdf code and let the user pass the grid inside. I feel that this will not be python specific though.

zalo:
So we’d probably want a “levelset_batch” version of the function, which calls f with an std::vector of points, and expects an std::vector of distances? I’d handle wrapping everything in NumPy arrays with nanobind just for Python.

Something similar probably for Javascript; I bet the JS JIT would also appreciate not having to jump the boundary too often.

I’ll look into how hard that is today.

zalo:
Ah… the grid is calculated implicitly and the gridVerts are stored sparsely in MortonCodes to save on memory, aren’t they…

Rather than serialize the whole grid to coordinates at once, I see an explicit Memory<->Compute Overhead tradeoff that can be tuned.   LevelSet_Batch should take in a ChunkSize arg that tunes how many grid indices it sends over with each evaluation.   We can then start to graph what the optimal speed<->memory usage tradeoff is.

This way we still get the advantages of vectorized function evaluation without destroying the RAM.

elalish:
I think you're on the right track, and I'm glad to hear you have experience with SDF systems. You've probably already realized this, but indeed the tetrahedral grid is regular: it's two sets of cubic lattice points offset from each other. Probably best to keep the grid implicit.

zalo:
> I think you're on the right track, and I'm glad to hear you have experience with SDF systems. You've probably already realized this, but indeed the tetrahedral grid is regular: it's two sets of cubic lattice points offset from each other. Probably best to keep the grid implicit.

I hear that, but I wonder if calculating the bi-grid in Morton order can be better vectorized in Python/Javascript, vs just precalculating those coordinate chunks in C++ and throwing the contiguous memory over the boundary…

Likewise, if the grid were to ever change, then implicitly calculating the grid in the client bindings would have to be redone as well…

I think incrementally explicitifying it can give a solid tradeoff…

I’m currently noodling over how to avoid recalculating the neighbor grid cells…. They overlap quite a lot, dont they?   Aren’t you visiting each grid point’s distance 8 times with this?
https://github.com/elalish/manifold/blob/b439603eb912014b7813403ffdf7ff9d891c3c18/src/sdf/src/sdf.cpp#L169

Not to mention that, since it’s Morton order, it’s frequently going to recalculate the same neighbor coordinates in consecutive grid samplings…

I suppose this might be faster when the SDF is cheap, but when it’s expensive, it might be faster to figure out some way of caching it…

pca006132:
yes, our sdf code is not very fast tbh and we do not have very complicated sdf test cases

btw maybe we should move the discussion to #561? feels like this is not something that can be done in 1 PR

zalo:
I don’t want to pollute that thread, and I’m using this one as a bit of a scratchpad 😅
My goal is also a little different than the direction that thread took; there might still be a future where passing the AST over for interval arithmetic evaluation makes sense, but my goal is to make arbitrary-code execution viable for my language binding 😅

I think I have a handle on how to avoid most of the repeat evaluations.   The batch of query points sent over will be the Morton Start -> End indices of a chunk, and then all of the neighbors (that fall outside of that interval) concatenated to the end (without deduplication).  Shouldn’t need a HashSet to keep track of that 💪

Since the morton code part is cheap (in C++), I’ll run that traversal logic once to populate the batch of coordinates, and again to properly consume the distances.

This shouldn’t look like too large of an addition to the existing sdf.cpp file (maybe 100 dense lines I hope?), so I don’t think it’s greater than one PR.

I’ll just change the scope of this PR to be for a Batch API to the sdf-subsystem (which Python will be one beneficiary of).

elalish:
That sounds like a reasonable approach to me; I look forward to reviewing it!

zalo:
The batched one will take a bit to troubleshoot, but the non-batched function appears to be pretty serviceable.

```Manifold Called sphere_implicit: 1,185,408 times in 397.50860002823174 ms```

```python
import math
import trimesh
import manifold3d
import numpy as np
from time import perf_counter

def sphere_implicit(x, y, z):
    global count
    count += 1
    return 0.5 - math.sqrt((x*x) + (y*y) + (z*z))

if __name__ == "__main__":
    t0 = perf_counter(); count = 0
    model = manifold3d.Manifold.levelset(sphere_implicit, [-0.5, -0.5, -0.5, 0.5, 0.5, 0.5], 0.025, 0.0)
    t1 = perf_counter()
    print("Manifold Called sphere_implicit:", count, "times in", (t1-t0)*1000, "ms")
```

image

zalo:
The batched function leads to a 2-6x speed increase in Python with NumPy:
```
Manifold Called sphere_implicit 1,185,408 times in 653.1 ms
Manifold Called batched_sphere_implicit once for 465,936 evaluations in 233.8 ms
```
The more complex the evaluation (and the more evaluations), the more the batching tends to shine, like with the resolution increased >2.5x on each axis in the sample below:
```
Manifold Called sphere_implicit 16,979,328 times in 7996.4473999571055 ms
Manifold Called batched_sphere_implicit once for 4,129,680 evaluations in 1235.2377999923192 ms
```

Here's the evaluation script:
```python
def sphere_implicit(x, y, z):
    sphere1 = 0.5 - math.sqrt((x*x) + (y*y) + (z*z))
    x += 0.25; y += 0.25; z -= 0.25
    sphere2 = 0.5 - math.sqrt((x*x) + (y*y) + (z*z))
    return min(sphere1, -sphere2)

def sphere_implicit_batched(coords):
    coords = coords.copy() # coords is read-only, so it needs to be copied to be modified
    x = coords[:, 0]; y = coords[:, 1]; z = coords[:, 2]
    sphere1 = 0.5 - np.sqrt((x*x) + (y*y) + (z*z))
    x += 0.25; y += 0.25; z -= 0.25 # modified for convenience's sake
    sphere2 = 0.5 - np.sqrt((x*x) + (y*y) + (z*z))
    return np.minimum(sphere1, -sphere2)

def run():
    model_slow = manifold3d.Manifold.levelset      (sphere_implicit,         [-0.5, -0.5, -0.5, 0.5, 0.5, 0.5], 0.025, 0.0)
    model_fast = manifold3d.Manifold.levelset_batch(sphere_implicit_batched, [-0.5, -0.5, -0.5, 0.5, 0.5, 0.5], 0.025, 0.0, False)
    return model_slow, model_fast
```

image

The number of batched evaluations should reduce by roughly half again once I figure out how to discard certain morton samples outside of the grid bounds via this check ( `(glm::any(glm::greaterThan(glm::ivec3(gridIndex), gridSize)))` ) again 😅 
(The indexing bookkeeping is easier without it)

I feel less compelled to evaluate the distances in chunks; `465,936` evaluations is only 6-7 megabytes of RAM with full Float32 coordinates and distances, and `4,129,680` evaluations is only ~64mb.    So once this simple "all-at-once" implementation is done and solid, that'll probably be good enough.

elalish:
Excellent work! From your Github I see you have a lot of experience with geometry and CAD. Might I ask what plans you have for manifold? I'd love to see more tools using SDFs for design.

zalo:
I'm a huge fan of OpenSCAD (my first exposure to modelling), so I've been trying different things out to see what the next generation of SCAD might look like...  The raw Manifold API on top of Python is, in many ways, cleaner than OpenSCAD itself (with the addition of NumPy and related libraries) and far faster (than the original), so now I'm a huge fan of Manifold 😁 .

Turbocharging OpenSCAD-style CSG-based CAD with SDFs could open up a lot of unique opportunities.   The Disney Research [AutoConnect](https://la.disneyresearch.com/publication/autoconnect/) paper hints at the power of such a fusion.    And I'm thinking that some of the new generative 3D modelling approaches could turbocharge this twice-over.  Implicits alone are a form of "content-aware-fill" that can bridge gaps and fill volumes with lattices, but with data-driven priors... they're like neuromorphic clay that can latch onto and mold itself to suit its environment.

So having a fast and efficient way of experimenting with manifold implicit construction is valuable to me :-D


(Also, I've found the bug that was causing the batch version to crash and I've switched it out to returning a Mesh instead of a Manifold... now to see what the deal is with Linux and Mac, and maybe what it'll take to add Emscripten bindings...)

elalish:
It sounds like we've had very similar paths - of course my "next SCAD" is manifoldCAD.org, mostly because personally I prefer JS to Python. Still Python seems to be the language of ML these days and it does have a lot of great math libraries, hence why I'm excited to have Python bindings now. 

I just read about [MWN](https://www.gradientspace.com/tutorials/2018/9/14/point-set-fast-winding) - very cool! We already have a parallel AABB BVH in our Collider; it seems like this could be an ideal follow-on to `Merge` that would allow us to quickly turn any non-manifold mesh into a `Manifold`. That would be huge - our biggest blocker right now is that most potential input data is not manifold.

zalo:
Heh, you've probably already seen [CascadeStudio](https://github.com/zalo/CascadeStudio), [LeapShape](https://github.com/leapmotion/LeapShape), and [libfive-unity](https://github.com/zalo/libfive-unity) on my profile.  

One thing I've just learned is that Winding Number Fields aren't continuous like SDFs; rather than linearly interpolating across crossings, you have to use bisection to find the surface.
In that article, Ryan did it with:
```
mc.RootMode = MarchingCubes.RootfindingModes.Bisection;
```

If you don't do it, then "repairing" this mesh:
image

Yields:
image

Whereas Ryan's "Solidify" remeshing implementation in [Cotangent](https://www.cotangent.io/download) yields:
image

zalo:
The whole Suzanne mesh was pretty slow to do with "Slow Winding Numbers", but it does seem like Generalized Winding Number of 3 Triangles is pretty fast:

TriangleWindingOrbit

```Manifold Called batched_triangle_implicit once for 58256 evaluations in 99.257599998964 ms```

```py
import trimesh
import manifold3d
import numpy as np
from time import perf_counter
from tqdm import tqdm

# Implementation from https://github.com/marmakoide/inside-3d-mesh/blob/master/is_inside_mesh.py#L28
def is_inside_turbo(triangles, X):
    # Compute euclidean norm along axis 1
    def anorm2(X):
        return np.sqrt(np.sum(X ** 2, axis = 1))
    # Compute 3x3 determinant along axis 1
    def adet(X, Y, Z):
        ret  = np.multiply(np.multiply(X[:,0], Y[:,1]), Z[:,2])
        ret += np.multiply(np.multiply(Y[:,0], Z[:,1]), X[:,2])
        ret += np.multiply(np.multiply(Z[:,0], X[:,1]), Y[:,2])
        ret -= np.multiply(np.multiply(Z[:,0], Y[:,1]), X[:,2])
        ret -= np.multiply(np.multiply(Y[:,0], X[:,1]), Z[:,2])
        ret -= np.multiply(np.multiply(X[:,0], Z[:,1]), Y[:,2])
        return ret
    # One generalized winding number per input vertex
    ret = np.zeros(X.shape[0], dtype = X.dtype)
    # Accumulate generalized winding number for each triangle
    for U, V, W in tqdm(triangles):
        A, B, C = U - X, V - X, W - X
        omega = adet(A, B, C)
        a, b, c = anorm2(A), anorm2(B), anorm2(C)
        k  = a * b * c 
        k += c * np.sum(np.multiply(A, B), axis = 1)
        k += a * np.sum(np.multiply(B, C), axis = 1)
        k += b * np.sum(np.multiply(C, A), axis = 1)
        ret += np.arctan2(omega, k)
    return ret # Job done

#holey_suzanne : trimesh.Trimesh = trimesh.load_mesh("suzanne_holes.obj")
#bounds = (holey_suzanne.bounds.flatten() * 1.1).tolist()
#def mesh_winding_number_batched(queries):
#    #holey_suzanne : trimesh.Trimesh = trimesh.load_mesh("suzanne_holes.obj")
#    return is_inside_turbo(holey_suzanne.triangles, queries)

bounds = [-0.5, -0.5, -0.5, 0.5, 0.5, 0.5]
def mesh_winding_number_batched(queries):
    global count; count += queries.shape[0]
    tris = np.array([[[ 0.5  ,  0.5,  0.0], # Top Right Tri
                      [ 0.125,  0.0,  0.5], 
                      [ 0.5  , -0.5,  0.0]],
                     [[-0.125,  0.0,  0.5], # Top Left Tri
                      [-0.5  ,  0.5,  0.0],
                      [-0.5  , -0.5,  0.0]],
                     [[ 0.0  ,  0.5, -0.5], # Bottom Tri
                      [ 0.5  , -0.5, -0.5], 
                      [-0.5  , -0.5, -0.5]]])
    return is_inside_turbo(tris, queries) - 1.8

def save_mesh(model, name):
    mesh = model.to_mesh()
    if mesh.vert_properties.shape[1] > 3:
        vertices = mesh.vert_properties[:, :3]
        colors = (mesh.vert_properties[:, 3:] * 255).astype(np.uint8)
    else:
        vertices = mesh.vert_properties
        colors = None
    meshOut = trimesh.Trimesh(vertices=vertices, faces=mesh.tri_verts,
                              vertex_colors=colors)
    trimesh.exchange.export.export_mesh(meshOut, name+'.stl', 'stl')
    print('Exported model to ', name, '.stl')

if __name__ == "__main__":
    t0 = perf_counter(); count = 0
    model = manifold3d.Mesh.levelset_batch(mesh_winding_number_batched, bounds, 0.05, 0.0, False)
    model = manifold3d.Manifold.from_mesh(model)
    save_mesh(model, "batched_triangle_implicit")
    t1 = perf_counter()
    print("Manifold Called batched_triangle_implicit once for", count, "evaluations in", (t1-t0)*1000, "ms")
```

zalo:
Alrighty; I've finally figured out why Linux/Mac/etc. was throwing fits during compilation: had too many `&` decorators for the compiler to feel comfortable casting my lambdas...    I assume passing `glm::vec3`'s by value is fine, and `std::vectors` are glorified pointers anyway(?), so I assume passing those by value is fine?

Either way, it builds and it's ready to for review!

elalish:
Yeah, that's interesting - the MWN is discontinuous where there's already a surface, and better behaved in the regions of holes. Ideally we wouldn't bother remeshing the input portions of the mesh anyway, but just use the SDF to fill in the missing bits. I wonder if there's some way to link up an open mesh and an SDF? Maybe compute the MWN at each input vertex too, perhaps with symbolic perturbation towards the inside?

bindings/python/examples/levelset_sdf.py
```diff
@@ -0,0 +1,41 @@
...
+def sphere_implicit_batched(coords):
+    coords = coords.copy()
+    x = coords[:, 0]; y = coords[:, 1]; z = coords[:, 2]
+    sphere1 = 0.5 - np.sqrt((x*x) + (y*y) + (z*z))
+    x += 0.25; y += 0.25; z -= 0.25
+    sphere2 = 0.5 - np.sqrt((x*x) + (y*y) + (z*z))
+    return np.minimum(sphere1, -sphere2)
+
+def run():
+    levelset_mesh_slow = Mesh.levelset      (sphere_implicit,         [-0.5, -0.5, -0.5, 0.5, 0.5, 0.5], 0.05, 0.0)
```
elalish, line 36: Are we checking formatting for Python? This seems weird, but I don't write a lot of Python...

bindings/python/examples/levelset_sdf.py
```diff
@@ -0,0 +1,41 @@
...
+from manifold3d import Manifold, Mesh
+
+def sphere_implicit(x, y, z):
+    sphere1 = 0.5 - math.sqrt((x*x) + (y*y) + (z*z))
+    x += 0.25; y += 0.25; z -= 0.25
+    sphere2 = 0.5 - math.sqrt((x*x) + (y*y) + (z*z))
+    return min(sphere1, -sphere2)
+
+def sphere_implicit_batched(coords):
+    coords = coords.copy()
```
elalish, line 28: What is the explicit copy for?

bindings/python/manifold3d.cpp
```diff
@@ -694,7 +695,97 @@ NB_MODULE(manifold3d, m) {
...
+          "grid."
+          ":param edgeLength: Approximate maximum edge length of the triangles "
+          "in the final result.This affects grid spacing, and hence has a "
+          "strong effect on performance."
+          ":param level: You can inset your Mesh by using a positive value, or "
+          "outset it with a negative value."
+          ":param canParallel: Parallel policies violate will crash language "
+          "runtimes with runtime locks that expect to not be called back by "
+          "unregistered threads.This allows bindings use LevelSet despite "
+          "being compiled with MANIFOLD_PAR active."
```
elalish, line 739: I believe this comes from the C bindings - considering we know we're using Python here, do we need this as an input?

src/sdf/src/sdf.cpp
```diff
@@ -356,5 +361,156 @@ Mesh LevelSet(std::function<float(glm::vec3)> sdf, Box bounds, float edgeLength,
...
+
+/**
+ * Similar to LevelSet, this constructs a level-set Mesh from the input
+ * Signed-Distance Function (SDF). This variant feeds an std:vector of points
+ * to the function, allowing the function to compute signed distances using
+ * its own threading/parallelization paradigm.
+ *
+ * @param sdf The signed-distance functor, containing this function signature:
+ * `float operator()(glm::vec3 point)`, which returns the
+ * signed distance of a given point in R^3. Positive values are inside,
```
elalish, line 373: Update?

src/sdf/src/sdf.cpp
```diff
@@ -356,5 +361,156 @@ Mesh LevelSet(std::function<float(glm::vec3)> sdf, Box bounds, float edgeLength,
...
+  // bindings use LevelSet despite being compiled with MANIFOLD_PAR
+  // active.
+  const auto pol = canParallel ? autoPolicy(maxMorton) : ExecutionPolicy::Seq;
+
+  int tableSize = glm::min(
+      2 * maxMorton, static_cast<Uint64>(10 * glm::pow(maxMorton, 0.667)));
+  HashTable<GridVert, identity> gridVerts(tableSize);
+  Vec<glm::vec3> vertPos(gridVerts.Size() * 7, glm::vec3(0.0));
+
+  // BEGIN DIVERGENCE ----------------------------------------------------------
```
elalish, line 412: Can we reduce duplication with `LevelSet`?

src/sdf/src/sdf.cpp
```diff
@@ -356,5 +361,156 @@ Mesh LevelSet(std::function<float(glm::vec3)> sdf, Box bounds, float edgeLength,
...
+
+  // Precompute the Coordinates for the desired Morton Indices
+  std::vector<glm::vec3> gridCoordinates(maxMorton + 1, glm::vec3(0.0));
+  for (Uint64 i = 0; i < maxMorton + 1; i++) {
+    const glm::ivec4 gridIndex = DecodeMorton(i);
+    // TODO: Handle grid indices outside of the traditional bounds...
+    // Indexing gets tough when coordinates are discarded...
+    // This halves the number of samples!!!
```
elalish, line 420: Still a TODO?

test/sdf_test.cpp
```diff
@@ -53,6 +61,162 @@ TEST(SDF, CubeVoid) {
...
+      [](std::vector<glm::vec3> pvec) {
+        std::vector<float> dvec(pvec.size());
+        for (int i = 0; i < pvec.size(); i++) {
+          dvec[i] = UnboundedGyroidSDF()(pvec[i]);
+        }
+        return dvec;
+      },
+      {glm::vec3(-size), glm::vec3(size)}, edgeLength);
+
+  // EXPECT_EQ(levelSet.vertPos .size(), levelSetBatch.vertPos .size());
```
elalish, line 80: Not working?

test/sdf_test.cpp
```diff
@@ -53,6 +61,162 @@ TEST(SDF, CubeVoid) {
...
+
+TEST(SDF, BatchSurface) {
+  const float size = 4;
+  const float edgeLength = 0.5;
+
+  Manifold cubeVoid(LevelSetBatch(
+      [](std::vector<glm::vec3> pvec) {
+        std::vector<float> dvec(pvec.size());
+        for (int i = 0; i < pvec.size(); i++) {
+          dvec[i] = CubeVoid()(pvec[i]);
```
elalish, line 152: Are we basically duplicating existing SDF tests here? Perhaps we should make a helper to automatically run both types and compare them.

elalish (changes requested):
Looking pretty good, but @pca006132 knows the Python side better than I do, so I imagine he'll chime in too.

bindings/python/examples/levelset_sdf.py
pca006132: no, we need autopep8/black for that

src/sdf/src/sdf.cpp
```diff
@@ -356,5 +361,156 @@ Mesh LevelSet(std::function<float(glm::vec3)> sdf, Box bounds, float edgeLength,
...
+    gridDistances[i] =
+        (onLowerBound || onUpperBound || onHalfBound) ? glm::min(d, 0.0f) : d;
+  }
+
+  // Retry Implicit Construction with progressively larger and larger hashtables
+  while (1) {
+    // Check each grid coordinate to see if its neighbors cross the 0 threshold
+    size_t vertIndex = 0;
+    Uint64 lastMorton = 0;
+    for (Uint64 i = 0; i < maxMorton + 1; i++) {
```
pca006132, line 447: is there a reason this is sequential? considering the sdf invocation (the thing that cannot be called in parallel) is done once only, the remaining stuff should be parallelizable.

pca006132:
and I wonder what would the performance be like if you just use the old sdf method, but the function just performs a lookup.

zalo:
All excellent points!  I’ll address these soon!

bindings/python/examples/levelset_sdf.py
zalo: I like to do vertical alignment to make the relationships between lines clear; should I get rid of the instances of this?

bindings/python/manifold3d.cpp
zalo: The batch function's `canParallel` just controls the triangulation, which _shouldn't_ be dependent on the SDF function... but, while developing this PR, I've noticed it crash sometimes with `canParallel` set to true... however the code was unstable and crashing for other reasons at the time, so I'm not sure if it was a threading race condition, or periodically getting lucky with uninitialized memory...

I'm tempted to make this bool control just the `ComputeVertices` step in both functions, and have the triangulation function always be parallel (at least until binary search modes are added for winding number triangulation).

src/sdf/src/sdf.cpp
zalo: Yep... I haven't been able to figure out how to get the routine to succeed when recording the valid morton indices to an array, and just iterating over that for future loops... I suspect perhaps it's not communicating to the triangulator in the same way?

The batched routine computes far fewer samples overall anyway, so it's still a net-gain over the other evaluator, but it's not the full 8x fewer than the naive method that it "should be".

I can change this routine too.

src/sdf/src/sdf.cpp
zalo: I'm not sure what the best way to do this is...
- The function signature is different (Though, I could make _all_ LevelSet calls ask for batched queries, and have a skeleton function overload that replicates the existing signature?  More perf/robustness-testing needed there.)
- It's tough to combine the logic because the changes touch almost every line, even if it's superficial...
- Open to thoughts here.

test/sdf_test.cpp
zalo: The triangulation is very slightly different between the batched and non-batched versions.

I suspect this is because of this check I had to add to avoid sampling out-of-bounds queries, which are hard to see ahead of time:
```cpp
        if (neighborCode >= gridDistances.size())
          continue;  // Neighbors outside of the bounds...
```
In practice, a mesh from the original function might have `95,078` triangles, while the batched function's might have `95,094` triangles.  It seems to fluctuate between having more and less than the original.

test/sdf_test.cpp
zalo: I added a for-loop to automatically run the same checks on each function...

src/sdf/src/sdf.cpp
zalo: Nope!  In theory, all the non-SDF call things can be done in parallel (but my intuition was that all the non-SDF stuff is negligible compared to the SDF evaluation... perhaps I'm mistaken!)

zalo:
> and I wonder what would the performance be like if you just use the old sdf method, but the function just performs a lookup.

It's a good question... though it can be naively simulated by just evaluating the batch function in a simple for-loop, a real-test would probably need the parallelism added back where it can be...

bindings/python/examples/levelset_sdf.py
zalo: The query points are fed in as read-only arrays, but the coordinates need to be offset for the second sphere, so a copy is made.   Hopefully the comment I added makes this clear...

bindings/python/examples/levelset_sdf.py
elalish: Yeah, I'd prefer we use some standard formatter, and ideally check the formatting in our CI. Otherwise every time a contributor comes with a different formatter it makes massive diffs every time they save a file.

bindings/python/manifold3d.cpp
elalish: @geoffder Mind lending your expertise here, since I believe you built this part?

test/sdf_test.cpp
```diff
@@ -96,204 +94,141 @@ TEST(SDF, BatchValidation) {
...
+                 edgeLength),
+        LevelSetBatch(
+            [](std::vector<glm::vec3> pvec) {
+              std::vector<float> dvec(pvec.size());
+              for (int i = 0; i < pvec.size(); i++) {
+                dvec[i] = CubeVoid()(pvec[i]);
+              }
+              return dvec;
+            },
+            {glm::vec3(-size / 2), glm::vec3(size / 2)}, edgeLength)}) {
```
elalish, line 121: Better, but I think we can still reduce a bunch more duplication in these examples by making a helper like `std::pair<Mesh, Mesh> LevelSet2(func, bounds, edgeLength)`.

test/sdf_test.cpp
elalish: That's a little troubling - I think of batch as being purely a performance optimization; I'm surprised it can come up with a different number of triangles or verts. Can you make a tiny example so we can see where the difference is? Do our tests at least verify that both versions are manifold and have the same volume and surface area?

src/sdf/src/sdf.cpp
pca006132: It depends on the grid size I guess. If you do something like a million points, the non-SDF stuff is still pretty time consuming.

In our SDF examples, the SDF we use are pretty simple so the performance largely depends on the non-SDF stuff. The time difference between parallel build and sequential build in that case is still pretty large.

zalo:
I'm going to close this for now, so as not to clutter up the PRs; I'll reopen it later as I have time to address the issues.

bindings/python/manifold3d.cpp
geoffder: The name canParallel doesn't seem familiar to me, but the part about preventing parallel calls to functions that live in another languages runtime is. The flag as it was when I added it prevented the closure from being run in parallel by thrust. Languages with runtime/domain locks like ocaml and python would not allow it.

elalish:
@zalo any interest in coming back to this? I had the impression you were pretty close and it's certainly beneficial. Also, we're in the midst of preparing v3.0, which will have a few breaking changes, so if you need any, now would be a great time.

zalo:
Thanks for the heads-up... I think the changes needed are _just_ beyond my level of comfort/familiarity with the code (for the time I have these days 🫠)...

elalish:
It seemed like you were awfully close. Would you mind putting a summary of what you think is left and where you ran into difficulty, if you can remember? Looking at this again, I think we probably should make level set always run in batch mode.

zalo:
iirc this PR boiled down to forking the LevelSet function to pre-generate a list of sampling points for a given volume, and then running it as usual.

I believe the big issues were:
- The triangulations came out non-deterministically somehow(?) and subtly disagreed with the original LevelSet function.
- Scope creep trying to add additional parallelism
- It wasn’t useful for my use-case without binary searching discontinuous fields 😅

Given that solving the disagreement between the two would mean going over the implementation with a fine-toothed comb, it might be better to rewrite it, merely taking inspiration from what I had 🙃

elalish:
>It wasn’t useful for my use-case without binary searching discontinuous fields 😅

Ah, then you may want to take a look at #866!

PR #603, opened 2023-11-10
RFC  to make SimplifyTopology() steps optional

ramcdona:
This is not intended for real inclusion.

Each major step in SimplifyTopology() was made separately optional.  While this is handy for debugging, it is unlikely the way to production.  In particular, some of the steps may depend on one another in ways that this can violate.

However, this gives a base for testing the optionality of steps in SimplifyTopology() and it provides a starting point for the discussion of how these options should be implemented in production.

### DIFF
--- a/src/manifold/src/edge_op.cpp
+++ b/src/manifold/src/edge_op.cpp
@@ -178,6 +178,10 @@ void Manifold::Impl::SimplifyTopology() {
   }
 #endif
 
+  if (!ManifoldParams().cleanupTriangles) {
+    return;
+  }
+
   std::vector<int> scratchBuffer;
   scratchBuffer.reserve(10);
   {
--- a/src/utilities/include/public.h
+++ b/src/utilities/include/public.h
@@ -613,6 +613,8 @@ struct ExecutionParams {
   bool suppressErrors = false;
   /// Deterministic outputs. Will disable some parallel optimizations.
   bool deterministic = false;
+  /// Perform optional but recommended triangle cleanups in SimplifyTopology()
+  bool cleanupTriangles = true;
 };
 
 #ifdef MANIFOLD_DEBUG


src/manifold/src/edge_op.cpp
```diff
@@ -152,100 +152,115 @@ void Manifold::Impl::SimplifyTopology() {
...
+      }
     }
-  }
 
 #ifdef MANIFOLD_DEBUG
-  if (ManifoldParams().verbose && numFlagged > 0) {
-    std::cout << "found " << numFlagged << " duplicate edges to split"
-              << std::endl;
-  }
+    if (ManifoldParams().verbose && numFlagged > 0) {
```
elalish, line 180: Check your clang-format

src/manifold/src/edge_op.cpp
```diff
@@ -152,100 +152,115 @@ void Manifold::Impl::SimplifyTopology() {
...
-    if (halfedge_[i].IsForward()) {
-      entries.push_back({halfedge_[i].startVert, halfedge_[i].endVert, i});
+  if ( ManifoldParams().splitDuplicateEdges )
+  {
+    Vec<SortEntry> entries;
+    entries.reserve(nbEdges / 2);
+    for (int i = 0; i < nbEdges; ++i) {
+      if (halfedge_[i].IsForward()) {
+        entries.push_back({halfedge_[i].startVert, halfedge_[i].endVert, i});
+      }
```
elalish, line 167: splitPinchedVerts and splitDuplicateEdges are not optional - they rarely make a change, but when they do, it's necessary for manifoldness.

Let's just put an early return after this section when `ManifoldParams().cleanupTriangles = false` and have that be the only new param.

elalish (changes requested):
Thanks! Let's simplify.

elalish (approved):
Thanks!

PR #608, opened 2023-11-11
Apparent infinite loop case for boolean operation

starseeker:
This comes from the FAA Generic Twin example model included with BRL-CAD - the glb files are written out from the input data structures that triggered the infinite loop behavior.

Note that on my local machine (Ubuntu Linux) this test does indeed cause the manifold_test executable to hang - I wasn't able to spot a way offhand to fail the test due to excessive runtime.  If there's a preferred way to do that please advise and I can update the pull request.


pca006132:
same comment with #609, also an issue in the triangulator @elalish

elalish (approved):
Thanks, I'll debug.

PR #609, opened 2023-11-11
Prepare a test input with failing boolean

starseeker:
Encountered this case when testing Manifold with the BRL-CAD example file - specifically, the r.glass8 region.  C++ exception "Vec out of range" is thrown.  Both inputs are the outputs of previous Manifold boolean evaluations.


elalish:
Funny that both this one and #608 only fail on Linux TBB and Mac builds.

pca006132:
maybe we did not enable MANIFOLD_EXPORT for the other CI?

pca006132:
if you enable `MANIFOLD_DEBUG`, the triangulator throws an error:
```
Error in file: /home/pca006132/code/manifold/src/polygon/src/polygon.cpp (73): 'halfedges.size() % 2 == 0' is false: Odd number of halfedges.
```

I think we can merge this after changing the test name to disable it temporarily.

starseeker:
Do you need me to make the name change?  If so, what name did you want?

pca006132:
I can do it, give me a few minutes

elalish (approved):
Thanks, I'll debug.

PR #610, opened 2023-11-12
Fix odd number halfedges

### LINKED ISSUE #606: Infinite loop in ear clipping
Hi!  I work on BRL-CAD (https://github.com/BRL-CAD/brlcad) which is an open source CSG modeling system.  We've recently wired Manifold into our facetize command (for converting implicit CSG boolean trees into evaluated meshes) and gotten excellent results.  However, on rare occasions I'm seeing what appears to be non-manifold output from a boolean operation, which is then fed back into the next operation and causing some sort of infinite loop (when I break gdb reports the code is in EarClip::Triangulate(), specifically EarClip::ProcessEar):

left: (((7863.1.t0.rcc.16_18 u 7863.1.t0.rcc.17_18) u (7863.1.t0.rcc.19_20 u 7863.1.t0.rcc.19_21)) u ((7863.1.t0.rcc.19_22 u 7863.1.t0.rcc.19_23) u (7863.1.t0.rcc.19_24 u 7863.1.t0.rcc.19_25)))
right: (((7863.1.t0.rcc.16_18 u 7863.1.t0.rcc.17_18) u (7863.1.t0.rcc.19_20 u 7863.1.t0.rcc.19_21)) u ((7863.1.t0.rcc.19_22 u 7863.1.t0.rcc.19_23) u (7863.1.t0.rcc.19_24 u 7863.1.t0.rcc.19_25)))
Manifold op: Union

The boolean strings aren't important here except to indicate that both the left and right inputs to the union operation are themselves output from other Manifold operations.  I've tried to capture the meshes themselves at:

http://brlcad.org/~starseeker/7863.1.t0_left.stl
http://brlcad.org/~starseeker/7863.1.t0_right.stl

I know this isn't the glb export from the Manifold routines used for tests, and if it should prove necessary or desirable I'm willing to try and wire up the exact Manifold methods for capturing glb files.  My own tooling suggests that the 7863.1.t0_right.stl input may be the problematic one, but I'm not completely sure of that.  If the necessary inputs for debugging need to come from the earlier mesh inputs that generated 7863.1.t0_right.stl I can try to capture those as well.

Thank you for making an excellent library available - overall, we saw a significant improvement in the success rate of our conversions when we wired Manifold into our pipeline, so it's quite an exciting development.

### LINKED ISSUE #607: Crashing Union case with Manifold
While I've got it set up, here are glb examples of a crashing Union op case with Manifold:

http://brlcad.org/~starseeker/havoc_crash_left.glb
http://brlcad.org/~starseeker/havoc_crash_right.glb

elalish:
Fixes #606 
Fixes #607 

Enables tests from #608, #609

Just a one-liner fixed both of these apparently. It was trying to connect a point that had already been clipped out because it didn't find any reflex corners, but the area calc wasn't precise enough to call it a degenerate polygon. Should be quite a bit more robust now.


starseeker:
I added the one-liner in the BRL-CAD integration of Manifold, and both of the failing cases succeeded there as well - thank you!

PR #612, opened 2023-11-14
Improve gltf-io and manifold-gltf

elalish:
I had noticed some problems with https://manifoldcad.org/make-manifold, where objects with properties were sometimes losing them. It turns out I wasn't reading and writing attributes correctly sometimes, partly because of my gltf-io API. 

Here I've:
1) Transitioned gltf-io and manifold-gltf to TS (which caught several errors on its own)
2) Changed the gltf-io API to use an id map and package each material with its attributes as `Properties`
3) Applied usage groups to remove the gltf-transform warnings
4) Fixed disposal and verified no unused or duplicate entries are being exported to the glTF
5) Simplified the examples and verified that all glTFs now pass validation
6) Added docs to gltf-io
7) Improved `Mesh` declaration by adding `MeshOptions` interface.

Ideally I should look into publishing gltf-io and manifold-gltf to npm.


pca006132:
I am not very familiar with this, so I guess you can just merge it if it works :)

PR #616, opened 2023-11-15
[FEAT] Add LevelSet Function to Python Bindings

zalo:
This PR adds the ability to construct meshes and manifolds from implicit function evaluations.

This PR does NOT include the LevelSet_Batch function, which still needs some work.


bindings/python/examples/levelset_sdf.py
```diff
@@ -0,0 +1,29 @@
...
+ distributed under the License is distributed on an "AS IS" BASIS,
+ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ See the License for the specific language governing permissions and
+ limitations under the License.
+ """
+
+import math
+from manifold3d import Manifold, Mesh
+
+def sphere_implicit(x, y, z):
```
elalish, line 20: This is fine for this PR, but I'd like to move toward consistent examples. We have the same examples in our C++ samples and in our JS examples, and they're designed to be both cute and exercise the various corners of the library. I haven't made Python versions yet mostly because I'm not great at Python, but it would be nice to have the same ones to act as a bit of a Rosetta stone for our API. It would probably help me with my Python too! If you have any interest in translating, it would be much appreciated.

bindings/python/manifold3d.cpp
```diff
@@ -694,7 +695,50 @@ NB_MODULE(manifold3d, m) {
...
+          ":param bounds: An axis-aligned box that defines the extent of the "
+          "grid."
+          ":param edgeLength: Approximate maximum edge length of the triangles "
+          "in the final result.This affects grid spacing, and hence has a "
+          "strong effect on performance."
+          ":param level: You can inset your Mesh by using a positive value, or "
+          "outset it with a negative value."
+          ":param canParallel: Parallel policies violate will crash language "
+          "runtimes with runtime locks that expect to not be called back by "
+          "unregistered threads.This allows bindings use LevelSet despite "
```
elalish, line 738: Based on this description I was hoping someone could answer whether Python has "runtime locks that expect not to be called back by unregistered threads" and therefore whether this should always be true or false. Is this really something we expect to take different values in different situations?

bindings/python/manifold3d.cpp
```diff
@@ -694,7 +695,50 @@ NB_MODULE(manifold3d, m) {
       .def_ro("merge_to_vert", &MeshGL::mergeToVert)
       .def_ro("run_index", &MeshGL::runIndex)
       .def_ro("run_original_id", &MeshGL::runOriginalID)
-      .def_ro("face_id", &MeshGL::faceID);
+      .def_ro("face_id", &MeshGL::faceID)
+      .def_static(
+          "levelset",
```
elalish, line 700: I think our naming convention would be `level_set`.

elalish (changes requested):
Thanks!

bindings/python/examples/levelset_sdf.py
zalo: Sure!   I'd already translated the gyroid one earlier to help isolate a crash on the batch function. 😄

bindings/python/manifold3d.cpp
zalo: I'll remove this for now, but I think we'll eventually want two `canParallel` args... one for the SDF sampling, and one for the triangulation (which doesn't call into the binding, afaict).

zalo:
@elalish Comments addressed; here's what I get for the ported Gyroid Example, does this look right?
image

bindings/python/examples/gyroid_module.py
```diff
@@ -0,0 +1,46 @@
...
+            math.cos(zi) * math.sin(xi))
+
+def gyroid_levelset(level, period, size, n):
+    return Manifold.from_mesh(
+       Mesh.level_set(gyroid, 
+                      [-period, -period, -period, period, period, period], 
+                      period / n, level)).scale(size / period)
+
+def rhombic_dodecahedron(size):
+  box = Manifold.cube(size * math.sqrt(2.0) * np.array([1, 1, 2]), True)
```
pca006132, line 36: maybe need to format this file to make sure the indentation is consistent.

zalo: Good spot, I regularized the indentation.

pca006132:
LGTM, but I will wait for @elalish to merge this in case he has further comments.

elalish (approved):
Excellent, that looks great! And if you ever want to quickly double-check how an example looks, just open it from the drop-down in manifoldCAD.org.

PR #620, opened 2023-11-16
added triangulator fuzzer

### LINKED ISSUE #110: suggestion: fuzzing
Perhaps we can use https://llvm.org/docs/LibFuzzer.html to test the library? Generate a number of primitive objects, perform random translation/scaling/rotation and boolean operations for them. There are a few things we can test for:

1. Memory usage.
2. Should not crash.
3. Compare with other libraries like CGAL? Not sure about if this is possible.

pca006132:
Added triangulator fuzzer using https://github.com/google/fuzztest

Closes #110

Note that
1. It seems that we must create a new compilation target containing the fuzzer. Adding the source file to `manifold_test` will not work, and I don't know why and the documentation is sparse...
2. Ugly preprocessor hack is done to share the polygon test cases between the unit test and fuzzer seed. This can help guide the fuzzer initial choice. I am not aware of any better way to do this.
3. Platform specific hack is done (POSIX specific, probably) to terminate the async thread when timeout is reached. If it is left running, the fuzzer will not notice that the timeout has already reached and will keep waiting, and we will not be able to get the error testcase. Not sure if this works on Mac, need further checking (and considering https://github.com/google/fuzztest/issues/179 is still open, it seems there is not much support for Mac, maybe virtualization will work).
4. To run the test and minimize the test case, you can try the minimizer script.

New testcase (hang indefinitely):
```cpp
TEST(Polygon, Looping1) {
  Polygons polys = {
      {{0.f, -2.f}, {2.f, 2.f}, {0.f, 4.f}, {-3.f, 3.f}},
      {{3.40282347e+38f, 3.40282347e+38f}, {-0.f, 0.f}, {0.f, 1.f}, {1.f, 0.f}},
      {{0.f, -1.f}, {-1.f, 1.f}, {1.f, 1.f}}
  };
  Triangulate(polys, -1);
}

TEST(Polygon, Looping2) {
  Polygons polys = {
      {{-0.277777791f, 0.376543224f}, {0.5f, -0.233333275f}, {0.285185158f, 0.214814827f}},
      {{0.325617313f, -0.055555556f}, {0.277777791f, 0.092592679f}, {0.328042328f, 0.0608465709f}},
      {{3.40282347e+38f, 0.283553332f}, {0.212963f, 0.092592597f}, {0.203703716f, 0.141975299f}, {0.212962985f, 0.138888896f}}
  };
  Triangulate(polys, -1);
}
```


pca006132:
Btw it seems that this fuzztest is mostly an internal tool in google, there is not much public documentation especially regarding how to use the minimizer.

pca006132:
We can also write a check to see if the input is geometrically valid, in case it throws invalid geometry error. Not sure if there is a simple way to do so for those that are only epsilon-valid.

test/polygon_fuzz.cpp
```diff
@@ -0,0 +1,99 @@
...
+#define TEST(_unused1, _unused2)
+
+std::vector<TestCase> SeedProvider() {
+  std::vector<TestCase> TestCases;
+  auto TestPoly = [&TestCases](Polygons polys, int _unused,
+                               float precision = -1.0f) {
+    TestCases.push_back({polys, precision});
+  };
+
+#include "polygon_corpus.cpp"
```
elalish, line 96: I'm thinking it might be more useful to make a test or two specific to fuzzing, rather than throwing all of our existing test cases at it. 

Two tests: 
1) overlapping input - ensure manifold and no crash
2) valid input - additionally ensure geometric validity (I expect this one to catch more bugs)

I don't think either of these need to depend on our current suite of specific polygons. For 1, I think effectively random input is fine. For 2, we can ensure validity by taking the same random input from 1 and running it through `CrossSection` to get the positive fill-rule version. 

To get a higher likelihood of generating problem polygons, I'd recommend rounding the coordinates severely to create lots of equality. To additionally test epsilon-valid geometry, the output of `CrossSection` could be perturbed randomly by less than precision.

test/polygon_fuzz.cpp
```diff
@@ -0,0 +1,99 @@
...
+          manifold::Triangulate(polys, precision);
+          faulted.store(false);
+        } catch (manifold::geometryErr e) {
+          // geometryErr is fine
+          faulted.store(false);
+        } catch (...) {
+          printf("got unexpected error\n");
+        }
+      });
+  if (asyncFuture.wait_for(std::chrono::milliseconds(
```
elalish, line 52: What is the async for? Is this so we can fuzz all the test cases in parallel? Perhaps if we reduce to just a few fuzzing test cases this won't be necessary anymore?

I also found `--gunit_fuzz_for=10m` in other documentation - would this help?

I haven't found *much* more documentation on this internally, but there are people I can ask. I'm not very familiar with any of this yet, but if you want to pass me a fairly specific question to ask, I can try to hunt down an answer.

elalish (commented):
This looks really interesting, thanks!

test/polygon_fuzz.cpp
elalish: Also, holes are the hardest part, so thinking of ways to tweak the generation such that many holes are formed would also be ideal.

minimizer.sh
```diff
@@ -0,0 +1,59 @@
+#!/usr/bin/env bash
+
+# Modified from https://github.com/google/fuzztest/blob/main/tools/minimizer.sh
+# If you want to use a specific reproducer, pass it in FUZZTEST_MINIMIZE_REPRODUCER
+# environment variable.
+# Otherwise, just run it as
+# $ ./minimizer.sh ./build/test/polygon_fuzz --fuzz=PolygonFuzz.TriangulationNoCrash
```
elalish, line 7: Considering you already made a nice `minimize_testcase` function, should we just use that instead of the generic one gfuzz has?

test/polygon_fuzz.cpp
pca006132: The way I understand seed is that seed aids the fuzzer in finding tricky inputs and improve test coverage (e.g. branch coverage). If we only provide one or two seeds, due to the large input space, it can be hard for the fuzzer to get large coverage. Typically, adding more seeds will not harm the fuzz result, but will improve the test result. So I think throwing all our existing test cases is quite useful.

For rounding coordinates, we can probably do https://github.com/google/fuzztest/blob/main/doc/domains-reference.md#reversiblemap-reversible-map. For generating valid geometries, we can define a CSG tree with various primitives, recursively generate the tree, and build the `CrossSection`. I think the CSG tree approach is also helpful for fuzzing our 3D operations. This will probably take a bit of time to do though.

test/polygon_fuzz.cpp
pca006132: No, this async is for implementing timeout. From their code, it seems that the `--gunit_fuzz_for` arg is used for `Runs all fuzz tests in fuzzing mode for the specified duration`, i.e. terminate after the specified duration instead of test timeout.

Parallel fuzzing requires support from their engine, I don't think I see anything related in their code for now.

Maybe you can ask about these two things: test timeout and parallel fuzzing.

minimizer.sh
pca006132: Probably not, `minimize_testcase` is mainly for geometrically valid polygons. Also, the way I wrote `minimize_testcase` is actually pretty similar to how fuzzer works, but not that general, so the fuzzer minimizer is strictly more powerful than our `minimize_testcase` code. We can probably remove the `minimize_testcase` code when the fuzzer integration is complete and easy to use.

test/polygon_fuzz.cpp
elalish: What is the difference between a test timeout and terminating a test after a specified duration?

test/polygon_fuzz.cpp
elalish: I would guess our branch coverage doesn't look too bad anyway, even without any fuzzing. To find new error cases, we need to find new problem situations, which are not that likely to hit a new branch anyway. I'm concerned that by starting with our existing test cases it may actually be harder to find novel problem polygons.

test/polygon_fuzz.cpp
pca006132: Test timeout is for individual test, while the option is for the whole fuzzing test.

test/polygon_fuzz.cpp
pca006132: I think we can think experiment with this when we cannot find error cases with the fuzzer. It seems too early to talk about this, and I think this is more empirical which requires more experiments to see the result. For now I think it works well.

pca006132:
@elalish what is the plan for this? Should we merge this for now and add those advanced fuzz targets later? i.e. valid geometries, rounding, etc.

elalish:
Yeah, you're driving this, I trust you. It's a holiday here so I haven't gotten to looking at your new polygon failure cases yet.

pca006132:
Did you try running it on Mac? Curious if it works, as there are issues in the fuzztest repo saying it does not work on Mac.

I think I will add a separate fuzz target with rounded coordinates, which is simple to implement, before merging.

pca006132:
I am not familiar with absl, will have a look at it later.

PR #621, opened 2023-11-17
fix performance regression for CreateProperties

### LINKED ISSUE #619: Perf regression with properties
This commit https://github.com/elalish/manifold/commit/a842de97c656c8890ae09d9008791857f8ba0d76
Caused by removal of this line https://github.com/elalish/manifold/blob/5e103931f1191eba8eb25d89b1039aa58a03e0db/src/manifold/src/boolean_result.cpp#L559

It creates tons of unique properties and the bins size gets very big (linear search...).
I don't have the full picture but it's a huge perf regression (unusable if polycount > ~70k).

Maybe using a map for the "miss prop" option? (don't really have time to make a PR or do some test atm)

```diff
diff --git a/src/manifold/src/boolean_result.cpp b/src/manifold/src/boolean_result.cpp
index 08ca4af..73f0665 100644
--- a/src/manifold/src/boolean_result.cpp
+++ b/src/manifold/src/boolean_result.cpp
@@ -532,7 +532,8 @@ void CreateProperties(Manifold::Impl &outR, const Manifold::Impl &inP,
 
   using Entry = std::pair<glm::ivec3, int>;
   int idMissProp = outR.NumVert();
-  std::vector<std::vector<Entry>> propIdx(outR.NumVert() + 1);
+  std::vector<std::vector<Entry>> propIdx(outR.NumVert());
+  std::unordered_map<glm::ivec3, int> propIdxMiss;
   outR.meshRelation_.properties.reserve(outR.NumVert() * numProp);
   int idx = 0;
 
@@ -577,17 +578,28 @@ void CreateProperties(Manifold::Impl &outR, const Manifold::Impl &inP,
         }
       }
 
-      auto &bin = propIdx[key.y];
       bool bFound = false;
-      for (int k = 0; k < bin.size(); ++k) {
-        if (bin[k].first == glm::ivec3(key.x, key.z, key.w)) {
-          bFound = true;
-          outR.meshRelation_.triProperties[tri][i] = bin[k].second;
-          break;
+      if (key.y == idMissProp) {
+        auto res =
+            propIdxMiss.try_emplace(glm::ivec3(key.x, key.z, key.w), idx);
+        if (!res.second) {
+          outR.meshRelation_.triProperties[tri][i] = res.first->second;
+          continue;
         }
+
+      } else {
+        auto &bin = propIdx[key.y];
+        for (int k = 0; k < bin.size(); ++k) {
+          if (bin[k].first == glm::ivec3(key.x, key.z, key.w)) {
+            bFound = true;
+            outR.meshRelation_.triProperties[tri][i] = bin[k].second;
+            break;
+          }
+        }
+        if (bFound) continue;
+        bin.push_back(std::make_pair(glm::ivec3(key.x, key.z, key.w), idx));
       }
-      if (bFound) continue;
-      bin.push_back(std::make_pair(glm::ivec3(key.x, key.z, key.w), idx));
+
       outR.meshRelation_.triProperties[tri][i] = idx++;
 
       for (int p = 0; p < numProp; ++p) {
```

pca006132:
Closes #619.

For some reason the proposed fix by @stephomi caused wrong index calculation and caused invalid index access later. I just do the minimum required to fix the performance issue, and it seems that it is not too slow. Perf doesn't report `CreateProperties` as a bottleneck.

The test case can check the performance regression. Basically, without this fix, the performance is so slow that I am not willing to wait for it to complete. It now takes 1200ms to complete.

### DIFF
--- a/src/manifold/src/boolean_result.cpp
+++ b/src/manifold/src/boolean_result.cpp
@@ -15,8 +15,6 @@
 #include <algorithm>
 #include <array>
 #include <map>
-#define GLM_ENABLE_EXPERIMENTAL
-#include <glm/gtx/hash.hpp>
 
 #if MANIFOLD_PAR == 'T' && __has_include(<tbb/concurrent_map.h>)
 #define TBB_PREVIEW_CONCURRENT_ORDERED_CONTAINERS 1
@@ -535,7 +533,7 @@ void CreateProperties(Manifold::Impl &outR, const Manifold::Impl &inP,
   using Entry = std::pair<glm::ivec3, int>;
   int idMissProp = outR.NumVert();
   std::vector<std::vector<Entry>> propIdx(outR.NumVert() + 1);
-  std::unordered_map<glm::ivec2, int> propMissIdx[2];
+  std::vector<int> propMissIdx(outR.NumVert() * 6, -1);
   outR.meshRelation_.properties.reserve(outR.NumVert() * numProp);
   int idx = 0;
 
@@ -581,12 +579,13 @@ void CreateProperties(Manifold::Impl &outR, const Manifold::Impl &inP,
       }
 
       if (key.y == idMissProp && key.z >= 0) {
-        auto entry = propMissIdx[key.x].find(glm::ivec2(key.z, key.w));
-        if (entry != propMissIdx[key.x].end()) {
-          outR.meshRelation_.triProperties[tri][i] = entry->second;
+        // only key.x/key.z matters
+        auto &entry = propMissIdx[key.z * 2 + key.x];
+        if (entry >= 0) {
+          outR.meshRelation_.triProperties[tri][i] = entry;
           continue;
         }
-        propMissIdx[key.x].insert({glm::ivec2(key.z, key.w), idx});
+        entry = idx;
       } else {
         auto &bin = propIdx[key.y];
         bool bFound = false;
--- a/src/manifold/src/boolean_result.cpp
+++ b/src/manifold/src/boolean_result.cpp
@@ -14,6 +14,8 @@
 
 #include <algorithm>
 #include <array>
+#define GLM_ENABLE_EXPERIMENTAL
+#include <glm/gtx/hash.hpp>
 #include <map>
 
 #if MANIFOLD_PAR == 'T' && __has_include(<tbb/concurrent_map.h>)
@@ -532,7 +534,7 @@ void CreateProperties(Manifold::Impl &outR, const Manifold::Impl &inP,
 
   using Entry = std::pair<glm::ivec3, int>;
   int idMissProp = outR.NumVert();
-  std::vector<std::vector<Entry>> propIdx(outR.NumVert() + 1);
+  std::vector<std::unordered_map<glm::ivec3, int>> propIdx(outR.NumVert() + 1);
   outR.meshRelation_.properties.reserve(outR.NumVert() * numProp);
   int idx = 0;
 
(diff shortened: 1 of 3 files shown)


pca006132:
While debugging this I also discovered another bug that cases the test to fail when circularSegments is set to 4096. There is no overflow. I am debugging to see why...

stephomi:
Yeah, the proposal was completely broken, can you try again with the new version, it should be fine.

```diff
diff --git a/src/manifold/src/boolean_result.cpp b/src/manifold/src/boolean_result.cpp
index 08ca4af..a6c5228 100644
--- a/src/manifold/src/boolean_result.cpp
+++ b/src/manifold/src/boolean_result.cpp
@@ -14,6 +14,7 @@
 
 #include <algorithm>
 #include <array>
+#include <iostream>
 #include <map>
 
 #if MANIFOLD_PAR == 'T' && __has_include(<tbb/concurrent_map.h>)
@@ -532,7 +533,8 @@ void CreateProperties(Manifold::Impl &outR, const Manifold::Impl &inP,
 
   using Entry = std::pair<glm::ivec3, int>;
   int idMissProp = outR.NumVert();
-  std::vector<std::vector<Entry>> propIdx(outR.NumVert() + 1);
+  std::vector<std::vector<Entry>> propIdx(outR.NumVert());
+  std::vector<int> propMissIdx(outR.NumVert() * 2, -1);
   outR.meshRelation_.properties.reserve(outR.NumVert() * numProp);
   int idx = 0;
 
@@ -577,17 +579,27 @@ void CreateProperties(Manifold::Impl &outR, const Manifold::Impl &inP,
         }
       }
 
-      auto &bin = propIdx[key.y];
-      bool bFound = false;
-      for (int k = 0; k < bin.size(); ++k) {
-        if (bin[k].first == glm::ivec3(key.x, key.z, key.w)) {
-          bFound = true;
-          outR.meshRelation_.triProperties[tri][i] = bin[k].second;
-          break;
+      if (key.y == idMissProp) {
+        // only key.z/PQ matters
+        auto &entry = propMissIdx[key.z * 2 + PQ];
+        if (entry >= 0) {
+          outR.meshRelation_.triProperties[tri][i] = entry;
+          continue;
         }
+        entry = idx;
+      } else {
+        auto &bin = propIdx[key.y];
+        bool bFound = false;
+        for (int k = 0; k < bin.size(); ++k) {
+          if (bin[k].first == glm::ivec3(key.x, key.z, key.w)) {
+            bFound = true;
+            outR.meshRelation_.triProperties[tri][i] = bin[k].second;
+            break;
+          }
+        }
+        if (bFound) continue;
+        bin.push_back(std::make_pair(glm::ivec3(key.x, key.z, key.w), idx));
       }
-      if (bFound) continue;
-      bin.push_back(std::make_pair(glm::ivec3(key.x, key.z, key.w), idx));
       outR.meshRelation_.triProperties[tri][i] = idx++;
 
       for (int p = 0; p < numProp; ++p) {

```

pca006132:
@stephomi No it still doesn't work, with similar error as the previous patch.
Also, I think the solution using hashmap is simple and efficient enough?

pca006132:
> While debugging this I also discovered another bug that cases the test to fail when circularSegments is set to 4096. There is no overflow. I am debugging to see why...

OK I was wrong, there is in fact overflow. I will add some checks to check for overflow.

pca006132:
The SparseIndices too large error is caused by me forgetting to add a translate for the sphere union test, so everything collided, and the spheres with 4096 circular segments are pretty large... It is now checked to report an error earlier.

stephomi:
Not efficient enough to me, a vector of hashmap is meh..
I'll see what's wrong with my proposal (I only tested with my code, I didn't checked the manifold test)

stephomi:
It adds +10% extra time to the whole operation.

> I wonder if the problem is due to the hashmap or the overall algorithm

I only tested in my codebase, I didn't run the manifold test yet.
I was ignoring the case of oldNumProp <=0, I fixed the snippet it should be good (hopefully).

---

(right now the manifold CMake fails on my machine, so I can't run the test... but I don't really want to investigate at the moment, CMake gets on my nerve, it's probably the FetchContent thing)
```
The link interface of target "utilities" contains:

    glm::glm

  but the target was not found.  Possible reasons include:

    * There is a typo in the target name.
    * A find_package call is missing for an IMPORTED target.
    * An ALIAS target is missing.

```

pca006132:
hmm, it seems that it is not correct yet.

pca006132:
seems that w is also important?

stephomi:
If w is set then y equals vert (and I assume vert is always < idMissProp).
I'm a bit in a blind since I can't run tests atm

pca006132:
indeed it makes sense, I should look into it more closely to see why it breaks...

stephomi:
Ah indeed, I used the wrong length.
I think it should be `std::vector<int> propMissIdx(std::max(inP.NumPropVert(),inQ.NumPropVert()) * 2, -1);`

pca006132:
done

elalish (approved):
Excellent, nice pair programming!

PR #624, opened 2023-11-19
Add animation to manifoldCAD

elalish:
This adds node-level animation support to manifoldCAD.org, along with an animation scrubber. This vaguely follows a similar feature in OpenSCAD, but is more powerful, particularly because the CSG tree doesn't need to be re-evaluated per animation frame, and it's exported properly to our GLB files. 

I updated our tetrahedron puzzle example to show off the feature:
image


hrgdavor:
that is awesome, gotta try it :)

elalish:
Thanks! This PR also gives the ability to change our default material, via `globalDefaults`, which also controls the animation.

PR #626, opened 2023-11-21
add thumbnail file to 3mf output

pca006132:
For some reason files without `_rel/.rels` cannot be imported into openscad. Seems like an issue with lib3mf, not sure if it is stated in the standard or not.

### DIFF
--- a/bindings/wasm/examples/worker.ts
+++ b/bindings/wasm/examples/worker.ts
@@ -14,7 +14,7 @@
 
 import {Accessor, Animation, Document, mat4, Material, Node, WebIO} from '@gltf-transform/core';
 import {KHRMaterialsUnlit, KHRONOS_EXTENSIONS} from '@gltf-transform/extensions';
-import {fileForContentTypes, to3dmodel} from '@jscadui/3mf-export';
+import {fileForContentTypes, fileForRelThumbnail, to3dmodel} from '@jscadui/3mf-export';
 import {strToU8, Zippable, zipSync} from 'fflate'
 import * as glMatrix from 'gl-matrix';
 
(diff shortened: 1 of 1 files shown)


t-paul:
I believe that comes from the OPC spec and is referenced in the 3MF spec in 2.1.1. 3D Parts and Payload Relationships.

This declares a mandatory relationship to the 3D Model part, so the minimal content of that file is a single entry like:

```
<?xml version="1.0" encoding="utf-8"?>
<Relationships xmlns="http://schemas.openxmlformats.org/package/2006/relationships">
        <Relationship Type="http://schemas.microsoft.com/3dmanufacturing/2013/01/3dmodel" Target="/3D/3dmodel.model" Id="rel0" />
</Relationships>
```

elalish (approved):
Thanks, I was trying to figure out why nothing but Cura wanted to import our 3MFs. Apparently I forgot this in the intervening 8 years since I was writing that spec...

elalish:
Huh, looks like Github is struggling today - our CI is mysteriously skipping jobs and my avatar icon is wrong. Hopefully it'll sort itself out...

pca006132:
I think the skipped jobs are due to this PR being a draft PR when I initially opened it. Should now be fixed.

hrgdavor:
This should not work. If you look at the example in
https://github.com/hrgdavor/jscadui/blob/main/file-format/3mf-export/testGen.js

you need to take image from canvas form util like: `cavassToPngA8`

```js

/** //example how to generate thumb from canvas and add it in fflate
const pngPreviewFile = new fflate.ZipPassThrough('Metadata/thumbnail.png');
zip.add(pngPreviewFile);
pngPreviewFile.push(cavassToPngA8(canvas), true);
*/
function cavassToPngA8(canvas) {
  let url = canvas.toDataURL('image/png')
  url = url.substring(url.indexOf(',') + 1)
  // strToU8 function from fflate
  return strToU8(url)
  // string to Uint8Array taken from stackoverflow, and should work in browser
  return new Uint8Array(
    atob(url)
      .split('')
      .map(c => c.charCodeAt(0)),
  )
}
```

notice 2 return return statements, use only one

pca006132:
Ah, I need to check to see if it works, I didn't run the js, I just figured out we need a file in that location to make the 3mf recognizable from other programs.

Thanks for the reminder.

PR #627, opened 2023-11-22
real fix for 3MF export

pca006132:
I forgot to actually test the code yesterday and it was buggy, now the problem is fixed. Thanks @hrgdavor for reminding me.


hrgdavor:
@pca006132 np, I assume after merge manifoldcad.org/ is updated, so I tried it and it did not work, I did not dig deeper to see why, but tried to guess.

pca006132:
waiting for https://github.com/hrgdavor/jscadui/issues/70 so we don't have to remove the thumbnail in an ugly way

elalish (approved):
Thanks!

PR #633, opened 2023-12-02
added fuzzer cases

elalish:
@pca006132 here I've added your two failing test cases from #620, but they aren't failing on my mac.

### DIFF
--- a/test/polygon_corpus.cpp
+++ b/test/polygon_corpus.cpp
@@ -119,6 +119,49 @@ TEST(Polygon, SimpleHoles2) {
   TestPoly(polys, 17);
 }
 
+TEST(Polygon, Looping1) {
+  Polygons polys;
+  polys.push_back({
+      {0.f, -2.f},  //
+      {2.f, 2.f},   //
+      {0.f, 4.f},   //
+      {-3.f, 3.f},  //
+  });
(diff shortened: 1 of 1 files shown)


pca006132:
interesting, will check them tonight.

elalish:
Yeah, since everything is passing, I guess this PR is not needed. But maybe the problem polygons were output wrong? Or maybe they need to be tested at a particular precision?

elalish:
Progress, but not quite there yet. It's late, I'll pick this up tomorrow.

src/polygon/src/polygon.cpp
```diff
@@ -286,7 +286,8 @@ class EarClip {
       VertItr center = tail;
       VertItr last = center;
 
-      while (nextL != nextR && tail != nextR) {
+      while (nextL != nextR && tail != nextR &&
+             nextL != (toLeft ? right : left)) {
```
elalish, line 290: This fixed the infinite loop

pca006132:
Btw, I tried running the two fuzzer cases for several minutes and was unable to find crashing/infinite loop cases. I guess the robustness is much better now. Will try to write the fuzzer for valid/ε-valid inputs.

PR #635, opened 2023-12-04
Fix precision

### LINKED ISSUE #634: precision calculation causes collider error
For the following code:
```c++
TEST(Boolean, Spiral) {
  ManifoldParams().deterministic = true;
  const int d = 2;
  std::function<Manifold(const int, const float, const float)> spiral = [&](const int rec, const float r, const float add) {
    const float rot = 360.0f / (glm::pi<float>() * r * 2) * d;
    const float rNext = r + add / 360 * rot;
    const Manifold cube = Manifold::Cube(glm::vec3(1), true).Translate({0, r, 0});
    if (rec > 0)
      return spiral(rec-1, rNext, add).Rotate(0, 0, rot) + cube;
    return cube;
  };
  const Manifold result = spiral(117, 25, 2);
}
```

It causes error for rec > 117. It seems that it triggered collider error:
```
'leafBB.size() == NumLeaves()' is false: must have the same number of updated boxes as original
```
and causes Vec out of range when MANIFOLD_DEBUG is not enabled.

elalish:
Fixes #634 

The only fix here for now is in the collider to keep it working with empty lists. The test is interesting - it is layering up rotations, which is indeed layering up rounding errors - it would be easy enough to avoid this. Still, it does feel like our precision calculations are too conservative. This is because we're using an H-infinity norm instead of an H-2 norm for our transform matrix. To get an H-2 norm, we need the largest singular value, which comes from an SVD op. I don't feel like pulling in a dependency on Eigen just for a simple 3x3 matrix op. Most promising I've seen so far is https://github.com/wi-re/tbtSVD, though I'd kind of like to simplify it away from CUDA. 

What do you think, @pca006132?

### DIFF
--- a/src/utilities/include/svd.h
+++ b/src/utilities/include/svd.h
@@ -26,6 +26,7 @@
 
 #include <algorithm>
 #include <cmath>
+#include <glm/glm.hpp>
 #include <tuple>
 #include <utility>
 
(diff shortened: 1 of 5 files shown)


pca006132:
It seems that there is a CPU implementation available: http://evjang.com/svd3/
But we may want to try some tests before using it, not sure about the quality of a random library...

pca006132 (approved):

elalish:
Yeah, I saw that too, but [this](https://github.com/ericjang/svd3/issues/7) gave me pause. Alas that GLM doesn't have these functions! The library I found is newer, but has no activity.

elalish:
Okay, I went ahead and pulled in that 3x3 SVD and simplified it a bit. It seems to be working, though you're right that some more testing would probably be a good idea. Any thoughts?

src/utilities/include/svd.h
```diff
@@ -0,0 +1,381 @@
...
+}
+// Helper class to contain a quaternion. Could be replaced with float4 (CUDA
+// based type) but this might lead to unintended conversions when using the
+// supplied matrices
+struct quaternion {
+  float x = 0.f, y = 0.f, z = 0.f, w = 1.f;
+  float& operator[](int32_t arg) { return ((float*)this)[arg]; }
+};
+// A simple 3x3 Matrix class
+struct Mat3x3 {
```
elalish, line 62: Should really replace this with glm::mat3 for consistency and to remove code. Also, I'm a little concerned I have my row-major/column-major backwards, but it's passing the tests... We should probably have more.

pca006132: I think eigenvalue is invariant under matrix transpose?

pca006132:
On my computer (4.4GHz i9 12900HK), it takes about 4.4us to compute the SVD on average, so it should be fast enough for our purpose :)

pca006132:
but yeah if the algorithm can be shorter it can still be nice.

src/utilities/include/svd.h
elalish: Oh right, that makes sense.

elalish:
@zalo, thanks, that was a good read. How do you get the singular value once you have the axis? I think I'll just stick with this one despite the complexity since some work has been done on verifying its accuracy and robustness to singular and edge cases. Also, I have the feeling a full SVD may come in handy down the road. 

@pca006132 Yeah, I'll go ahead and refactor this to use glm in this PR. Thanks for verifying!

src/utilities/include/svd.h
```diff
@@ -0,0 +1,310 @@
...
+  // R is now set to desired value
+  R[0][0] = B[0][0];
+  R[1][0] = B[1][0];
+  R[2][0] = B[2][0];
+  R[0][1] = fmaf(a, B[0][1], b * B[0][2]);
+  R[1][1] = fmaf(a, B[1][1], b * B[1][2]);
+  R[2][1] = fmaf(a, B[2][1], b * B[2][2]);
+  R[0][2] = fmaf(-b, B[0][1], a * B[0][2]);
+  R[1][2] = fmaf(-b, B[1][1], a * B[1][2]);
+  R[2][2] = fmaf(-b, B[2][1], a * B[2][2]);
```
elalish, line 247: I could simplify this stuff too, using GLM vector ops, but since it's operating on rows instead of columns, I'd need to transpose and I decided it was simpler to just leave it for now. And who knows, maybe this really is slightly faster?

src/utilities/include/svd.h
elalish: Okay, I verified that we now have the correct transpose for the actual decomposition too.

Issue #192, opened 2022-09-04
Mesh Offsetting


pca006132:
Mesh offsetting seems very useful for adding tolerance to parts, especially when combined with face selection. 

I wonder if I can just offset each vertices according to their tangent? Although we may need to recompute the tangent if we only selected a subset of faces connecting to that vertex.

pentacular:
You run the risk of self-intersection.

Also, offsetting by the tangent of the vertex won't offset the adjacent face by a fixed amount, which may give uneven tolerance across the face.

pca006132:
So maybe we can compute the offset of each selected faces and add them to the vertices? Not sure about self intersection though.

elalish:
Yeah, I was thinking about adding the vertex normal as a const input to the `Warp` function to allow this kind of thing, but indeed, it is very easy for a warping function to introduce self-intersections. The "right" way to offset a mesh is to do a Minkowski operation with a sphere (OpenSCAD supports this) but it's very slow and dramatically explodes the number of faces.

thehans:
To avoid self intersections, you need to guarantee that the input is convex.  So what is needed is a convex decomposition operation (then offset each convex piece, then union to get the final result).

> I wonder if I can just offset each vertices according to their tangent? Although we may need to recompute the tangent if we only selected a subset of faces connecting to that vertex. OK this will not work.

I don't think a vertex has any specific "tangent", maybe you mean the vertex normal?
[Vertex normal](https://en.wikipedia.org/wiki/Vertex_normal) makes a bit more sense as that is a concept from the field of graphics.  But i'd consider it more like a hack/fudged values that looks ok to the eye, than something of rigorously defined geometry.
Imagine translating each face by the offset, along its normal;  then conceptually extend each face until it once again intersects at new edges/vertices.  That's where you want the vertex to end up, but I'm not entirely sure if the straight average of face normals, applied to the vertex results in this transformation or if there is more to it.

elalish:
The reason a Minkowski sum is the "right" way is that for a general vertex (not symmetric, more than three faces), when you offset the faces, they will no longer meet up at that same single vertex (in any position). The simplest alternative is just to offset the triangles exactly along their face normals, then stitch the edge and vertex gaps with new faces, but again it only works on convex polyhedra and it explodes the number of faces rapidly. 

[...]

Issue #426, opened 2023-05-07
Get CrossSection from Manifold and plane


harmanpa:
There doesn't appear to be a way of creating a CrossSection from a Manifold. This would be a useful feature for creating a 2D slice from a 3D mesh.

pca006132:
Yes we did think about this in #127, but it seems that we forgot it and no one remember to implement it.

elalish:
Agreed, I've been wanting this, but haven't gotten around to implementing it. Shouldn't be too hard; we can use a similar API to `trimByPlane`.

elalish:
Any interest in contributing @harmanpa?

johnnybigert:
Plus one on this one, would be very useful to have :)

harmanpa:
Awesome, thank you @elalish

PR #636, opened 2023-12-05
[FEAT] Add Sweep Operation

zalo:
This PR is an attempt to break out a successful experiment from Python.
https://github.com/elalish/manifold/issues/192#issuecomment-1823327943

- Does this make sense to move to Impl with parallel for-loops?
- Should I add full transform (translation/rotation/scale) variant with a `$fn` for more interesting sweep trajectories?
- Maybe Catmull-Rom Spline Interpolation between Segments? (Could also be left to the user)

### DIFF
--- a/bindings/c/include/manifoldc.h
+++ b/bindings/c/include/manifoldc.h
@@ -120,7 +120,7 @@ ManifoldManifold *manifold_hull_pts(void *mem, ManifoldVec3 *ps, size_t length);
 // Sweeps
 
 ManifoldManifold *manifold_sweep(void *mem, ManifoldManifold *m, float x,
-                                     float y, float z);
+                                 float y, float z);
 
 // Manifold Transformations
 
--- a/bindings/c/include/manifoldc.h
+++ b/bindings/c/include/manifoldc.h
@@ -117,6 +117,11 @@ ManifoldManifold *manifold_hull(void *mem, ManifoldManifold *m);
 ManifoldManifold *manifold_batch_hull(void *mem, ManifoldManifoldVec *ms);
 ManifoldManifold *manifold_hull_pts(void *mem, ManifoldVec3 *ps, size_t length);
 
+// Sweeps
+
+ManifoldManifold *manifold_sweep(void *mem, ManifoldManifold *m, float x,
+                                     float y, float z);
+
 // Manifold Transformations
 
 ManifoldManifold *manifold_translate(void *mem, ManifoldManifold *m, float x,
--- a/bindings/c/manifoldc.cpp
+++ b/bindings/c/manifoldc.cpp
@@ -222,6 +222,13 @@ ManifoldManifold *manifold_hull_pts(void *mem, ManifoldVec3 *ps,
   return to_c(new (mem) Manifold(hulled));
 }
 
+ManifoldManifold *manifold_sweep(void *mem, ManifoldManifold *m, float x,
+                                 float y, float z) {
+  auto v = glm::vec3(x, y, z);
+  auto swept = from_c(m)->Sweep(v);
+  return to_c(new (mem) Manifold(swept));
+}
+
 ManifoldManifold *manifold_translate(void *mem, ManifoldManifold *m, float x,
                                      float y, float z) {
   auto v = glm::vec3(x, y, z);
--- a/bindings/python/manifold3d.cpp
+++ b/bindings/python/manifold3d.cpp
@@ -154,6 +154,21 @@ NB_MODULE(manifold3d, m) {
           },
           "Compute the convex hull enveloping a set of 3d points.")
       .def(
+          "sweep",
+          [](Manifold &self, float x = 0.0f, float y = 0.0f, float z = 0.0f) {
+            return self.Sweep(glm::vec3(x, y, z));
+          },
+          nb::arg("x") = 0.0f, nb::arg("y") = 0.0f, nb::arg("z") = 0.0f,
+          "Sweep this Manifold through space."
+          "\n\n"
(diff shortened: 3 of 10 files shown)


zalo:
I think Swept profiles are called "Pipe"s in CAD; having a sweep on the `CrossSection` would be fun too.

Swept Volumes aren't used suuper often, but there also aren't many libraries that can do them (I think [Siemens NX](https://www.youtube.com/watch?v=yOmabwGMQgw) is the only CAD package with it).
In modelling, I believe they're typically used for cutting out the profile of one body against another that needs to slide or rotate near it.

Like, cutting funny shaped gears:
https://www.shadertoy.com/view/ldsBRr
Or the interesting animations shown in the last example here:
https://www.youtube.com/watch?v=6iLqMQ3kd24

It seems like the interesting stuff requires full transform support, so I'll see about adding that...

elalish:
Based on your examples, it sounds like another way to achieve this is through #426 - particularly if we support both of OpenSCAD's [options](https://en.wikibooks.org/wiki/OpenSCAD_User_Manual/Using_the_2D_Subsystem#3D_to_2D_Projection). For a sliding part, you could extrude the projection with cut = false. This feels like a more intuitive and flexible API to me, which is part of why I think this is a rare function.

I'm thinking we'd probably make this two functions: `Project()` and `Slice(float height)`, since they'd be implemented differently anyway. Would you be interested in implementing `Project()`? I think it would be just a 2D convex hull.

pca006132:
But if `project` is implemented as 2D convex hull, it will not work for 2D shape that is not convex, and what about things like a torus that has a hole inside?

elalish:
Oh right, I didn't think about that long enough. Hmm, I know how to do a slice, but what's the algorithm for a projection?

pca006132:
I think openscad just project every triangle into 2D and union them.

elalish:
@zalo I don't mean to be discouraging - that video is cool. Still, for `n` triangles, we're talking `n` convex hulls all unioned together? And that only gets us one step in the sweeps they're talking about, and doesn't even include rotation. How computationally feasible is this algorithm?

pca006132:
I feel that the idea of using 2D projection for sweep, union that with the initial and final position should be faster.

zalo:
2D projection + Extrude won’t work, see linked comment gif for counter example shape:
https://github.com/elalish/manifold/issues/192#issuecomment-1823327943
(Also, CrossSections lose 3D information?)

Also, I’m not familiar enough with Manifold’s internals to implement project and slice operators efficiently 😅
I’d probably try to do slicing by splitting by a plane, and then deleting all the vertices that are collocated with the original part, and projection by unioning the triangles in 2D as @pca006132 said…

Like Minkowski Sums, this might be part of a class of more expensive mesh algorithms…
Winding Number fills, Morphological Dilation/Erosion/Opening/Closing, Retopologizing, etc.

Though (save for the morphological ops), they tend to veer away from the “Exactish-CSG” spirit of the library (does simple things fast), so I’d understand if you’d rather tell users to use a different library, or write their own extension kernels in the client bindings languages…

pca006132:
@zalo I am deep inside a debug session of something else, will response later tmr about the slice thing. I think slower operation is fine, we just want to make sure we are not missing some obvious better implementation.

pca006132:
And I see why this sweep is not that simple now. Thanks.

pca006132:
Thinking about it, I think we can have a fast and simple algorithm for this.

Say the direction is v, for every face we compute the dot product of its normal with v, positive means that we should offset the face, and negative means that it should stay. To avoid self-intersection, we use the collider to compute if the swept volume of a face (positive dot product) will intersect with another face, mark them and do not move them. After moving the faces, we stitch the gaps by checking the neighbors of the moved faces. Because the marked faces have a positive dot product with v, we can compute the sweep result of each connected component of marked faces and union with the previous result.

Just thought of this, not sure if it makes sense.

elalish:
@zalo Yeah, this particular op feels like a combination of too specialized and not really any slower to implement outside of the core library. I think a polygon-sweep-along-path op will be more generally useful, but I want to achieve this via #289, because then it will be very fast and that will also solve a number of other important problems.

zalo:
> Thinking about it, I think we can have a fast and simple algorithm for this.
>  Because the marked faces have a positive dot product with v, we can compute the sweep result of each connected component of marked faces and union with the previous result.
> 
> Just thought of this, not sure if it makes sense.

This actually sounds pretty good (though, I’m not sure it’ll work for the more general Translation+Rotation+Scale Matrix Sweep…)

My mesh-fu is a little weak… if we delete the backwards triangles, and isolate just the connected component regions, is there a straight-forward way to step along the component boundary?

I guess the boundary edges can be marked and remembered during the “backwards triangle deleting” process… That could work!  I’ll think about this a little more…

zalo:
Oh, there's a couple more use-cases for sweeps that might be really fun:
- Drawing/Sculpting/Modelling Tools; Dreams has a "Smear" tool (which is Sweeping along VR Controller paths)
https://youtu.be/0BbxHRZB9UA?t=1289
https://youtu.be/_PNUVFYgI8w?t=153
https://youtu.be/v27LZivpffQ?t=208
https://youtu.be/v27LZivpffQ?t=1025

- CNC Toolpath Planning/Simulation:
image
image

pca006132:
> Do you want to attack those or should I? I also have an idea for offset I want to try out.

You can do it. I am working on other stuff right now and don't have enough time for this.

zalo:
Alright; I'll close for now since the 90% use-case is on Convex Tools/Solids, which can just be Hulled.

PR #637, opened 2023-12-06
Added morph animation to ManifoldCAD.org

elalish:
Since I already added rigid animations, I figured it wouldn't take too much more to handle morph target glTF animations as well. This is a nice way to visually demonstrate what a `Warp` function is doing. Took me a bit to figure out an API that would actually work in general, so I've finally settled on `setMorphStart` and `setMorphEnd`. Of course glTF morph targets can do much more than this, but I think a simple 2-shape interpolation is good for simplicity.

### DIFF
--- a/bindings/wasm/examples/public/editor.d.ts
+++ b/bindings/wasm/examples/public/editor.d.ts
@@ -49,14 +49,37 @@ declare const globalDefaults: {
 }
 
 /**
- * Set material properties on the input manifold. They will be carried along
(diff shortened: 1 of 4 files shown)


pca006132:
Interesting, do you have an animated GIF for that?

elalish:
https://github.com/elalish/manifold/assets/1649964/2995963e-a964-443d-b015-2954abb66a09

PR #638, opened 2023-12-08
Added Project()

elalish:
Addresses one part of #426 

Pretty simple algorithm. However, I'm getting lots of micro holes in the bracelet projection, probably from rounding errors, which are messing up the triangulation. I haven't determined if the CrossSection is actually epsilon-valid, and either way, I'm surprised that Clipper's positive fill rule isn't getting rid of them.

### DIFF
--- a/src/cross_section/src/cross_section.cpp
+++ b/src/cross_section/src/cross_section.cpp
@@ -146,6 +146,18 @@ void decompose_hole(const C2::PolyTreeD* outline,
   }
 }
 
+void flatten(const C2::PolyTreeD* tree, C2::PathsD& polys, int i) {
+  auto n_outlines = tree->Count();
+  if (i < n_outlines) {
+    auto outline = tree->Child(i);
+    flatten(outline, polys, 0);
+    polys.push_back(outline->Polygon());
+    if (i < n_outlines - 1) {
+      flatten(tree, polys, i + 1);
+    }
+  }
+}
+
 bool V2Lesser(glm::vec2 a, glm::vec2 b) {
   if (a.x == b.x) return a.y < b.y;
   return a.x < b.x;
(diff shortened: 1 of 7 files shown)


src/manifold/src/face_op.cpp
```diff
@@ -194,17 +196,12 @@ void Manifold::Impl::Face2Tri(const Vec<int>& faceEdge,
  * For the input face index, return a set of 2D polygons formed by the input
```
pca006132, line 196: need to update documentation

pca006132:
Yeah I envisioned this epsilon-valid issue when converting our 3D model into 2D, I think similar things will happen for slice as well. Not sure how clipper handles this.

src/cross_section/src/cross_section.cpp
```diff
@@ -585,7 +597,28 @@ CrossSection CrossSection::Warp(
...
+  C2::PathsD filtered;
+  for (C2::PathD poly : polys) {
+    auto area = C2::Area(poly);
+    Rect box;
+    for (auto vert : poly) {
+      box.Union(glm::vec2(vert.x, vert.y));
+    }
+    glm::vec2 size = box.Size();
+    if (glm::abs(area) > glm::max(size.x, size.y) * epsilon) {
+      filtered.push_back(poly);
```
elalish, line 617: This is how I took care of the tiny sliver polygons we were getting in the bracelet.

src/third_party/clipper2
elalish, line 1: I also went ahead and updated Clipper2 to the latest release - I didn't actually notice any difference, but they said they fixed some bugs and I don't think we were on a release at all before.

pca006132 (approved):
Feel like we may want to add more simplification to CrossSection later. LGTM for now.

PR #640, opened 2023-12-08
Misc

pca006132:
1. Makes tracy support better.
2. Remove some warnings from clang and clang-tidy.

image


src/polygon/src/polygon.cpp
```diff
@@ -13,26 +13,17 @@
 // limitations under the License.
 
 #include "polygon.h"
-#if MANIFOLD_PAR == 'T'
-#include "tbb/tbb.h"
```
elalish, line 17: 💯

src/utilities/include/svd.h
```diff
@@ -32,22 +32,22 @@
...
 // Threshold value
-constexpr float _SVD_EPSILON = 1e-6f;
+inline constexpr float _SVD_EPSILON = 1e-6f;
 // Iteration counts for Jacobi Eigen Analysis, influences precision
-constexpr int JACOBI_STEPS = 12;
+inline constexpr int JACOBI_STEPS = 12;
 
 // Helper function used to swap X with Y and Y with  X if c == true
-void CondSwap(bool c, float& X, float& Y) {
+inline void CondSwap(bool c, float& X, float& Y) {
```
elalish, line 44: Ah yes, I forgot about this.

src/utilities/include/svd.h
```diff
@@ -289,7 +289,7 @@ struct SVDSet {
  *
  * @param A The matrix to decompose.
  */
-SVDSet SVD(glm::mat3 A) {
+inline SVDSet SVD(glm::mat3 A) {
```
elalish, line 292: Should we move this into a .cpp? I'm not completely clear on the tradeoffs.

src/utilities/include/utils.h
```diff
@@ -26,6 +26,15 @@
 #include "par.h"
 #include "vec.h"
 
+#if __has_include(<tracy/Tracy.hpp>)
+#include <tracy/Tracy.hpp>
+#else
+#define FrameMarkStart(x)
+#define FrameMarkEnd(x)
+#define ZoneScoped
+#define ZoneScopedN(name)
```
elalish, line 35: Maybe a good place to document how we should use `ZoneScoped`? I have a vague idea from your usage, but I'm not entirely sure what it does.

README.md
```diff
@@ -144,6 +144,7 @@ Contributions are welcome! A lower barrier contribution is to simply make a PR t
 
 There is now basic support for the [Tracy profiler](https://github.com/wolfpld/tracy) for our tests.
 To enable tracing, compile with `-DTRACY_ENABLE=on` cmake option, and run the test with Tracy server running.
+To enable memory profiling in addition to tracing, compile with `-DTRACY_MEMORY_USAGE=ON` in addition to `-DTRACY_ENABLE=ON`.
```
elalish, line 147: 👍

elalish (approved):
Thanks! Did this find anything interesting so far?

src/utilities/include/svd.h
pca006132: Probably not much pros and cons, more like ideological issue. If this is complicated and used in multiple places, putting it in a `.cpp` will reduce compilation time a bit, but it is not the case here.

src/utilities/include/utils.h
pca006132: Sure, it is just something to mark the current zone (scope) for the profiler to see.

pca006132:
No, clang-tidy did not find anything interesting.

PR #642, opened 2023-12-10
change extrude behavior

pca006132:
#632 

This change the extrude behavior to match openscad linear extrude.

### DIFF
--- a/src/manifold/src/constructors.cpp
+++ b/src/manifold/src/constructors.cpp
@@ -233,6 +233,7 @@ Manifold Manifold::Sphere(float radius, int circularSegments) {
  * bottom, interpolated linearly for the divisions in between.
  * @param scaleTop Amount to scale the top (independently in X and Y). If the
  * scale is {0, 0}, a pure cone is formed with only a single vertex at the top.
+ * Note that scale is applied after twist.
  * Default {1, 1}.
  */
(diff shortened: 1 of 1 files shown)


pca006132:
Not yet addressing the shortest diagonal part. I think we should probably also use this shortest diagonal thing for rotate extrude.

elalish (approved):
Do you have an image showing the before/after difference?

pca006132:
I had, but it is not here with me now. Will reply here tonight.

pca006132:
```
const a = 110;
const a2 = 150;
const h = 30;
const R = 20;
const r = 2;
const result = Manifold.extrude(
  CrossSection.circle(r).translate([R*Math.sin(a2), R*Math.cos(a2)]), h, 100, a, [0, 1])
  .add(Manifold.cube([1, 1, h]));

```

Before:
image


After:
image

pca006132:
@elalish btw this also changes the extrude direction, not sure if we should do this as some users may rely on the extrude direction.

pca006132:
In terms of openscad compatibility, we can just negate the sign of the twist angle.

pca006132:
OK I reverted the direction change, it is not worth breaking existing code.

elalish (approved):
Thanks, yeah CCW seems more logical to me anyway.

PR #643, opened 2023-12-10
example url

pca006132:
1. Allows choosing the script via URL hash, e.g. `#Intro`.
2. Removed the annoying newline at the start of examples.

### DIFF
--- a/bindings/wasm/examples/editor.js
+++ b/bindings/wasm/examples/editor.js
@@ -207,16 +207,19 @@ function addEdit(button) {
 }
 
 const newButton = document.querySelector('#new');
-function newItem(code) {
-  const name = uniqueName('New Script');
+function newItem(code, scriptName = undefined) {
+  const name = uniqueName(scriptName ?? 'New Script');
   setScript(name, code);
   const nextButton = createDropdownItem(name);
   newButton.insertAdjacentElement('afterend', nextButton.parentElement);
   addEdit(nextButton);
-  nextButton.click();
+  return {
+    button: nextButton,
+    name
+  };
 };
 newButton.onclick = function() {
-  newItem('');
+  newItem('').button.click();
 };
 
 const runButton = document.querySelector('#compile');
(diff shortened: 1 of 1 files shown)


pca006132:
I think we can also allow users to import raw code from another website, e.g. gist, but not sure what is the best way of embedding that into the url.

bindings/wasm/examples/editor.js
```diff
@@ -299,7 +301,14 @@ require(['vs/editor/editor.main'], async function() {
       addEdit(button);
     }
   }
-  switchTo(currentName);
+  if (window.location.hash.length > 0) {
+    const name = unescape(window.location.hash.substring(1));
+    if (exampleFunctions.get(name) != null || getScript(name) != null)
+      switchTo(name);
```
elalish, line 307: Is this `if` necessary?

elalish (approved):
Thanks!

bindings/wasm/examples/editor.js
pca006132: yeah it is not really needed, will remove

pca006132:
OK nice, will do that.

pca006132:
I thought about this, I don't think this is usually in the threat model. For example, if you think about sites like codepen, they also allow user submitted js. The typical issue is usually things like stealing credentials or cookies, but we don't have that here. We do need to think if we consider user code as confidential.

pca006132:
Added script loading code, now users can load code via `?url=${url}&name=${scriptName}`.

bindings/wasm/examples/editor.js
```diff
@@ -299,7 +302,25 @@ require(['vs/editor/editor.main'], async function() {
       addEdit(button);
     }
   }
-  switchTo(currentName);
+  const params = new URLSearchParams(window.location.search);
+  if (window.location.hash.length > 0) {
+    const name = unescape(window.location.hash.substring(1));
+    switchTo(name);
+  } else if (params.get('url') != null) {
+    console.log(`Fetching ${params.get('url')}`);
```
elalish, line 310: It seems weird to call a url param `url`, especially when it's not actually a URL. How about `script`?

bindings/wasm/examples/editor.js
```diff
@@ -299,7 +302,25 @@ require(['vs/editor/editor.main'], async function() {
...
+    autoExecute = false;
+    const response = await fetch(params.get('url'));
+    const code = await response.text();
+    const name = uniqueName(params.get('name') ?? 'New Script');
+    setScript(name, code);
+    const nextButton = createDropdownItem(name);
+    newButton.insertAdjacentElement('afterend', nextButton.parentElement);
+    addEdit(nextButton);
+    switchTo(name);
```
elalish, line 319: Is some of this duplicated code that could be a function?

elalish (commented):
Yeah, I guess I'm just used to Google-level security review, which our site certainly doesn't need. Might be cool to add a little share button that created this link with the proper escaping.

bindings/wasm/examples/editor.js
pca006132: sure

bindings/wasm/examples/editor.js
pca006132: yes

pca006132:
yeah I wanted to add a UI control for this, but I don't have a good design for this.

elalish:
> yeah I wanted to add a UI control for this, but I don't have a good design for this.

I'll take care of it eventually. Basically just a little circle button with a share icon somewhere.

pca006132:
Nice. Will do the changes later this week as I have a deadline this Thursday night. You can also take over this if you like.

elalish:
No rush, sounds good.

elalish (approved):
Thanks!

PR #644, opened 2023-12-11
added Slice()

### LINKED ISSUE #426: Get CrossSection from Manifold and plane

elalish:
Fixes #426 

This gets us proper cross-sections of a Manifold. I gave it a height parameter to make it cheaper to do a stack of slices as a 3D printer might need, rather than having to transform the whole model each time like in OpenSCAD. This will return the bottom face if the height is equal to the bottom of the bounding box, and return empty when at the top of the bounding box. 

~~Now we just need to add the various bindings for this and `Project`.~~ Done

### DIFF
--- a/bindings/c/manifoldc.cpp
+++ b/bindings/c/manifoldc.cpp
@@ -202,6 +202,17 @@ ManifoldManifold *manifold_trim_by_plane(void *mem, ManifoldManifold *m,
   return to_c(new (mem) Manifold(trimmed));
 }
 
+ManifoldCrossSection *manifold_slice(void *mem, ManifoldManifold *m,
+                                     float height) {
+  auto poly = from_c(m)->Slice(height);
+  return to_c(new (mem) CrossSection(poly));
+}
+
+ManifoldCrossSection *manifold_project(void *mem, ManifoldManifold *m) {
+  auto poly = from_c(m)->Project();
+  return to_c(new (mem) CrossSection(poly));
+}
+
 ManifoldManifold *manifold_hull(void *mem, ManifoldManifold *m) {
   auto hulled = from_c(m)->Hull();
   return to_c(new (mem) Manifold(hulled));
--- a/bindings/wasm/examples/worker.ts
+++ b/bindings/wasm/examples/worker.ts
@@ -61,7 +61,7 @@ const manifoldStaticFunctions = [
 const manifoldMemberFunctions = [
   'add', 'subtract', 'intersect', 'decompose', 'warp', 'transform', 'translate',
   'rotate', 'scale', 'mirror', 'refine', 'setProperties', 'asOriginal',
-  'trimByPlane', 'split', 'splitByPlane', 'hull'
+  'trimByPlane', 'split', 'splitByPlane', 'slice', 'project', 'hull'
 ];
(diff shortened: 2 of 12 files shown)


src/manifold/src/face_op.cpp
```diff
@@ -226,6 +227,68 @@ PolygonsIdx Manifold::Impl::Face2Polygons(VecView<Halfedge>::IterC start,
...
 
+CrossSection Manifold::Impl::Slice(float height) const {
+  Box plane = bBox_;
+  plane.min.z = plane.max.z = height;
+  Vec<Box> query;
+  query.push_back(plane);
+  const SparseIndices collisions =
+      collider_.Collisions<false, false>(query.cview());
+
+  std::set<int> tris;
```
pca006132, line 238: what about `std::unordered_set`?

pca006132 (approved):
We may also want to clarify that triangles that does not *cross* the target height may not be sliced. E.g. `Cube().Slice(1.0)` is empty while `Cube().Translate({0, 0, 1.0}).Slice(1.0)` is a square, because we require `z.min <= height && z.max > height`.

pca006132:
Btw we should also add Slice and project to js, C and python bindings. Should be simple though.

bindings/c/manifoldc.cpp
```diff
@@ -202,6 +202,17 @@ ManifoldManifold *manifold_trim_by_plane(void *mem, ManifoldManifold *m,
...
 
+ManifoldCrossSection *manifold_slice(void *mem, ManifoldManifold *m,
+                                     float height) {
+  auto poly = from_c(m)->Slice(height);
+  return to_c(new (mem) CrossSection(poly));
+}
+
+ManifoldCrossSection *manifold_project(void *mem, ManifoldManifold *m) {
+  auto poly = from_c(m)->Project();
+  return to_c(new (mem) CrossSection(poly));
```
elalish, line 213: @geoffder Does this look right?

bindings/wasm/examples/worker.ts
```diff
@@ -95,15 +95,13 @@ function addMembers(
   const cls = module[className];
   const obj = areStatic ? cls : cls.prototype;
   for (const name of methodNames) {
-    if (name != 'cylinder') {
```
elalish, line 98: @geoffder Can I assume this was a debugging leftover of some kind?

bindings/c/manifoldc.cpp
geoffder: Yes (late I know, already merged). Sorry I've been swamped in trying to wrap up things with my PhD (and will be for a while) so I've been skimming too many of my manifold GitHub email notifications (though I've still been trying to read most of them to stay up to date).

bindings/wasm/examples/worker.ts
geoffder: Yea looks like it sorry!

bindings/c/manifoldc.cpp
elalish: No prob! I know how it is with wrapping up a PhD - good luck! It's better on the other side, I promise!

PR #646, opened 2023-12-11
[python] add revolveDegrees to revolve() API

wrongbad:
I tested a simple example locally. I built & installed the python package with this change. Generated a simple manifold by revolving a polygon 220 degrees, and previewed the result did the right thing.


elalish (approved):
Thanks for helping to keep our bindings up to date!

PR #647, opened 2023-12-11
[python] fix JoinType.Round and simplify(epsilon)

wrongbad:
3 more minor fixes for python API


elalish (approved):
Excellent, thanks for looking carefully!

PR #650, opened 2023-12-13
[python] manifold3d.triangulate and CrossSection.to_polygons_numpy

wrongbad:
I'm not super stoked about the name `CrossSection.to_polygons_numpy()`, so I'm open to suggestions there. 

Also added some helpers to make `vector<glm::*>` - numpy conversions pretty clean.

Side note: I spotted a random memory leak in `halfedge_tangent` and fixed it. Can move to a separate branch if desired.

### DIFF
--- a/bindings/python/manifold3d.cpp
+++ b/bindings/python/manifold3d.cpp
@@ -24,6 +24,7 @@
 #include "nanobind/stl/optional.h"
 #include "nanobind/stl/tuple.h"
 #include "nanobind/stl/vector.h"
+#include "polygon.h"
 #include "sdf.h"
 
 template <>
@@ -65,6 +66,36 @@ struct nanobind::detail::type_caster<glm::vec3> {
 };
 
 namespace nb = nanobind;
+
+// helper to convert std::vector<glm::vec*> to numpy
+template <class T, int N, glm::qualifier Precision>
+nb::ndarray<nb::numpy, T, nb::shape<nb::any, N>> to_numpy(
+    std::vector<glm::vec<N, T, Precision>> const &vecvec) {
+  // transfer ownership to PyObject
+  T *buffer = new T[vecvec.size() * N];
+  nb::capsule mem_mgr(buffer, [](void *p) noexcept { delete[](T *) p; });
+  for (int i = 0; i < vecvec.size(); i++) {
+    for (int j = 0; j < N; j++) {
+      buffer[i * N + j] = vecvec[i][j];
+    }
+  }
+  return {buffer, {vecvec.size(), N}, mem_mgr};
+}
+
+// helper to convert numpy to std::vector<glm::vec*>
+template <class T, size_t N, glm::qualifier Precision = glm::defaultp>
+std::vector<glm::vec<N, T, Precision>> to_glm_vector(
+    nb::ndarray<nb::numpy, T, nb::shape<nb::any, N>> const &arr) {
+  std::vector<glm::vec<N, T, Precision>> out;
+  out.reserve(arr.shape(0));
+  for (int i = 0; i < arr.shape(0); i++) {
+    out.emplace_back();
+    for (int j = 0; j < N; j++) {
+      out.back()[j] = arr(i, j);
+    }
+  }
+  return out;
+}
 
 using namespace manifold;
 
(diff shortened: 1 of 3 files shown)


wrongbad:
Also `CrossSection.to_polygons_numpy()` won't be the last function that I'll want a returns-numpy version of, so now would be a good time to think about how to deal with this in general. Another option might be adding an arg+default like `to_polygons(return_numpy=False)`.

bindings/python/manifold3d.cpp
```diff
@@ -1014,6 +1056,17 @@ NB_MODULE(manifold3d, m) {
           },
           "Returns the vertices of the cross-section's polygons as a "
           "List[List[Tuple[float, float]]].")
+      .def(
+          "to_polygons_numpy",
```
elalish, line 1060: This shows my lack of Python knowledge - I think we made `mesh` use numpy, but maybe not the rest? I'd love to understand the tradeoffs better from our Python users; @johnnybigert @axel-angel @zalo, any thoughts? Which APIs could use an update? I am leaning towards a boolean to provide numpy output rather than duplicating everything.

wrongbad: I'm also leaning toward `return_numpy` arg over the separate function names. 

As far as which functions next, personally I'd prioritize anywhere that uses open-ended lists/vectors to transfer possibly large numbers of scalars, as these will see the largest performance benefits. Other functions which take small tuples, or x,y,z args are probably not worth messing with - as python gives very clean syntax to convert between ndarray, tuple, and multi-args.

wrongbad: And a subtle point is that numpy wants uniform dimensions, so in cases like this one which are a list of variable length polygon points, I have it return a list of ndarrays as a compromise, to make the data significantly _more_ contiguous and compact even if it's not 100% contiguous.

zalo: I think it’s pretty safe to remove the bool, make NumPy a hard requirement, and to return it for everything all the time.  They’re functionally equivalent to traditional lists, but they add a ton of functionality (and operator overloads for batch math!)  For anything involving any math/numbers, It’s probably the most used/depended on Python package of all time by a wide margin.  It’s one of the few “libraries” that makes everything it touches just faster and better, and every Python user (who isn’t just scraping websites) already has it.

For ragged arrays, a list of numpy arrays is the best compromise.

I can go in and add the other APIs if that makes sense 👍
(I can also come up with a list of extremely popular libraries that also rely entirely on Numpy if you’re not convinced)

wrongbad: Some of the other APIs are already numpy-only so it's not really a question of dependency. It's more a question of introducing breaking API changes. The bool arg allows existing users to upgrade without breaking.

I'm a new user of manifold, so I personally have no problem with breaking API changes :P

wrongbad: I pushed the `return_numpy` version, since it's better than the previous version at least.

If we do decide to break the APIs to switch fully to numpy, we should try to bundle them all under a clear version bump, so people can pin to old version if not wanting to migrate.

wrongbad: > They’re functionally equivalent to traditional lists (so the change won’t break most code!)

Oh I missed this point earlier. That is a good point, many common interactions would still work, for example:

```
x = my_cross_section.to_polygons()
for poly in x:
  for x, y in poly
    # do stuff
```

wrongbad: Turns out one of the `run_all.py` test cases breaks with this change, see the new diff in run_all.py for details.
Arguably, creating a `set()` from point tuples is more of a contrived test fixture than a realistic use though.

elalish: Thanks @zalo, that's a good point about this being a nearly transparent change. And yeah, I'm not worried about that test change, @wrongbad. Honestly, I think I'm comfortable calling this part of v2.3 - I don't think we have quite enough users to be quite that picky about small breaking changes. Does that sound reasonable @pca006132?

pca006132: I think using numpy arrays directly is fine.

elalish (approved):
Looks good, thanks! Please hunt through our Python API and see if you find anything else that needs the numpy treatment - I'd like to make sure we're consistent everywhere before our next release.

PR #651, opened 2023-12-14
add WarpBatch - array process warp with single callback

wrongbad:
I introduce `WarpBatch` for both `Manifold` and `CrossSection`. 

The new callback signature is `void(manifold::VecView<glm::vec3>)` to allow for arbitrary contiguous storage. This moves the per-vertex tight-loop fully into the callback, where math vectorizations can now apply.

This is especially helpful for python users where scalar math is vastly slower, but array math can use compiled libs like numpy.

### DIFF
--- a/src/cross_section/src/cross_section.cpp
+++ b/src/cross_section/src/cross_section.cpp
@@ -567,21 +567,42 @@ CrossSection CrossSection::Transform(const glm::mat3x2& m) const {
  */
 CrossSection CrossSection::Warp(
     std::function<void(glm::vec2&)> warpFunc) const {
-  auto paths = GetPaths();
-  auto warped = C2::PathsD();
-  warped.reserve(paths->paths_.size());
-  for (auto path : paths->paths_) {
-    auto sz = path.size();
-    auto s = C2::PathD(sz);
-    for (int i = 0; i < sz; ++i) {
-      auto v = v2_of_pd(path[i]);
-      warpFunc(v);
-      s[i] = v2_to_pd(v);
+  return WarpBatch([&warpFunc](VecView<glm::vec2> vecs) {
+    for (glm::vec2& p : vecs) {
+      warpFunc(p);
     }
-    warped.push_back(s);
+  });
+}
+
+/**
+ * Same as CrossSection::Warp but calls warpFunc with
+ * a VecView which is roughly equivalent to std::span
+ * pointing to all vec2 elements to be modified in-place
+ *
+ * @param warpFunc A function that modifies multiple vertex positions.
+ */
+CrossSection CrossSection::WarpBatch(
+    std::function<void(VecView<glm::vec2>)> warpFunc) const {
+  std::vector<glm::vec2> tmp_verts;
+  C2::PathsD paths = GetPaths()->paths_;  // deep copy
+  for (C2::PathD const& path : paths) {
+    for (C2::PointD const& p : path) {
+      tmp_verts.push_back(v2_of_pd(p));
+    }
   }
+
+  warpFunc(VecView<glm::vec2>(tmp_verts.data(), tmp_verts.size()));
+
+  auto cursor = tmp_verts.begin();
+  for (C2::PathD& path : paths) {
+    for (C2::PointD& p : path) {
+      p = v2_to_pd(*cursor);
+      ++cursor;
+    }
+  }
+
   return CrossSection(
-      shared_paths(C2::Union(warped, C2::FillRule::Positive, precision_)));
+      shared_paths(C2::Union(paths, C2::FillRule::Positive, precision_)));
 }
 
 /**
--- a/src/utilities/include/vec_view.h
+++ b/src/utilities/include/vec_view.h
@@ -1,0 +1,112 @@
+// Copyright 2023 The Manifold Authors.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#pragma once
+
+#include <stdexcept>
+
+namespace manifold {
+
+/**
+ * View for Vec, can perform offset operation.
+ * This will be invalidated when the original vector is dropped or changes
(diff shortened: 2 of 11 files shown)


elalish:
This looks good to me - would you mind adding a python example for this? Perhaps convert our Heart or Torus Knot examples? It would also be a good excuse to see a performance comparison between the two warp approaches, which it would be nice to record here. 

I'm also curious if we should do the same thing for our JS bindings. Though perhaps WASM has less overhead in that tight loop, but I'm not sure. @pca006132 this also seems related to some discussions we've had around SDF. Does this seem like a useful pattern there as well?

pca006132:
SDF: https://github.com/elalish/manifold/pull/598

Yes this is also a useful pattern for SDF, but I think we have more things to consider for SDF in the previous PR.
This PR looks good to me as is.

wrongbad:
I added torus_knot.py showing fully vectorized warp math. We're gonna convert you to a python lover :P 

> I'm also curious if we should do the same thing for our JS bindings

If you have the time, sketch it up and run some timings on torus knot with high segment numbers. I'm sure the JIT helps a lot but even fully native code is punished by high numbers of function-pointer calls.

wrongbad:
Is there a helper script somewhere to auto-format everything? I've been reverse-engineering the commands to run from the CI outputs, and it's a bit of a pain.

pca006132:
Similar thing can help for js, the problem is that JIT is not good at optimizing across language boundaries.

We don't have a script right now, but it is just using the black formatter for python files.

wrongbad:
On my laptop, using the exact same numpy warp code but running 1 point at a time (Manifold.warp) it takes ~800ms. Running warp_batch takes about 40-50ms.

pca006132:
I think black is asking you to reformat `bindings/python/examples/torus_knot.py`

wrongbad:
Yeah, I think I have it figured out now. I just mean having to download black and learn how to use it based on failure reports. It was worse with clang-format - I figured out I needed to downgrade versions to get consistent with CI. A script would save some effort, especially for new people getting set up.

pca006132:
Interesting, I thought clang-format should have consistent settings across versions. Agreed, we should pin the detailed format settings and provide a script to format everything. Can be addressed in another PR.

pca006132 (approved):

wrongbad:
I'd like to change the function signature to `void(VecView<glm::vec3>)` - it supports range-for-loops, and also plays better with python bindings auto-casting to/from numpy.

src/cross_section/src/cross_section.cpp
```diff
@@ -567,21 +567,43 @@ CrossSection CrossSection::Transform(const glm::mat3x2& m) const {
...
+  return WarpBatch([&warpFunc](VecView<glm::vec2> vecs) {
+    for (glm::vec2& p : vecs) {
+      warpFunc(p);
+    }
+  });
+}
+
+/**
+ * Same as CrossSection::Warp but calls warpFunc with begin and end
+ * iterators to all the vertices to be warped.
```
elalish, line 579: Update docs for VecView?

src/cross_section/src/cross_section.cpp
```diff
@@ -567,21 +567,43 @@ CrossSection CrossSection::Transform(const glm::mat3x2& m) const {
...
+ * Note that warpFunc begin/end pointers follow c++ iterator
+ * conventions - the end is exclusive and not to be dereferenced
+ *
+ * @param warpFunc A function that modifies multiple vertex positions.
+ */
+CrossSection CrossSection::WarpBatch(
+    std::function<void(VecView<glm::vec2>)> warpFunc) const {
+  std::vector<glm::vec2> tmp_verts;
+  C2::PathsD paths = GetPaths()->paths_;  // deep copy
```
elalish, line 588: If this line the deep copy, or the for loops following?

src/cross_section/src/cross_section.cpp
```diff
@@ -567,21 +567,43 @@ CrossSection CrossSection::Transform(const glm::mat3x2& m) const {
...
+  for (C2::PathD& path : paths) {
+    for (C2::PointD& p : path) {
```
elalish, line 590: Our style is to use `const` anywhere is possibly can be (mostly for readability). Mind taking a pass through the PR to constify it?

src/utilities/include/vec_view.h
```diff
@@ -0,0 +1,112 @@
+// Copyright 2021 The Manifold Authors.
```
elalish, line 1: 2023

elalish (changes requested):
Looks good, just some nits.

elalish:
Regarding formatting, we generally use VSCode and their related extensions. I think that keeps things fairly synced, probably since MS also owns Github and Actions. What dev environment are you using?

pca006132:
No, vscode extension only calls the clang-format in your environment, so microsoft does not control what formatter you use :)

elalish:
I'm on clang-format v17 on my mac and it's working fine. What was the problem version you had, @wrongbad?

src/cross_section/src/cross_section.cpp
wrongbad: I copy it up front, then use it to update in-place after the warp func. Mainly because the code is simpler. Are you concerned with the initialization overhead doing it this way? I could optimize it if you think the complexity is worth it.

wrongbad:
I was also using v17 on mac and it was formatting this line in a way that didn't pass the CI, but v11 was working for me.
```
  nb::capsule mem_mgr(buffer, [](void *p) noexcept { delete[](T *) p; });
```

src/cross_section/src/cross_section.cpp
wrongbad: I fixed these and don't see any more, but let me know if I missed any.

pca006132:
Yes I also saw the same problem with clang-format 15. I think it changes when upgrading from 14 to 15.

src/cross_section/src/cross_section.cpp
pca006132: I think this is fine, I guess Emmett is just not sure about where the deep copy comment belongs to (the single line or including the loop after it).

elalish: Yeah, it kind of looked like it was just a one-line deep copy followed by a loop deep copy - I'm still not quite sure why.

elalish: Nevermind, I think I understand now. It's all good.

elalish (approved):
Thanks! Minor nit - for ease of review I'd appreciate it if you refrain from force-pushing. I know lots of people use rebase to avoid merge commits, but we do a squash and merge anyway so it makes no difference. If we stick to regular merging, then we can more easily see what changed since the last review.

PR #653, opened 2023-12-14
emscripten tbb stuff

pca006132:
This is mainly to reproduce the issues, not intended for merging. There are still weird issues with pthread enabled build, and the performance is not that good when running from the browser (it can run significantly slower).

1. Enabling the build is actually very easy, just need to specify `-pthread` to `CMAKE_CXX_FLAGS` and `CMAKE_EXE_LINKER_FLAGS`. We also need to set the cross origin policy bla bla bla to use SharedArrayBuffer, which is required for shared memory between workers.
2. It turns out the vite issue is https://github.com/vitejs/vite/issues/7015 and https://github.com/emscripten-core/emscripten/issues/20580, and there is also an issue from emscripten that uses `require` from the worker which does not work with ES6 (is this https://github.com/emscripten-core/emscripten/issues/17664 ? they said pthread supports ES6 though). I managed to fix by an ugly script.
3. For some reason I am unable to get anything run with node when pthread and ES6 module is used. This is the major issue and I am unable to get this work with the latest version of emscripten (3.1.51). You may need to modify the patch function to use `.\/built\/manifold.worker.js` instead of `built\/manifold.worker.js`. Somehow vite is only happy with the latter while node is only happy with the former.
4. For extra fun, try
   ```diff
   diff --git a/CMakeLists.txt b/CMakeLists.txt
   index 881182d..a1137d6 100644
   --- a/CMakeLists.txt
   +++ b/CMakeLists.txt
   @@ -57,9 +57,10 @@ endif()
    
    if(EMSCRIPTEN)
      message("Building for Emscripten")
   +  set(CMAKE_EXECUTABLE_SUFFIX_CXX ".mjs")
      set(MANIFOLD_FLAGS -fexceptions -D_LIBCUDACXX_HAS_THREAD_API_EXTERNAL -D_LIBCUDACXX_HAS_THREAD_API_CUDA)
      # ALLOW_MEMORY_GROWTH is slow, so we just allow 2gb of memory usage
   -  set(CMAKE_EXE_LINKER_FLAGS "${CMAKE_EXE_LINKER_FLAGS} -sINITIAL_MEMORY=2gb -sALLOW_MEMORY_GROWTH=0 -fexceptions -sDISABLE_EXCEPTION_CATCHING=0")
   +  set(CMAKE_EXE_LINKER_FLAGS "${CMAKE_EXE_LINKER_FLAGS} -sMODULARIZE=1 -sEXPORT_ES6=1 -sINITIAL_MEMORY=2gb -sALLOW_MEMORY_GROWTH=0 -fexceptions -sDISABLE_EXCEPTION_CATCHING=0")
      set(MANIFOLD_PYBIND OFF)
      if(MANIFOLD_PAR STREQUAL "TBB")
        set(CMAKE_CXX_FLAGS ${CMAKE_CXX_FLAGS} -pthread)
   ```
   and then you will find that nodejs somehow will not run `test/manifold_test.mjs` no matter how hard you try... Probably similar issue with 3.


kripken:
Regarding Emscripten's support for the combination of ES6+node+pthreads, it looks like it works but it does need a little boilerplate. Here is a PR in Emscripten to add a test for it (this was not tested before, it seems, even though it worked), which contains the working boilerplate, which is basically to create another `.mjs` file that does
```js
import initModule from "./hello.mjs";
initModule();
```
So the module needs to be imported, and also executed/called, I think.

pca006132:
Thanks, will give this a try later.

pca006132:
I think the main problem now is why manifold worker does not work on vitest.

elalish:
> I think the main problem now is why manifold worker does not work on vitest.

Can you elaborate a little? Are 1-4 no longer current? Can you update this PR so that the CI fails in the way you're describing?

pca006132:
No, this just means that 4 is no longer an issue. I just have some idea about a potential workaround, will try and update this PR.

CMakeLists.txt
```diff
@@ -58,8 +58,13 @@ endif()
 if(EMSCRIPTEN)
   message("Building for Emscripten")
   set(MANIFOLD_FLAGS -fexceptions -D_LIBCUDACXX_HAS_THREAD_API_EXTERNAL -D_LIBCUDACXX_HAS_THREAD_API_CUDA)
-  set(CMAKE_EXE_LINKER_FLAGS "${CMAKE_EXE_LINKER_FLAGS} -sALLOW_MEMORY_GROWTH=1 -fexceptions -sDISABLE_EXCEPTION_CATCHING=0")
+  # ALLOW_MEMORY_GROWTH is slow, so we just allow 2gb of memory usage
+  set(CMAKE_EXE_LINKER_FLAGS "${CMAKE_EXE_LINKER_FLAGS} -sMODULARIZE=1 -sEXPORT_ES6=1 -sINITIAL_MEMORY=2gb -fexceptions -sDISABLE_EXCEPTION_CATCHING=0")
   set(MANIFOLD_PYBIND OFF)
+  if(MANIFOLD_PAR STREQUAL "TBB")
+    set(CMAKE_CXX_FLAGS ${CMAKE_CXX_FLAGS} -pthread)
+    set(CMAKE_EXE_LINKER_FLAGS "${CMAKE_EXE_LINKER_FLAGS} -pthread -sPTHREAD_POOL_SIZE=\"(typeof window == 'undefined')?4:(()=>navigator.hardwareConcurrency)()\"")
```
pca006132, line 66: If we don't set it ourselves, `PTHREAD_POOL_SIZE` defaults to zero and I think it is a link-time constant, so we must set it to some sensible value here...

bindings/wasm/examples/vite.config.js
```diff
@@ -7,6 +7,18 @@ export default defineConfig({
   worker: {
     format: 'es',
   },
+  plugins: [
+    {
+      name: 'configure-response-headers',
+      configureServer: (server) => {
+        server.middlewares.use((_req, res, next) => {
+          res.setHeader('Cross-Origin-Embedder-Policy', 'require-corp');
```
pca006132, line 15: This is the cross-origin policy required for SharedArrayBuffer to work, there is no way around this. Hope that github pages supports this.

bindings/wasm/examples/worker.test.js
```diff
@@ -14,19 +14,11 @@
...
 import {expect, suite, test} from 'vitest';
 
-import Module from './built/manifold.js';
-import {readMesh, setupIO} from './gltf-io';
 import {examples} from './public/examples.js';
 import ManifoldWorker from './worker?worker';
 
-const io = setupIO(new WebIO());
-
-const wasm = await Module();
```
pca006132, line 27: This is the weird part. For some reason if we load `manifold.js` multiple times, the subsequent loads will not complete.

Perhaps this is caused by pthread worker loader implementation. No idea about this, will try to come up with a minimal example and submit the issue to emscripten.

I currently workaround this issue by having only a single manifold worker, and move the get properties and genus calls to `worker.ts`, but this is not ideal.

patch-emscripten.sh
```diff
@@ -0,0 +1,7 @@
+#/usr/bin/env bash
+if test -f $1/manifold.js; then
+  # I still have no idea why this is needed...
+  sed -i 's/new Worker/new (ENVIRONMENT_IS_NODE ? global.Worker : Worker)/g' $1/manifold.js
```
pca006132, line 4: I have no idea why `global.Worker != Worker`, considering there are no `let Worker` or `var Worker` anywhere...

Also, `Worker` is something like this:
```
class extends EventTarget {
    constructor(url, options2) {
      super();
      this._vw_workerTarget = new EventTarget();
      this._vw_insideListeners = /* @__PURE__ */ new Map();
      this._vw_outsideListeners = /* @__PURE__ */ new Map();
      this._vw_messageQueue = [];
      this.onmessage = null;
      this.onmessageerror = null;
      this.onerror = null;
      const context = {
        onmessage: null,
        name: options2 == null ? void 0 : options2.name,
        close: () => this.terminate(),
        dispatchEvent: (event) => {
          return this._vw_workerTarget.dispatchEvent(event);
        },
```

while `global.Worker` is this:
```
class Worker extends EventEmitter {
  constructor(filename, options = kEmptyObject) {
    super();
    debug(`[${threadId}] create new worker`, filename, options);
    if (options.execArgv)
      validateArray(options.execArgv, 'options.execArgv');

    let argv;
    if (options.argv) {
      validateArray(options.argv, 'options.argv');
      argv = ArrayPrototypeMap(options.argv, String);
    }

    let url, doEval;
    if (options.eval) {
      if (typeof filename !== 'string') {
        throw new ERR_INVALID_ARG_VALUE(
          'options.eval',
          options.eval,
          'must be false when \'filename\' is not a string',
        );
      }
      url = null;
      doEval = 'classic';
```

`global.Worker` is the correct target in this case.

patch-emscripten.sh
```diff
@@ -0,0 +1,7 @@
...
+fi
+sed -i 's/var nodeWorkerThreads=require("worker_threads");/const{createRequire:createRequire}=await import("module");var require=createRequire(import.meta.url);var nodeWorkerThreads=require("worker_threads");/g' $1/*.worker.js
```
pca006132, line 6: Just some ES6 compatibility issue. This is something upstream should fix, and pretty easy to fix.

patch-emscripten.sh
pca006132: it seems that nodejs dynamic import will define `Worker` for the module being imported, but I am still not sure why the Worker defined here is the wrong one.

pca006132:
and I do occasionally observe this kind of weird index out of bound issue when using pthread on the browser...

```
 ❯ worker.test.js  (9 tests | 1 failed) 8329ms
   ❯ worker.test.js > Examples > Gyroid Module
     → no glbURL
worker sent an error! undefined:undefined: table index is out of bounds
```

so even if we can make it build and work, not sure if we want to really use this in production for now.

elalish:
@kripken I think we have some more specific emscripten issues to look at now. Do you have any insights on why we're randomly getting `undefined:undefined: table index is out of bounds` with pthreads? And perhaps you can comment on `patch-emscripten.sh`?

kripken:
`table index is out of bounds` is unfortunately a very general error, something like a segfault. It's a little more specific than that, it's basically a segfault in the space of function pointers, but still it can be caused by any sort of memory corruption. Emscripten supports several [sanitizers](https://emscripten.org/docs/debugging/Sanitizers.html), which might help here. See also the [function pointers debugging issues](https://emscripten.org/docs/porting/Debugging.html#function-pointer-issues) section.

I'm not familiar enough with the worker global scope stuff to have a guess at the issues there. Is there a chance those can be reduced to standalone testcases? I ask because the first at least seems like it should be a general node.js issue with pthreads, but we do test that heavily in upstream, so there must be some difference in how you build or run the code (and maybe reducing it to a standalone testcase will lead us to find the difference).

pca006132:
OK I figured out the potential reason: `Worker` is defined by `vitest`.

```js
// a.test.js
import '@vitest/web-worker';
import {test} from 'vitest';
import b from './b';

test('foo', () => { b() })

// b.mjs
export default () => console.log(`worker: ${Worker}`)
```

this prints

```
worker: class extends EventTarget {
    constructor(url, options2) {
      super();
      this._vw_workerTarget = new EventTarget();
      this._vw_insideListeners = /* @__PURE__ */ new Map();
      this._vw_outsideListeners = /* @__PURE__ */ new Map();
      this._vw_messageQueue = [];
      this.onmessage = null;
      this.onmessageerror = null;
      this.onerror = null;
      const context = {
        onmessage: null,
        name: options2 == null ? void 0 : options2.name,
        close: () => this.terminate(),
        dispatchEvent: (event) => {
          return this._vw_workerTarget.dispatchEvent(event);
        },
        addEventListener: (...args) => {
          if (args[1])
            this._vw_insideListeners.set(args[0], args[1]);
          return this._vw_workerTarget.addEventListener(...args);
        },
        removeEventListener: this._vw_workerTarget.removeEventListener,
        postMessage: (...args) => {
          if (!args.length)
            throw new SyntaxError('"postMessage" requires at least one argument.');
          debug("posting message %o from the worker %s to the main thread", args[0], this._vw_name);
          const event = createMessageEvent(args[0], args[1], cloneType());
          this.dispatchEvent(event);
        },
```

No idea why `global.Worker` is different from `Worker`. Perhaps some weird scoping issue.

pca006132:
With the ES6 module issue being fixed in the upstream, we are now left with the function pointer thing and the Worker variable issue.

pca006132:
For some reason address sanitizer and ub sanitizer are really really slow (waited for several minutes for one test case and it did not finish, while it should normally finish within a few seconds) for the js binding, but seems to work fine (albeit slow) with the native test (wasm). Building with `-sASSERTIONS=1` did not reveal anything.

pca006132:
Interestingly when I tried to test with `-sMALLOC=mimalloc` on emscripten 3.1.51, the build succeeds but nodejs somehow refuses to utilize multiple threads. This is the only change and removing it makes pthread works again.

pca006132:
testing on browser with address sanitizer enables shows "RuntimeError: indirect call signature mismatch" without any address sanitizer error (except reporting leaks from tbb)

pca006132:
OK the function pointer issue is fixed: it turns out we were calling the javascript function from multiple threads and somehow it will cause problem only in the subsequent runs... weird

pca006132:
The performance on the browser is not bad: 

| Example | no pthread | with pthread |
| - | - | - |
| Rounded Frame | **0.16s** | 0.19s |
| Scallop | 0.69s | **0.62s** |
| Torus Knot | 0.44s | **0.41s** |
| Menger Sponge | **0.54s** | 0.82s |
| Stretchy Bracelet | 3.19s | **1.27s** |
| Gyroid Module | 3.02s | **2.54s** |

I am not entirely sure why menger sponge is significantly slower with pthread, probably overhead from parallelization being too fine?

pca006132:
found the culprit: emscripten does not like parallel copy. the performance is faster in all cases now.

pca006132:
@elalish it is ready now, the only problem here is the weird worker variable issue, remaining part of the patching script is to fix something that the upstream already fixes.

CMakeLists.txt
```diff
@@ -58,8 +58,13 @@ endif()
 if(EMSCRIPTEN)
   message("Building for Emscripten")
   set(MANIFOLD_FLAGS -fexceptions -D_LIBCUDACXX_HAS_THREAD_API_EXTERNAL -D_LIBCUDACXX_HAS_THREAD_API_CUDA)
-  set(CMAKE_EXE_LINKER_FLAGS "${CMAKE_EXE_LINKER_FLAGS} -sALLOW_MEMORY_GROWTH=1 -fexceptions -sDISABLE_EXCEPTION_CATCHING=0")
+  # ALLOW_MEMORY_GROWTH is slow, so we just allow 2gb of memory usage
```
elalish, line 61: Interesting, how slow (roughly)?

bindings/wasm/CMakeLists.txt
```diff
@@ -30,7 +30,14 @@ file(MAKE_DIRECTORY ${CMAKE_CURRENT_SOURCE_DIR}/examples/built)
...
 add_custom_target(js_deps ALL
         DEPENDS manifoldjs
-               ${CMAKE_CURRENT_SOURCE_DIR}/manifold*.d.ts)
+               ${CMAKE_CURRENT_SOURCE_DIR}/manifold*.d.ts
+               ${CMAKE_SOURCE_DIR}/patch-emscripten.sh)
+
+if(MANIFOLD_PAR STREQUAL "TBB")
+add_custom_command(
+        TARGET js_deps POST_BUILD
+        COMMAND ${CMAKE_SOURCE_DIR}/patch-emscripten.sh ${CMAKE_CURRENT_BINARY_DIR})
```
elalish, line 39: Why is the patch applied both here and in the manifold.yml file?

bindings/wasm/examples/worker.test.js
elalish: Did you ever submit this to emscripten?

bindings/wasm/examples/worker.test.js
```diff
@@ -58,29 +50,16 @@ async function runExample(name) {
...
-          reject('no glbURL)');
-        }
-        const docIn = await io.read(glbURL);
-        const nodes = docIn.getRoot().listNodes();
-        for (const node of nodes) {
-          const docMesh = node.getMesh();
-          if (!docMesh) {
-            continue;
-          }
-          const {mesh} = readMesh(docMesh);
```
elalish, line 70: So, the one problem I see with this refactor is we're no longer testing `readMesh`, and we're not really testing `writeMesh` either since the test isn't validating its output at all. These functions aren't actually dependent on `Manifold`, so perhaps we can slot them back in?

patch-emscripten.sh
elalish: Is this one that already got fixed upstream?

patch-emscripten.sh
elalish: From your other comment it sounds like you think this might be a `vitest` issue? Have we opened anything on their forums?

elalish (changes requested):
Amazing work on this! The only change I'd like to make sure gets in before this merges is to make sure our JS gltf I/O is still getting tested.

CMakeLists.txt
pca006132: Not sure, this is a compiler warning (https://github.com/WebAssembly/design/issues/1271), and it does not hurt to make it a fixed memory size. This probably affects our js callbacks as they perform quite a lot of memory read/write.

I think a fixed memory size should be fine, it does not occupy physical memory if not that much memory is actually used, and because wasm by default is using 32 bit pointers we cannot use more than 4gb of memory anyway. Forgot why I set it to 2gb though, perhaps just a random choice.

bindings/wasm/CMakeLists.txt
pca006132: The patch applied in `manifold.yml` is for patching the `test/manifold_test.js`, while the patch here is to patch the binding files. I can remove the `package.json` so running `test/manifold_test.js` does not require any patch. I added that mainly for testing and forgot to remove it.

bindings/wasm/examples/worker.test.js
pca006132: I forgot to try to reproduce this, will try later.

patch-emscripten.sh
pca006132: I think it is related to `vitest`, but I cannot make a minimal example yet and don't quite understand why it causes an issue, so I have not yet open anything there.

patch-emscripten.sh
pca006132: Yes, but I think they have not yet released a new version.

bindings/wasm/examples/worker.test.js
pca006132: Actually we do test writeMesh, in the `worker.ts` code I am reading the output of `writeMesh` for the property.

CMakeLists.txt
```diff
@@ -60,13 +60,14 @@ endif()
 if(EMSCRIPTEN)
   message("Building for Emscripten")
   set(MANIFOLD_FLAGS -fexceptions -D_LIBCUDACXX_HAS_THREAD_API_EXTERNAL -D_LIBCUDACXX_HAS_THREAD_API_CUDA)
-  # ALLOW_MEMORY_GROWTH is slow, so we just allow 2gb of memory usage
-  set(CMAKE_EXE_LINKER_FLAGS "${CMAKE_EXE_LINKER_FLAGS} -sMODULARIZE=1 -sEXPORT_ES6=1 -sINITIAL_MEMORY=2gb -fexceptions -sDISABLE_EXCEPTION_CATCHING=0")
```
elalish: Did you intend to remove MODULARIZE and EXPORT_ES6?  Or are they already set somewhere else?

flake.nix
```diff
@@ -103,10 +103,7 @@
                 emmake make -j''${NIX_BUILD_CORES}
               '';
               checkPhase = ''
-                cd test
-                echo '{"type":"module"}' > package.json
-                node --experimental-wasm-threads -e '(async () => {const module = await import("./manifold_test.js"); await module.default();})()'
```
elalish: I guess this is causing the nix failure somehow?

CMakeLists.txt
pca006132: yes, they are already set for the js binding

flake.nix
pca006132: probably, will check

bindings/wasm/examples/worker.test.js
elalish: Ah, I see it now - okay, that's perfect. Thanks!

elalish (approved):
I think I'm ready when you are for this one.

pca006132:
For some reason I am unable to get nodejs20 and nodejs21 to run with multiple threads... The `--experimental-wasm-threads` option is removed (people said it is now the default after v12) but I have trouble getting it working. Need more time to investigate...

pca006132:
I still don't have any idea about why nodejs20/21 fail to use multiple cores. I think we should probably just build everything and look into the nodejs thing later? what do you think @elalish 
I can clean this up and prepare for merging.

elalish:
SG to me - basically it's faster unless we have the latest node? That sounds like something we can reasonably follow up on on.

pca006132:
Another question: should we publish the version with tbb on npm as a separate package or replacing the current one?

elalish:
I would prefer to just have one version - what are the pros/cons?

pca006132:
I think the version with tbb is slower when only a single thread is used. As I don't know how to make nodejs use multiple threads for newer versions, this will likely be the case for users using it on npm. Also, shared array buffer requires `Cross-Origin-Embedder-Policy` which some websites may not be able to set?

The cons of sticking with the current single threaded implementation is that it is slower than multithreaded one when multiple threads are available, which does work on browser and older version of nodejs for now.

elalish:
Hmm, yeah that COEP thing is an issue - Safari doesn't even support `credentialless` yet. Okay, maybe two versions is best for now.

pca006132:
This is stalled for quite some time, I don't think people (e.g. me) are interested in this for now. Maybe we should wait until things are more mature (PSTL, oneTBB on emscripten).

PR #656, opened 2023-12-14
Python more numpy

wrongbad:
Note: this branch is on top of WarpBatch commit - that will disappear when that other PR merges.

Changed `hull_points` to accept numpy point arrays. Also removed several redundant numpy shape checks - these shape constraints were already enforced by the shape template parameters.

`sdf` / `level_set` could benefit from numpy batching but requires deeper refactoring (#598)

`set_properties` also could benefit, and I'd like to look into it more here.

That's all that I see. Let me know if I missed anything.

While I was auditing the API, I noticed something else needing cleanup. Many functions need  a single vector/point, but the convention is all over the map. Some take 3 floats, some take a tuple, some take a `glm::vec3`, some take a numpy(3), and several have overloads to support some random subset of those. We should probably converge on 1 way. My preference would be tuple, but I'm curious to hear other opinions.


pca006132:
There is a caster for `glm::vec3` that can accept anything that is a sequence with 3 elements, e.g. list, tuple and numpy arrays.

pca006132:
I think there are also some other tiny fixes that can be added into this PR.

Missing `\n` in line 371:
https://github.com/elalish/manifold/blob/eba866d266f1954f81211c71d9f5329477606e5e/bindings/python/manifold3d.cpp#L368-L375

This should be an int instead of a double:
https://github.com/elalish/manifold/blob/eba866d266f1954f81211c71d9f5329477606e5e/bindings/python/manifold3d.cpp#L997

wrongbad:
> Not sure if arguments accepting numpy arrays can accept normal python list.

A quick test suggests the answer is no. I guess the auto casting approach has the best upside. I'll tinker with it some more.

wrongbad:
I removed almost all of the lambda functions in the bindings. A few remain for overload-resolution, dealing with stray types like `Rect` or `Smoothness`, or doing something non-trivial.

wrongbad:
I also removed some overloads like `cube(float, float, float)` when `cube(vec3)` exists. I think right now a primary obstacle for maintainability and cleanliness is the sheer quantity of code in the manifold3d.cpp, so anything that reduces that code footprint is good.

pca006132:
Thanks for the cleanup. I think much of the code comes from docstrings, it would be nice if we can find a way to unify the documentation for the main C++ library and the 3 different bindings, but I guess this is a hard problem (e.g. name inconsistencies, specifics for certain binding).

elalish:
This looks awesome, thanks! I always appreciate cleanups that will ease maintenance going forward. I'll review more closely after the other PR is merged and this comes out of draft.

pca006132:
I think you probably need to push a dummy commit to trigger the CI build. We should fix this CI problem later...

bindings/python/manifold3d.cpp
```diff
@@ -168,106 +243,69 @@ NB_MODULE(manifold3d, m) {
...
-              new (self) Manifold(first);
-            } else {
-              new (self) Manifold();
-            }
-          },
-          "Construct manifold as the union of a set of manifolds.")
+      .def(nb::init<>(), "Construct empty Manifold object")
+      .def(nb::init<const MeshGL &>(),
+           "Convert a MeshGL into a Manifold, retaining its properties and "
+           "merging only"
```
pca006132, line 256: issues with space characters here.

bindings/python/manifold3d.cpp
```diff
@@ -168,106 +243,69 @@ NB_MODULE(manifold3d, m) {
...
-            return self.Transform(mat_glm);
-          },
-          nb::arg("m"),
+          "transform", &Manifold::Transform, nb::arg("mat"),
           "Transform this Manifold in space. The first three columns form a "
           "3x3 matrix transform and the last is a translation vector. This "
           "operation can be chained. Transforms are combined and applied "
           "lazily.\n"
           "\n\n"
           ":param m: The affine transform matrix to apply to all the vertices.")
```
pca006132, line 303: parameter name is different.

bindings/python/manifold3d.cpp
```diff
@@ -294,73 +332,35 @@ NB_MODULE(manifold3d, m) {
...
           "the user to choose their function with discretion."
           "\n\n"
           ":param f: A function that modifies a given vertex position.")
+      .def("warp_batch", &Manifold::WarpBatch, nb::arg("f"),
+           "Same as Manifold.warp but calls `f` with a "
+           "ndarray(shape=(N,3), dtype=float) and expects an ndarray "
+           "of the same shape and type in return. The input array can be "
+           "modified and returned if desired. "
+           "\n\n"
+           ":param f: A function that modifies multiple vertex positions.")
```
pca006132, line 349: thinking about it, we may want documentation about the type signature of the function. same for `warp` above and `set_properties` below.

bindings/python/manifold3d.cpp
```diff
@@ -498,47 +481,29 @@ NB_MODULE(manifold3d, m) {
...
+           "\n\n"
+           ":param cutter: This is the manifold to cut by.\n")
+      .def("split_by_plane", &Manifold::SplitByPlane, nb::arg("normal"),
+           nb::arg("origin_offset"),
+           "Convenient version of Split() for a half-space."
+           "\n\n"
+           ":param normal: This vector is normal to the cutting plane and its "
+           "length does not matter. The first result is in the direction of "
+           "this vector, the second result is on the opposite side.\n"
+           ":param originOffset: The distance of the plane from the origin in "
```
pca006132, line 502: parameter name

bindings/python/manifold3d.cpp
```diff
@@ -547,16 +512,13 @@ NB_MODULE(manifold3d, m) {
           "vector from the plane.\n"
           ":param originOffset: The distance of the plane from the origin in "
```
pca006132, line 513: parameter name

bindings/python/manifold3d.cpp
```diff
@@ -568,28 +530,24 @@ NB_MODULE(manifold3d, m) {
            "as empty. Likewise, empty meshes may still show NoError, for "
            "instance if they are small enough relative to their precision to "
            "be collapsed to nothing.")
-      .def_prop_ro(
+      .def(
```
pca006132, line 533: we should probably state that this is a breaking change, i.e. property vs method.

bindings/python/manifold3d.cpp
```diff
@@ -568,28 +530,24 @@ NB_MODULE(manifold3d, m) {
...
       .def_static(
           "smooth",
           [](const MeshGL &mesh,
-             const std::vector<std::tuple<int, float>> &sharpenedEdges = {}) {
-            std::vector<Smoothness> vec(sharpenedEdges.size());
-            for (int i = 0; i < sharpenedEdges.size(); i++) {
-              vec[i] = {std::get<0>(sharpenedEdges[i]),
-                        std::get<1>(sharpenedEdges[i])};
-            }
+             std::vector<std::pair<int, float>> sharpenedEdges) {
```
pca006132, line 545: I'm thinking if we want to make this numpy-friendly, maybe by splitting it into an integer index array and smoothness array.

bindings/python/manifold3d.cpp
```diff
@@ -656,49 +592,27 @@ NB_MODULE(manifold3d, m) {
...
+          nb::arg("circular_segments") = 0, nb::arg("center") = false,
           "A convenience constructor for the common case of extruding a "
           "circle. Can also form cones if both radii are specified."
           "\n\n"
           ":param height: Z-extent\n"
           ":param radiusLow: Radius of bottom circle. Must be positive.\n"
           ":param radiusHigh: Radius of top circle. Can equal zero. Default "
           "(-1) is equal to radiusLow.\n"
-          ":param circularSegments: How many line segments to use around the "
+          ":param circular_segments: How many line segments to use around the "
```
pca006132, line 605: need to change `radius_low` and `radius_high` as well.

bindings/python/manifold3d.cpp
```diff
@@ -946,91 +831,59 @@ NB_MODULE(manifold3d, m) {
...
-              throw std::runtime_error("Invalid matrix shape, expected (2, 3)");
-            glm::mat3x2 mat_glm;
-            for (int i = 0; i < 2; i++) {
-              for (int j = 0; j < 3; j++) {
-                mat_glm[j][i] = mat(i, j);
-              }
-            }
-            return self.Transform(mat_glm);
-          },
+          "transform", &CrossSection::Transform, nb::arg("mat"),
```
pca006132, line 866: `m` vs `mat`

bindings/python/manifold3d.cpp
```diff
@@ -1143,12 +972,8 @@ NB_MODULE(manifold3d, m) {
           "Default is calculated by the static Defaults.\n"
           ":param revolve_degrees: rotation angle for the sweep.")
       .def_static(
-          "square",
-          [](Float2 dims, bool center) {
-            return CrossSection::Square({std::get<0>(dims), std::get<1>(dims)},
-                                        center);
-          },
-          nb::arg("dims"), nb::arg("center") = false,
+          "square", &CrossSection::Square, nb::arg("dims"),
```
pca006132, line 975: `dims` vs `size`

pca006132 (changes requested):
some nitpicks.

btw, not sure if we want `Manifold const &self`, I think we typically write `const Manifold &self`, and order does not matter for references (`T&` is like `T const *` anyway).

wrongbad:
What do you think about `bounding_box` -> `bounding_box()` and `get_surface_area()` -> `surface_area()` and `get_volume()` -> `volume()`?

pca006132:
I think every methods on `Manifold` may require evaluating the underlying CSG graph, so they can be really expensive and should probably be methods instead of attributes. Removing the `get_` prefix seems fine to me.

bindings/python/manifold3d.cpp
wrongbad: Makes sense to me. Done

bindings/python/examples/all_apis.py
```diff
@@ -0,0 +1,96 @@
...
+    c = c.simplify()
+    c = CrossSection.square((1, 1))
+    p = c.to_polygons()
+    c = c.transform([[1, 0, 0], [0, 1, 0]])
+    c = c.translate((1, 1))
+    c = c.warp(lambda p: (p[0] + 1, p[1] / 2))
+    c = c.warp_batch(lambda ps: ps * [1, 0.5] + [1, 0])
+
+
+def all_manifold():
```
elalish, line 47: Thanks - this big basic test file is a great idea. Nice as an API index too.

elalish (approved):
That's a lot of great clean-up. LGTM; I'll leave it to @pca006132 for final approval since I'm no Python master.

elalish:
Looks like a pretty simple error the python test caught, but are we only running Python tests on some of our builds? Is that intentional @pca006132?

pca006132:
Forgot if this is intentional or not, we can just enable it for every build. We can also enable all tests even for draft PR, so we will not get skipped tests when we change the state from draft to ready for review.

I think I will wait for @wrongbad to signal this ready for review, not sure if he is done with it now.

wrongbad:
At this point I'm just hoping to get all the tests passing, and resolve any review concerns. I'll look at generating doc-strings in a follow-up.

pca006132:
I will have a look at it tmr. We can leave the documentation generation thing later. I feel that it should not be too complicated, but may require parsing the documentation to get a list of parameters for substitution. Can probably be done together with the python stub.

bindings/python/manifold3d.cpp
```diff
@@ -22,107 +22,180 @@
...
+    if (arr_cast.from_python(src, flags, cleanup)) {
+      int num_vec = arr_cast.value.shape(0);
+      value.resize(num_vec);
+      for (int i = 0; i < num_vec; i++) {
+        for (int j = 0; j < N; j++) {
+          value[i][j] = arr_cast.value(i, j);
+        }
+      }
+    } else {
+      int num_vec = PyObject_Size(src.ptr());  // negative on failure
```
pca006132, line 141: This is surprising. Just curious, any reference on this?

pca006132 (approved):
I think I still see some inconsistencies in parameter name for the docstring, but I guess we can pass this for now and embrace automation :)

bindings/python/manifold3d.cpp
wrongbad: So this function is `noexcept` and if you call `nb::len` on a `float` it throws, and terminates the program. So I looked into source for `nb::len` and found this function which gives a softer fallback mode.

pca006132:
So I guess this can be merged?

PR #659, opened 2023-12-15
use double instead of floats (#542)

### LINKED ISSUE #542: allow double precision
While it is not our main usecase, it might be desirable to be able to handle large meshes which requires using 64 bit indices. Note that the mesh itself may not be that large, but there can be many collisions that makes intermediate operation exceed the 32 bit integer range.

Apart from refactoring, we may hit https://github.com/NVIDIA/cccl/issues/744. Doing NVIDIA/thrust#541 first may help.

Also, using larger integer indices will probably come with a performance cost due to the additional memory transfers. I wonder if we should dynamically dispatch the operations to select the best size, int16/32/64 depending on the data size. This will make the code a bit more complicated (additional logic for dispatch, maybe we can make it something like the existing par.h API), but may be worth it.

pca006132:
Closes #542.

This changes all floats into doubles. There is a slight performance hit, but I don't think it is large enough to warrant allowing users to choose between floats and doubles.

Note that this is a breaking change, existing users need to change to using doubles, e.g. `float -> double`, `glm::vec3 -> glm::dvec3`, `glm::mat3 -> glm::dmat3`. Not sure if it is worth it to keep the older float API around.

Currently, we get the following test failures:
```
Polygon orientation issue:
[  FAILED  ] Polygon.CoincidentHole
[  FAILED  ] Polygon.Ugly
[  FAILED  ] Polygon.Sponge4a
[  FAILED  ] Polygon.SpongeThin
[  FAILED  ] Polygon.Sliver6
[  FAILED  ] Polygon.TouchingHole2
[  FAILED  ] Polygon.Degenerate2
[  FAILED  ] Polygon.Degenerate3
[  FAILED  ] Polygon.Degenerate4
[  FAILED  ] Polygon.Hole3
[  FAILED  ] Polygon.Comb
[  FAILED  ] Polygon.Comb2
[  FAILED  ] Polygon.Sponge4

[ RUN      ] Manifold.MeshRelation
/home/pca006132/code/manifold/test/test_main.cpp:398: Failure
Expected: (volumeP) <= (area * out.Precision()), actual: 1.3453593853989385e-15 vs 1.3353731364036562e-15

[ RUN      ] Samples.Bracelet
/home/pca006132/code/manifold/test/samples_test.cpp:193: Failure
  rect.min.x
    Which is: -35.88292583078146
  box.min.x
    Which is: -35.882925828651445

/home/pca006132/code/manifold/test/samples_test.cpp:194: Failure
  rect.min.y
    Which is: -35.88292583078146
  box.min.y
    Which is: -35.882925828651445

/home/pca006132/code/manifold/test/samples_test.cpp:195: Failure
  rect.max.x
    Which is: 35.88292583078146
  box.max.x
    Which is: 35.882925828651445

/home/pca006132/code/manifold/test/samples_test.cpp:196: Failure
  rect.max.y
    Which is: 35.88292583078146
  box.max.y
    Which is: 35.882925828651445

/home/pca006132/code/manifold/test/samples_test.cpp:198: Failure
  projection.NumContour()
    Which is: 7
  2

/home/pca006132/code/manifold/test/samples_test.cpp:201: Failure
  extrusion.Genus()
    Which is: 4
  1

[ RUN      ] Samples.GyroidModule
/home/pca006132/code/manifold/test/samples_test.cpp:224: Failure
The difference between bounds.min.z and 0 is 2.4203234083586267e-07, which exceeds precision, where
bounds.min.z evaluates to 2.4203234083586267e-07,
0 evaluates to 0, and
precision evaluates to 5.656854152679441e-13.

/home/pca006132/code/manifold/test/samples_test.cpp:225: Failure
The difference between bounds.max.z and size * glm::sqrt(2.0) is 2.4203234261221951e-07, which exceeds precision, where
bounds.max.z evaluates to 28.284271005429559,
size * glm::sqrt(2.0) evaluates to 28.284271247461902, and
precision evaluates to 5.656854152679441e-13.
```

The default `kTolerance` is now 1e-14. It seems that the polygon issue is not entirely due to kTolerance, changing kTolerance back to 1e-6 does not fix the issue. Changing the polygon code to use float internally + change kTolerance back to 1e-6 fixes those issues, so probably some precision handling thing in polygon?

Btw, I wished I can make this PR smaller, but I really need to find-and-replace so many things to get it compile.


pca006132:
Using this in the CCW function reduces the number of failures.
```
  if (area * area * 4 <= base2 * tol * tol || area * area * 4 <= kTolerance)
```

But there seems to be legitimate errors:
```
[ RUN      ] Polygon.Sliver6
area: -0.000004371138970, precision: 0.000000000000200
/home/pca006132/code/manifold/test/polygon_test.cpp:63: Failure
Expected: triangles = Triangulate(polys, precision) doesn't throw an exception.
  Actual: it throws manifold::geometryErr with description "triangulation is not entirely CCW!".

[ RUN      ] Polygon.SpongeMore
area: -0.000457247209419, precision: 0.000000000000008
/home/pca006132/code/manifold/test/polygon_test.cpp:66: Failure
Expected: triangles = Triangulate(Turn180(polys), precision) doesn't throw an exception.
  Actual: it throws manifold::geometryErr with description "triangulation is not entirely CCW!".

[  FAILED  ] Polygon.SpongeMore (0 ms)
[ RUN      ] Polygon.Sponge3a
[       OK ] Polygon.Sponge3a (1 ms)
[ RUN      ] Polygon.Sponge4
area: -0.000406441100005, precision: 0.000000000000010
/home/pca006132/code/manifold/test/polygon_test.cpp:66: Failure
Expected: triangles = Triangulate(Turn180(polys), precision) doesn't throw an exception.
  Actual: it throws manifold::geometryErr with description "triangulation is not entirely CCW!".

[  FAILED  ] Polygon.Sponge4 (0 ms)
[ RUN      ] Polygon.BigSponge
area: -0.000152415604760, precision: 0.000000000000010
/home/pca006132/code/manifold/test/polygon_test.cpp:66: Failure
Expected: triangles = Triangulate(Turn180(polys), precision) doesn't throw an exception.
  Actual: it throws manifold::geometryErr with description "triangulation is not entirely CCW!".
```

wrongbad:
Since you have a "types.h" already, I'd recommend something like this.
```
namespace manifold {
using float_t = double;
using vec3 = glm::vec<float_t, 3, glm::defaultp>;
...
}
```
And then replace `glm::vec3` with `manifold::vec3` all over.
I promise it will come in handy again in the future to quickly toggle float type for some experiments.

elalish:
Looking at e.g. Samples.Bracelet, it looks like a difference between Clipper's rounding errors and ours (2D vs 3D). I'd guess the issue is that Clipper uses integer coordinates (and we're not handling the scaling of those as well as we could be), while we're using doubles. When it was all getting truncated to floats it was close enough that things ended up equal. Since Clipper uses 64-bit ints and doubles only have 53 bits of mantissa, Clipper should still be more precise, but only if it's actually scaled properly. Maybe that would fix it? I think that means we'd need to do the conversion from PathD to Path64 ourselves before calling any Clipper functions, but that doesn't cost anything because that's what happens internally anyway. 

Either that or we just accept a precision's worth of tolerance in our tests, which is also reasonable. Also, it sounds like there are probably more problems than just this. It's worth checking if some of the failing polygon tests were epsilon-valid with their old default tolerance, but not with the new one. We may just need to set their precision manually in the `TestPoly` calls.

pca006132:
Indeed, forgot about checking those polygons.

pca006132:
We may also want to think about how to merge this breaking change. Maybe wait until all the recent python PRs are merged, release a version first, and then work on this?

pca006132:
> Since you have a "types.h" already, I'd recommend something like this.
> 
> ```
> namespace manifold {
> using float_t = double;
> using vec3 = glm::vec<float_t, 3, glm::defaultp>;
> ...
> }
> ```
> 
> And then replace `glm::vec3` with `manifold::vec3` all over. I promise it will come in handy again in the future to quickly toggle float type for some experiments. And when the API switches to `manifold::vec3` things like this stop being breaking changes.

I thought about this, I just don't think there will be future changes that requires changing back to float, the performance different is not worth the complicated code.

wrongbad:
> I just don't think there will be future changes that requires changing back to float, the performance difference is not worth the complicated code.

It's not that you predict changing it again, but It's more of an API best practice, and insurance policy against things you might find a compelling reason to change later. Aside from float/double, you'd also have abstraction around the `glm` implementation detail, and unblock yourself from exploring different tensor math libraries in the future.

I'd argue there's no real complexity doing this first pass with typedefs, you just have like 5-10 extra lines in types.h, and all the same find-replace touch-points as before.

pca006132:
Sounds good. My point about complexity is not really about adding a type alias, but is about switching between different precision, which does not apply here.
Will do the change later, after the master branch is stable (after merging your python PR and probably release a new version).

zalo:
Excited to see what the timing differences are after the types.h change makes it easy to try them out!

The voronoi fracture stuff all happens in `double` space by default, so it might also be interesting to have it use whatever manifold is configured to use 💪

pca006132:
Here are some results running in i5 13600KF @ 3.5GHz with turbo boost disabled, `MANIFOLD_DEBUG=OFF`, `MANIFOLD_PAR=TBB` and using the [mimalloc](https://github.com/microsoft/mimalloc) allocator (which improves performance over the system default glibc allocator significantly).

Parallel:

| Sample            | float (ms) | double (ms) |
| ----------------- | ---------- | ----------- |
| Knot13            | 12         | 14          |
| Knot42            | 8          | 8           |
| Scallop           | 51         | 57          |
| TetPuzzle         | 6          | 6           |
| FrameReduced      | 1          | 2           |
| Frame             | 12         | 11          |
| Bracelet          | 217        | 260         |
| GyroidModule      | 288        | 321         |
| Sponge1           | 0          | 0           |
| Sponge4           | 3352       | 3536        |
| CondensedMatter16 | 276        | 292         |
| CondensedMatter64 | 4420       | 4743        |

Single threaded:

| Sample            | float (ms) | double (ms) |
| ----------------- | ---------- | ----------- |
| Knot13            | 26         | 29          |
| Knot42            | 16         | 17          |
| Scallop           | 198        | 235         |
| TetPuzzle         | 18         | 18          |
| FrameReduced      | 1          | 1           |
| Frame             | 16         | 17          |
| Bracelet          | 1301       | 1371        |
| GyroidModule      | 2360       | 2776        |
| Sponge1           | 0          | 0           |
| Sponge4           | 16581      | 16896       |
| CondensedMatter16 | 2014       | 2030        |
| CondensedMatter64 | 37295      | 37208       |

perf stat result shows that there are more page faults and memory access, causing slowdown, which is expected.

elalish:
Yeah, that perf difference looks quite minor compared to getting so much more precision. This looks excellent. Probably v3.0? May be a good time for everyone to make a request list for their favorite breaking changes.

pca006132:
Sounds good.

pca006132:
At least Polygon.CoincidentHole is not epsilon-valid under the new precision:
image

Wonder if there is a way to automatically determine if some polygon is epsilon-vaid given an epsilon, that would can be very helpful.

elalish:
Well, you could try triangulating it at various epsilon values, but that would involve trusting our triangulator. I would just set the precision of all failing polygon tests to the auto-calculated float value. No need to optimize them too much.

pca006132:
I will close this for now and open another PR later.

PR #660, opened 2023-12-15
add formatting script

pca006132:
This also fix the version for clang-format to 14. I think this script should work for other distros if they have clang-14 installed, would be nice if someone can help me test it.

### DIFF
--- a/.github/workflows/check_format.yml
+++ b/.github/workflows/check_format.yml
@@ -17,7 +17,7 @@ jobs:
         source: '.'
         exclude: '*/third_party'
         extensions: 'h,cpp,js,ts,html'
-        clangFormatVersion: 12
+        clangFormatVersion: 11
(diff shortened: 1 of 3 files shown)


wrongbad:
On MacOS, it's not super convenient to get clang-14. Brew has v8, v11, and v17. `pip install 'clang-format==14.*'` works though, but it just puts a `clang-format` in the virtual env and on the `PATH` so it won't be found the way your script works.
I was using v11 because it wasn't diverging from the v12 in the CI.

All that to say I'm not sure what the answer is here.

pca006132:
Interesting, I thought the CI is using v14.
I think the script can add a fallback to pip if clang-14 is not found?

pca006132:
I think I can just make it use v11.

elalish (approved):
Thanks!

PR #666, opened 2023-12-20
Add Naive Minkowski Sum Implementation

zalo:
This is the minimal code to get the naive minkowksi sum working, with basic affordances for multithreading.
This is essentially the other PR, but without the in-progress experimentation or the controversial `voro++` dependency.

The benefits of the Naive Technique are that:
1) The output mesh topology looks much nicer  without any need for surface simplification.
2) It doesn't need any additional dependencies.

The drawbacks to the Naive Technique are that:
1) Sometimes, it's a bit slower than the general technique (especially for Non-Convex/Non-Convex).
2) Sometimes, it appears to run out of memory when performing the boolean (especially during Non-Convex/Non-Convex).

There are two basic versions of the function, one with significant threading and the other without.   
I expect we'll consolidate to one or the other as their respective pros/cons make themselves apparent; the speed appears to be pretty comparable on my windows machine (but perhaps that's not true on Linux?).

PR #668, opened 2023-12-20
Add Offset

elalish:
Inspired by and related to #666, this is an Offset function, so a Minkowski only with a sphere (positive or negative). I used convexity to apply fewer edge and vert Booleans.

### DIFF
--- a/src/manifold/src/manifold.cpp
+++ b/src/manifold/src/manifold.cpp
@@ -52,13 +52,13 @@ struct ConvexEdge {
     const Halfedge edge = halfedge[idx];
     if (!edge.IsForward()) return false;
 
+    const glm::vec3 normal0 = faceNormal[edge.face];
+    const glm::vec3 normal1 = faceNormal[halfedge[edge.pairedHalfedge].face];
+    if (glm::all(glm::equal(normal0, normal1))) return false;
+
     const glm::vec3 edgeVec = vertPos[edge.endVert] - vertPos[edge.startVert];
-    const bool convex =
-        inset ^
-        (glm::dot(edgeVec,
-                  glm::cross(faceNormal[edge.face],
-                             faceNormal[halfedge[edge.pairedHalfedge].face])) >
-         0);
+    const bool convex = ((inset ? -1 : 1) *
+                         glm::dot(edgeVec, glm::cross(normal0, normal1))) > 0;
     if (convex) {
       vertConvex[edge.startVert] = true;
       vertConvex[edge.endVert] = true;
@@ -612,10 +612,11 @@ Manifold Manifold::Offset(float delta, int circularSegments) const {
   }
 
   const bool inset = delta < 0;
+  const float radius = glm::abs(delta);
   const int n = circularSegments > 0 ? (circularSegments + 3) / 4
                                      : Quality::GetCircularSegments(delta) / 4;
-  const Manifold sphere = Manifold::Sphere(delta, circularSegments);
-  const Manifold cylinder = Manifold::Cylinder(1, delta, delta, 4 * n);
+  const Manifold sphere = Manifold::Sphere(radius, circularSegments);
+  const Manifold cylinder = Manifold::Cylinder(1, radius, radius, 4 * n);
   const SimplePolygon triangle = {{-1, -1}, {1, 0}, {0, 1}};
   const Manifold block = Manifold::Extrude(triangle, 1);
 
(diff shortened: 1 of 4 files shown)


elalish:
As @zalo also found, this is basically a fuzz test for the Boolean due to using so many coplanar objects. The outset looks reasonable:
image

but it has a terrible genus, and the inset makes it clearer what is going wrong internally:
image

Most of these shards are due to bad symbolic perturbation (I think). Then there's lots of trouble with CW triangulation, due to self-intersecting polygons - I have a hunch about this problem too. And of course it's highly order-dependent and not infrequently seg-faults, which is the worst part. Since `Vec` isn't catching it, it must be somewhere unusual. 

Anyway, even if this doesn't eventually become our API, it's serving as a good test case.

zalo:
Nice!  The inset parameter makes a lot of sense; I feel silly for not including it.

Do you notice any seams/artifacts between the extruded triangles and the low-resolution spheres?  If so, I feel like it might be worth rotating the spheres/cylinders so they have at least one vertex aligned with the adjacent extruding triangle 🤔

Also, I wonder at the speed difference between hulling two spheres and unioning two spheres and a cylinder…   I’ll look at the implementation more carefully tomorrow when I’m on a real PC again 😄

elalish:
Yes, there are little dimples - I'm considering tweaking the precision to see if I can just decimate them out. Our spheres are tessellated octahedra, which I think already have the nicely aligned verts you're thinking of. I don't think it's worth rotating though, since having only one vert aligned isn't much better than none. 

The idea with unioning these separately is that with the hulls you're effectively unioning the same sphere 6 times (since that vert is part of 6 triangles), where I'm just doing it once. Plus, unioning shapes with identical faces is the slowest part, since you get lots of collisions to check. Still, I'm curious how much difference it makes in practice; would love to see some comparisons!

zalo:
I added just the Naive NonConvex-Convex method to your branch as a point of comparison:
https://github.com/elalish/manifold/compare/offset...zalo:manifold:offset-comparison?w=1
```
[ RUN      ] Boolean.Offset
C:\Repositories\manifold\test\boolean_test.cpp(601): error: Expected equality of these values:
  fat.Genus()
    Which is: -57
  cutout.Genus()
    Which is: 5

C:\Repositories\manifold\test\boolean_test.cpp(602): error: Expected equality of these values:
  thin.Genus()
    Which is: -78
  -7

[  FAILED  ] Boolean.Offset (5930 ms)
[ RUN      ] Boolean.OffsetHull
[       OK ] Boolean.OffsetHull (10095 ms)
```

Seems like it's twice as slow, but it returns the correct Genus 🤔 

I believe the issue is that the triangle extrusions probably don't match up perfectly with the edges of the spheres and cylinders.
image

The hulling method gets around this by just sweeping the same low-res sphere everywhere, but of course, that's where all of the slowness comes from.

Perhaps the hull method can be sped up by sweeping just a few vertices of the sphere on triangle faces, but the whole thing on convexEdges/Points?

elalish:
Yeah, you get the right genus because you aren't unioning barely-touching extrusions like I am. I'm going to focus on fixing the Boolean problems this has revealed.

zalo:
If you'd like some mesh inspiration from OpenCascade's Offset Operation:
https://zalo.github.io/CascadeStudio/#code=y0ktUchPSytOLQlKTMksLVawVQjOyUxJLdJQgggo6SiY65nqKBgo6CiYGmhac%2BUAdRQXZKQWpYLUghkaJqZQieTSJJCwU36FhjlQEwyXFJWmomjNSS0udoaoDSlKzCvOSSxJ1Yg2AFqjo2AMtE8b2U2xOgoumWlpQG15yakaICt0FKIh5sRqQo2FqA9GN9wfLKyBaqkOio81rQE%3D&gui=q1ZKzs8tyM9LzSvxS8xNVbJSSk4sTk5MSQ3LTC1X0lHyTS3OCEotVrIy0DPUUXJOTM5ItVeyKikqTdVRci%2FKL81LCchJzAOJpSXmFIMFM1PgKoISUzJLgZrN9UxrAQ%3D%3D 

It's kind of unfair given the amount of code OpenCascade needs to achieve this, but look at that topology! 😄
[SpherelessCubeOffset.stl.txt](https://github.com/elalish/manifold/files/13732692/SpherelessCubeOffset.stl.txt)

pca006132:
Nice work! I have less capacity currently as I am working on a new openscad interpreter (https://github.com/pca006132/sscad). It is basically a minimal implementation of openscad's script, with cleaner code, less dependencies and hopefully faster speed. The goal is to use it in manifoldcad, python (maybe @wrongbad will be interested), and eventually as experimental alternative code backend for openscad. I want to get something working before next year, so I am pretty busy with it. Will have a look at offset and minkowski later after I get something running.

elalish:
Closing in favor of #666.

src/manifold/src/manifold.cpp
```diff
@@ -569,6 +595,92 @@ Manifold Manifold::Warp(std::function<void(glm::vec3&)> warpFunc) const {
...
+               const Halfedge halfedge = pImpl->halfedge_[convexEdges[idx]];
+               glm::vec3 edge = pImpl->vertPos_[halfedge.endVert] -
+                                pImpl->vertPos_[halfedge.startVert];
+               const float length = glm::length(edge);
+               // Reverse RotateUp
+               edge.x *= -1;
+               edge.y *= -1;
+               batch[edgeOffset + idx] =
+                   cylinder.Scale({1, 1, length})
+                       .Transform(RotateUp(edge))
```
elalish, line 671: @starseeker This might help you.

PR #674, opened 2023-12-26
[python] fix warp callback

wrongbad:
I recently broke this making it a no-op due to the vec reference becoming a by-value copy in python callback. 

This fix provides a simple adapter like we had before allowing python to return modified values from the callback.

### DIFF
--- a/.github/workflows/manifold.yml
+++ b/.github/workflows/manifold.yml
@@ -23,7 +23,7 @@ jobs:
       run: |
         sudo apt-get -y update
         DEBIAN_FRONTEND=noninteractive sudo apt install -y libglm-dev libgtest-dev libassimp-dev git libtbb-dev libthrust-dev pkg-config libpython3-dev python3 python3-distutils python3-pip lcov
-        pip install trimesh
+        pip install trimesh pytest
     - uses: actions/checkout@v3
       with:
         submodules: recursive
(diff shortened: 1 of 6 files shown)


elalish:
Can you add a test that shows the problem you're solving? Ideally using one of our warp sample models?

wrongbad:
I updated torus_knot to expose both warp modes, so we can then test it in a ipynb like this:
```py
import sys
sys.path.append('bindings/python/examples')
import torus_knot
m = torus_knot.run(warp_single=True)

from badcad import Solid
Solid(m) # visualize in jupyter / vscode
```
Before:
Screen Shot 2023-12-26 at 6 34 12 PM
After:
Screen Shot 2023-12-26 at 6 35 21 PM

wrongbad:
Did you want an automated regression test specifically? I guess we could check `len(m.split_by_plane().decompose())` as it would be different when the warp actually works.

wrongbad:
I might need your help hooking up the CI side. I added those checks to a pytest setup. 

To run it, the scripts needs to pip install `manifold3d` package, and `pytest` package.
Then run `pytest` or `python -m pytest` from the `manifold` root.

Right now it will search python/examples for `test_*` files, and run all `test_*` functions, noting any assert failures.

elalish:
I *think* you just need to add `pytest` [here](https://github.com/elalish/manifold/blob/master/.github/workflows/manifold.yml#L26).

elalish (approved):
Excellent, thank you! I appreciate the assertion pattern - we should definitely expand on that to match our JS testing.

Issue #679, opened 2023-12-31
No way to opt-out of SimplifyTopology when creating Manifold from MeshGL


wrongbad:
So I'm stuck again getting my verts and triangles auto-deleted before I can use them.

This time I prototyped a `refine_to_length()` function in python using MeshGL interface - if the longest edge is >threshold, cut it in half.

```py
from badcad import *
t = text('yo', size=10, font='monospace')
yo = t.offset(1,'round').extrude(2) - t.extrude(2)
yom = yo.to_mesh()
verts = yom.vert_properties.astype(np.float32)
tris = yom.tri_verts.astype(np.uint32)

def refine_to_length(verts, tris, target=1):
    verts = verts.tolist()
    tris = tris.tolist()
    mids = {}
    done = False
    while not done:
        done = True
        i = 0
        while i < len(tris):
            tri = tris[i]
            i += 1
            v = [verts[i] for i in tri]
            dv = v - np.roll(v, 1, 0)
            l = np.linalg.norm(dv, axis=-1)
            mi = np.argmax(l)
            ml = l[mi]
            if ml > target:
                key = (min(tri[mi],tri[mi-1]), max(tri[mi],tri[mi-1]))
                if key not in mids:
                    mididx = len(verts)
                    midv = [(v[mi][j] + v[mi-1][j])/2 for j in [0,1,2]]
                    verts += [midv]
                    mids[key] = mididx
                else:
                    mididx = mids[key]
                tri2 = [*tri]
                tri2[mi-1] = mididx
                tris += [tri2]
                tri[mi] = mididx
                done = False
    return np.array(verts), np.array(tris)

verts, tris = refine_to_length(verts, tris, 1)

display((verts, tris), wireframe=True)
```
<img width="549" alt="Screen Shot 2023-12-31 at 11 59 10 AM" src="https://github.com/elalish/manifold/assets/10201449/9fd87036-4c6a-40f1-9c45-82221e9cddcb">

But when I import it into a `Manifold` in order to use for CSG and Warp, it forcibly simplifies back

```py
display(Manifold(manifold3d.Mesh(verts, tris)), wireframe=True)
```
<img width="540" alt="Screen Shot 2023-12-31 at 12 00 08 PM" src="https://github.com/elalish/manifold/assets/10201449/9214089e-6889-42cb-b294-2a3aa86503a4">

wrongbad:
[...]

PR #680, opened 2023-12-31
no simplify in Manifold from Mesh

wrongbad:
see https://github.com/elalish/manifold/issues/679


wrongbad:
I see this breaks a lot of test cases. I guess this is not the solution.

PR #687, opened 2024-01-07
Install libpolygon

cjmayo:
libmanifold depends on libpolygon, if installed it needs libpolygon.

### DIFF
--- a/src/polygon/CMakeLists.txt
+++ b/src/polygon/CMakeLists.txt
@@ -26,3 +26,14 @@ target_link_libraries( ${PROJECT_NAME}
 
 target_compile_options(${PROJECT_NAME} PRIVATE ${MANIFOLD_FLAGS})
 target_compile_features(${PROJECT_NAME} PUBLIC cxx_std_17)
+
+install(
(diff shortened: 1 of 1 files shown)


elalish:
@pca006132 I think you added the `install` to collider - what does that do? Should we make similar changes like this in other folders?

pca006132:
The original idea is to make them independent libraries that can be used without libmanifold. I guess I forgot to expose libpolygon.

Not sure why object library is needed. I think you can already link everything statically.

pca006132:
I think they are usable independently, but not sure if there are actual users that want to use them independently.

I will look into the object library thing today and see if it is easy to use, to be honest I am still not familiar with cmake after writing a bunch of cmake code here, too complicated for mortals to understand.

pca006132:
I think we should probably just merge everything into a single library file, that way it should be easier for users. I am thinking about how to deal with clipper2 dependency (if the system can provide it) and how to make something like `manifoldConfig.cmake`. I have some progress now, will open a PR later today, which should close this and #690.

PR #690, opened 2024-01-10
Build collider and polygon as object libraries

cjmayo:


elalish:
This seems reasonable to me, but CMake is not at all my forte. Would you mind explaining what effect this change has?

Issue #691, opened 2024-01-10
Bug in Python API


briansturgill:
Essentially you made Extrude and Revolve non-static members of CrossSection, when they
are in fact static members of Manifold.
I "fixed this" thus, but I really don't understand nanobind, it works, but I'm guessing here:
(Basically, I moved the routines up with Manifold routines, changed them to .def_static and added a crossSection arg.)

```
diff --git a/bindings/python/manifold3d.cpp b/bindings/python/manifold3d.cpp
index b8d2d57..544c8f3 100644
--- a/bindings/python/manifold3d.cpp
+++ b/bindings/python/manifold3d.cpp
@@ -384,6 +384,14 @@ NB_MODULE(manifold3d, m) {
       .def_static("tetrahedron", &Manifold::Tetrahedron, manifold__tetrahedron)
       .def_static("cube", &Manifold::Cube, nb::arg("size") = glm::vec3{1, 1, 1},
                   nb::arg("center") = false, manifold__cube__size__center)
+      .def_static(
+          "extrude", &Manifold::Extrude, nb::arg("crossSection"), nb::arg("height"),
+          nb::arg("n_divisions") = 0, nb::arg("twist_degrees") = 0.0f,
+          nb::arg("scale_top") = std::make_tuple(1.0f, 1.0f),
+          manifold__extrude__cross_section__height__n_divisions__twist_degrees__scale_top)
+      .def_static("revolve", &Manifold::Revolve, nb::arg("crossSection"), nb::arg("circular_segments") = 0,
+           nb::arg("revolve_degrees") = 360.0,
+           manifold__revolve__cross_section__circular_segments__revolve_degrees)
       .def_static(
           "cylinder", &Manifold::Cylinder, nb::arg("height"),
           nb::arg("radius_low"), nb::arg("radius_high") = -1.0f,
@@ -649,14 +657,6 @@ NB_MODULE(manifold3d, m) {
           nb::arg("pts"), cross_section__hull__pts)
       .def("decompose", &CrossSection::Decompose, cross_section__decompose)
       .def("to_polygons", &CrossSection::ToPolygons, cross_section__to_polygons)
-      .def(
-          "extrude", &Manifold::Extrude, nb::arg("height"),
-          nb::arg("n_divisions") = 0, nb::arg("twist_degrees") = 0.0f,
-          nb::arg("scale_top") = std::make_tuple(1.0f, 1.0f),
[...]

PR #692, opened 2024-01-10
Extrude and Resolve should be static on Manifold.

briansturgill:
OK, the python binding has the Extrude and Revolve API calls as being member functions on CrossSection, when
in fact they are static on Manifold.
This PR fixes the problem.
#691


elalish:
@wrongbad would you like to take a look at this PR? Technically this is a breaking change, and it's a touch arbitrary. We leave this as a static method in the C++ to keep `CrossSection` from having a dependency on `Manifold`, but they're all combined in the Python lib anyway. Still, consistency does seem nice from a docs perspective. 

@briansturgill I believe the CLA is an automatic click through if you sign as an individual. It's pretty standard stuff, though slightly annoying.

briansturgill:
@elalish CLA has now been signed.
Sorry, I didn't realize you had automated testing for Python and didn't fix the test.
Can I amend a pull request?
The issue isn't entirely arbitrary. It's kind of an accident that it was working before.
Because the first argument was a CrossSection and "this" is passed in most C++ implementations
as a first argument. It accidently worked as a member function... but it seems unwise to leave it that way.

elalish:
Ah, good call, thanks for the explanation. Yes, just add another commit to your branch and push - the tests will automatically rerun.

briansturgill:
I have no idea what to do about the code formatting. It is very unclear from the "detail" what is being complained about.

elalish:
Have you tried our formatting script? https://github.com/elalish/manifold?tab=readme-ov-file#formatting

wrongbad:
From my POV...
pro: consistency with c++, and c++ documentation
con: less chainable (e.g. `CrossSection(pts).offset(2).mirror((0,1)).extrude(10)`) but that can be worked around.
I'm fine with it.

geoffder:
You pass in a function to use for the binding, which just happens to be a static method on `Manifold`, I wouldn't say that is "on accident".

Also I don't really see the point of simultaneously removing it from CrossSection since it is convenient to have there. We only don't have it that way in CPP because of the dependency.

briansturgill:
@geoffder
The file instructed nano bind that a static method was a member method.
Nanobind caused the call to be cs.member(...)
However, that was really a call to another class which was static, fortunately it needed cs as it's first argument.
It _accidentally_ worked because clang does member calls by pushing "this" on the stack as the first argument.
There is nothing in the C++ standard which says that must work.
Different compiler, potentially a different result.
Regardless, it is invalid code and probably shouldn't be used.

geoffder:
Then the solution should rather be to write a wrapper function, than to remove the functionality entirely.

wrongbad:
> The file instructed nano bind that a static method was a member method.
> Nanobind caused the call to be cs.member(...)
> However, that was really a call to another class which was static, fortunately it needed cs as it's first argument.

Nanobind is a strongly typed c++ library. The only way to do what you're saying is to cast the function pointer through a `void*`. Much more likely they have a template/overload which picks up non-member functions (e.g. all the lambda wrappers are non-member functions) accepting the object as the first argument. There's nothing unsafe or undefined about the code.

pca006132:
I don't think it is an issue, and it is probably not very useful to make it a static function on manifold for the python binding considering it is much easier to use it as method chaining.

pca006132:
Any consensus on this? I don't think this PR is needed.

elalish:
Perhaps just add the static method to the Python as well? Then it's not breaking, but also matches our C++ better.

pca006132:
sounds good
@briansturgill do you want to do that in this PR?

briansturgill:
@pca006132 
Yes, I'll do it today. I'll get compose added as well.

bindings/python/examples/test_torus_knot.py
```diff
@@ -81,9 +81,9 @@ def func_single(v):
         return pts[0]
 
     if warp_single:
-        return circle.revolve(int(m)).warp(func_single)
+        return Manifold.revolve(circle, int(m)).warp(func_single)
     else:
-        return circle.revolve(int(m)).warp_batch(func)
+        return Manifold.revolve(circle, int(m)).warp_batch(func)
```
elalish, line 86: Maybe leave one example using `circle.extrude()` for variety? Unless there's already one that you didn't change.

elalish (approved):
Looks great, thanks!

bindings/python/examples/test_torus_knot.py
pca006132: I think if we are having it in the `all_apis.py` it should be fine.

PR #700, opened 2024-01-13
CrossSection::Compose was accidently left out of Python API.

briansturgill:
@pca006132 fixed the nanobind issue, so this works now too!


elalish (approved):
Excellent, thank you!

PR #704, opened 2024-01-14
Install Python module into site-packages

cjmayo:
Allows `import manifold3d` to work from the REPL or a script using the installed system interpreter.

### DIFF
--- a/bindings/python/CMakeLists.txt
+++ b/bindings/python/CMakeLists.txt
@@ -44,7 +44,7 @@ install(
 else()
 install(
   TARGETS manifold3d
-  LIBRARY DESTINATION ${CMAKE_INSTALL_FULL_LIBDIR}
+  LIBRARY DESTINATION ${Python_SITEARCH}
   COMPONENT bindings
 )
 endif()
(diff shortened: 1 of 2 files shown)


pca006132:
I think the typical way is not using cmake to install the python package, but use pip.

pca006132:
Well you can install from source using pip build, this is how the sdist and bdist are created. The sdist issue you got is just a bug we should fix.

I think the current install path is needed for scikit-build-core to get the required library files, not sure about that.

pca006132:
After looking at the code more carefully, I think this is fine but needs some tweaking for nix (which is normal, nix is... not a common distro).

```diff
diff --git a/flake.nix b/flake.nix
index af6d989..f728097 100644
--- a/flake.nix
+++ b/flake.nix
@@ -55,6 +55,10 @@
                 "-DFETCHCONTENT_SOURCE_DIR_THRUST=${thrust-src}"
                 "-DMANIFOLD_PAR=${pkgs.lib.strings.toUpper parallel-backend}"
               ];
+              prePatch = ''
+                substituteInPlace bindings/python/CMakeLists.txt \
+                  --replace 'DESTINATION ''${Python_SITEARCH}' 'DESTINATION "${placeholder "out"}/${pkgs.python3.sitePackages}"'
+              '';
               checkPhase = ''
                 cd test
                 ./manifold_test
```

elalish (approved):
Thanks!

PR #707, opened 2024-01-15
Added BatchBoolean to Python API, both CS & M.

briansturgill:
Starting to find this easy. :-)
Simple add of functions, with tests.
Had to add a comment in manifold.cpp (used a similar one to that in cross_section.cpp).


pca006132 (approved):
LGTM

Issue #712, opened 2024-01-19
Post boolean op Status() crash


starseeker:
I'm trying to subtract the cray_plate.glb mesh from cray_cloud.glb.  cloud_plate.glb is generated from a half space primitive - what I'm trying to do is make a really big cuboid mesh to stand in for the half space, and use that for the subtraction.  It's quite possible I'm not generating that properly yet, but the status checks for both inputs pass and the boolean op itself seems to run.  However, when I do the status check to validate the output:

bool_out = lm->Boolean(*rm, manifold_op);
if (bool_out.Status() != manifold::Manifold::Error::NoError) {
   ...
}

I get a crash in the Status() call stack, but not consistently on all platforms.  One curious part is dumping bool_out on the crashing platform before calling Status() does produce a glb file, which I've included in the below directory.

I am testing both inputs with checks of their status, which pass - presumably that's why Boolean runs.

lm->Status() != manifold::Manifold::Error::NoError
rm->Status() != manifold::Manifold::Error::NoError

I've uploaded the glb dumps at https://brlcad.org/~starseeker/glb/cray_cloud, although I've not yet tried to reproduce the crash in a self-contained test case and I'm not sure if it will reproduce consistently across platforms, depending on what the cause is.

I can use smaller numbers for the half space cuboid mesh too, if I'm exceeding numerical limits for good behavior in Manifold with that input...

elalish:
Strange! Regardless of what you do, it shouldn't crash, so I'd call this a bug. The Status thing may be a red-herring - that may just be triggering the CSG tree to evaluate. Please make a PR with a `TEST` - our CI runs on quite a few platforms, so hopefully one will catch it.

Regarding your goal of cutting with a half-space, we actually already have functions to make that easy: `SplitByPlane()` and `TrimByPlane()`.

starseeker:
Pull request #713 submitted.  As expected it doesn't fail on my local machine, so we'll see what happens with the others.  In case it's useful, I managed to get  a gdb backtrace of the observed crash:

```
#0  0x00007ffff217cacf in raise () from /lib64/libc.so.6
#1  0x00007ffff214fea5 in abort () from /lib64/libc.so.6
#2  0x00007ffff21bdcc7 in __libc_message () from /lib64/libc.so.6
#3  0x00007ffff21c4fcc in malloc_printerr () from /lib64/libc.so.6
#4  0x00007ffff21c6b54 in _int_free () from /lib64/libc.so.6
[...]

PR #713, opened 2024-01-19
Set up test for subtraction of a very large cuboid

starseeker:
Test case for #712 - these glb inputs were generated from manifolds that succeeded in producing a Boolean output, but (on one platform but not another) the Status() call on the output boolean manifold crashed.

(A test for calling Status() on the output captured on the crashing platform (in case there was something specifically bad about that particular output on that platform) was tried, but the result was simply a successful Status() call even with that specific input - hence, that additional test is not included in this commit.)

Right now we're using the same GetMeshGL() call as the other tests, but perhaps it would be preferable to use Status() instead since that was the observed failure?  I wasn't sure if the former was preferred as a more rigorous test of the Manifold structure.


elalish:
Looks like we caught the problem here, thanks! I'll take a look.

elalish:
I have a local repro of this problem - for once it's not the triangulator! @pca006132 Might be good to look at this case to get an idea of how to construct fuzzer tests. Speaking of, is there a way to use our fuzzer to minimize this test case?

pca006132:
No, this is not something fuzzer can do. We can probably write something that simplifies the triangular mesh though.

elalish:
Closing in favor of #714. Thanks for a good edge case test! For future reference, please make your PRs on a new branch - we can't merge if you're using master already.

PR #717, opened 2024-01-23
fix cube performance

### LINKED ISSUE #710: Manifold::Cube is slow.
Cube is nearly half the speed of a function I wrote in python calling square/extrude.
I can do a PR if you want.

```
 1.292 100,000 creations of M with cube member function
 0.858 100,000 creations of M with mesh from cube
 0.719 100,000 creations of M with python function
```

```python
import time
import manifold3d as m
from random import random
import numpy as np

ts = time.time()
for i in range(100000):
    c = m.Manifold.cube((10.0, 10.0, 10.0))
    c.num_vert()
te = time.time()
tdiff = te - ts
print(f"{tdiff: 5.3f}", "100,000 creations of M with cube member function")

mesh = c.to_mesh()
mesh = m.Mesh(
    tri_verts=np.array(mesh.tri_verts.tolist().copy(), np.int32),
    vert_properties=np.array(mesh.vert_properties.tolist().copy(), np.float32),
)
ts = time.time()
for i in range(100000):
    c = m.Manifold(mesh)
    c.num_vert()
te = time.time()
tdiff = te - ts
print(f"{tdiff: 5.3f}", "100,000 creations of M with mesh from cube")


def cube(dim):
    x, y, z = dim
    c = m.CrossSection.square([x, y])
    return m.Manifold.extrude(c, z)


ts = time.time()
for i in range(100000):
    c = cube((10.0, 10.0, 10.0))
    c.num_vert()
te = time.time()
tdiff = te - ts
print(f"{tdiff: 5.3f}", "100,000 creations of M with python function")
```

pca006132:
Removed the need for using transforms and `AsOriginal` for cube, which improves performance quite a bit.
Closes #710.

@elalish do you know why MeshGLRoundTrip now breaks?

```c++
TEST(Boolean, MeshGLRoundTrip) {
  Manifold cube = Manifold::Cube(glm::vec3(2));
  ASSERT_GE(cube.OriginalID(), 0);
  const MeshGL original = cube.GetMeshGL();

  Manifold result = cube + cube.Translate({1, 1, 0});

  ASSERT_LT(result.OriginalID(), 0);
  ExpectMeshes(result, {{18, 32}});
  RelatedGL(result, {original});

  MeshGL inGL = result.GetMeshGL();
  ASSERT_EQ(inGL.runOriginalID.size(), 2);
  const Manifold result2(inGL);

  ASSERT_LT(result2.OriginalID(), 0);
  ExpectMeshes(result2, {{16, 28}}); // <--- need this to pass the test
  RelatedGL(result2, {original});

  const MeshGL outGL = result2.GetMeshGL();
  ASSERT_EQ(outGL.runOriginalID.size(), 2);
}
```

I assume it simplifies the mesh when manifold is created from the MeshGL, but I have no idea why  it does this. Changing cube to `Manifold::Cube(glm::vec3(2)).AsOriginal()` doesn't help.

### DIFF
--- a/src/manifold/src/constructors.cpp
+++ b/src/manifold/src/constructors.cpp
@@ -153,10 +153,9 @@ Manifold Manifold::Cube(glm::vec3 size, bool center) {
       glm::length(size) == 0.) {
     return Invalid();
   }
-  auto cube = Manifold(std::make_shared<Impl>(Impl::Shape::Cube));
-  cube = cube.Scale(size);
-  if (center) cube = cube.Translate(-size / 2.0f);
-  return cube.AsOriginal();
+  glm::mat4x3 m =
+      glm::translate(center ? (-size / 2.0f) : glm::vec3(0)) * glm::scale(size);
+  return Manifold(std::make_shared<Impl>(Manifold::Impl::Shape::Cube, m));
 }
 
 /**
--- a/src/manifold/src/impl.cpp
+++ b/src/manifold/src/impl.cpp
@@ -534,7 +534,7 @@ Manifold::Impl::Impl(const Mesh& mesh, const MeshRelationD& relation,
  * Create either a unit tetrahedron, cube or octahedron. The cube is in the
  * first octant, while the others are symmetric about the origin.
  */
-Manifold::Impl::Impl(Shape shape) {
+Manifold::Impl::Impl(Shape shape, const glm::mat4x3 m) {
   std::vector<glm::vec3> vertPos;
   std::vector<glm::ivec3> triVerts;
   switch (shape) {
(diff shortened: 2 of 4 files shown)


elalish:
Hmm, that's odd. I almost wonder if the test was broken before? I feel like both `ExpectMeshes` calls should ask for `{16, 28}`. Does that fix it? What are we actually getting for `result` and `result2`?

pca006132:
`result = {18, 32}`, `result2 = {16, 28}`, thinking about it `{16, 28}` makes more sense, we union two identical cubes, the mesh should be simplified.

pca006132:
@elalish what do you think about the value of `result`? I feel that `{16, 28}` makes more sense but somehow we are not getting that...

elalish (approved):
Thanks, that test looks better now!

PR #722, opened 2024-01-28
optimize triangulator for msvc

pca006132:
It seems that msvc does not optimize `glm::determinant(glm::mat2(a, b))` well, it requires an intermediate struct that is stored in the stack, causing performance issues. This patch improves the performance of the test in #718 from 5s to 2s on Windows, and no performance degradation with clang/gcc.

@briansturgill maybe you can test this as well, I am using a windows vm so the performance number may not be that accurate.

### DIFF
--- a/src/polygon/src/polygon.cpp
+++ b/src/polygon/src/polygon.cpp
@@ -15,12 +15,8 @@
 #include "polygon.h"
 
 #include <algorithm>
-#include <list>
 #include <map>
-#include <numeric>
-#include <queue>
 #include <set>
-#include <stack>
 
 #include "optional_assert.h"
 #include "utils.h"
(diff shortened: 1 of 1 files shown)


briansturgill:
@pca006132 You want just Windows, or Windows and Mac.

pca006132:
@briansturgill just windows is fine

briansturgill:
@pca006132 Good news... several runs, results in range 1500-1600 over 4x performance boost.

pca006132:
Nice! Thanks a lot. Btw feel free to open discussions if you find other potential performance issues, even if they are platform specific. Tracy is a useful tool to find such issues.

elalish:
Good find!

Issue #701, opened 2024-01-14
header file naming/directory structure


pca006132:
We want to avoid conflicting with header files from other packages.

pca006132:
https://github.com/elalish/manifold/pull/901#issuecomment-2295314295

@hzeller what do you think if we put all headers inside the `manifold` subdirectory, and access them as `#include "manifold/xxx.h"`?

hzeller:
yes, I think that would be good. That way, it is easy to immediately see where headers are coming from and there are no conflicts.

PR #729, opened 2024-01-31
Install headers into manifold subdirectory

cjmayo:
Basically this goes back to the way things were. I think it addresses #701 OK, don't know if there are other ideas.

```
/usr/include/manifold/meshIO.h
/usr/include/manifold/sdf.h
/usr/include/manifold/manifold.h
/usr/include/manifold/polygon.h
/usr/include/manifold/cross_section.h
/usr/include/manifold/collider.h
/usr/include/manifold/vec_view.h
/usr/include/manifold/public.h
```

Because of the CMake config files I am able to build OpenSCAD (modified to use `find_package(manifold)`) with either master or this PR.

### DIFF
--- a/meshIO/CMakeLists.txt
+++ b/meshIO/CMakeLists.txt
@@ -17,7 +17,7 @@ project(meshIO)
 add_library(${PROJECT_NAME} src/meshIO.cpp)
 
 target_include_directories(${PROJECT_NAME} PUBLIC
-    $<INSTALL_INTERFACE:include>
+    $<INSTALL_INTERFACE:include/${CMAKE_PROJECT_NAME}>
     $<BUILD_INTERFACE:${PROJECT_SOURCE_DIR}/include>)
 
 target_link_libraries(${PROJECT_NAME}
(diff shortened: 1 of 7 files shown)


elalish:
Thanks, sounds good to me, but I don't understand library installs well enough to review this with any confidence. If @pca006132 is happy and it works for OpenSCAD, then I'm happy.

pca006132:
Looks fine to me, but ultimately I think we should change headers such as `public.h` to something else, or it will conflict with others...

elalish:
Can you help me understand this better? I *think* what's happening here is that instead of installing `usr/include/public.h`, which might collide with anything called `public.h`, we're installing `usr/include/manifold/public.h`, which shouldn't collide with anything except maybe another library called manifold. Or am I misunderstanding how header conflicts happen?

pca006132:
Yeah this avoids conflict during installation, but if a user uses two packages with public.h, the include will have conflict as well.

elalish:
Oh, I thought you'd still have to put `#include <manifold/public.h>`.

pca006132:
That requires changing our current directory structure and includes.

PR #730, opened 2024-02-01
Fix to 2D HullImpl problem.

### LINKED ISSUE #728: Python API: batch_hull fails bizarrely
batch_hull from Python is failing very strangely.

Further experimentation suggests that the number of segments being a multiple
of the y length, is heavily related to making it occur.

```python
import manifold3d as m

def rounded_rectangle(x, y, radius, segments):
    circ = m.CrossSection.circle(radius, segments)
    o = m.CrossSection.batch_hull([
        circ.translate((radius, radius)),
        circ.translate((x - radius, radius)),
        circ.translate((x - radius, y - radius)),
        circ.translate((radius, y - radius)),
    ])
    return o

# You might find this helpful to view the problem
def _save_svg(filename, obj):
    txt = []
    bb = obj.bounds()
    width = round(bb[2] - bb[0], 5)
    height = round(bb[3] - bb[1], 5)
    color = (0, 128, 255)
    units = "mm"

    txt.append("<?xml version=\"1.0\" encoding=\"UTF-8\"?>")
    txt.append("<!-- Created by Piecad. -->")
    txt.append("<!DOCTYPE svg PUBLIC \"-//W3C//DTD SVG 1.1 Tiny//EN\" \"http://www.w3.org/Graphics/SVG/1.1/DTD/svg11-tiny.dtd\">")
    txt.append(f"<svg width=\"{width}{units}\" height=\"{height}{units}\" viewBox=\"0 0 {width} {height}\" fill=\"rgb({color[0]},{color[1]},{color[2]})\" xmlns=\"http://www.w3.org/2000/svg\" fill-rule=\"evenodd\">")

    txt.append(f"<g><path d=\"")

    off_x = 0 - bb[0]
    off_y = 0 - bb[1]
    paths = obj.to_polygons()
    for path in paths:
        for i in range(len(path)):
            vert = path[i]
            x = round(vert[0] + off_x, 5)
            y = round(vert[1] + off_y, 5)
            if i == 0:
                txt.append(f"M{x} {y}")
            else:
                txt.append(f"L{x} {y}")

    txt.append("\"/></g>\n")

    txt.append("</svg>")

    with open(filename, "w") as f:
        f.write("\n".join(txt))


radius = 9.0
x = 51
y = 36
segments = 36

# These don't work
o = rounded_rectangle(x, y, radius, segments)
print(o.num_vert())
_save_svg("test.svg", o)
o = rounded_rectangle(x, y, radius+3, segments)
print(o.num_vert())
o = rounded_rectangle(x+1, y, radius, segments)
print(o.num_vert())
o = rounded_rectangle(x-1, y, radius, segments)
print(o.num_vert())

# These work fine
o = rounded_rectangle(x, y, radius-6, segments)
print(o.num_vert())
o = rounded_rectangle(x, y, radius-3, segments)
print(o.num_vert())
o = rounded_rectangle(x, y, radius-1, segments)
print(o.num_vert())
o = rounded_rectangle(x, y, radius+1, segments)
print(o.num_vert())
o = rounded_rectangle(x, y+1, radius, segments)
print(o.num_vert())
o = rounded_rectangle(x, y+2, radius, segments)
print(o.num_vert())
o = rounded_rectangle(x, y-1, radius, segments)
print(o.num_vert())
o = rounded_rectangle(x+1, y+1, radius, segments)
print(o.num_vert())
```

briansturgill:
Fixes #728 

I did a refactor of the current Andrew/Monotone Chain algorithm,
using [this reference](https://en.wikibooks.org/wiki/Algorithm_Implementation/Geometry/Convex_hull/Monotone_chain)
I think the main problem was in the old check of the turn direction.
At least in my python timings for rounded_rectangle, the new version is slightly faster.

### DIFF
--- a/src/cross_section/src/cross_section.cpp
+++ b/src/cross_section/src/cross_section.cpp
@@ -187,7 +187,7 @@ C2::PathD HullImpl(SimplePolygon& pts) {
     lower.push_back(pts[i]);
   }
   auto upper = std::vector<glm::vec2>{};
-  for (int i = len-1; i >= 0; i--) {
+  for (int i = len - 1; i >= 0; i--) {
     auto sz = upper.size();
     while (sz >= 2 && Cross(upper[sz - 2], upper[sz - 1], pts[i]) <= 0) {
       upper.pop_back();
@@ -204,7 +204,7 @@ C2::PathD HullImpl(SimplePolygon& pts) {
     path[i] = v2_to_pd(lower[i]);
   }
   auto llen = lower.size();
-	int sz = upper.size(); // "fix" -Waggressive-loop-optimizations warning.
+  int sz = upper.size();  // "fix" -Waggressive-loop-optimizations warning.
   for (int i = 0; i < sz; i++) {
     path[i + llen] = v2_to_pd(upper[i]);
   }
--- a/src/cross_section/src/cross_section.cpp
+++ b/src/cross_section/src/cross_section.cpp
@@ -163,8 +163,12 @@ bool V2Lesser(glm::vec2 a, glm::vec2 b) {
   return a.x < b.x;
 }
 
-double Cross(glm::vec2 a, glm::vec2 b, glm::vec2 c) {
-  return (a.x - c.x) * (b.y - c.y) - (a.y - c.y) * (b.x - c.x);
+void HullBacktrack(const glm::vec2& pt, std::vector<glm::vec2>& stack) {
+  auto sz = stack.size();
+  while (sz >= 2 && CCW(stack[sz - 2], stack[sz - 1], pt, 1e-18) <= 0) {
+    stack.pop_back();
+    sz = stack.size();
+  }
 }
 
 // Based on method described here:
@@ -179,20 +183,12 @@ C2::PathD HullImpl(SimplePolygon& pts) {
 
   auto lower = std::vector<glm::vec2>{};
   for (int i = 0; i < len; i++) {
-    auto sz = lower.size();
-    while (sz >= 2 && Cross(lower[sz - 2], lower[sz - 1], pts[i]) <= 0) {
-      lower.pop_back();
-      sz = lower.size();
-    }
+    HullBacktrack(pts[i], lower);
     lower.push_back(pts[i]);
   }
   auto upper = std::vector<glm::vec2>{};
   for (int i = len - 1; i >= 0; i--) {
-    auto sz = upper.size();
-    while (sz >= 2 && Cross(upper[sz - 2], upper[sz - 1], pts[i]) <= 0) {
-      upper.pop_back();
-      sz = upper.size();
-    }
+    HullBacktrack(pts[i], upper);
     upper.push_back(pts[i]);
   }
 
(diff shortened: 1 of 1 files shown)


src/cross_section/src/cross_section.cpp
```diff
@@ -163,50 +163,46 @@ bool V2Lesser(glm::vec2 a, glm::vec2 b) {
...
-                   std::vector<int>& keep, const int hold) {
-  const int stop = keep.size() - hold;
-  int i = 0;
-  while (i < stop && !IsCw(pts[idx], pts[keep[keep.size() - 1]],
-                           pts[keep[keep.size() - 2]])) {
-    keep.pop_back();
-    i++;
+void HullBacktrack(const glm::vec2& pt, std::vector<glm::vec2>& stack) {
+  auto sz = stack.size();
+  while (sz >= 2 && CCW(stack[sz - 2], stack[sz - 1], pt, 1e-18) <= 0) {
```
elalish, line 168: glm::vec2 is two floats, so they're not usually precise down to 1e-18. Since you aren't making use of the colinear case, I think it's safe to just the tolerance to 0 here.

briansturgill: The precision value came from the previous code that was there. I have no idea what it should be.
However, the code IS making use of the colinear case as both colinear and CW points are dropped.

briansturgill: @elalish I need a hint here... is it OK to just use the zero, or should I use something else.
The code really is making use of the colinear case.

elalish: It's really not - at 1e-18, the collinear case is either never being hit, or if it is, it's garbage. Agreed the same error was in the previous code. Changing it to zero shouldn't make any difference, but will make the code less confusing to read.

elalish: Oh, I see where 1e-18 came from now: he squared both sides, so he was really using a tolerance of 1e-9. I'm still not really sure what that tolerance does, since there's no third case for collinear in the algorithm. You might try a test with a bunch of collinear points to see what difference it makes. I would guess even if you switched from `<= 0` to `< 0` it wouldn't matter much. Probably a *very* slightly concave and convex line of points would be a good test.

briansturgill: @elalish OK, after looking at numerous implementations of both Graham and Andrew/Monotone algorithms, I only
found that bothered with it at all. Looking strictly at how the algorithm works, you might actually break things
if you don't use 0. So I used zero and  pushed.

elalish (approved):
Looks great, thanks! Would you mind adding your test case?

briansturgill:
@elalish All testing was done from Python.
Would you like me to do a C++ one...
Also, do you mean the ones that were failing earlier?
(I use Hull over the place because I use rounded rectangles in many places.)

elalish:
Yes, I mean adding a C++ test of your rounded rectangle that wasn't working. I think a simple area check would suffice. Ideally check that it fails as expected without your fix here.

elalish:
something like this: https://github.com/elalish/manifold/blob/master/test/cross_section_test.cpp#L55

PR #732, opened 2024-02-05
Added test for hull problem fixed in #730.

briansturgill:
Test was confirmed to fail in earlier versions that had the faulty HullImpl.


elalish (approved):
Thanks for the test!

PR #733, opened 2024-02-06
Editor buttons

elalish:
Follow-up to #643 

Adds undo, redo, format, and share buttons to ManifoldCAD UI.


pca006132:
the buttons are currently not aligned properly, at least on firefox:

image

Adding `vertical-align: middle; ` to `.top-icon` fixes it.

image

but still, the vertical alignment is still slightly off... The alignment would be perfect if the text is `x`, but it is not. Anyway probably not that important.

pca006132:
btw, for the share functionality, we can only share gists/web pages? new script for example cannot be shared.

elalish:
> btw, for the share functionality, we can only share gists/web pages? new script for example cannot be shared.

What do you mean? Here's a new script example from the share button: http://localhost:5173/?script=const+result+%3D+Manifold.cube%28%29%3B%0A#New%20Script%202

pca006132:
Oh, sorry I overlooked that, I was creating a new empty script and look at the url. This should be fine then.

pca006132 (approved):

elalish:
So, I am getting "Error: URI too long" with larger shares. Apparently this is server config, so my best guess is Github pages goes with the Apache default max of 8190 characters. I can't think of a way around that, but at least this works for medium-sized scripts.

hrgdavor:
> I can't think of a way around that,

we implemented gzip just recently for data url https://github.com/hrgdavor/jscadui/issues/76

pca006132:
1. We probably want a warning about url too long when the user click on the share button.
2. Not sure if base64 + gzip can have very high compression ratio to worth the effort. I feel that we should add a separate button to allow script url similar to jscad.app

elalish:
Agreed, though didn't @hrgdavor just say that the way jscad does it is with base64+gzip? On text that might give ~10x compression, which would probably be plenty.

hrgdavor:
I just took a script from jscad examples  (gears)

| content     | size bytes |
| ----------- | ---- |
| script raw  | 5.320 |
| urlencoded  | 7.906 |
| base64      | 7.096 |
| gzip&base64 | 2.344 |

I was surprised to see urlencoding (7.9) is even more overhead than base64 (7.1 k)

it is not 10x, but looks worth it

elalish:
@hrgdavor Thanks for that bit about the hash not getting sent to the server. @pca006132 It looks like I was exactly [wrong](https://github.com/elalish/manifold/pull/643#issuecomment-1849560385). So, if we switch back from query params to hashes, then it probably doesn't matter whether it's zipped or not. I'll take a look.

PR #739, opened 2024-02-12
[python] add MeshGL::Merge() to python API

wrongbad:
Small diff to add MeshGL::Merge to python API, so .stl import can be done via python API


elalish (approved):
Good catch, thanks!

elalish:
Would you mind adding it to our `all_apis.py` file to get a least a trivial test of it?

Issue #738, opened 2024-02-12
GetMesh core dump with SparseIndices too large


starseeker:
I'm getting a core dump trying to retrieve a mesh, in a rather surprising situation - unlike the other SparceIndices report, this isn't (or at least I wouldn't expect it to qualify as) a large mesh. (For my own reference, this is coming from doing a plate to volume mode bot operation on 7081.1.t0 in BRL-CAD's GenericTwin.g example).  The initial backtrace is here:

```
#9  0x00007ffff3eb3914 in manifold::Boolean3::Boolean3(manifold::Manifold::Impl const&, manifold::Manifold::Impl const&, manifold::OpType) [clone .cold] () from libmanifold.so
#10 0x00007ffff3f41536 in manifold::CsgOpNode::BatchBoolean(manifold::OpType, std::vector<std::shared_ptr<manifold::Manifold::Impl const>, std::allocator<std::shared_ptr<manifold::Manifold::Impl const> > >&) () from libmanifold.so
#11 0x00007ffff3f43b5d in manifold::CsgOpNode::BatchUnion() const () from libmanifold.so
#12 0x00007ffff3f445fd in manifold::CsgOpNode::ToLeafNode() const () from libmanifold.so
#13 0x00007ffff3f6864c in manifold::Manifold::GetCsgLeafNode() const () from libmanifold.so
#14 0x00007ffff3f6ba40 in manifold::Manifold::GetMesh() const () from libmanifold.so
```

I tried rebuilding with MANIFOLD_DEBUG, but it doesn't seem to have an impact with this case - I still get:

```
terminate called after throwing an instance of 'std::out_of_range'
  what():  SparseIndices too large
Aborted (core dumped)
```

I've isolated the inputs that lead to this into two .glb files and a test case, which fails (at least on my local box).

starseeker:
Oh, reading the comments, this may be a similar situation to #720 after all.

elalish:
yeah, closely related, but your PR still picked up and interesting difference. When you say "doesn't qualify as a large mesh", how many tris to each of these models have?

starseeker:
Left should have on the order of 20k faces, and right about 11.5k triangles.  They're not small, certainly, but by the standards of some of the meshes we deal with they didn't strike me as huge.

starseeker:
[...]

PR #740, opened 2024-02-12
Prepare a test case for "SparseIndices too large" core dump

starseeker:
This is a pull request pertaining to https://github.com/elalish/manifold/issues/738


elalish:
Wow, super strange that only some platforms repro the problem. Maybe that implies some kind of undefined behavior somewhere? @pca006132 What do you make of this?

It seems to be hitting this check:
```
if (size >= std::numeric_limits<int32_t>::max() / sizeof(int64_t))
      throw std::out_of_range("SparseIndices too large");
```
Is `sizeof(int64_t)` really different on different platforms? Also, I'm a touch confused on where this limit comes from - it makes it sound like we're going to access individual bytes by `int` indices, but I think we're only accessing `int` or `long`? Should it be `if (size >= std::numeric_limits<int32_t>::max() * sizeof(int32_t) / sizeof(int64_t))`?

pca006132:
weird, will have a look at this, I don't think this is hard to fix.

PR #741, opened 2024-02-14
Refine To Length

elalish:
This is still WIP, but it's getting close. It's generating good triangulations, though I will continue making a few improvements before this gets merged. ~~Right now I'm stuck because `./manifold_test --gtest_filter=Manifold.SmoothSphere` is giving me `malloc: *** error for object 0x3f0000003ec000ef: pointer being freed was not allocated`, but the address sanitizer isn't catching anything. I'm not sure how I could be causing this, since I'm just using `Vec` - do you have any ideas @pca006132?~~


pca006132:
interesting, I will try to figure it out when I have time.

pca006132:
ah, I should have noted this: Vec does not support things with non-trivial constructor/destructor...

pca006132:
I can either make it work, or use template magic to ban it. What do you think? @elalish

pca006132:
the simplest way is to change `subTris` into `std::vector<Partition>`.

elalish:
Switching to std::vector indeed fixed the allocation error, but I'm still doing something seriously wrong I think. I'm pretty sure I have undefined behavior here somewhere; running `Manifold.SmoothSphere` gets different results from run to run in parallel mode, and consistent results in sequential mode that change when it gets compiled differently (like running the address sanitizer). It even gets different results (sometimes correct) when running the loop just once instead of several times. Do you notice any other problems @pca006132?

src/manifold/src/smoothing.cpp
```diff
@@ -308,6 +191,296 @@ struct InterpTri {
...
+    if (sortedDiv[1] > sortedDiv[0]) {
+      std::swap(sortedDiv[1], sortedDiv[0]);
+      std::swap(triIdx[1], triIdx[0]);
+      if (sortedDiv[2] > sortedDiv[1]) {
+        std::swap(sortedDiv[2], sortedDiv[1]);
+        std::swap(triIdx[2], triIdx[1]);
+      }
+    }
+
+    Partition partition = GetCachedPartition(sortedDiv);
```
elalish, line 224: The triangulation is purely topological - it depends only on the number of divisions of the three sides of the triangle. This allows them to be cached and reused for similar triangles. The shape of the final surface is defined by the tangents and the barycentric coordinates of the new verts.

src/manifold/src/smoothing.cpp
```diff
@@ -308,6 +191,296 @@ struct InterpTri {
     pos = HNormalize(posH);
   }
 };
+
+class Partition {
+ public:
+  glm::ivec3 idx;
```
elalish, line 197: The cached partitions don't have this - it's added to the copy returned from `GetPartition` that contains the mapping of the input divisions into the sorted divisions that are uniquely cached.

src/manifold/src/smoothing.cpp
```diff
@@ -308,6 +191,296 @@ struct InterpTri {
...
+      const int next = edgeOffset + i;
+      triVert.push_back({last, next, cornerVerts[2]});
+      last = next;
+    }
+    triVert.push_back({last, cornerVerts[1], cornerVerts[2]});
+  }
+
+  // Partitions are parallel to the first edge unless two consecutive edgeAdded
+  // are zero, in which case a terminal triangulation is performed.
+  static void PartitionQuad(Vec<glm::ivec3>& triVert, Vec<glm::vec3>& vertBary,
```
elalish, line 363: The algorithm actually focuses on partitioning quads recursively - triangles are turned into quads by clipping one sub-triangle off of a corner.

src/manifold/src/smoothing.cpp
```diff
@@ -308,6 +191,296 @@ struct InterpTri {
...
+    for (int i = 0; i < triVert.size(); ++i) {
+      for (const int j : {0, 1, 2}) {
+        newTriVert[i][outTri[j]] = newVerts[triVert[i][j]];
+      }
+    }
+    return newTriVert;
+  }
+
+ private:
+  static std::unordered_map<glm::ivec3, Partition> cache;
```
elalish, line 267: I was expecting I would need a bit of help making this safe in parallel, but I'm surprised I'm having trouble in single-threaded mode.

pca006132: interesting, will have a look

pca006132:
Not sure what exactly is the bug, but it seems that in `CompactProps`, before the last `for_each`, `NumTri()` is wrong. I got
```
propOld2New: 8
triProperties.size() = 14, NumTri = 224
```
which cannot be true because there are only 7 vertices (`propOld2New.size() - 1`)...

pca006132:
btw here are some suggested changes, one adds a static check to avoid using `Vec` with non-trivial destructors, and the other one fixes the compilation error.

```diff
diff --git a/src/manifold/src/smoothing.cpp b/src/manifold/src/smoothing.cpp
index 54faa08..5923f64 100644
--- a/src/manifold/src/smoothing.cpp
+++ b/src/manifold/src/smoothing.cpp
@@ -264,7 +264,7 @@ class Partition {
   }
 
  private:
-  static std::unordered_map<glm::ivec3, Partition> cache;
+  static inline auto cache = std::unordered_map<glm::ivec3, Partition>();
 
   // n[0] >= n[1] >= n[2] > 0
   static Partition GetCachedPartition(glm::ivec3 n) {
@@ -479,8 +479,6 @@ class Partition {
                   newEdgeAdded, newEdgeFwd);
   }
 };
-
-auto Partition::cache = std::unordered_map<glm::ivec3, Partition>();
 }  // namespace
 
 namespace manifold {
diff --git a/src/utilities/include/vec.h b/src/utilities/include/vec.h
index 5054cc6..5409843 100644
--- a/src/utilities/include/vec.h
+++ b/src/utilities/include/vec.h
@@ -237,6 +237,8 @@ class Vec : public VecView<T> {
 
  private:
   size_t capacity_ = 0;
+
+  static_assert(std::is_trivially_destructible<T>::value);
 };
 /** @} */
 }  // namespace manifold
```

elalish:
I haven't actually hooked up the props yet, so I don't think it's that, but I'll take a look.

elalish:
Okay, it's almost working now - I need to handle the props and do a little cleanup on the triangulation.

src/manifold/src/smoothing.cpp
```diff
@@ -632,9 +632,10 @@ Vec<Barycentric> Manifold::Impl::Subdivide(
                const int v1 = Next3(v0);
                const int tri = edge.halfedgeIdx / 3;
                for (int i = 0; i < n; ++i) {
-                 glm::vec3& uvw = vertBary[offset + i].uvw;
+                 glm::vec3 uvw(0);
```
pca006132, line 635: I don't think the previous behavior is ub? The `vertBary` vector should be initialized...

elalish: It appeared to be initialized with `-1`s, but only after the first run (it was initialized with `0`s the first time). Not sure why.

pca006132:
```diff
diff --git a/src/manifold/src/smoothing.cpp b/src/manifold/src/smoothing.cpp
index 137293f..06f1bcb 100644
--- a/src/manifold/src/smoothing.cpp
+++ b/src/manifold/src/smoothing.cpp
@@ -267,7 +267,9 @@ class Partition {
   }
 
  private:
-  static inline auto cache = std::unordered_map<glm::ivec3, Partition>();
+  static inline auto cacheLock = std::mutex();
+  static inline auto cache =
+      std::unordered_map<glm::ivec3, std::unique_ptr<Partition>>();
 
   // This triangulation is purely topological - it depends only on the number of
   // divisions of the three sides of the triangle. This allows them to be cached
@@ -275,9 +277,12 @@ class Partition {
   // by the tangents and the barycentric coordinates of the new verts. The input
   // must be sorted: n[0] >= n[1] >= n[2] > 0
   static Partition GetCachedPartition(glm::ivec3 n) {
-    auto cached = cache.find(n);
-    if (cached != cache.end()) {
-      return cached->second;
+    {
+      auto lockGuard = std::lock_guard<std::mutex>(cacheLock);
+      auto cached = cache.find(n);
+      if (cached != cache.end()) {
+        return *cached->second;
+      }
     }
     Partition partition;
     partition.sortedDivisions = n;
@@ -349,7 +354,8 @@ class Partition {
       }
     }
 
-    cache.insert({n, partition});
+    auto lockGuard = std::lock_guard<std::mutex>(cacheLock);
+    cache.insert({n, std::make_unique<Partition>(partition)});
     return partition;
   }
 
```

This may help to fix the compilation error, and make the cache thread-safe.

pca006132:
wow the test results are beautiful, will have a look at the code in details these two days.

elalish:
Thank you! I'm just adding properties support now, but still needs some debug.

elalish:
Okay, I think this one is good to go now - property support seems to be working well, the triangulations look decent, and I've added all the bindings.

Leeow76:
I just want to say I love you so much for this!! :1st_place_medal:

src/manifold/src/smoothing.cpp
```diff
@@ -53,131 +53,14 @@ void FillRetainedVerts(Vec<Barycentric>& vertBary,
...
-        int b = Vert(i + 1, j, tri);
-        int c = Vert(i, j + 1, tri);
-        triVerts[pos++] = glm::ivec3(c, a, b);
-        if (j < n - 1 - i) {
-          int d = Vert(i + 1, j + 1, tri);
-          triVerts[pos++] = glm::ivec3(b, d, c);
-        }
-      }
-    }
+    half2Edge[halfedges[halfedge].pairedHalfedge] = edge;
```
pca006132, line 63: is this a bug fix or needed due to the new feature?

src/manifold/src/smoothing.cpp
```diff
@@ -308,6 +191,312 @@ struct InterpTri {
...
+
+    return partition;
+  }
+
+  Vec<glm::ivec3> Reindex(glm::ivec3 tri, glm::ivec3 edgeOffsets,
+                          glm::bvec3 edgeFwd, int interiorOffset) const {
+    Vec<int> newVerts;
+    newVerts.reserve(vertBary.size());
+    glm::ivec3 triIdx = idx;
+    std::vector<int> outTri = {0, 1, 2};
```
pca006132, line 238: maybe `std::array<int, 3>` or `glm::ivec3`? seems like a fixed length array to me, so no need to use `vector`.

src/manifold/src/smoothing.cpp
```diff
@@ -308,6 +191,312 @@ struct InterpTri {
...
+    if (idx[1] != Next3(idx[0])) {
+      triIdx = {idx[2], idx[0], idx[1]};
```
pca006132, line 240: it seems that this `triIdx` is just rotated instead of reversed? not sure if this is what you want.

src/manifold/src/smoothing.cpp
```diff
@@ -308,6 +191,312 @@ struct InterpTri {
...
+      const int n = sortedDivisions[i] - 1;
+      int offset = edgeOffsets[idx[i]] + (edgeFwd[idx[i]] ? 0 : n - 1);
+      for (int j = 0; j < n; ++j) {
+        newVerts.push_back(offset);
+        offset += edgeFwd[idx[i]] ? 1 : -1;
+      }
+    }
+    const int offset = interiorOffset - newVerts.size();
+    for (int i = newVerts.size(); i < vertBary.size(); ++i) {
+      newVerts.push_back(i + offset);
```
pca006132, line 257: probably depends on the size of `vertBary` compared to `newVerts`, but it will be faster if we reserve first and then do `push_back`, or just set the length of `newVerts` to `vertBary`'s length and set the new elements accordingly.

src/manifold/src/smoothing.cpp
```diff
@@ -308,6 +191,312 @@ struct InterpTri {
...
+    }
+
+    Vec<glm::ivec3> newTriVert(triVert.size());
+    for (int i = 0; i < triVert.size(); ++i) {
```
pca006132, line 261: seems like a good parallelization opportunity.

src/manifold/src/smoothing.cpp
```diff
@@ -308,6 +191,312 @@ struct InterpTri {
...
+  static inline auto cacheLock = std::mutex();
+  static inline auto cache =
+      std::unordered_map<glm::ivec3, std::unique_ptr<Partition>>();
+
+  // This triangulation is purely topological - it depends only on the number of
+  // divisions of the three sides of the triangle. This allows them to be cached
+  // and reused for similar triangles. The shape of the final surface is defined
+  // by the tangents and the barycentric coordinates of the new verts. The input
+  // must be sorted: n[0] >= n[1] >= n[2] > 0
+  static Partition GetCachedPartition(glm::ivec3 n) {
```
pca006132, line 279: Maybe we will want a limit on the size of the cache, and make it a LRU or something. Otherwise there may be memory leak in long running applications. Perhaps we can define the limit in `ManifoldParams`?

src/manifold/src/smoothing.cpp
```diff
@@ -421,73 +610,258 @@ void Manifold::Impl::CreateTangents(
...
-  vertPos_.resize(triVertStart + numTri * VertsPerTri(n - 2));
-  Vec<Barycentric> vertBary(vertPos_.size());
-  FillRetainedVerts(vertBary, halfedge_);
-
-  MeshRelationD oldMeshRelation = std::move(meshRelation_);
-  meshRelation_.triRef.resize(n * n * numTri);
-  meshRelation_.originalID = oldMeshRelation.originalID;
-
+Vec<Barycentric> Manifold::Impl::Subdivide(
+    std::function<int(glm::vec3)> edgeDivisions) {
```
pca006132, line 620: Wonder if the division function can be more powerful. For example instead of accepting the edge as a vector, maybe we can get the edge as an index in the original mesh, so we can implement something like remeshing that produce delaunay triangulation? (although iirc those are usually iterative algorithms)

src/manifold/src/smoothing.cpp
```diff
@@ -308,6 +191,312 @@ struct InterpTri {
...
+  // Partitions are parallel to the first edge unless two consecutive edgeAdded
+  // are zero, in which case a terminal triangulation is performed.
+  static void PartitionQuad(Vec<glm::ivec3>& triVert, Vec<glm::vec3>& vertBary,
+                            glm::ivec4 cornerVerts, glm::ivec4 edgeOffsets,
+                            glm::ivec4 edgeAdded, glm::bvec4 edgeFwd) {
+    auto GetEdgeVert = [&](int edge, int idx) {
+      return edgeOffsets[edge] + (edgeFwd[edge] ? 1 : -1) * idx;
+    };
+
+    // std::cout << "added: " << edgeAdded << std::endl;
```
pca006132, line 384: remove these debugging code?

src/manifold/src/smoothing.cpp
elalish: I used to only need that mapping for the forward halfedges, but now I need it for both.

src/manifold/src/smoothing.cpp
elalish: I think the more powerful thing I want to do is also use the edge's tangent vectors - that's my plan for `RefineToPrecision`. Delaunay is great, but you still need to place the interior verts, plus it's geometric, so no chance to cache like this one.

src/manifold/src/smoothing.cpp
elalish: Let's leave that for a follow-on PR. I think it's pretty unlikely the cache will be that huge, and if it is, it means it's being used.

src/manifold/src/smoothing.cpp
elalish: Yeah, it's `outTri` that gets reversed. This is rotated because the association of verts and edges ends up off by one.

src/manifold/src/smoothing.cpp
elalish: newVerts corresponds to vertBary, so they're the same length. It's reserved on line 236.

PR #742, opened 2024-02-15
Fix sparseindices

### LINKED ISSUE #720: Large meshes result in SparseIndicie too large error - Python
Using the Python bindings of the manifold library in-conjunction with the trimesh Python library,  there is the following error when working with extremely large meshes (+200mb)

`IndexError: SparseIndices too large`

This is encountered after conversion to mesh following the boolean operation.

### LINKED ISSUE #738: GetMesh core dump with SparseIndices too large

pca006132:
The changes are everywhere... Basically this PR does two things:

1. Use `size_t` for `Vec` indices when possible.
2. Catch `Vec` that are too large (`>= INT32_MAX`) so other parts that rely on int32 indices will not break.

1 is easy, 2 is hard, and I am pretty sure that there are missing cases or cases that can be relaxed a bit, but still it is better than the status quo.

resolves #720, resolves #738, resolves #740.

Btw @starseeker can you verify that the result is correct? The result is like 3 really long rods that are connected at the end or something, the mesh is extremely slim making it hard to screenshot...


pca006132:
For some reason the macos sdf case keeps failing, no idea why...

starseeker:
@pca006132 That sounds right - the original inputs were indeed long, super-thin cylinders.

starseeker:
Yes - exporting the mesh and visually inspecting it, it appears to be correct.

pca006132:
@elalish btw not sure if we should use the generic `std::out_of_range` or define a custom exception. I just used that because I am lazy.

elalish:
Good question. I think I only defined custom exceptions for things that are unique to manifold. Out of range sounds reasonable for using the standard.

pca006132:
one issue though: std::out_of_range is usually what you get when there is a bug, so it is generally hard to recover. However, in our case it means the operation on large meshes is not supported.

elalish:
Fair, though I'm not really sure how you'd recover from a too-large mesh either, besides just cancel everything and post a message to the user.

pca006132:
The way I view this is similar to rust: if it is a software bug, e.g. the normal index out of bound thing, we should just abort (although we are not that strict now). Otherwise, it is an user error and user can cancel the operation.

src/utilities/include/sparse.h
```diff
@@ -71,33 +68,33 @@ class SparseIndices {
     stable_sort(autoPolicy(size()), view.begin(), view.end());
   }
 
-  void Resize(int size) { data_.resize(size * sizeof(int64_t), -1); }
+  void Resize(size_t size) { data_.resize(size * sizeof(int64_t), -1); }
 
-  inline int& Get(int i, bool use_q) {
+  inline int& Get(size_t i, bool use_q) {
```
elalish, line 73: These look like the critical changes to fix the bug, right?

test/samples_test.cpp
```diff
@@ -306,6 +306,17 @@ TEST(Samples, SelfIntersect) {
   manifold::PolygonParams().processOverlaps = false;
 }
 
+TEST(Samples, GenericTwinBooleanTest7081) {
+  manifold::PolygonParams().processOverlaps = true;
```
elalish, line 310: Is this necessary for this test to pass the CI?

src/sdf/src/sdf.cpp
```diff
@@ -334,10 +334,16 @@ Mesh LevelSet(std::function<float(glm::vec3)> sdf, Box bounds, float edgeLength,
...
+
+      if (tableSize >= std::numeric_limits<int>::max())
+        throw std::out_of_range("mesh too large");
+
       if (ratio > 1000)  // do not trust the ratio if it is too large
         tableSize *= 2;
       else
         tableSize *= ratio;
+      if (tableSize > std::numeric_limits<int>::max())
+        tableSize = std::numeric_limits<int>::max();
```
elalish, line 346: If we've changed it to index with `size_t`, then shouldn't we be able to handle a larger max length?

src/manifold/src/boolean3.cpp
```diff
@@ -578,6 +584,14 @@ Boolean3::Boolean3(const Manifold::Impl &inP, const Manifold::Impl &inQ,
       Intersect12(inQ, inP, s20, p2q0, s11, p1q1, z20, xyzz11, p2q1_, false);
   PRINT("x21 size = " << x21_.size());
 
+  if (p0q2.size() >= std::numeric_limits<int>::max() ||
+      p2q0.size() >= std::numeric_limits<int>::max() ||
```
elalish, line 588: Why should 0-2 (vertices shadowing triangles) have to be fewer than INT_MAX? That's still worst case an O(n^2) combination problem. The other four below this are actually contributing verts, so the check makes more sense (though shouldn't it be their sum?).

elalish (changes requested):

src/manifold/src/boolean3.cpp
```diff
@@ -92,9 +92,12 @@ SparseIndices Filter11(const Manifold::Impl &inP, const Manifold::Impl &inQ,
                        const SparseIndices &p1q2, const SparseIndices &p2q1) {
   ZoneScoped;
   SparseIndices p1q1(3 * p1q2.size() + 3 * p2q1.size());
-  for_each_n(autoPolicy(p1q2.size()), zip(countAt(0), countAt(0)), p1q2.size(),
-             CopyFaceEdges<false>({p1q2, p1q1, inQ.halfedge_}));
-  for_each_n(autoPolicy(p2q1.size()), zip(countAt(p1q2.size()), countAt(0)),
+  for_each_n(
+      autoPolicy(p1q2.size()),
+      zip(countAt(static_cast<size_t>(0)), countAt(static_cast<size_t>(0))),
```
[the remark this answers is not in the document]
pca006132: Yes if we want to change everything into using `size_t`. Thinking about it, we can also use `0ll` to replace the `static_cast`. I used to use `0l` but it errors on Windows, because `long` is 32-bits on Windows and we need `long long`, so I just switched everything into `static_cast`...

src/utilities/include/hashtable.h
```diff
@@ -123,16 +123,21 @@ class HashTableD {
 template <typename V, hash_fun_t H = hash64bit>
 class HashTable {
  public:
-  HashTable(uint32_t size, uint32_t step = 1)
-      : keys_{1 << (int)ceil(log2(size)), kOpen},
-        values_{1 << (int)ceil(log2(size)), {}},
-        table_{keys_, values_, used_, step} {}
+  HashTable(size_t size, uint32_t step = 1)
+      : keys_{size == 0 ? 0 : static_cast<size_t>(1) << (int)ceil(log2(size)),
```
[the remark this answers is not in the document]
pca006132: Not compile error, but by default it cannot go over int32 max, so we need to make it `size_t`. Can also use `1ll`.

src/manifold/src/face_op.cpp
```diff
@@ -155,13 +155,15 @@ void Manifold::Impl::Face2Tri(const Vec<int>& faceEdge,
...
+                 triCount.begin(), static_cast<size_t>(0));
+  if (triCount.back() >= std::numeric_limits<int>::max())
+    throw std::out_of_range("too many triangles");
   triVerts.resize(triCount.back());
   triNormal.resize(triCount.back());
   triRef.resize(triCount.back());
 
   auto processFace2 = std::bind(
-      processFace, [&](int face) { return std::move(results[face]); },
+      processFace, [&](size_t face) { return std::move(results[face]); },
```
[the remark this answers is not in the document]
pca006132: Ah this was to please the compiler, because we count with `size_t` below. Can change to `int` and do a cast below.

src/utilities/include/sparse.h
pca006132: yes, but not only here because we need to make sure that code in `boolean3.cpp` handles this correctly.

test/samples_test.cpp
pca006132: No, will remove.

src/sdf/src/sdf.cpp
pca006132: Ah, this was because `used` is `uint32_t`. Thinking about it, we can just use a size_t, don't even need the vec.

src/manifold/src/boolean3.cpp
pca006132: Indeed, I was thinking that `p0`/`q0` have the same size as `p0q2`/`p2q0`, and thought that they affect the size of `w03_` and `w30_`. Will address this.

src/manifold/src/boolean3.cpp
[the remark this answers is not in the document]
pca006132: ah, nevermind, https://en.cppreference.com/w/cpp/language/integer_literal size_t suffix is only supported on c++23, we can use something user defined though.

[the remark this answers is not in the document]
pca006132: now we can use `0_z` where the `_z` suffix specifies the number is a `size_t`.

src/manifold/src/boolean3.cpp
pca006132: fixed, need to check if the current one is correct.

src/manifold/src/boolean3.cpp
```diff
@@ -351,8 +351,8 @@ std::tuple<Vec<int>, Vec<float>> Shadow02(const Manifold::Impl &inP,
   Vec<float> z02(p0q2.size());
 
   auto vertNormalP = forward ? inP.vertNormal_ : inQ.vertNormal_;
-  for_each_n(autoPolicy(p0q2.size()), zip(countAt(0), s02.begin(), z02.begin()),
-             p0q2.size(),
+  for_each_n(autoPolicy(p0q2.size()),
+             zip(countAt(0ll), s02.begin(), z02.begin()), p0q2.size(),
```
elalish, line 355: Is there a reason this one isn't `0_z`?

src/manifold/src/boolean3.cpp
```diff
@@ -578,6 +578,10 @@ Boolean3::Boolean3(const Manifold::Impl &inP, const Manifold::Impl &inQ,
       Intersect12(inQ, inP, s20, p2q0, s11, p1q1, z20, xyzz11, p2q1_, false);
   PRINT("x21 size = " << x21_.size());
 
+  if (x12_.size() + x21_.size() >= std::numeric_limits<int>::max() ||
+      w03_.size() + w30_.size() >= std::numeric_limits<int>::max())
```
elalish, line 582: Thinking about this, we can probably remove w03 and w30 too: they always just have size of P.NumVert() and Q.NumVert() respectively.

src/sdf/src/sdf.cpp
```diff
@@ -318,14 +318,14 @@ Mesh LevelSet(std::function<float(glm::vec3)> sdf, Box bounds, float edgeLength,
...
-  int tableSize = glm::min(
+  size_t tableSize = glm::min(
       2 * maxMorton, static_cast<Uint64>(10 * glm::pow(maxMorton, 0.667)));
   HashTable<GridVert, identity> gridVerts(tableSize);
   Vec<glm::vec3> vertPos(gridVerts.Size() * 7);
 
   while (1) {
     Vec<int> index(1, 0);
-    for_each_n(pol, countAt(0), maxMorton + 1,
+    for_each_n(pol, countAt(0ul), maxMorton + 1,
```
elalish, line 328: `0_z`?

src/utilities/include/public.h
```diff
@@ -31,6 +31,8 @@
 #include <sstream>
 #endif
 
+constexpr std::size_t operator"" _z(unsigned long long n) { return n; }
```
elalish, line 34: This is cool, but it looks like emscripten doesn't like it?

elalish (changes requested):
Thanks, looks like we're close now.

src/manifold/src/boolean3.cpp
pca006132: ah, make sense

elalish (approved):
Thanks!

PR #747, opened 2024-02-19
prefer find_package for tbb

pca006132:
> `pkg_check_modules(TBB tbb)` bypasses CMake's find system, and pollutes the global namespace with `TBB_*` variables

this avoids failing OpenSCAD reconfigure as they use `find_package`.


kintel (approved):

kintel:
This fixes a config mismatch we were having in OpenSCAD :)

Issue #750, opened 2024-02-20
Better documentation for FillRule


briansturgill:
Straightening this out will require breaking changes.

Two `CrossSection` constructors have a `FillRule` parameter and this is used in the single object union.
In all other instances, the `FillRule` is hard-coded to `FillRule.Positive`.
Thus, even though all FillRules are listed, only FillRule.Positive can actually be used.

In Clipper2 the 'FillRule' is used only in booleans and other clipping functions.
There is no way to provide this information in Manifold's 2d API.
Additionally, the odd use of +, -, and ^ for large, expensive operations makes solving this even harder.

One idea would be to require the use of FillRule.Positive. The FillRule parameter would eventually go
away and would throw an error in the meantime if some other FillRule was used.
This probably breaks the least existing code, but is also the least flexible.

I want to raise the caveat that  FillRule.EvenOdd requires no knowledge of winding on the part of end users
(it all gets sorted out automatically). That was my original plan for use in Piecad, though
I have ways around this if needed.

A second idea is to have a globally set a FillRule variable that is then used by all functions.
Similar to how OpenSCAD uses $fn.
This maintains flexibility, while breaking very little existing code.
I'm not sure what Manifold policy is for global static vars.
Enums (ie. FillRule) are allowed to have static variables. Perhaps it could go there?

A third idea is to go through each 2D call and add the ability to set FillRule.
For +, - and ^ one could default to FillRule.Positive and corresponding union/subtract/intersect functions if you
want to use something other than FillRule.Positive.
This option best matches Clipper's idea of how things should work.

I guess I favor the second option. I'd like to use FillRule.EvenOdd if I can.

I'll gladly do a PR for this, but a direction needs to be set first.

geoffder:
The fill rule parameter is only exposed in the constructors intentionally, as once the initial "cleanup" union is performed on user provided paths, all winding thereafter is positive.

elalish:
Lots of confidence in this title... would you mind providing an example of what is broken, i.e. a returned vs expected result?

EvenOdd loses winding direction information and results lots of "kissing" polygons, which is not terribly helpful for solid geometry.

pca006132:
In my understanding, everything in manifold is done in `Positive` fill rule. The constructor is just for users to convert their polygons using alternate fill rules into `Positive`. I think as long as the representation is consistent, the fill rule used for boolean operation should not matter?

Maybe we should document this better though, e.g. the result of `ToPolygons` is `Positive`.

[...]

Issue #723, opened 2024-01-28
triangulator optimization


pca006132:
The triangulator takes `O(n^2)` time, which performs poorly when the number of points of a polygon increases. For example, for the woodgrain svg we had for #502, there are around 60k points in the polygon. The old triangulator (not as robust) takes 35ms to triangulate the 60k points, while our current triangulator takes around 11500ms (11.5s!). We should try to use bvh to reduce the number of points to check, to avoid quadratic complexity in the general case (for degenerate cases bvh can't help much I guess).

- [ ] Make `Collider` generic and support 2D `Rect`. This is not necessary, but will probably help with performance and is not hard to do.
- [ ] Add `Remove` method to `Collider` to support point removal. We don't have to care about rebalancing because the height is bounded (at least in terms of complexity it does not matter, not sure about that in practice, and rebalancing is costly).
- [ ] Avoid calculating `DelaunayCost`. We can just categorize ears into different categories: definitely valid, potentially valid, definitely invalid, according to the max cost over query results.

@elalish what do you think?

elalish:
Yeah, that sounds good. And actually, I think Delaunay cost should work too - we only care within a bounded region: Delaunay only matters for points closer to the edge midpoint than its half-length. So we should still be able to safely screen out more distant points with BHV.

pca006132:
btw wonder if it is possible to move some of the sort by morton code logic into the collider.

elalish:
> btw wonder if it is possible to move some of the sort by morton code logic into the collider.

Or in its own header file. Still, it's important the sort the actual underlying data structures, since that gives us cache coherence. So the Morton codes are for more than just the collider.

onlynadeem:
I know how to do triangulator optimization very well, but where to import its source file?

pca006132:
https://github.com/elalish/manifold/blob/master/src/polygon/src/polygon.cpp

pca006132:
Adding a comment from #751:

> Optimizations related to triangulator is interesting, I am open to adding fast paths, but a more general one (e.g. monotone triangulation, but doesn't handle epsilon), and directly into the triangulator rather than limited to `extrude`.

kintel:
This is starting to hurt for some OpenSCAD usage too. It's also surprisingly slow in debug mode:
* Manifold debug: 5m50s
* Manifold release: 12s
* CGAL constrained delaunay triangulator: 0.3s

^ not very scientific, just measuring OpenSCAD e2e processing time.

[...]

PR #751, opened 2024-02-25
Changes that support piecad.

briansturgill:
This pull request contains several things I've worked on when experimenting
with improving Manifold's support for Python and especially 2d objects.
These are the ones that have provided the most speed improvments.

NOTE: in benchmarks `_mo_` indicates a direct test of Manifold objects from Python.

<details><summary>Benchmarks overview.</summary>

```
Benchmark Name (times are means in us)             Current         New
---
test_mo_square                                        0.69        0.71
test_mo_circle_10                                     1.05        1.02
test_mo_rounded_rectangle                                -        1.17
test_mo_circle_100                                    4.94        4.74
test_mo_cube                                          5.64        5.44
test_cube_from_create_from_verts_and_triangles       18.90        8.23
test_rounded_rectangle                               14.74        9.09
test_mo_cylinder_100_centered                       521.90      161.38
test_mo_cylinder_100                                283.59      161.54
test_extrude_transforming_circle_100_fan                 -      155.68
test_extrude_transforming_circle_100_earcut              -      276.72

```
</details>

## One

The number one issue is extrusion speed. Manifold uses an earcut derivative,
which like earcut, is rather slow.

As you can see, currently, cylinders are extruded and centered cylinders are just sad.

```
Benchmark Name (times are means in us)             Current         New
---
test_mo_cylinder_100_centered                       521.90      161.38
test_mo_cylinder_100                                283.59      161.54
test_extrude_transforming_circle_100_fan                 -      155.68
test_extrude_transforming_circle_100_earcut              -      276.72
```

I fixed this issue by making a function called ExtrudeTransforming, which
contains all the code from Extrude, with two additions.

One, if you set isConvex to true, then fan triangulation is used.

Two, I added initialZ, which allows one to use extrude to properly handle centering the cylinder.

The Extrude routine is now just a call to ExtrudeTransforming without the two new arguments.
Leaving Manifold's API the same.

The two `test_extrude_trans..." test are tests ONLY of extrude time, not circle creation time.
It shows clearly how much time is wasted when unnecessarily using earcut.

## Two

The second issue is the unnecessary overhead that is endured by Python programs needing to
make primitives.

```
Benchmark Name (times are means in us)             Current         New
---
test_rounded_rectangle                               14.74        9.09
test_cube_from_create_from_verts_and_triangles       18.90        8.23
```

The new versions skip Unions (2D) and use a very cut down Extrude called CreateFromVertsAndTriangles,
that simply pases through the verts and triangles (3D).

(The CrossSection addition is called CreateFromPolygonsUnchecked.)

## Three

The third issue issue isn't really an issue, it's just that a built in RoundedRectangle works much faster.
RoundedRectangles cannot be scaled without deforming the corners, so I cache (by circularSegments)
the unit trig ops and it works about as fast as scaling a unit circle does.

```
Benchmark Name (times are means in us)             Current         New
---
test_mo_rounded_rectangle                                -        1.17
```

Compare that against the Python versions:

```
test_rounded_rectangle                               14.74        9.09
```

### DIFF
--- a/src/cross_section/src/cross_section.cpp
+++ b/src/cross_section/src/cross_section.cpp
@@ -305,6 +305,129 @@ std::shared_ptr<const PathImpl> CrossSection::GetPaths() const {
 }
 
 /**
+ * Constructs a CrossSection from a vector of polygons.
+ * No checking is done, so don't have self-intersecting paths
+ * Polygon contours should be wound CCW.
+ * Polygon holes should be wound CW.
+ *
+ * @param polygons A vector of a vector of path points.
+ */
+CrossSection CrossSection::CreateFromPolygonsUnchecked(
+    std::vector<std::vector<glm::vec2>> polygons) {
+  if (polygons.size() < 1) {
+    return CrossSection();
+  }
+  for (int i = 0; i < polygons.size(); i++) {
+    if (polygons[i].size() < 3) {
+      return CrossSection();
+    }
+  }
+
+  auto ps = C2::PathsD(polygons.size());
+
+  for (int i = 0; i < ps.size(); i++) {
+    auto path = polygons[i];
+    auto p = C2::PathD(polygons[i].size());
+    for (int j = 0; j < p.size(); j++) {
+      auto v = path[j];
+      p[j] = C2::PointD(v.x, v.y);
+    }
+    ps[i] = p;
+  }
+  return CrossSection(shared_paths(ps));
+}
+
+/**
+ * Constructs a rounded rectangle from given size and radius.
+ *
+ * @param size The X, and Y dimensions of the square.
+ * @param radius Radius of the circle. Must be positive.
+ * @param circularSegments Number of segments along its diameter. Default is
+ * calculated by the static Quality defaults according to the radius.
+ * @param center Set to true to shift the center to the origin.
+ */
+CrossSection CrossSection::RoundedRectangle(const glm::vec2 size, float radius,
+                                            int circularSegments, bool center) {
+  if (size.x < 0.0f || size.y < 0.0f || glm::length(size) == 0.0f) {
+    return CrossSection();
+  }
+  if (radius <= 0.0f) {
+    return CrossSection();
+  }
+  auto n = circularSegments > 2 ? circularSegments
+                                : Quality::GetCircularSegments(radius);
+  int segs_per_arc = std::trunc(n / 4) + 1;
+  float degs_per_seg = 90.0f / segs_per_arc;
+  auto arc = [segs_per_arc, degs_per_seg](float start) {
+    std::vector<glm::vec2> l;
+    auto end = start + 90.0f;
+    auto deg = start;
+    for (auto i = 0; i < segs_per_arc - 1; i++) {
+      l.push_back(glm::vec2(cosd(deg), sind(deg)));
+      deg += degs_per_seg;
+    }
+    l.push_back(glm::vec2(cosd(end), sind(end)));
+    return l;
+  };
+  static std::unordered_map<int, std::vector<std::vector<glm::vec2>>> arc_map =
+      {};
+
+  std::vector<std::vector<glm::vec2>> arcs;
+  auto search = arc_map.find(n);
+  if (search != arc_map.end()) {
+    arcs = search->second;
+  } else {
+    arcs = std::vector<std::vector<glm::vec2>>();
+    arcs.push_back(arc(180));  // Bottom left
+    arcs.push_back(arc(270));  // Bottom right
+    arcs.push_back(arc(0));    // Top right
+    arcs.push_back(arc(90));   // Top left
+    arc_map[n] = arcs;
+  }
+  auto bl = arcs[0];  // Bottom left
+  auto br = arcs[1];  // Bottom right
+  auto tr = arcs[2];  // Top right
+  auto tl = arcs[3];  // Top left
+
+  auto len_arc = bl.size();
+
+  auto x = size.x;
+  auto y = size.y;
+
+  auto x_off = 0.0f;
+  auto y_off = 0.0f;
+
+  if (center) {
+    x_off = -x / 2.0;
+    y_off = -y / 2.0;
+  }
+
+  auto rr = C2::PathD(4 * len_arc);
+  int pdidx = 0;
+  for (auto vec : bl) {  // Bottom left
+    rr[pdidx] = C2::PointD(radius + radius * vec.x + x_off,
+                           radius + radius * vec.y + y_off);
+    pdidx++;
+  }
+  for (auto vec : br) {  // Bottom right
+    rr[pdidx] = C2::PointD((x - radius) + radius * vec.x + x_off,
+                           radius + radius * vec.y + y_off);
+    pdidx++;
+  }
+  for (auto vec : tr) {  // Top right
+    rr[pdidx] = C2::PointD((x - radius) + radius * vec.x + x_off,
+                           (y - radius) + radius * vec.y + y_off);
+    pdidx++;
+  }
+  for (auto vec : tl) {  // Top left
+    rr[pdidx] = C2::PointD(radius + radius * vec.x + x_off,
+                           (y - radius) + radius * vec.y + y_off);
+    pdidx++;
+  }
+  return CrossSection(shared_paths(C2::PathsD{rr}));
+}
+
+/**
  * Constructs a square with the given XY dimensions. By default it is
  * positioned in the first quadrant, touching the origin. If any dimensions in
  * size are negative, or if all are zero, an empty Manifold will be returned.
--- a/src/manifold/include/manifold.h
+++ b/src/manifold/include/manifold.h
@@ -146,9 +146,17 @@ class Manifold {
                            float radiusHigh = -1.0f, int circularSegments = 0,
                            bool center = false);
   static Manifold Sphere(float radius, int circularSegments = 0);
+  static Manifold CreateFromVertsAndTriangles(
+      std::vector<glm::vec3> verts, std::vector<glm::ivec3> triangles);
   static Manifold Extrude(const CrossSection& crossSection, float height,
                           int nDivisions = 0, float twistDegrees = 0.0f,
                           glm::vec2 scaleTop = glm::vec2(1.0f));
+  static Manifold ExtrudeTransforming(const CrossSection& crossSection,
+                                      float height, int nDivisions = 0,
+                                      float twistDegrees = 0.0f,
+                                      glm::vec2 scaleTop = glm::vec2(1.0f),
+                                      float initialZ = 0.0f,
+                                      bool isConvex = false);
   static Manifold Revolve(const CrossSection& crossSection,
                           int circularSegments = 0,
                           float revolveDegrees = 360.0f);
--- a/src/manifold/src/constructors.cpp
+++ b/src/manifold/src/constructors.cpp
@@ -183,10 +183,9 @@ Manifold Manifold::Cylinder(float height, float radiusLow, float radiusHigh,
 
   CrossSection circle = CrossSection::Circle(radiusLow, n);
   Manifold cylinder =
-      Manifold::Extrude(circle, height, 0, 0.0f, glm::vec2(scale));
-  if (center)
-    cylinder =
-        cylinder.Translate(glm::vec3(0.0f, 0.0f, -height / 2.0f)).AsOriginal();
+      Manifold::ExtrudeTransforming(circle, height, 0, 0.0f, glm::vec2(scale),
+                                    center ? -height / 2.0f : 0.0f, true);
+
   return cylinder;
 }
 
@@ -217,6 +216,37 @@ Manifold Manifold::Sphere(float radius, int circularSegments) {
 }
 
 /**
+ * Constructs a manifold from lists of vertices and triangles.
+ *
+ * @param vertices A vector of glm::vec3.
+ * @param triangles A vector of glm::ivec3.
+ *
+ */
+Manifold Manifold::CreateFromVertsAndTriangles(
+    std::vector<glm::vec3> verts, std::vector<glm::ivec3> triangles) {
+  ZoneScoped;
+
+  auto pImpl_ = std::make_shared<Impl>();
+  auto& vertPos = pImpl_->vertPos_;
+  Vec<glm::ivec3> triVertsDH;
+  auto& triVerts = triVertsDH;
+
+  for (auto v : verts) {
+    vertPos.push_back(v);
+  }
+  for (auto tri : triangles) {
+    triVerts.push_back(tri);
+  }
+
+  pImpl_->CreateHalfedges(triVertsDH);
+  pImpl_->Finish();
+  pImpl_->meshRelation_.originalID = ReserveIDs(1);
+  pImpl_->InitializeOriginal();
+  pImpl_->CreateFaces();
+  return Manifold(pImpl_);
+}
+
+/**
  * Constructs a manifold from a set of polygons by extruding them along the
  * Z-axis.
  * Note that high twistDegrees with small nDivisions may cause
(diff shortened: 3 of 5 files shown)


src/manifold/include/manifold.h
```diff
@@ -146,9 +146,17 @@ class Manifold {
                            float radiusHigh = -1.0f, int circularSegments = 0,
                            bool center = false);
   static Manifold Sphere(float radius, int circularSegments = 0);
+  static Manifold CreateFromVertsAndTriangles(
+      std::vector<glm::vec3> verts, std::vector<glm::ivec3> triangles);
   static Manifold Extrude(const CrossSection& crossSection, float height,
                           int nDivisions = 0, float twistDegrees = 0.0f,
                           glm::vec2 scaleTop = glm::vec2(1.0f));
+  static Manifold ExtrudeTransforming(const CrossSection& crossSection,
```
elalish, line 154: In general, I consider it a very high bar to accept a change that enlarges our public API. Significant new functionality that can't easily be implemented externally to the library: good chance. Performance improvement: best done while avoiding new APIs, as this makes it harder for new users to discover the optimal way to use the library.

src/cross_section/src/cross_section.cpp
```diff
@@ -304,6 +304,129 @@ std::shared_ptr<const PathImpl> CrossSection::GetPaths() const {
...
+/**
+ * Constructs a rounded rectangle from given size and radius.
+ *
+ * @param size The X, and Y dimensions of the square.
+ * @param radius Radius of the circle. Must be positive.
+ * @param circularSegments Number of segments along its diameter. Default is
+ * calculated by the static Quality defaults according to the radius.
+ * @param center Set to true to shift the center to the origin.
+ */
+CrossSection CrossSection::RoundedRectangle(const glm::vec2 size, float radius,
```
elalish, line 349: This is quite a bit of extra code to maintain for a pretty narrow use case. Also, did you intend to include tests?

src/cross_section/src/cross_section.cpp
```diff
@@ -304,6 +304,129 @@ std::shared_ptr<const PathImpl> CrossSection::GetPaths() const {
...
 
+/**
+ * Constructs a CrossSection from a vector of polygons.
+ * No checking is done, so don't have self-intersecting paths
+ * Polygon contours should be wound CCW.
+ * Polygon holes should be wound CW.
+ *
+ * @param polygons A vector of a vector of path points.
+ */
+CrossSection CrossSection::CreateFromPolygonsUnchecked(
```
elalish, line 315: I understand the idea here, but I'm not convinced. This is mostly because "checked" is not an easily-testable condition that I can trust external developers have done in the same way as the library. Anything related to floating-point geometry has this trouble. I think it leaves open too many foot-guns.

src/manifold/src/constructors.cpp
```diff
@@ -231,13 +261,38 @@ Manifold Manifold::Sphere(float radius, int circularSegments) {
...
+ * @param nDivisions Number of extra copies of the crossSection to insert into
+ * the shape vertically; especially useful in combination with twistDegrees to
+ * avoid interpolation artifacts. Default is none.
+ * @param twistDegrees Amount to twist the top crossSection relative to the
+ * bottom, interpolated linearly for the divisions in between.
+ * @param scaleTop Amount to scale the top (independently in X and Y). If the
+ * scale is {0, 0}, a pure cone is formed with only a single vertex at the
+ * top. Note that scale is applied after twist. Default {1, 1}.
+ */
+Manifold Manifold::ExtrudeTransforming(const CrossSection& crossSection,
```
elalish, line 292: Same problem - `isConvex` is not as trivial to test as you might think. If you're certain of your answer, you probably aren't doing it right.

src/manifold/src/constructors.cpp
```diff
@@ -216,6 +215,37 @@ Manifold Manifold::Sphere(float radius, int circularSegments) {
...
 }
 
+/**
+ * Constructs a manifold from lists of vertices and triangles.
+ *
+ * @param vertices A vector of glm::vec3.
+ * @param triangles A vector of glm::ivec3.
+ *
+ */
+Manifold Manifold::CreateFromVertsAndTriangles(
```
elalish, line 225: Thank you for providing benchmarks, but the reason these functions are not as deeply optimized is that they are not generally performance-limiting. So while your point-benchmarks show the differences narrowly, I would be more convinced by a realistic end-to-end test - look at our Samples tests for instance. Where will you be calling so many constructors and so few Boolean ops that the constructors actually dominate? I am honestly interested, as you may well have a use case I haven't considered.

elalish (commented):
Please answer the discussions before adding more code - I am yet to be convinced this will be mergable.

pca006132:
I think adding fast path is fine, but adding fast paths that have really narrow usecases is not a good idea. It will be more likely to be merged if the performance improvement related to using alternative triangulation algorithm can be applied in general, not just to extrude, and not with special flags.

E.g. Maybe we can have a fast monotone triangulation that does not attempt to use epsilon, and fallback to the robust version if it fails. I think this can benefit boolean operation as well.

---

I am more curious as to why `test_mo_cylinder_100_centered` is 1x slower than `test_mo_cylinder_100`. Can you show me the actual benchmark code? I think the way you implemented the optimization by adding `initialZ` can just avoid triggering the slow path *in this particular case* instead of actually fixing the issue. It would be nice to find the root cause and fix that.

src/manifold/src/constructors.cpp
briansturgill: @elalish Possible mis-understanding here.  The isConvex flag is set by the caller, THE CALLER determines whether or not the input is convex.  For example, Cylinder, the circles involved OBVIOUSLY are convex, and thus the version of Cylinder I made sets that flag. The default is False, and will work just fine, but slower.

src/cross_section/src/cross_section.cpp
briansturgill: It's really not that narrow of a use case. It was one of the most asked questions with OpenSCAD (which did not have a good way to do this). Note I would not need to ask if you actually provided a way to make this from outside Manifold. Currently I lose an order of magnitude due to artificial overhead.

briansturgill:
@pca006132 Here is my pytest-benchmark for Manifold objects.
It's so I can make sure I'm not losing too much in overhead.
For example I do extensive argument error checking.

<details><summary>Test code.</summary>

```python
import manifold3d as _m
import pytest
import numpy as _np
from piecad import cos, sin


def _square(sz):
    o = _m.CrossSection.square(sz)
    o.num_vert()
    return o


def _circle(r, s):
    c = _m.CrossSection.circle(r, s)
    c.num_vert()
    return c


def _cylinder(h, r, s, centered):
    o = _m.Manifold.cylinder(h, r, r, s, centered)
    o.num_vert()
    return o


def _cube(sz):
    o = _m.Manifold.cube(sz)
    o.num_vert()
    return o


def test_mo_circle_10(benchmark):
    o = benchmark(_circle, 3, 10)


def test_mo_circle_100(benchmark):
    o = benchmark(_circle, 3, 100)


def test_mo_square(benchmark):
    o = benchmark(_square, (10, 10))


def test_mo_cylinder_100(benchmark):
    o = benchmark(_cylinder, 15, 10, 100, False)


def test_mo_cylinder_100_centered(benchmark):
    o = benchmark(_cylinder, 15, 10, 100, True)


def test_mo_cube(benchmark):
    o = benchmark(_cube, (15, 10, 36))


def _from_verts_and_triangles(v, f):
    if hasattr(_m.Manifold, "create_from_verts_and_triangles"):
        o = _m.Manifold.create_from_verts_and_triangles(v, f)
    else:
        v = _np.array(v, _np.float32)
        f = _np.array(f, _np.uint32)
        mesh = _m.Mesh(v, f)
        o = _m.Manifold(mesh)
    o.num_vert()
    return o


def test_cube_from_create_from_verts_and_triangles(benchmark):
    w = 10.0
    d = 10.0
    h = 10.0
    vertices = [
        (0.0, 0.0, 0.0),
        (0.0, 0.0, h),
        (0.0, d, 0.0),
        (0.0, d, h),
        (w, 0.0, 0.0),
        (w, 0.0, h),
        (w, d, 0),
        (w, d, h),
    ]
    triangles = [
        (1, 0, 4),
        (2, 4, 0),
        (1, 3, 0),
        (3, 1, 5),
        (3, 2, 0),
        (3, 7, 2),
        (5, 4, 6),
        (5, 1, 4),
        (6, 4, 2),
        (7, 6, 2),
        (7, 3, 5),
        (7, 5, 6),
    ]

    out = benchmark(_from_verts_and_triangles, vertices, triangles)
    assert out.num_vert() == 8


_arc_trig_vals_map = {}


def _rounded_rectangle(
    size: list[float, float],
    rounding_radius: float = 2.0,
    segments: int = 36,
    center: bool = False,
) -> _m.CrossSection:
    segs_per_arc = segments // 4 + 1
    deg_per_seg = 90.0 / segs_per_arc
    pts = []
    x, y = size

    def make_arc_trig_vals(deg):
        end = deg + 90
        l = []
        for i in range(0, segs_per_arc - 1):
            l.append((cos(deg), sin(deg)))
            deg += deg_per_seg

        l.append((cos(end), sin(end)))
        return l

    if segments in _arc_trig_vals_map:
        arc_trig_vals = _arc_trig_vals_map[segments]
    else:
        arc_trig_vals = (
            make_arc_trig_vals(180),  # Bottom left
            make_arc_trig_vals(270),  # Bottom right
            make_arc_trig_vals(0),  # Top right
            make_arc_trig_vals(90),  # Top left
        )
        _arc_trig_vals_map[segments] = arc_trig_vals

    rr = rounding_radius

    c_x_off = -x / 2.0 if center else 0.0
    c_y_off = -y / 2.0 if center else 0.0

    bl, br, tr, tl = arc_trig_vals

    x_off = rr + c_x_off
    y_off = rr + c_y_off
    for c, s in bl:  # Bottom left
        pts.append((x_off + rr * c, y_off + rr * s))

    x_off = x - rr + c_x_off
    y_off = rr + c_y_off
    for c, s in br:  # Bottom right
        pts.append((x_off + rr * c, y_off + rr * s))

    x_off = x - rr + c_x_off
    y_off = y - rr + c_y_off
    for c, s in tr:  # Top right
        pts.append((x_off + rr * c, y_off + rr * s))

    x_off = rr + c_x_off
    y_off = y - rr + c_y_off
    for c, s in tl:  # Top left
        pts.append((x_off + rr * c, y_off + rr * s))

    if hasattr(_m.CrossSection, "create_from_polygons_unchecked"):
        o = _m.CrossSection.create_from_polygons_unchecked([pts])
    else:
        o = _m.CrossSection([pts], _m.FillRule.EvenOdd)
    o.num_vert()
    return o


def test_rounded_rectangle(benchmark):
    o = benchmark(_rounded_rectangle, (10, 12))
    assert o.num_vert() == 40


if hasattr(_m.CrossSection, "rounded_rectangle"):

    def _mo_rounded_rectangle(sz, r, segs):
        o = _m.CrossSection.rounded_rectangle(sz, r, segs)
        o.num_vert()
        return o

    def test_mo_rounded_rectangle(benchmark):
        o = benchmark(_mo_rounded_rectangle, (10.0, 12.0), 2.0, 36)
        assert o.num_vert() == 40


if hasattr(_m.Manifold, "extrude_simple"):

    def _mo_extrude_simple(cs, isConvex):
        o = _m.Manifold.extrude_simple(cs, 10, is_convex=isConvex)
        o.num_vert()
        return o

    def test_extrude_simple_circle_100_earcut(benchmark):
        c = _m.CrossSection.circle(5, 100)
        o = benchmark(_mo_extrude_simple, c, False)
        assert o.num_vert() == 200

    def test_extrude_simple_circle_100_fan(benchmark):
        c = _m.CrossSection.circle(5, 100)
        o = benchmark(_mo_extrude_simple, c, True)
        assert o.num_vert() == 200


if hasattr(_m.Manifold, "extrude_transforming"):

    def _mo_extrude_transforming(cs, isConvex):
        o = _m.Manifold.extrude_transforming(cs, 10, is_convex=isConvex)
        o.num_vert()
        return o

    def test_extrude_transforming_circle_100_earcut(benchmark):
        c = _m.CrossSection.circle(5, 100)
        o = benchmark(_mo_extrude_transforming, c, False)
        assert o.num_vert() == 200

    def test_extrude_transforming_circle_100_fan(benchmark):
        c = _m.CrossSection.circle(5, 100)
        o = benchmark(_mo_extrude_transforming, c, True)
        assert o.num_vert() == 200
```
</details>

briansturgill:
@pca006132 As to centered cylinder... you extrude it (slowly with EarCut), then Translate it, then do the dreaded AsOriginal(). That's where the problem is. The initialZ is a variable that should be on almost every extrude.
The need to center is common.  Take a look at my Cylinder code in the PR.

pca006132:
I think the solution should be to change the internal API instead of adding `initialZ` to the external API...

src/cross_section/src/cross_section.cpp
briansturgill: @elalish As to tests, for the parts you think you might want to accept, I'll gladly add tests.
I frankly am skeptical that you'll accept any of it.
But I had done the work, I'm simply offering it for your potential use.

src/cross_section/src/cross_section.cpp
pca006132: From what I know so far, no one really complained about the lack of primitives because they implement their own primitives (from code years ago), and just construct the manifold object from their mesh object.
And our checks are really needed to make sure they don't blow their legs off...

Also, I think most of their concerns are about the performance of boolean operations. Mesh construction performance is rarely the bottleneck AFAIK. Maybe the performance for extruding complex SVG is, but those are really complicated and are not convex.

src/manifold/src/constructors.cpp
pca006132: @briansturgill The issue here is that very often users don't (exactly) know what they are doing, they often just copy some example code or read some comments about *HOW TO MAKE YOUR CODE 3 TIMES FASTER* and just set `isConvex = true`. In those cases our internal data structure will silently corrupt, then users will come and ask us why our library is outputting problematic meshes. This is why we are hesitant in adding this kind of `unchecked` APIs.

E.g. `convexity` parameter in openscad. A lot more complicated than just `isConvex`, but the idea is the same...

src/cross_section/src/cross_section.cpp
pca006132: btw, isn't `CreateFromVertsAndTriangles` just the manifold constructor that constructs from a mesh?

briansturgill: > I understand the idea here, but I'm not convinced. This is mostly because "checked" is not an easily-testable condition that I can trust external developers have done in the same way as the library. Anything related to floating-point geometry has this trouble. I think it leaves open too many foot-guns.

Why do you need to "trust external developers"? When I use these routines I "shoot myself in the foot" a lot.
The world doesn't end.  Surely, it's my problem if I fail to provide proper input to those routines.

pca006132:
@briansturgill Here are my views:

1. Optimization to improve the performance of cylinder when `centered=true` is welcome, but we should probably move the transform into internal API similar to #717. This way the external API is cleaner.
2. Optimizations related to triangulator is interesting, I am open to adding fast paths, but a more general one (e.g. monotone triangulation, but doesn't handle epsilon), and directly into the triangulator rather than limited to `extrude`. @elalish what do you think about this one? We can probably discuss in #723 
3.  Unchecked APIs will probably not be added, those checks are worth the performance overhead as it prevents users from blowing their legs off (it catches many many manifoldness issues in openscad). If you find performance to be a serious issue, we can look into optimizing things including manifoldness checks. In general though we recommend implementing some caching mechanism, so repeated calls to the same primitive is not necessary (we should probably implement this in manifold instead).

elalish:
@pca006132, I'm ambivalent to 1, agree with 2 (though adding BVH to earclip is another option), and certainly agree with 3. 

Yes, if you're going to use a bunch of cylinders, you should certainly create one and copy it rather than constructing them over and over. That seems simple enough for users we shouldn't need to add a caching layer for it, but I guess we'll see.

src/manifold/src/constructors.cpp
elalish: @briansturgill This is the most important comment to address.

src/cross_section/src/cross_section.cpp
elalish: Accepting external code is not quite like getting mana from heaven. That you say this makes me think you have not yet accepted external code into one of your projects?

src/cross_section/src/cross_section.cpp
elalish: @briansturgill Please understand that we don't know anything about you or your history, only what you have written to us on Github. The first thing I remember you saying to me was that you weren't planning to use Manifold because you were going to write your own Mesh Boolean in "a couple of weeks". You also recently blamed us for poor design for what turned out to be your own lack of understanding of the basic underpinnings of computational geometry (winding numbers): #750. None of this gives me much confidence in your opinions. 

I'm being harsher here than usual because I don't feel you're giving my collaborators the respect they deserve. Maybe you're the best developer the world has ever known, but to prove it to me, you need to start by being humble, listening well, asking questions, and leading with evidence instead of opinions.

elalish:
I'm going to close this - if you want to continue down this path, go ahead and do it the traditional way: split each separate idea out instead of combining everything together. Start with an issue that describes a realistic use case and how one of these issues causes a significant fraction of the performance time, and propose one or several solutions. Anything requiring a public API change, get agreement in the issue before starting the PR. Then make a PR with tests and samples to demonstrate the improvement.

PR #762, opened 2024-03-05
v2.4.5

elalish:
Well, I have no idea why `github.ref == 'refs/heads/master'` was false, but I don't think that guard is especially important. At this point I just want to remove anything that might stop a release action from running, since they really never want to finish running.


pca006132:
wow, just found that there are 5 versions in 1 day! I think we can try to manually trigger the actions to debug deployment issues. These actions are quite painful to debug...

elalish:
Yeah, I also found [Act](https://github.com/nektos/act/discussions/2238), which might help.

Issue #186, opened 2022-08-14
Add minGap function


elalish:
A convenient function would be to measure the minimum gap between two Manifolds (good for checking that mechanisms have the proper tolerances). The signature would be: `float Manifold.MinGap(const Manifold& other, float searchLength) const`

It would always return a value between zero and `searchLength`. It would perform an intersection and if that was non-empty, it returns zero (overlapping). If not, it will collide the triangles of each mesh with the verts of the other, expanded into boxes `2*searchLength` wide, then for those collisions, calculate point-to-triangle distances. The minimum is returned.

pca006132:
@elalish I updated the GSoC description to calculate triangle-to-triangle distance, because it is different from point-to-triangle distance. Consider two triangles where a pair of edges touch each other, but on two orthogonal planes. The point-to-triangle distance from any vertex to the opposite triangle is non-zero, but the triangle-to-triangle distance is zero.

elalish:
Good call, thanks!

elalish:
Added by #765

Issue #667, opened 2023-12-20
more fuzz tests


pca006132:
We only have tests for the triangulator, and the test is only checking for segfault/memory corruption and infinite loops. We should add more tests.

1. [ ] Add tests for the triangulator with valid polygons.
2. [ ] Add tests for CSG.

CSG can be fuzzed with a simple grammar.

pca006132:
probably good dataset: https://ten-thousand-models.appspot.com/results.html?q=is+manifold%2C+is+oriented%2C+with+degenerate

dasoya:
Dear Emmett Lalish,

Thank you for considering my proposal.

I am passionate about working in the entertainment industry, such as
digital fashion using computer graphics. During my internship with NCsoft's
Digital Human team, I utilized C++ in Unreal Engine and TensorFlow to
develop 3D modeling AI. I developed games and had the opportunity to be a
PM and lead the release. My strength is that I take responsibility and get
things done. My technical skills include implementing ray tracing using C++
based on OpenGL. I have a strong interest in the technology that Manifold
implements and am willing to make improvements where needed.

Thank you for your consideration,

Kind regards
Daso Jung.


2024년 4월 17일 (수) 오전 2:13, Emmett Lalish ***@***.***>님이 작성:

elalish:
@dasoya That's wonderful, and we'd certainly appreciate your contributions! From the GSoC side, they're much more willing to accept those who have demonstrated successful contribution to the code base. If you'd be willing to open a simple PR we can review and merge, it'll go a long way. If you're not sure where to start, a simple task we've been meaning to get around to is splitting our tests into more files - there are so many now that the granularity is much too coarse. If you'd like to take a stab at organizing some of them a bit, we'd certainly appreciate it.

dasoya:
Thank you for your suggestion. I’ll try to contribute to the code base!

Daso Jung

dasoya:
@elalish  As you suggested, I thought boolean_test.cpp has high complexity, so I divided it into two files.
I'm going to make additional boolean tests in the complex_boolen_test.cpp or new fuzz test file.

https://github.com/elalish/manifold/compare/master...dasoya:manifold:boolean_fuzz

Is this the right way? Let me know what you think.

elalish:
Yes, that's a great start - go ahead and open a PR. It's always best to have many, smaller PRs anyway, and it gives us a good place to discuss.

zmerlynn:
[...]

PR #765, opened 2024-03-06
Add MinGap function

mleleszi:
implements https://github.com/elalish/manifold/issues/186 / https://github.com/opencax/GSoC/issues/82


elalish:
Looking at the CI - is there a chance you're working from an out-of-date branch? You might want to try `git pull origin master`. Let us know when you're ready for some review.

pca006132:
the CI seems fine, the function `TriangleDistance` indeed is only declared but undefined.

mleleszi:
Added the missing definition, so the build should work fine now. (tests are still gonna fail though)

src/manifold/src/manifold.cpp
```diff
@@ -881,4 +882,120 @@ Manifold Manifold::Hull() const { return Hull(GetMesh().vertPos); }
...
+
+  std::transform(faceBoxOther.begin(), faceBoxOther.end(), faceBoxOther.begin(),
+                 [searchLength](Box box) {
+                   return Box(box.min - glm::vec3(searchLength),
+                              box.max + glm::vec3(searchLength));
+                 });
+
+  Collider collider(faceBox, faceMorton);
+
+  SparseIndices collisions = collider.Collisions(faceBoxOther.cview());
```
mleleszi, line 995: @elalish @pca006132 
Got a basic implementation working, the test are passing! Now I only need to figure out how to use the collider for the broad phase. Can I get some pointers here?

I looked how the collider is used in [Manifold::Impl::EdgeCollisions](https://github.com/elalish/manifold/blob/master/src/manifold/src/impl.cpp#L868), am I supposed to implement something similar, but for the expanded faces? Also looked at the how the broad phase is implemented [here](https://github.com/elalish/manifold/blob/master/src/manifold/src/boolean3.cpp#L499).

Am I completely off base with this code snippet? Also, I could also use a few pointers on working with SparseIndices, like getting the vertices out of them to calculate the tri-tri distances.

pca006132: looks fine, but I think you forgot to sort them according to the morton code.

mleleszi: oh got it, thanks

mleleszi:
@elalish @pca006132 
I probably still need to refactor/clean up a bit but I think It's ready for an initial review.

src/manifold/src/manifold.cpp
```diff
@@ -881,4 +882,73 @@ Manifold Manifold::Hull() const { return Hull(GetMesh().vertPos); }
...
+
+    manifold.GetCsgLeafNode().GetImpl()->GetFaceBoxMorton(faceBox, faceMorton);
+
+    transform(autoPolicy(faceBox.size()), faceBox.begin(), faceBox.end(),
+              faceBox.begin(), [searchLength](const Box& box) {
+                return Box(box.min - glm::vec3(searchLength),
+                           box.max + glm::vec3(searchLength));
+              });
+
+    manifold.GetCsgLeafNode().GetImpl()->SortFaceBoxMorton(faceBox, faceMorton);
```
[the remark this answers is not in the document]
elalish: Since `SortFaceBoxMorton` only seems to be called once in the whole library, I think I'd prefer to just inline it here instead of making a new `Impl` member.

elalish (changes requested):
Thanks, this is looking pretty good! Since the public API looks stable and good you can also go ahead and add the various bindings as well - look at one of our recent PRs that added an API and you'll see which files to change.

src/utilities/include/tri_dist.h
```diff
@@ -0,0 +1,262 @@
...
+
+#pragma once
+
+#include <array>
+#include <glm/glm.hpp>
+
+namespace manifold {
+
+template <typename T>
+const T& clamp(const T& v, const T& lo, const T& hi) {
```
elalish, line 23: Try `glm::saturate` instead.

src/utilities/include/tri_dist.h
```diff
@@ -0,0 +1,262 @@
...
+    u = (t * ADotB - BDotT) / BDotB;
+
+    // if u is on segment (q, b), t and u correspond to closest points,
+    // otherwise, clamp u, recompute and clamp t
+    if (u < 0.0f) {
+      u = 0.0f;
+      if (ADotA != 0.0f)
+        t = clamp(ADotT / ADotA, 0.0, 1.0);
+      else
+        t = 0.0f;
```
elalish, line 80: Let's convert these to ternaries where possible - makes the code cleaner in my opinion.

src/utilities/include/tri_dist.h
```diff
@@ -0,0 +1,262 @@
...
+ * Returns the distance between two line segments.
+ *
+ * @param[out] cq Closest point on line segment pa.
+ * @param[out] cp Closest point on line segment qb.
+ * @param[in]  p  One endpoint of the first line segment.
+ * @param[in]  q  Other endpoint of the first line segment.
+ * @param[in]  p  One endpoint of the second line segment.
+ * @param[in]  q  Other endpoint of the second line segment.
+ */
+void EdgeEdgeDist(glm::vec3& x, glm::vec3& y,  // closest points
```
elalish, line 43: I think the CI compilers are complaining because header implementations should be declared with `inline`.

src/utilities/include/tri_dist.h
```diff
@@ -0,0 +1,262 @@
...
+                  const glm::vec3& p,
+                  const glm::vec3& a,  // seg 1 origin, vector
+                  const glm::vec3& q,
+                  const glm::vec3& b)  // seg 2 origin, vector
+{
+  const glm::vec3 T = q - p;
+  const double ADotA = glm::dot(a, a);
```
elalish, line 50: The inputs are all `float`, so you probably don't need to bother with `double` internally. Our whole library is pretty much float for now, but there's an open issue to migrate.

src/manifold/src/manifold.cpp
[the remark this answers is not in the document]
mleleszi: You may be right, do we need to sort here at all? The Morton codes are based on the centers of the bounding boxes and we are just expanding them here, which should leave their centres untouched. The tests pass without the sorting as well

src/manifold/src/manifold.cpp
```diff
@@ -881,4 +882,73 @@ Manifold Manifold::Hull() const { return Hull(GetMesh().vertPos); }
...
+  auto [faceBox, faceMorton] = getSortedExpandedFaceBoxMorton(*this);
+  auto [faceBoxOther, faceMortonOther] = getSortedExpandedFaceBoxMorton(other);
+
+  Collider collider{faceBox, faceMorton};
+
+  SparseIndices collisions = collider.Collisions(faceBoxOther.cview());
+
+  float minDistanceSquared = searchLength * searchLength;
+
+  auto vertPos = GetMesh().vertPos;
```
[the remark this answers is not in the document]
mleleszi: Yea you're right, there's no reason to get the mesh here

src/utilities/include/tri_dist.h
mleleszi: Ah got it, I wasn't sure why the builds were failing

src/utilities/include/tri_dist.h
mleleszi: Or I could just use `glm::clamp`

src/utilities/include/tri_dist.h
mleleszi: Done!

src/utilities/include/tri_dist.h
```diff
@@ -0,0 +1,262 @@
...
+
+/**
+ * Returns the minimum squared distance between two triangles.
+ *
+ * @param[out] cq Closest point on the first triangle.
+ * @param[out] cp Closest point on the second triangle.
+ * @param[in]  p  First  triangle.
+ * @param[in]  q  Second triangle.
+ */
+float DistanceTriangleTriangleSquared(glm::vec3& cp, glm::vec3& cq,
```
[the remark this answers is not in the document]
mleleszi: I don't we can fully remove them internally as the algorithm relies heavily on them.  We can't use a default argument for a non const reference either. But if we don't intend to use these we could just remove the out params and just use them inside the function only. Maybe something like [this](https://github.com/elalish/manifold/pull/765/commits/ee073a1fe6088e846157f50ad265ce5b012e01ed)? Or we could just leave them as is.

src/manifold/src/manifold.cpp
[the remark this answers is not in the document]
elalish: Please don't take my word for it - have a look at `GetFaceBoxMorton` and make sure. You can test this by rotating a sphere - I have the feeling your newly calculated Morton codes won't match the sorting that was done before rotation.

src/utilities/include/tri_dist.h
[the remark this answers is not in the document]
elalish: Looks good to me.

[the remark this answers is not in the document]
elalish: Go ahead and re-request review when you've addressed the open comments and you're happy with the state.

test/mingap_test.cpp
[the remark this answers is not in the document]
pca006132: I think the typical way of doing this is to have parameterized tests. I can have a look at that next week. A bit busy recently.

src/manifold/src/manifold.cpp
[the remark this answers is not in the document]
pca006132: I think it should be fine if you check against different test cases. I can run some larger tests with openscad later, when everything else is addressed. Please ping me at that point, I am a bit busy recently and may not respond quickly. (grading assignments...)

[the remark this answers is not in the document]
elalish: Good to know it's correct - but the point of the BVH is speed. To ensure it hasn't messed up the tree, I'd take a very high-res sphere and check the rotated against the non-rotated for timing. If they're pretty similar, then it's probably fine.

bindings/python/manifold3d.cpp
```diff
@@ -315,6 +315,14 @@ NB_MODULE(manifold3d, m) {
       .def("calculate_curvature", &Manifold::CalculateCurvature,
            nb::arg("gaussian_idx"), nb::arg("mean_idx"),
            manifold__calculate_curvature__gaussian_idx__mean_idx)
+      .def(
+          "min_gap",
+          [](const Manifold &self, const Manifold &other, float searchLength) {
+            return self.MinGap(other, searchLength);
+          },
```
elalish, line 322: Is this function necessary, or can you replace it with `&Manifold::MinGap`?

test/mingap_test.cpp
```diff
@@ -80,18 +80,32 @@ TEST(MinGap, SphereSphereOutOfBounds) {
 
 TEST(MinGap, ClosestPointOnEdge) {
   auto a = Manifold::Cube();
-  auto b = Manifold::Sphere(1.0f).Translate({3.0f, 0.0f, 0.5f});
+  auto b = Manifold::Cube().Translate({2.0f, 1.0f, 0.5f});
```
elalish, line 83: The closest approach here comes at multiple locations including at a vertex. To test edge-to-edge you'll need to apply rotation as well so that the unique closest point comes in the interior of two edges.

test/mingap_test.cpp
```diff
@@ -80,18 +80,32 @@ TEST(MinGap, SphereSphereOutOfBounds) {
...
+  EXPECT_FLOAT_EQ(distance, 1.0f);
 }
 
 TEST(MinGap, ClosestPointOnTriangleFace) {
-  auto a = Manifold::Cube().Translate({0.0f, -0.25f, 0.0f});
-  auto b = Manifold::Sphere(1.0f).Translate({3.0f, 0.0f, 0.5f});
+  auto a = Manifold::Cube();
+  auto b = Manifold::Cube()
+               .Scale({10.0f, 10.0f, 10.0f})
+               .Translate({2.0f, -5.0f, -1.0f});
```
elalish, line 94: 👍

test/mingap_test.cpp
```diff
@@ -80,18 +80,32 @@ TEST(MinGap, SphereSphereOutOfBounds) {
...
+
+  EXPECT_FLOAT_EQ(distance, 1.0f);
+}
+
+TEST(MinGap, AfterTransformations) {
+  auto a = Manifold::Sphere(1.0f, 100);
+  auto b = Manifold::Sphere(1.0f, 100)
+               .Scale({3.0f, 1.0f, 1.0f})
+               .Rotate(0, 90, 45)
+               .Translate({3.0f, 0.0f, 0.0f});
```
elalish, line 106: Nice test! I don't think all the `.0f` are necessary. I should really clean those up in the rest of our tests.

src/manifold/src/impl.cpp
```diff
@@ -902,4 +903,57 @@ SparseIndices Manifold::Impl::VertexCollisionsZ(
...
+                    return Box(box.min - glm::vec3(searchLength),
+                               box.max + glm::vec3(searchLength));
+                  });
+
+        // impl.SortFaceBoxMorton(faceBox, faceMorton);
+        return std::pair{faceBox, faceMorton};
+      };
+
+  auto [faceBox, faceMorton] = getSortedExpandedFaceBoxMorton(*this);
+  auto [faceBoxOther, faceMortonOther] = getSortedExpandedFaceBoxMorton(other);
```
elalish, line 931: If you expand them both by the search length, I believe you're actually searching twice the search length.

test/mingap_test.cpp
```diff
@@ -80,18 +80,32 @@ TEST(MinGap, SphereSphereOutOfBounds) {
 
 TEST(MinGap, ClosestPointOnEdge) {
   auto a = Manifold::Cube();
-  auto b = Manifold::Sphere(1.0f).Translate({3.0f, 0.0f, 0.5f});
+  auto b = Manifold::Cube().Translate({2.0f, 1.0f, 0.5f});
 
   float distance = a.MinGap(b, 10);
 
-  EXPECT_FLOAT_EQ(distance, 1);
+  EXPECT_FLOAT_EQ(distance, 1.0f);
```
elalish, line 87: Might be good to cut all the search distances down to say, 10% more than the expected distance, to tighten the tests regarding BVH culling.

bindings/python/manifold3d.cpp
```diff
@@ -315,6 +315,14 @@ NB_MODULE(manifold3d, m) {
       .def("calculate_curvature", &Manifold::CalculateCurvature,
            nb::arg("gaussian_idx"), nb::arg("mean_idx"),
            manifold__calculate_curvature__gaussian_idx__mean_idx)
+      .def(
+          "min_gap",
```
elalish, line 319: Can you add this to `all_apis.py` as well, for at least a trivial test?

elalish (changes requested):
Looking great, thanks! Just a few small things.

test/mingap_test.cpp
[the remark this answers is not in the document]
pca006132: http://google.github.io/googletest/advanced.html#value-parameterized-tests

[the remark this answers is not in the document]
elalish: Hmm, that's interesting, but I was more thinking that right now we have `manifold_test.cpp`, `boolean_test.cpp`, `samples_test.cpp` that are each hundreds or more lines, while in the PR we have a couple of new test files just for individual methods. That feels inconsistent in terms of file scope.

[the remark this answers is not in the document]
pca006132: I would prefer a finer granularity, maybe module level, e.g. split `manifold_test.cpp` into multiple files. For `boolean_test.cpp` I think the size is due to how complicated boolean is, and it is hard to split into meaningful categories.

src/manifold/src/manifold.cpp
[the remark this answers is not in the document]
mleleszi: Looks like we might need to sort after all. If I check the min gap between two rotated spheres without sorting, I get a Vec out of range exception.

test/mingap_test.cpp
mleleszi: Ah, good idea. I cut them down and looks like it's not working correctly, still need to figure out why though.

src/manifold/src/impl.cpp
mleleszi: Oh, you're right!

bindings/python/manifold3d.cpp
mleleszi: Done!

pca006132:
Oh and btw, remember the GSoC proposal thing, iirc the deadline is 2nd of April.

pca006132:
but maybe you want to apply for another project considering how fast you are doing!

bindings/python/manifold3d.cpp
mleleszi: Oh right, looks like that works as well

test/mingap_test.cpp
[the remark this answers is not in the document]
elalish: Yes, I tend to agree. As such, I think these tests can be moved into `manifold_test.cpp` for now, pending a broader reorganization.

test/mingap_test.cpp
mleleszi: Fixed, [this](https://manifoldcad.org/#EdgeEdge%3Ccode%3E%2F%2F%20Write%20code%20in%20JavaScript%20or%20TypeScript%20and%20this%20editor%20will%20show%20the%0A%2F%2F%20API%20docs.%20Type%20e.g.%20%22box.%22%20to%20see%20the%20Manifold%20API.%20Type%0A%2F%2F%20%22CrossSection.%22%20or%20%22Manifold.%22%20to%20list%20the%202D%20and%203D%20constructors%2C%0A%2F%2F%20respectively.%20Type%20%22module.%22%20to%20see%20the%20static%20API%20-%20these%20functions%0A%2F%2F%20can%20also%20be%20used%20bare.%20Use%20console.log()%20to%20print%20output%20(lower-right).%0A%2F%2F%20This%20editor%20defines%20Z%20as%20up%20and%20units%20of%20mm.%0Aconst%20%7Bcube%2C%20sphere%7D%20%3D%20Manifold%3B%0Aconst%20s%20%3D%20cube().rotate(%5B0%2C%200%2C%2045%5D)%3B%0Aconst%20s2%20%3D%20cube().rotate(%5B0%2C%2045%2C%200%5D).translate(%5B1%2C%200%2C%200.5%5D)%0A%0A%0Aconst%20result%20%3D%20s2.add(s)%3B) should do it!

test/mingap_test.cpp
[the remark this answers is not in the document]
mleleszi: I'll move the MinGap tests to `manifold_test.cpp` then. What about the triangle distance stuff?

test/mingap_test.cpp
mleleszi: @elalish @pca006132 
Can you check out [these](https://github.com/elalish/manifold/pull/765/files#diff-fb49545a59b0ea198e1005fcf8f5390dffff8f1823090d521b1abcd85a8cbb3eR102-R127) two test cases?

For this more complex test case, if I put in a search distance 10% more than the actual distance, the method returns the search distance, even though it's larger than the actual distance. 

However, If I put in a dummy boolean operation before calling MinGap, it seems to work correctly. I suspect something might be wrong with my sorting? What could these boolean ops be doing that make it work?

elalish: The BVH paper I implemented is here: https://github.com/elalish/manifold/blob/master/docs/ParallelBVH.pdf

src/manifold/src/impl.cpp
```diff
@@ -902,4 +903,55 @@ SparseIndices Manifold::Impl::VertexCollisionsZ(
...
+                         box.max + glm::vec3(searchLength));
+            });
+
+  Vec<Box> faceBoxOther;
+  Vec<uint32_t> faceMortonOther;
+
+  other.GetFaceBoxMorton(faceBoxOther, faceMortonOther);
+  other.SortFaceBoxMorton(faceBoxOther, faceMortonOther);
+
+  Collider collider{faceBox, faceMorton};
```
elalish, line 931: Actually, maybe this is a lot simpler - Try following how the Boolean op works, which also starts with the collider. The collider should already exist - try just using it instead of creating a new one. That seems faster and less likely to be wrong.

test/mingap_test.cpp
mleleszi: Seems like sorting the other manifold's boxes were causing the issue. If I leave them as is and only sort the original one it works.

src/manifold/src/impl.cpp
mleleszi: Oh right, I initially thought I needed to create a new collider because the boxes needed to be expanded/sorted. (now I can see that I could've updated the boxes of the collider as well). But I could just expand the other manifold's boxes, and no extra sorting is needed at all then.

test/manifold_test.cpp
```diff
@@ -833,3 +834,165 @@ TEST(Manifold, EmptyHull) {
...
+  EXPECT_FLOAT_EQ(distance, 2 * sqrt(2) - 2);
+}
+
+TEST(Manifold, MinGapSphereSphere2) {
+  auto a = Manifold::Sphere(1);
+  auto b = Manifold::Sphere(1).Translate({2, 2, 0});
+
+  float distance = a.MinGap(b, 0.85f);
+
+  EXPECT_FLOAT_EQ(distance, 2 * sqrt(2) - 2);
```
elalish, line 889: Is this test identical to the one above?

test/manifold_test.cpp
```diff
@@ -833,3 +834,165 @@ TEST(Manifold, EmptyHull) {
...
+}
+
+TEST(Manifold, MinGapSphereSphereOutOfBounds) {
+  auto a = Manifold::Sphere(1);
+  auto b = Manifold::Sphere(1).Translate({2, 2, 0});
+
+  float distance = a.MinGap(b, 0.8f);
+
+  EXPECT_FLOAT_EQ(distance, 0.8f);
```
elalish, line 898: I think one out of bounds test is enough.

test/manifold_test.cpp
```diff
@@ -833,3 +834,165 @@ TEST(Manifold, EmptyHull) {
...
+}
+
+TEST(Manifold, MinGapClosestPointOnEdge) {
+  auto a = Manifold::Cube().Rotate(0, 0, 45);
+  auto b = Manifold::Cube().Rotate(0, 45, 0).Translate({1, 0, 0.5f});
```
elalish, line 903: I think this is right, but it might be nice to use `center = true` on those cubes to make the motions a little more intuitive.

test/manifold_test.cpp
```diff
@@ -833,3 +834,165 @@ TEST(Manifold, EmptyHull) {
...
+
+TEST(Manifold, MingapAfterTransformations) {
+  auto a = Manifold::Sphere(1, 256).Rotate(30, 30, 30);
+  auto b =
+      Manifold::Sphere(1, 256).Scale({3, 1, 1}).Rotate(0, 90, 45).Translate(
+          {3, 0, 0});
+
+  float distance = a.MinGap(b, 1.1f);
+
+  ASSERT_NEAR(distance, 1, 0.001f);
```
elalish, line 927: Nice test 👍

src/utilities/include/tri_dist.h
```diff
@@ -0,0 +1,225 @@
...
+
+/**
+ * Returns the distance between two line segments.
+ *
+ * @param[out] cq Closest point on line segment pa.
+ * @param[out] cp Closest point on line segment qb.
+ * @param[in]  p  One endpoint of the first line segment.
+ * @param[in]  q  Other endpoint of the first line segment.
+ * @param[in]  p  One endpoint of the second line segment.
+ * @param[in]  q  Other endpoint of the second line segment.
```
elalish, line 35: check param names.

src/utilities/include/tri_dist.h
```diff
@@ -0,0 +1,225 @@
...
+            glm::vec3 cp = pIndex;
+            glm::vec3 cq = pIndex + Tn * Sp[index] / Tnl;
+            return glm::dot(cp - cq, cp - cq);
+          }
+        }
+      }
+    }
+  }
+
+  return shown_disjoint ? mindd : 0.0f;
```
elalish, line 223: This is where I find code coverage to be a useful check - it's a little surprising that execution never reaches this point. I wouldn't consider this a blocker, but it's interesting - sometimes it means we just haven't picked enough variety in our tests, but sometimes it points to vestigial code that actually can't run. It might be worth throwing in a larger, more complicated test object  (using a large search radius) just to see if we can hit all the corners of this algorithm.

elalish (approved):
A few minor things, but this looks good to me. Let's see what @pca006132 thinks.

elalish:
By the way, excellent work on this feature! How would you like to approach your GSOC application? Do you know what you'd like to work on next?

test/manifold_test.cpp
mleleszi: Done!

test/manifold_test.cpp
mleleszi: Yep, removed it

src/utilities/include/tri_dist.h
mleleszi: I'll check it out!

src/manifold/src/impl.cpp
```diff
@@ -902,4 +903,47 @@ SparseIndices Manifold::Impl::VertexCollisionsZ(
...
+            [searchLength](const Box& box) {
+              return Box(box.min - glm::vec3(searchLength),
+                         box.max + glm::vec3(searchLength));
+            });
+
+  SparseIndices collisions = collider_.Collisions(faceBoxOther.cview());
+
+  float minDistanceSquared = searchLength * searchLength;
+
+  for (int i = 0; i < collisions.size(); ++i) {
```
pca006132, line 929: maybe we can do parallelization here?

pca006132:
Overall this looks great!

src/manifold/src/impl.cpp
mleleszi: I'll have a look!

mleleszi:
@elalish @pca006132 
Thank you, I've had a lot of fun working on this! However, I've realized that I might not have enough time over the summer to commit to GSOC. But that doesn't mean I'll stop contributing, I plan to keep at this long term! I'm already looking at what pick up next. I might take a look at https://github.com/elalish/manifold/issues/723 or one of the GSOC tasks that don't get picked for GSOC (maybe https://github.com/elalish/manifold/issues/667). I'll take a look at the other open issues as well. (I'm open to suggestions as well)

elalish:
@mleleszi It's great to hear that you have a realistic understanding of your time commitments. Good expectation setting is one of the most important and least common traits of a software engineer. We would certainly welcome more contributions as your time allows! 

I think #723 would be great - it's nice that we have such a clear example where we know there are orders of magnitude speed increase available. And since you already have some familiarity with our Collider from this PR, that should give you a leg up. Perf PRs are nice because you already have an algorithm that's getting the right answer with lots of tests - all you need to do is not let its answers change.

src/manifold/src/impl.cpp
mleleszi: Great suggestion, it does speed it up a lot! The high res sphere [test case](https://github.com/elalish/manifold/pull/765/files#diff-9befa6164adaa593fe572551c8e23da7a628a42e0053623f1bd22f725dfabc50R902) goes from `11791 ms` to `2690 ms`!

src/utilities/include/tri_dist.h
mleleszi: Added a test case with you stretchy bracelet design and it seems to cover almost everything!

src/manifold/src/impl.cpp
```diff
@@ -902,4 +903,59 @@ SparseIndices Manifold::Impl::VertexCollisionsZ(
   else
     return collider_.Collisions<false, false>(vertsIn);
 }
+
+struct GetDistanceTriangleTriangleSquared {
```
elalish, line 907: These private functors actually need to live up above in the anonymous namespace to avoid potential name collisions. However, we've actually been transitioning the code base away from these functors to lambdas, especially for simple functions like this. The functors were mainly because CUDA support of lambdas was iffy, but now that we've moved off of CUDA, lambdas are generally easier to read.

elalish (changes requested):
Nice job parallelizing! One minor change and let's merge this.

src/manifold/src/impl.cpp
mleleszi: Done!

elalish (approved):
Beautiful, thank you!

PR #771, opened 2024-03-12
CalculateNormals and Smooth

elalish:
This is the next iteration on improving our mesh smoothing, now that we have `RefineToLength`: we now have a `Smooth` member function that is slightly higher-level than our `Manifold::Smooth` constructor. Rather than specifying smoothness at particular edges, now you just specify an angle beyond which edges become sharpened. You can also specify a `minSmoothness` to soften those sharp edges uniformly. 

Additionally, I've added `CalculateNormals`, which works like `CalculateCurvature`, but makes vert normals for smooth reflections and e.g. glTF export. It uses the same sharp angle specification as the above `Smooth` and automatically duplicates prop verts as necessary to create sharp edges.

Both of these functions take special care with flat faces of three or more triangles - these are always considered flat and so have all their vert normals match their face normal. This also makes it easy to have proper curvature applied when e.g. a circular fillet meets a flat face, in which case traditional pseudo-normals give poor results. 

I'm not considering quads to be flat as I plan to give them special handling in our `Smooth`/`Refine` functions soon.


elalish:
I should mention this is technically a breaking change in that the behavior of `Manifold::Smooth` has been updated to match the new `Smooth` (handling flat faces correctly and some other small tweaks), but I'm going to consider this bug fixes since I never liked the previous behavior and I doubt it has much usage yet. More changes will be coming in the next PRs.

bindings/c/include/manifoldc.h
```diff
@@ -188,6 +188,9 @@ ManifoldManifold *manifold_set_properties(
     void (*fun)(float *new_prop, ManifoldVec3 position, const float *old_prop));
 ManifoldManifold *manifold_calculate_curvature(void *mem, ManifoldManifold *m,
                                                int gaussian_idx, int mean_idx);
+ManifoldManifold *manifold_calculate_normals(void *mem, ManifoldManifold *m,
```
elalish, line 191: @geoffder I tried adding `manifold_smooth`, but you already have that for the static version and apparently overloads are not allowed?? I really hate C... Anyway, how would you like to resolve the naming?

bindings/python/manifold3d.cpp
```diff
@@ -315,6 +315,12 @@ NB_MODULE(manifold3d, m) {
       .def("calculate_curvature", &Manifold::CalculateCurvature,
            nb::arg("gaussian_idx"), nb::arg("mean_idx"),
            manifold__calculate_curvature__gaussian_idx__mean_idx)
+      .def("calculate_normals", &Manifold::CalculateNormals,
+           nb::arg("normal_idx"), nb::arg("min_sharp_angle"),
+           manifold__calculate_normals__normal_idx__min_sharp_angle)
+      .def("smooth", &Manifold::Smooth, nb::arg("min_sharp_angle"),
```
elalish, line 321: @wrongbad I appear to have messed up these python bindings somehow or maybe just the docs? Any thoughts on what I did wrong?

pca006132:
Will have a look tmr. Btw, any idea why the test failed?

src/manifold/src/smoothing.cpp
```diff
@@ -501,6 +501,269 @@ class Partition {
...
 
+// sharpenedEdges are referenced to the input Mesh, but the triangles have
+// been sorted in creating the Manifold, so the indices are converted using
+// meshRelation_.
+std::vector<Smoothness> Manifold::Impl::UpdateSharpenedEdges(
+    const std::vector<Smoothness>& sharpenedEdges) const {
+  std::unordered_map<int, int> oldHalfedge2New;
+  for (int tri = 0; tri < NumTri(); ++tri) {
+    int oldTri = meshRelation_.triRef[tri].tri;
+    for (int i : {0, 1, 2}) oldHalfedge2New[3 * oldTri + i] = 3 * tri + i;
```
pca006132, line 512: curious if a vector can be used instead, and maybe we can have parallelism here?

src/manifold/src/smoothing.cpp
```diff
@@ -501,6 +501,269 @@ class Partition {
...
+      const int neighborTri =
+          halfedge_[halfedge_[3 * tri + j].pairedHalfedge].face;
+      const TriRef& jRef = meshRelation_.triRef[neighborTri];
+      if (jRef.SameFace(ref)) {
+        ++faceNeighbors;
+        faceTris[j] = neighborTri;
+      }
+    }
+    if (faceNeighbors > 1) {
+      triIsFlatFace[tri] = true;
```
pca006132, line 540: can parallelize this as well?
if we want to avoid repeated work, we should check `faceTris[tri]` before the start of the `j` loop.

src/manifold/src/smoothing.cpp
```diff
@@ -501,6 +501,269 @@ class Partition {
...
+      const int startEdge = current;
+      const int vert = halfedge_[current].startVert;
+
+      if (vertNumSharp[vert] < 2) {
+        const glm::vec3 normal = vertFlatFace[vert] >= 0
+                                     ? faceNormal_[vertFlatFace[vert]]
+                                     : vertNormal_[vert];
+        int lastProp = -1;
+        do {
+          current = NextHalfedge(halfedge_[current].pairedHalfedge);
```
pca006132, line 653: thinking about it, maybe we should just make a halfedge iterator that loops through all the halfedges connected to a vertex, to make this pattern more readable to beginners.

src/manifold/src/smoothing.cpp
```diff
@@ -501,6 +501,269 @@ class Partition {
...
+  return vertFlatFace;
+}
+
+std::vector<Smoothness> Manifold::Impl::SharpenEdges(
+    float minSharpAngle, float minSmoothness) const {
+  std::vector<Smoothness> sharpenedEdges;
+  for (int e = 0; e < halfedge_.size(); ++e) {
+    if (!halfedge_[e].IsForward()) continue;
+    const int pair = halfedge_[e].pairedHalfedge;
+    const float dihedral = glm::degrees(
```
pca006132, line 576: maybe it will be slightly faster if we convert the angle into radian and only compare using radian, instead of doing conversion of degree to radian in a loop. same for the dihedral calculation below as well.

src/manifold/src/smoothing.cpp
```diff
@@ -501,6 +501,269 @@ class Partition {
...
+        do {
+          current = NextHalfedge(halfedge_[current].pairedHalfedge);
+          const int thisTri = current / 3;
+          const int j = current - 3 * thisTri;
+          const int prop = oldTriProp[thisTri][j];
+          meshRelation_.triProperties[thisTri][j] = prop;
+          if (prop == lastProp) continue;
+          lastProp = prop;
+          for (int i = 0; i < oldNumProp; ++i)
+            meshRelation_.properties[prop * numProp + i] =
```
pca006132, line 661: `std::copy`?

pca006132 (commented):
Just some little comments for now, has to sleep. Also, I think for the python issue you can use the following patch:

```diff
diff --git a/bindings/python/manifold3d.cpp b/bindings/python/manifold3d.cpp
index c85b61c..a4f0bf7 100644
--- a/bindings/python/manifold3d.cpp
+++ b/bindings/python/manifold3d.cpp
@@ -318,9 +318,13 @@ NB_MODULE(manifold3d, m) {
       .def("calculate_normals", &Manifold::CalculateNormals,
            nb::arg("normal_idx"), nb::arg("min_sharp_angle"),
            manifold__calculate_normals__normal_idx__min_sharp_angle)
-      .def("smooth", &Manifold::Smooth, nb::arg("min_sharp_angle"),
-           nb::arg("min_smoothness"),
-           manifold__smooth__min_sharp_angle__min_smoothness)
+      .def(
+          "smooth",
+          [](const Manifold &self, float minSharpAngle, float minSmoothness) {
+            return self.Smooth(minSharpAngle, minSmoothness);
+          },
+          nb::arg("min_sharp_angle") = 60, nb::arg("min_smoothness") = 0,
+          manifold__smooth__min_sharp_angle__min_smoothness)
       .def("refine", &Manifold::Refine, nb::arg("n"), manifold__refine__n)
       .def("refine_to_length", &Manifold::RefineToLength, nb::arg("length"),
            manifold__refine_to_length__length)
```

I think nanobind is just confused about the 3 overloads we have for `Smooth`.

bindings/python/manifold3d.cpp
```diff
@@ -315,6 +315,16 @@ NB_MODULE(manifold3d, m) {
       .def("calculate_curvature", &Manifold::CalculateCurvature,
            nb::arg("gaussian_idx"), nb::arg("mean_idx"),
            manifold__calculate_curvature__gaussian_idx__mean_idx)
+      .def("calculate_normals", &Manifold::CalculateNormals,
+           nb::arg("normal_idx"), nb::arg("min_sharp_angle"),
+           manifold__calculate_normals__normal_idx__min_sharp_angle)
+      .def(
+          "smooth",
+          [](const Manifold &self, float minSharpAngle, float minSmoothness) {
+            return self.Smooth(minSharpAngle, minSmoothness);
```
elalish, line 324: @pca006132 Alas, no luck here. I wish the error wasn't such a pile of spew. Also, even when I build with MANIFOLD_PYBIND=ON I don't get the error - is something else required to make it build?

wrongbad:
When I try to build your branch it passes for me. Maybe you should push your branch with this in `pyproject.toml` and see what the CI build says then
```
cmake.args = ["-DCMAKE_BUILD_TYPE=Debug", . . .
```

bindings/python/manifold3d.cpp
pca006132: weird... will debug this

pca006132: It seems that nanobind is unhappy with `smooth` being both a static and instance method for Manifold`.

```
Critical nanobind error: nb::detail::nb_func_new("smooth"): mismatched static/instance method flags in function overloads!
```

src/manifold/src/smoothing.cpp
elalish: I tried that first, but it's made difficult because there are potentially more triangles in the old indices than the new and I'm not sure how many.

src/manifold/src/smoothing.cpp
elalish: I think for correctness we can't do the early-out.

bindings/python/manifold3d.cpp
elalish: Okay, I'm going with `SmoothOut` - slightly more verb sounding.

src/manifold/src/smoothing.cpp
pca006132: It should be possible to make an iterator out of this. I will open an issue about this.

Issue #775, opened 2024-03-20
vertex halfedge iterator


pca006132:
https://github.com/elalish/manifold/pull/771#discussion_r1523743688

Implement a halfedge iterator that loops through all the halfedges connected to a vertex.

PR #776, opened 2024-03-25
Smooth By Normals

elalish:
Follow-on to #771, that allows smoothing based on input normal properties. Has the convenient side effect of making a solid test since we now have two different paths to create the same type of smoothing, in the case of sharp edges. 

I also added `ForVert`, which I think is a good start toward addressing #775.

### DIFF
--- a/src/manifold/src/impl.h
+++ b/src/manifold/src/impl.h
@@ -137,6 +137,7 @@ struct Manifold::Impl {
   void SplitPinchedVerts();
 
   // smoothing.cu
+  glm::vec3 GetNormal(int halfedge, int normalIdx) const;
   std::vector<Smoothness> UpdateSharpenedEdges(
       const std::vector<Smoothness>&) const;
   Vec<bool> FlatFaces() const;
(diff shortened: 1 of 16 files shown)


hrgdavor:
mm, this look interesting, could you add a screenshot or two ?

elalish:
This is still WIP, even once I merge this. I still need to add quad support, as right now it's still pretty ugly.

pca006132:
Will take a bit of time for me to review this, I have a hundred assignments to grade, probably will review this weekend.

src/manifold/src/impl.h
```diff
@@ -66,6 +66,29 @@ struct Manifold::Impl {
...
+    do {
+      func(current);
+      current = NextHalfedge(halfedge_[current].pairedHalfedge);
+    } while (current != halfedge);
+  }
+
+  template <typename T>
+  void ForVert(int halfedge, std::function<T(int halfedge)> transform,
+               std::function<void(int halfedge, const T& here, const T& next)>
+                   binaryOp) {
```
elalish, line 80: @pca006132 Don't worry about a serious review - I was mostly curious if you thought this was a reasonable way to handle #775.

PR #781, opened 2024-03-28
Added : VHACD & karimnaaji qhull implementation

Kushal-Shah-03:
For [issue 89](https://github.com/opencax/GSoC/issues/89) Optimize Convex Hull Algorithm, I have added the Karimanaaji implementation : https://github.com/karimnaaji/3d-quickhull/tree/master (and the VHACD implementation from zalo's reference is also added, but I just added it as an attempt to get used to the codebase), I have written comments, as to where the algorithm is having issues with larger hulls, there are segfaults that occur in the functions of the header files, I have tried to identify the faulty function and have written that in comments as well.


elalish:
Thanks; what are your next steps? Do you need anything from us right now? And have you managed to sign the CLA (the individual one should be instant)?

Kushal-Shah-03:
In this PR I have not made any changes to any other file, except adding the two new implementations, as well as adding some testcases, I am not sure as to why some testcases are failing, could you suggest what I can try to solve those issues in my PR?

Also, for my further plan I am looking into adding some other implementations as suggested by zalo like https://github.com/tomilov/quickhull https://github.com/andreacasalino/Fast-Quick-hull , I am almost done with another implementation (I will prioritize this first). I plan of adding implementations of Graham's Scan and Chan's algorithm, as they seem like suitable alternatives. I also plan to optimize the piece of code in quickhull using parallelization.

And yes I signed the individual CLA now.

elalish:
It looks like you've already made progress - most of it is building with tests passing. It looks like MSVC may be a bit pickier about implicit type conversions. And the fact the WASM is running out of memory is not a great sign - this probably implies the memory usage of that algorithm doesn't scale well. 

It looks like you're investigating several different approaches to Convex Hull, but at some point you'll want to pick one to focus on optimizing. What is your sense so far? How do you plan to choose between them?

Kushal-Shah-03:
For the WASM, I realised, I wasn't freeing up memory at one place, so I will try to see if that solves the issue, apart from that, I'm not sure as to why build_windows is failing the error message isn't clear to me, since I haven't made any changes to minimize_testcase.cpp, maybe it's an issue with the memory as well, I will fix that and see.

So, yeah right now as suggested, I am exploring different approaches and I think some part of this project is about exploring approaches so that we can narrow down our options to say 2 or 3, and then I think I'll spend some time understanding the theory of the algorithm in depth as well as the code, to account for edge cases, and once we have the options, we can study them properly to identify bottlenecks to optimize the algorithm. About choosing, the algorithms I was reading about some common algorithms, and I found that Graham's scan and Chan's algorithm were fairly fast and handled some edge cases that quickhull couldn't. I feel while identifying which 2 or 3 to narrow it down to, we can consider some complex testcases to identify time constrains of algorithms as well as look at certain cases where the quickhull algorithm suffers, to see if we can identify an algorithm which performs better in those, and certainly we should also consider the cases where the new alogirthm we plan to use suffers, to understad the tradeoff between the algorithms better. I would appreciate your input on my thoughts as well, so that I can better understand your expectations.

elalish:
That sounds good - more speed with less maintenance is ideal. As far as correctness, I'd love to see some examples of incorrect solutions. Keep in mind that our library does inexact geometry, so I'm less picky about how nearly coplanar points are handled, so long as the result is sane. You'll notice we tend to check our test cases on volume and surface area - I think this is a good approach for Hull as well. If all the points are coplanar, I'm equally okay with returning a flat manifold or an empty manifold - same volume after all.

Kushal-Shah-03:
Yeah, regarding when I said incorrect solution, I was mainly referring to coplanar points, but I understand your arguments, as to why that isn't an issue, the other case I was considering was deleted point deletions cases I can't quite recall where I read about it, but since we are mainly considering volume and surface area that shouldn't be a major problem as well. I will still try to find cases, to see if we can identify some edge case. 

Also regarding this PR, I have now passed all cases but build_windows and it's TBB, and I am working on identifying the issues but it's taking some time. I will update you once it's done so you can review it, so that I can use that feedback to improve

Kushal-Shah-03:
@elalish , so I attempted to fix the issues with build_windows, but haven't succeeded yet I have switched to windows to fix the issue, I hope to get it done, but with the proposal deadline coming closer, I wanted to prioritize writing a good proposal, to convey my thoughts better. I was wondering if you could review my code (I know it doesn't pass all checks yet, but as far as I understand it's merely an implicit typecasting issue, and I would appreciate it if you could go through the code) and give me some feedback, so that I can consider that in my proposal. I also went through the guidelines on the opencax gsoc website, and since they were quite general ( for example need for IRC username even though it's not mentioned as mode of communication used by Manifold), I was wondering if there are separate guidelines that I need to consider for this project (I will follow those on opencax as well).

PS: Sorry have been busy with institute exams lately, will certainly get back to fixing this issue soon.

pca006132:
We are first year in GSoC, so we don't really have any additional guideline on that. I think as long as your proposal can demonstrate your interest as long as show some related experience in courses/other work, it should be fine. And you can replace the irc username with github handle, it should be fine.

Kushal-Shah-03:
I have a submitted a draft of my GSoC proposal on the website and would appreciate it if you could go through it once and give me feedback on it.

pca006132:
which website did you submit to?

Kushal-Shah-03:
GSoC website - https://summerofcode.withgoogle.com/proposals , under BRL-CAD organization

pca006132:
I think the simplest way to let us read it is to email the mentors directly, so we don't have to ask BRL-CAD admin to let us read the proposal. I think you can just me, @elalish and @zalo directly.

Kushal-Shah-03:
Alright, sorry I was not aware of that, I will send an email now

Kushal-Shah-03:
I cannot find @elalish's email address, could you share that?

pca006132:
I can forward the email to him, not sure if he wants his email to be in the public.

Kushal-Shah-03:
I got the build_windows check to pass now, turns out the issue was just using a static array
`qh_vertex_t input_pts[numVert];`
When I allocated memory in the heap
`qh_vertex_t* input_pts = new qh_vertex_t[numVert];`
The errors get fixed.

I am not sure, as to why that was an issue though, is there some sort of explanation?

pca006132:
You never have variable length static array, it is allocated in the stack. C++ does not support variable length array, and you rarely want it anyway because it is easy to have stack overflow with that.

Kushal-Shah-03:
I am aware about the memory, being allocated in the stack. I had initially done it, because the example code for the algorithm, had declared a static array, but I understand why that was giving error when the value was large. But my question, was it was being shown as a typecasting error as @elalish had suggested, shouldn't it have been a memory limit exceeded error, or did we misinterpret the error message.

pca006132:
Can you show me that error message? the typecasting error.

Kushal-Shah-03:
Screenshot from 2024-04-06 19-50-30
Screenshot from 2024-04-06 19-50-35
Screenshot from 2024-04-06 19-50-38
Sorry for the late reply.

pca006132:
I think those are warnings. This is the error: https://github.com/elalish/manifold/actions/runs/8544112595/job/23409427303#step:5:242

```
D:\a\manifold\manifold\src\manifold\src\manifold.cpp(1000,25): error C2131: expression did not evaluate to a constant
```

Kushal-Shah-03:
Oh, I missed that, I just assumed since we treat warnings as errors too, they were errors, thanks for the help, I'll look out for this in the future. Also, I am done with another implementation, I should add it to this PR only right?

pca006132:
Yeah it is hard to read these warnings/errors, we don't treat warnings as errors for msvc because we don't use that :).

For the other implementation, sure please add it here so we can discuss in this PR. I think the most important thing though is not to have good code or fix build errors on other platforms, but rather to try to evaluate the hull quality and performance of the implementation. Try to filter out bad choices first so you can spend less effort.

Kushal-Shah-03:
Alright, actually during integrating the implementation of the algorithm, I do try to look at the performance of the implementation, but I will also start filtering out the choices before implementation, because implementation does end up taking a decent amount of time. Also by evaluating hull quality, what do you expect? (I have currently looked at the outputs produced in blender, to verify the implementation, but I was wondering what all things, should I look at?)

Kushal-Shah-03:
Alright, I will compare them with CGAL solutions, and will look into the benchmarks as well.

starseeker:
A note about implementations - you'll want to make sure any existing implementation you look at using has a compatible license.  For example, Fast-Quick-hull initially didn't have a license, and when I inquired they opted (at least as an initial selection) for GPL, which (at least for BRL-CAD's use case) would be too restrictive.  If an implementation looks particularly good and doesn't have an appropriate license you might inquire if they'd be willing to change it.

starseeker:
>  when I inquired they opted (at least as an initial selection) for GPL, which (at least for BRL-CAD's use case) would be too restrictive.

Update:  the author is willing to change the license :-) 
 https://github.com/andreacasalino/Fast-Quick-hull/issues/5

Kushal-Shah-03:
Thanks for bringing that to my notice, I will keep this in mind while looking for implementations. What exactly do you mean by a compatible license though, what points do I have to look for, sorry if it's trivial, but I am not aware about what license to look for?

Kushal-Shah-03:
I have added a third implementation https://github.com/leomccormack/convhull_3d. Now that I feel comfortable with adding implementations, I plan to shift my focus to testing these implementations extensively as well. I have also started to look into how we can run benchmarks, but as my current semester is near completion, I won't get a lot of time to start testing right now, I will try to test with whatever time I get, but will only be properly able to get back to testing after 15-20 days.

pca006132:
I think you can try to search for compatible license with Apache. For example, GPL or LGPL is probably too restrictive here.

zalo:
It seems like we already have basic timing results from [the unit tests](https://github.com/elalish/manifold/actions/runs/8637644799/job/23680306289?pr=781#step:6:375):
```
[       OK ] Manifold.TictacHull (Current) (2159 ms)
[       OK ] Manifold.TictacHull2 (VHACD) (80 ms)
[       OK ] Manifold.TictacHull3 (QUICKHULL) (38 ms)
[       OK ] Manifold.TictacHull4 (CONVHULL) (42 ms)

[       OK ] Manifold.HollowHull (Current) (411 ms)
[       OK ] Manifold.HollowHull2 (VHACD) (17580 ms)
??? Missing Test for HollowHull3 ???
[       OK ] Manifold.HollowHull4 (CONVHULL) (594 ms)
```

Seems like Hull3 is missing a test for HollowHull.

starseeker:
> I think you can try to search for compatible license with Apache. For example, GPL or LGPL is probably too restrictive here.

For BRL-CAD GPL definitely won't work - LGPL can if handled correctly but I doubt an LGPL licensed lib is something Manifold would want to depend on?  @elalish should weigh in here, but I'd think Apache 2, MIT and BSD are probably safe bets...

Of the possibilities listed in https://github.com/opencax/GSoC/issues/89 ... taking a quick look it appears most of them are under liberal licenses - https://github.com/melax/sandbox/blob/master/include/hull.h doesn't have a license called out in the header file, but the top level sandbox repo has an MIT license.  (If we were to wind up incorporating that one the license should be added as a comment to the top of the hull.h file, but based on the comments the author put in there I'd be a little surprised if that one was the winner anyway...)

starseeker:
@pca006132 When it comes to the robustness question... what are your thoughts on what might be a good way to check for issues?  I'm not well versed in fuzz testing, but perhaps one way to go at it would be to take a tetrahedron mesh of four triangles as a starting point and see if there are any near-degenerate or degenerate cases where the hull algorithms fail?  Then one of the triangles could be split into three or an edge could be split in the middle to allow for minimalist concave tests to be similarly fuzzed for edge cases?

elalish:
Agreed that sticking to Apache 2, MIT, BSD is generally the way to go. I find it pretty rare someone uses something else that we want to touch. Basically if the license is more than 3 paragraphs, be wary. 

As for testing, one easy way to check lots of degenerates is to take something with piles of exact degenerates, like our Menger Sponge, and then give it some rotations. The floating point error on irrational numbers is a great way to generate tons of near degenerates that's a pretty realistic use case.

Kushal-Shah-03:
> It seems like we already have basic timing results from [the unit tests](https://github.com/elalish/manifold/actions/runs/8637644799/job/23680306289?pr=781#step:6:375):
> 
> ```
> [       OK ] Manifold.TictacHull  (Current)   (2159 ms)
> [       OK ] Manifold.TictacHull2 (VHACD)     (80 ms)
> [       OK ] Manifold.TictacHull3 (QUICKHULL) (38 ms)
> [       OK ] Manifold.TictacHull4 (CONVHULL)  (42 ms)
> 
> [       OK ] Manifold.HollowHull  (Current)   (411 ms)
> [       OK ] Manifold.HollowHull2 (VHACD)     (17580 ms)
> ??? Missing Test for HollowHull3 ???
> [       OK ] Manifold.HollowHull4 (CONVHULL)  (594 ms)
> ```
> 
> Seems like Hull3 is missing a test for HollowHull.

Regarding this I was actually having issues with larger Hulls with implementation 4, I was going through the code of that implementation I found that they had set a limit on the maximum number of Faces, which was giving empty hulls for large hull tests, but even after removing that, the time taken for the larger Hull's is a lot and thus I don't think it's a viable implementation, I tried to find some part of the code that was not implemented optimally leading to these high times, but haven't been able to find it yet. And for the Hull 3, I was having a similar issue iirc, I will look into it as well, to see if I can find out an issue. And for all future implementations, I plan on first going through the flow of the implementations, before adding the implementations, to avoid wasting time on implementations, that end up taking such significantly larger times even when implemented correctly.
PS: Sorry for the inactivity, had been busy with project submission lately, will try to get back to the task of testing now, and I will look into the suggestions regarding testing mentioned on this thread as well. For the testing for degenerates, how would I go about testing with the Menger Sponge, could you guide me on that a bit?

Kushal-Shah-03:
Sorry I clicked on Close with Comment by mistake

elalish:
>For the testing for degenerates, how would I go about testing with the Menger Sponge, could you guide me on that a bit?

The hull of our Menger Sponge sample is just a cube, but it has lots of exactly coplanar points on the surface, so it might make for interesting input. A large, high-res sphere is also interesting - technically every input point should be in the hull, but as it gets locally flatter, some may be removed. Probably fine as long as the resulting shape is within tolerance.

Kushal-Shah-03:
So I had a couple of queries, 
https://github.com/elalish/manifold/blob/master/test/samples_test.cpp?plain=1#L252-L291
Does this test utilize the quickhull algorithm in it's underlying implementation?

In case it does not I have currently just converted the output to a mesh and used the pts to form a convex hull using the algorithm for testing, but wanted to verify it once.

So currently I am using this piece of code to get the output
image
I am getting the desired cube in blender, I will also add tests for the volume and surface area as discussed, but wanted to verify if I'm going in the right direction before proceeding

And I also wanted to ask in the cases where the current quick hull implementation does drop points, are we showing warning/error messages on the terminal while running the command?

pca006132:
1. Yeah you are in the correct direction.
2. Yeah there are messages in stderr when the current algorithm encounters a failure.

Kushal-Shah-03:
I've pushed the code for the menger sponge rotation test as well as the high quality sphere test.
I've implemented it in a way, that would make testing all implementations fairly simple, since we would just need to change the name of the Hull functions to test the other hulls.

Just to clarify, when we meant a high quality sphere, a sphere with a very large number of circular segments was expected right?

Also, while testing for a large number of circular segments, my local machine couldn't handle more than 6000 circular segments, would that be enough, and similarly for checking lots of values of rotations in menger sponge, the time taken for each iteration is not very feasible, so I was wondering, if there's a way I can run it on some external server, which is faster than my machine, so as to test a higher quality sphere and test several rotations? (Like one sometimes uses google collab to perform ML tasks)

pca006132:
I have a spare machine that I rarely use, it is not a high-end machine, just i5-13600 with 32GB of RAM, but I can add more RAM if needed. The problem though is that the department policy does not allow others to ssh into it (and it is behind a firewall, I have no proxy server to bridge that), but you can send me the script to run.

And can you complete the CGAL perf test? I think it is precise enough when we consider 32 bit floating points.

Kushal-Shah-03:
Yeah, I'm able to run the CGAL perf test (the number of circular segments was 4096 for that test), I am able to run till 6000. Regardless, your spare machine has higher specifications than mine, so I can maybe increase the count to 10000, and then ask you to run it once. Also, for the menger sponge, I'll modify the code for the test, and ask you to run it, considering it can run faster.

Also, I was thinking, since we can execute bash commands in collab, like cloning a repository, maybe I can get our code to run as well. I will try to see if I can get that to work. I will get back to you regarding this.

Kushal-Shah-03:
I was able to make the script run on colab using TPU (It crashed on CPU and T4 GPU due to memory limitations I think), Tthere were no drop points message in the stderror. So, we can assume that the algorithm doesn't drop points in this case.

We can reduce it to 5000, for testing, since it took around 10 mins and 50GB ram for the 10,000 case, and since around 5000 is sufficient for floating point precision, as you suggested, we can use it while comparing algorithms to save time, over the 10,000.

I will continue using colab for compute intensive testing for now. Since, it does seem to function without a lot of issues.

I also wanted to share the output of Surface area and volume of the spheres test

This is the one with 6,000 circular segs
6000Spheres

This is the output for the case with 10,000 circular segs
10000Spheres


All the outputs for 6,000 segs are quite accurate.

The outputs for 10,000 for SA and Vol for CGAL (for both using the hull as well as normal implementation) are very accurate, but those from our implementation are not that accurate, in fact the area computed directly from the manifold is off by 0.1. And I think we should look into why when we increased the quality of the sphere, did the answer become less accurate.

elalish:
With the spheres specifically, it may be more helpful to look at vertex count, so you can tell how many points are being dropped. The surface area and volume calculations may be getting some rounding error since it's adding lots of small values.

pca006132:
Note that CGAL probably uses precise arithmetic, while we only use floating points. When things differ by orders of magnitude, floating point precision issue is more apparent.

pca006132:
btw thinking about it, is this really using the TPU or just the TPU configuration has more RAM? I know nothing about TPUs, thought they are just accelerators which requires special kernels to work, similar to GPU programming.

Kushal-Shah-03:
I'm not sure, the TPU configuration has significantly more RAM, and it did end up using around 50GB of RAM, which wasn't there in either the GPU or CPU configuration. So it probably doesn't use the TPU, just the extra RAM.

Kushal-Shah-03:
Note : Hull2 is VHACD implementation, and Hull is our current implementation

1750623.stl
```
VolHull_hull2 : 0.9883278400351572
AreaHull_hull2 : 0.9136602346123665

VolHull_CGAL : 0.9883278400351572
AreaHull_CGAL : 0.9136602346123665
```

Unnormalized values:
1750623.stl
```
VolHull : 113775
AreaHull : 14909.7
VolHull_hull2 : 112447
AreaHull_hull2 : 13622.4
VolHull_CGAL : 112447
AreaHull_CGAL : 13622.4
```

39202.stl
```

AreaHull_hull2 : 0.936631098033676
AreaHull_CGAL : 0.936631098033676
```

Unnormalized values:
39202.stl

```
VolHull : 274765
AreaHull : 29023.7
VolHull_hull2 : 273110
AreaHull_hull2 : 27184.5
VolHull_CGAL : 273110
AreaHull_CGAL : 27184.5
```
These are two cases where the outputs for volume and area for our current implementation varies from VHACD and CGAL outputs (which give the same outputs), Hull3 and Hull4 give errors on these hulls (although that is a fault of the implementations)

I was able to identify these cases when I was looking for errors in volume and area over the range of 1%, there are more occurrences when checking for 0.1% but I think going through these cases could be a good starting point for trying to identify why there's a difference.

Kushal-Shah-03:
The 5 cases of segfaults that were happening in all algorithms, were due to corrupted files or files with incorrect format, I verified that the error was in the ImportMesh function call. So, I think we can safely ignore them in further analysis.

Also, for the cases where our algorithm showed the warning of dropping horizon edges, I checked the outputs of those cases and they gave the same outputs as the other algorithms, so the outputs are accurate even when warnings are present.

For reference,
'_hull2' - refers to VHACD implementation
'_hull3' -  refers to karimnaaji qhull implementation (https://github.com/karimnaaji/3d-quickhull)
'_hull4' - refers to leomccormack convex hull (https://github.com/leomccormack/convhull_3d)
'_CGAL' - refers to CGAL implementation for convex hull algorithm
and the reference that we are comparing with is our current quickhull implementation


For normalized starter statistics for the algorithms
```
                       mean     median      mode            max       min  Success_Count
VolHull            1.000000   1.000000  1.000000       1.000000  1.000000         9995.0
VolHull_hull2      0.999998   1.000000  1.000000       1.000135  0.988328         9993.0
VolHull_hull3      0.799843   1.000138  0.000000       1.635197  0.000000         7088.0
VolHull_hull4      0.303321   0.000000  0.000000       1.000067  0.000000         9854.0
VolHull_CGAL       0.999999   1.000000  1.000000       1.000941  0.988328         9995.0
AreaHull           1.000000   1.000000  1.000000       1.000000  1.000000         9995.0
AreaHull_hull2     0.999984   1.000000  1.000000       1.000039  0.913660         9993.0
AreaHull_hull3     0.796163   1.000016  1.000000       1.246588  0.000000         7088.0
AreaHull_hull4     0.303377   0.000000  0.000000       1.000006  0.000000         9854.0
AreaHull_CGAL      0.999984   1.000000  1.000000       1.000072  0.913660         9995.0
HullTri            1.000000   1.000000  1.000000       1.000000  1.000000         9995.0
HullTri_hull2      0.998383   1.000000  1.000000       1.605794  0.714286         9993.0
HullTri_hull3      1.144347   1.017440  1.000000      19.833333  0.000000         7088.0
HullTri_hull4      0.305715   0.000000  0.000000       2.000000  0.000000         9854.0
HullTri_CGAL       1.071703   1.000750  1.000000      11.333333  0.864198         9995.0
Time               1.000000   1.000000  1.000000       1.000000  1.000000         9995.0
Time_hull2         5.576693   2.588005  0.032369     372.584085  0.032369         9993.0
Time_hull3        56.266342  21.619789  0.050973   12634.812889  0.050973         7088.0
Time_hull4      1262.230941  17.375242  0.073636  147236.434200  0.073636         9854.0
Time_CGAL          2.699975   2.181629  0.010641     119.581236  0.010641         9995.0
```

Mainly for the time metric, we discussed that it's helpful to omit the cases with very small values of time, to more accurately identify the difference in time by various algorithms, so I've filtered out times less than 1ms (0.001 seconds)
```
                      mean     median      mode           max       min  Success_Count
VolHull           1.000000   1.000000  1.000000      1.000000  1.000000         3516.0
VolHull_hull2     0.999995   1.000000  1.000000      1.000135  0.988328         3514.0
VolHull_hull3     0.559386   1.000047  0.000000      1.317467  0.000000          995.0
VolHull_hull4     0.133583   0.000000  0.000000      1.000003  0.000000         3466.0
VolHull_CGAL      0.999996   1.000000  1.000000      1.000941  0.988328         3516.0
AreaHull          1.000000   1.000000  1.000000      1.000000  1.000000         3516.0
AreaHull_hull2    0.999956   1.000000  1.000000      1.000039  0.913660         3514.0
AreaHull_hull3    0.556907   1.000009  0.000000      1.011987  0.000000          995.0
AreaHull_hull4    0.133583   0.000000  0.000000      1.000005  0.000000         3466.0
AreaHull_CGAL     0.999956   1.000000  1.000000      1.000072  0.913660         3516.0
HullTri           1.000000   1.000000  1.000000      1.000000  1.000000         3516.0
HullTri_hull2     0.996553   1.000000  1.000000      1.605794  0.804674         3514.0
HullTri_hull3     0.568876   0.900530  0.000000      2.703125  0.000000          995.0
HullTri_hull4     0.134017   0.000000  0.000000      1.089385  0.000000         3466.0
HullTri_CGAL      1.045920   1.010335  1.000000      5.833333  0.932990         3516.0
Time              1.000000   1.000000  1.000000      1.000000  1.000000         3516.0
Time_hull2       11.741470   4.908870  0.684735    372.584085  0.684735         3514.0
Time_hull3       96.975058  42.616741  3.628361  12634.812889  3.628361          995.0
Time_hull4      301.975446  37.398160  0.073636  17927.832958  0.073636         3466.0
Time_CGAL         2.879137   2.285675  0.556763     65.602728  0.556763         3516.0
```

From these results it's clear that Hull3 and Hull4 are not suitable alternatives, despite Hull4 showing a lot of "success" counts, it actually returns an empty hull if it's not able to compute the output, and thus it's volume and area result in many zeroes, ultimately affecting the mean volume and hull metric. Despite, several attempts to fix this, I've not been able to, so I think it's safe to say we can drop Hull4 from our list of options. 
Similarly for Hull3, it gave segfaults on many testcases, so I think we should drop Hull3 as well.

Hull2 is certainly slower but is quite accurate, so is still a suitable alternative, especially considering the discrepancy in volume and area in the cases in the prior comment https://github.com/elalish/manifold/pull/781#issuecomment-2178105438

With a basic idea of how to compare the algorithms, I now plan to integrate the other implementations we had discussed and run tests, so that we have a larger set of algorithms to compare with.

Kushal-Shah-03:
> Note : Hull2 is VHACD implementation, and Hull is our current implementation
> 
> 1750623.stl
> 
> ```
> VolHull_hull2 : 0.9883278400351572
> AreaHull_hull2 : 0.9136602346123665
> 
> VolHull_CGAL : 0.9883278400351572
> AreaHull_CGAL : 0.9136602346123665
> ```
> 
> Unnormalized values: 1750623.stl
> 
> ```
> VolHull : 113775
> AreaHull : 14909.7
> VolHull_hull2 : 112447
> AreaHull_hull2 : 13622.4
> VolHull_CGAL : 112447
> AreaHull_CGAL : 13622.4
> ```
> 
> 39202.stl
> 
> ```
> 
> AreaHull_hull2 : 0.936631098033676
> AreaHull_CGAL : 0.936631098033676
> ```
> 
> Unnormalized values: 39202.stl
> 
> ```
> VolHull : 274765
> AreaHull : 29023.7
> VolHull_hull2 : 273110
> AreaHull_hull2 : 27184.5
> VolHull_CGAL : 273110
> AreaHull_CGAL : 27184.5
> ```
> 
> These are two cases where the outputs for volume and area for our current implementation varies from VHACD and CGAL outputs (which give the same outputs), Hull3 and Hull4 give errors on these hulls (although that is a fault of the implementations)
> 
> I was able to identify these cases when I was looking for errors in volume and area over the range of 1%, there are more occurrences when checking for 0.1% but I think going through these cases could be a good starting point for trying to identify why there's a difference.

Keeping this in mind, we discussed how to simplify the problem to identify what could be the issue, As suggested by @pca006132 I plan to experiment with a sort of binary search where I remove half points randomly and check the volume and area to identify the set of points that could have caused the issue and then we can evaluate it better.

I also plan to subtract the two convex hulls produced and then look at the output to see if there's a clear visual indicator of what could be the issue

@elalish, these are a couple of ideas we had to better identify the issue, if there are some other ideas that you think we can look into for the same, I will try to test them as well.

PS : @pca006132 were u able to find something about the epsilon value you were mentioning earlier

pca006132:
No, I haven't yet have time to do that.

For randomly removing points to reduce the test case, it was about the warning from quickhull that says failure to resolve horizon edges.

elalish:
Excellent work! So it looks like our current hull is the fastest, but may have some significant accuracy problems. I very much like the idea of taking the difference of the two hulls - save it to a GLB and link it here. I think that will tell us a lot quickly.

Kushal-Shah-03:
> Excellent work! So it looks like our current hull is the fastest, but may have some significant accuracy problems. I very much like the idea of taking the difference of the two hulls - save it to a GLB and link it here. I think that will tell us a lot quickly.

Sorry for the delay, I've added the glb files to [https://github.com/Kushal-Shah-03/manifold/tree/master/extras/GLB_Files](url). [ Note : hull1 refers to our current implementation, hull2 refers to VHACD implementation and hull1_hull2 is the file obtained from subtracting hull2 from hull1 and similarly for hull2_hull1]

elalish:
Great, do you see where the problem is occurring? Would be nice to put some renders up here perhaps with a colored overlay so we can see where the difference is.

Kushal-Shah-03:
As discussed, after testing the other algorithms (Hull 5 and Hull 6 didn't give satisfactory results in our tests and for Hull 7 (https://github.com/melax/sandbox/blob/master/include/hull.h) the time complexity wasn't feasible in theory, and Hull 8 (https://github.com/andoresu47/ConvexHull3D) had constraints like no 3 points are colinear, which are not feasible), we have decided to continue with our current implementation and VHACD implementation. 

As for the cases that gave different volumes and surface areas for our implementation and VHACD implementation, here are the important observations:

39202.stl:

Screenshot from 2024-06-27 19-16-47
Screenshot from 2024-06-27 20-56-05
Screenshot from 2024-06-27 21-13-10
Screenshot from 2024-06-28 13-56-12
Screenshot from 2024-06-28 13-56-38

From these images it's clear to see that the excess volume and area is caused due to the piece contained inside the object, this extra piece is produced by our current implementation, and is clearly something that should not be present. The VHACD implementation does not have this issue.

Similary for 1750623.stl : 
 
Screenshot from 2024-06-28 13-58-16
Screenshot from 2024-06-28 13-58-25
Screenshot from 2024-06-28 13-58-46
Screenshot from 2024-06-28 14-00-23

The extra piece is contained inside the object, and it is also produced by our implementation but not the VHACD implementation.

I hope these renders highlight the problem clearly.

I plan to slowly reduce the number of points that persist with this issue, to investigate what caused this issue,


On a side note, for the "Failed to solve Horizon edge" warning message in our current algorithm. I've added two .bin files that contain the points that caused this issue, and their glb files. I'm attaching the photos of the hull produced in blender as well. I don't think the renders help us a lot in identifying what could have caused the issue, so I plan to use gbd to run the code to better identify what causes the warning message, and if it actually affects the output in anyway (by comparing the outputs of VHACD implementation and CGAL).

points_1411700.bin:
Screenshot from 2024-06-29 16-12-21
Screenshot from 2024-06-29 16-12-53
image

points_1411702.bin:

Screenshot from 2024-06-29 16-14-52
Screenshot from 2024-06-29 16-14-59
Screenshot from 2024-06-29 16-15-05

Also, I was thinking I should post the same example for the volume discrepancy in an issue of the repo of our current implementation, to get some ideas on what could cause the issue, while I work on reducing the number of points, does that sound alright?

pca006132:
Nice figures. Yeah, you should open an issue upstream and ping us.

For debugging, you can probably try to add some print statements to print the edge/point IDs that the algorithm is processing, as well as a list of neighbors (or something similar in their algorithm). This can allow you to highlight specific edges/points in the mesh for visualization, and that should help with debugging. Think of the sorting algorithm visualization, you probably want something similar but with a mesh. I am not sure if directly using GDB to debug this kind of algorithmic issue can help you much.

elalish:
Yeah, so the bug is that our quickHull implementation is generating a very non-convex shape for this convex hull operation. Good find! Does it have any interior verts? It kind of looks like it only generated incorrect triangles. Any idea where the algorithm is going off the rails? Have you filed this as a bug upstream yet? We should try to get their opinion sooner than later.

Kushal-Shah-03:
After some testing and a bit of brute force I was able to reduce the number of points for our problematic volume down to 18. This will make it significantly easier for testing.
I've attached the .bin file containing the points (18points.bin) to prevent any loss of data from floating point conversion.

The points are added to the hull in each algorithm in this order 
Note : Base refers to the points added to the base tetrahedron, Hull1 is current implementation, Hull2 is VHACD implementation

Hull1 Point Base -18.2869 31.6738 2.175
Hull1 Point Base -17.6332 -17.342 89.9628
Hull1 Point Base 49.8656 -0 55.5071
Hull1 Point Base -20.4426 -35.4077 8.275
Hull1 Point -23.0164 39.8656 79.0839
Hull1 Point 10.2294 -14.7178 10.508
Hull1 Point -24.9832 43.2722 52.7107
Hull1 Point -18.6273 -39.5444 55.5071
Hull1 Point 13.2675 -19.98 28.1179
Hull1 Point -24.9832 -40.2722 52.7107
Hull1 Point 11.1761 -22.3575 45.2756
Hull1 Point -24.9492 1.5 54.5981
Hull1 Point -25 21.8857 49.9071
Hull1 Point 18.4196 -18.2153 52.4501
Hull1 Point -25 -12.7727 49.9071
Hull1 Point -24.9832 -43.2722 52.7107
Hull1 Point 9.20583 -23.4785 55.334
Hull1 Point -1.62324 -29.7942 48.3949

Hull2 Point Base -17.6332 -17.342 89.9628
Hull2 Point Base 49.8656 -0 55.5071
Hull2 Point Base -23.0164 39.8656 79.0839
Hull2 Point Base -18.2869 31.6738 2.175
Hull2 Point -20.4426 -35.4077 8.275
Hull2 Point -24.9832 43.2722 52.7107
Hull2 Point -24.9832 -43.2722 52.7107
Hull2 Point 10.2294 -14.7178 10.508
Hull2 Point 13.2675 -19.98 28.1179
Hull2 Point -18.6273 -39.5444 55.5071
Hull2 Point -25 21.8857 49.9071
Hull2 Point -24.9492 1.5 54.5981
Hull2 Point 11.1761 -22.3575 45.2756
Hull2 Point -25 -12.7727 49.9071
Hull2 Point 18.4196 -18.2153 52.4501
Hull2 Point -1.62324 -29.7942 48.3949

The first 5 points forming the hull are same for the both algorithms, so we can safely check for the points after it.

Also, I have highlighted the two differing points (these points lead to the excess volume),
One of them "-24.9832 -40.2722 52.7107", is clearly colinear (since -24.9832 43.2722 52.7107, -24.9832 -43.2722 52.7107 are a part of the convex hull)
The other point is "9.20583 -23.4785 55.334"

These 5 points are coplanar for our case
(-24.9832, -43.2722, 52.7107) (-25, -12.7727, 49.9071) (-24.9832, -40.2722, 52.7107) (-25, 21.8857, 49.9071) (-24.9832, 43.2722, 52.7107) 

I will make the upstream issue tonight after our discussion, I wanted to reduce the testcase to a smaller size and have some basic idea before I posted this issue, thus the delay.
PS : I have an interesting observation that I want to discuss with you today, it may help us identify the issue.

Kushal-Shah-03:
1.
Hull1
Screenshot from 2024-07-03 12-30-55
Hull2
Screenshot from 2024-07-03 12-30-57

2.
Hull1
Screenshot from 2024-07-03 12-31-33
Hull2
Screenshot from 2024-07-03 12-31-40

3.
Hull1
Screenshot from 2024-07-03 12-31-52
Hull2
Screenshot from 2024-07-03 12-31-55

Note: Hull1 is our current implementation, Hull2 is VHACD implementation
These 6 images should give you an idea of where the points lie, and you can clearly see the location of the two differing points (one of them is collinear to one of the edges of the convex hull)

Kushal-Shah-03:
I've created the issue upstream https://github.com/akuukka/quickhull/issues/23, and will now start working on a check for whether the intermediate output of the algorithm is a convex hull. I will create a pull request upstream, when I've made some progress

starseeker:
Could the quickhull implementation be ported to use Manifold's existing data structures directly, if Manifold is going to have its own version that logic?  Curious how much duplication their is conceptually between Quickhull and Manifold in that regard.

starseeker:
I realize I'm very late in spotting this, but maybe it would be worth checking out this one if we haven't already?  I've had pretty good luck with a couple other algorithmic implementations from GTE over the years:

https://www.geometrictools.com/Samples/Geometrics.html#ConvexHull3D

elalish:
Ha! I worked with David Eberly at a startup some years back - great guy. It's true, his code is pretty solid; I should have thought to look there. @starseeker If you happen to have his stuff running already, would you mind giving it a quick speed test against against ours? None of the other algorithms we've tested have been faster than quickhull so far. If it looks like it's as fast or faster, then we should probably look at it more deeply.

starseeker:
I don't have it running at the moment, but I might be able to set it up in the next week or two (we're working on getting a release out right now, so it'll be a little bit before I can dig into it.)

starseeker:
@Kushal-Shah-03 What would you suggest as the best way to get a quick comparison between quickhull and the GTE ConvexHull3.h?

Kushal-Shah-03:
> @Kushal-Shah-03 What would you suggest as the best way to get a quick comparison between quickhull and the GTE ConvexHull3.h?

@starseeker , sorry for the late reply. I have written a script that runs the hull algorithm for the Thingi10K dataset and saves it in a csv file, so we can run it for the current implementation and the new implementation to compare the time and accuracy, I've cleaned it up, and added it to a new PR, I would recommend using that. I've explained how to use the testing script there.
You would have to integrate the code for the new implementation into manifold though for it to work (i.e just put the hull implementation inside a wrapper Hull function in our code, I've implemented several implementations that you can take as an example you can look at this https://github.com/Kushal-Shah-03/manifold/blob/master/src/manifold/src/manifold.cpp#L961C1-L997C2 and then some declarations at https://github.com/Kushal-Shah-03/manifold/blob/6448874ed9adba9c88470a0b8447643676da0647/src/manifold/src/manifold.cpp#L1391C1-L1391C70, https://github.com/Kushal-Shah-03/manifold/blob/master/src/manifold/src/manifold.cpp#L1428C1-L1430C2, https://github.com/Kushal-Shah-03/manifold/blob/master/src/manifold/include/manifold.h#L295C1-L297C60 would get it to work (the declarations are very simple though shouldn't be any issues)). If you do face issues in integrating the code do tag me, I'll try my best to explain the process.

starseeker:
@Kushal-Shah-03 I'll be juggling a lot for the next week or two here, so I don't have a lot of bandwidth - if you're still set up to do it, any chance you could put GTE through the comparison?  Hopefully the above does most of the "how to hook it in" part...

elalish:
@Kushal-Shah-03 has enough to work on for now, so I think we'll leave this one for the moment. Besides, Geometric Tools uses exact arithmetic, which isn't how Manifold operates, and it tends to cost in performance.

starseeker:
Is there anything in this pull request that should be merged, or is it OBE now?

elalish:
Oh, good call - forgot to close this one. It's been very much superseded now.

PR #782, opened 2024-03-29
fix glm install

pca006132:
Fixes #777


elalish:
Huh, well now the CI is reproducing exactly the error from the issue on Mac. Go figure.

pca006132:
Hmm, probably know what is happening her. The main problem is that we are using glm but it is not in our exported targets. But putting it in our exported target will make cmake complain as well. My previous "fix" is removing it from our install interface, but in that case probably the header path is messed up when using our library.

Anyway, will try to fix.

pca006132:
I need to check if the glm installation works on ubuntu as well, will do tmr.

PR #789, opened 2024-04-13
Use forVert

### LINKED ISSUE #775: vertex halfedge iterator

### LINKED ISSUE #784: Vec out of Range
I started getting strange "Vec out of Range" errors on some of my models after recently updating. Happen with some hulls operations. They're not that large though, should not be blowing past size limits. It is oddly state dependent. I haven't been able to isolate it to a small case yet.  @pca006132 perhaps you could illuminate when this error in VecView might occur?

Possibly related to #742

I'm currently two commits behind main on 32eefe9157616aaa4d3b817cadcdf9d0c5b33491.

elalish:
Re-land #778 

Fixes #775 
Fixes #784 

Let's get a test added and fix whatever problem I introduced.


src/manifold/src/edge_op.cpp
```diff
@@ -325,12 +323,10 @@ void Manifold::Impl::DedupeEdge(const int edge) {
...
     if (vertNormal_.size() > 0) vertNormal_.push_back(vertNormal_[endVert]);
 
-    do {
+    ForVert(current, [this, newVert](int current) {
       halfedge_[current].endVert = newVert;
-      current = NextHalfedge(current);
-      halfedge_[current].startVert = newVert;
-      current = halfedge_[current].pairedHalfedge;
-    } while (current != edge);
+      halfedge_[halfedge_[current].pairedHalfedge].startVert = newVert;
```
pca006132, line 328: This should be `halfedge_[NextHalfedge(current)].startVert = newVert`?

src/manifold/src/edge_op.cpp
```diff
@@ -350,12 +346,10 @@ void Manifold::Impl::DedupeEdge(const int edge) {
...
     if (vertNormal_.size() > 0) vertNormal_.push_back(vertNormal_[endVert]);
 
-    do {
+    ForVert(current, [this, newVert](int current) {
       halfedge_[current].endVert = newVert;
-      current = NextHalfedge(current);
-      halfedge_[current].startVert = newVert;
-      current = halfedge_[current].pairedHalfedge;
-    } while (current != pair);
+      halfedge_[halfedge_[current].pairedHalfedge].startVert = newVert;
```
pca006132, line 351: same as above

src/manifold/src/smoothing.cpp
```diff
@@ -947,18 +947,17 @@ void Manifold::Impl::CreateTangents(std::vector<Smoothness> sharpenedEdges) {
...
-          tangent[current] = smoothness * tangent[current];
-          current = NextHalfedge(halfedge_[current].pairedHalfedge);
-        }
-      };
-
-      SmoothHalf(first, second,
-                 (vert[0].second.smoothness + vert[1].first.smoothness) / 2);
-      SmoothHalf(second, first,
-                 (vert[1].second.smoothness + vert[0].first.smoothness) / 2);
+      float smoothness =
```
pca006132, line 950: should we run `ForVert` twice here?

elalish: No, I changed the function a bit so it could finish in one pass instead.

src/manifold/src/edge_op.cpp
elalish: Not quite, but that was definitely the bug. Turns out I was orbiting the wrong end of the edge.

elalish:
@cartesian-theatrics Thank you for the excellent test case! I'm honestly a little troubled we didn't have that case covered in our tests before...

Issue #790, opened 2024-04-13
Triangulate bug: Two separate polygons


kintel:
Adding two separate outlines to `Polygons` and tessellating sometimes causes very odd triangulations.

Example test case:
```
TEST(Polygon, Separate_Quad_And_Tri) {
  Polygons polys;
  polys.push_back({{3, 1}, {3, 0}, {2, 0}});
  polys.push_back({{0.8, 0.8}, {1, 1}, {0, 1}, {0.2, 0.8}});
  TestPoly(polys, 3, 0.00145);
}
```

This should yield 3 triangles (one triangulated quad plus one triangle), but:
```
Expected equality of these values:
  triangles.size()
    Which is: 10
  2 * expectedNumTri
    Which is: 6
```

Side note: These tests only measure the number of triangles, which may make the tests very brittle.

elalish:
When we run our tests, we run them with `params.intermediateChecks = true`, which asserts the manifoldness of each triangulation, and with `params.processOverlaps = false`, which checks that every triangle is CCW. Those should catch this well before the number of triangles check - would you mind adding this test as a PR; then we should see that behavior on the CI. 

This seems like a pretty basic error - I'm surprised our existing corpus doesn't catch it.

elalish:
Not that it should matter, but that's a surprisingly wide precision you've specified - is that necessary to repro?

kintel:
I opened a PR with the failure case. The precision setting isn't needed, I just inherited the argument from surrounding tests in my first repro.

I experience lots of errors like this in OpenSCAD (first attempt at using the triangulator). This is the most trivial one I could find, but it's probably good to start off fixing this before potentially registering a bunch of duplicates.

elalish:
Closing as expected behavior since this is an invalid polygon.

kintel:
Thanks for taking a look - I'll fix my upstream issue and run my larger tests :)

PR #791, opened 2024-04-14
Add failing polygon test

kintel:
Add failing test for #790

### DIFF
--- a/test/polygon_corpus.cpp
+++ b/test/polygon_corpus.cpp
@@ -4747,3 +4747,10 @@ TEST(Polygon, Woodgrain2) {
   });
   TestPoly(polys, 9, 0.00145);
 }
+
+TEST(Polygon, Separate_Quad_And_Tri) {
+  Polygons polys;
+  polys.push_back({{3, 1}, {3, 0}, {2, 0}});
+  polys.push_back({{0.8, 0.8}, {1, 1}, {0, 1}, {0.2, 0.8}});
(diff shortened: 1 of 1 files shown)


test/polygon_corpus.cpp
```diff
@@ -4747,3 +4747,10 @@ TEST(Polygon, Woodgrain2) {
   });
   TestPoly(polys, 9, 0.00145);
 }
+
+TEST(Polygon, Separate_Quad_And_Tri) {
+  Polygons polys;
+  polys.push_back({{3, 1}, {3, 0}, {2, 0}});
```
elalish, line 4753: I should have noticed this earlier - these polygons are not valid (not even epsilon-valid). The triangle is CW and outside of the quad. So I guess I would say this is expected behavior, in that the returned triangulation is manifold, but not entirely CCW. 

I'm even seeing that it's producing your expected 3 triangles - only in the duplicate case (where we put a second copy right next to the first) is it producing 10 tris instead of 6, since it's attempting to fit holes into contours.

What would be your expected behavior?

kintel: Oh, I didn't notice that at first either. OpenSCAD is way too forgiving about winding order (we use clipper to sanitize input polygons). I still got odd results in OpenSCAD, but the Manifold test now passes. I'll go back and debug where that odd behavior originates.

elalish:
closing as expected behavior.

PR #793, opened 2024-04-24
Smooth quads

elalish:
Adding quad smoothing support, for proper cylinders, etc.


CMakeLists.txt
```diff
@@ -91,7 +91,7 @@ else()
   else()
     set(WARNING_FLAGS -Werror -Wall -Wno-sign-compare -Wno-unused)
   endif()
-  set(MANIFOLD_FLAGS ${MANIFOLD_FLAGS} ${WARNING_FLAGS} -O3)
```
elalish, line 94: Removing this lets release and debug mode work by default - debug mode actually works great now that we're not on CUDA.

elalish (commented):
@pca006132 The CI is reproducing what I see locally - these smooth/refine tests are now throwing Vec out of range, but when I run each test in isolation, they pass just fine. I tried address sanitizer, and it finds an integer overflow, but a check in that function finds nothing. Any ideas?

CMakeLists.txt
pca006132: Sounds like something caused by `-O3`? Perhaps we got some undefined behavior... I should look into this later.

elalish: No, the undefined behavior is independent of the amount of optimization.

pca006132: Ah, so removing this line does not make it work, just allows you to compile in debug mode?

elalish: exactly - and actually I just solved the problem. Turns out I had some extra unused verts that weren't being used, but were being processed and contained undefined stuff. Address sanitizer is great, but even though it doesn't slow down running a debug build that much, it makes the actual debugger unusably slow. Oh well, went back to print statements and figured it out.

elalish:
@hrgdavor, since you were interested, smoothing is working better now. From our `Manifold.SmoothFlat` test, we start with this low-res truncated cone:
image

We can use `CalculateNormals` to create vertex normals, designed to make everything smooth except for edges beyond a sharpness threshold, in this case only the lower circle. The renderer can make the reflections look smooth, sort of, but it doesn't match the geometry:
image

With refined geometry and flat facets (based either on vertex normals or the geometry directly - same results in this case):
image
Note the quads on the sides are refined as quads - the edge between the two triangles is ignored, allowing a symmetric solution. Note also that the polygonal face on top remains perfectly flat, and the sides adopt its tangents to interpolate smoothly. Also note that the lower edge is sharp (as requested) but still smoothly circular. 

With refined geometry and interpolated normals:
image
Now the normals match the geometry better and the shape looks real.

elalish:
However, our interpolation is not perfectly circular. It never will be, but I may try to improve it a bit still:

https://github.com/elalish/manifold/assets/1649964/4e8569b2-f98b-4796-b8bd-a1fe25968a99

elalish:
Cylinders are perfect at least.

hrgdavor:
@elalish thanks for sharing the visuals. I really admire you work on manifold. I have finally started to play a bit with manifold, so looking forward to trying out these things after I spend some time with basics.

PR #794, opened 2024-04-25
Ensure properties maintained through Decompose operation

### LINKED ISSUE #786: Manifold Decompose doesn't preserve vertex properties
Calling Decompose on a Manifold made up of multiple disconnected meshes divides it into the separate Manifolds, but only containing vertex positions and not including other vertex properties if defined (unless I'm mistaken?).

harmanpa:
Fixes #786 by setting numProps in GatherFaces method. Added test for Decompose with properties, and ExpectProperties function in test_main in order to check properties in a mesh match those set by SetProperties.

### DIFF
--- a/test/manifold_test.cpp
+++ b/test/manifold_test.cpp
@@ -161,8 +161,7 @@ TEST(Manifold, Decompose) {
 }
 
 TEST(Manifold, DecomposeProps) {
-  auto posNorm = [](float* newProp, glm::vec3 pos,
-                           const float* oldProp) {
+  auto posNorm = [](float* newProp, glm::vec3 pos, const float* oldProp) {
     newProp[0] = hypotf(hypotf(pos.x, pos.y), pos.z);
   };
   std::vector<Manifold> manifoldList;
--- a/test/manifold_test.cpp
+++ b/test/manifold_test.cpp
@@ -160,6 +160,32 @@ TEST(Manifold, Decompose) {
   RelatedGL(manifolds, input);
 }
 
+TEST(Manifold, DecomposeProps) {
+  auto posNorm = [](float* newProp, glm::vec3 pos,
+                           const float* oldProp) {
+    newProp[0] = hypotf(hypotf(pos.x, pos.y), pos.z);
+  };
+  std::vector<Manifold> manifoldList;
+  manifoldList.emplace_back(Manifold::Tetrahedron().SetProperties(1, posNorm));
+  manifoldList.emplace_back(Manifold::Cube().SetProperties(1, posNorm));
+  manifoldList.emplace_back(Manifold::Sphere(1, 4).SetProperties(1, posNorm));
+  Manifold manifolds = Manifold::Compose(manifoldList);
+
+  ExpectMeshes(manifolds, {{8, 12, 1}, {6, 8, 1}, {4, 4, 1}});
+  ExpectProperties(manifolds.GetMeshGL(), 1, posNorm);
+
+  std::vector<MeshGL> input;
+
+  for (const Manifold& manifold : manifoldList) {
+    EXPECT_GE(manifold.OriginalID(), 0);
+    auto mesh = manifold.GetMeshGL();
+    input.emplace_back(mesh);
+    ExpectProperties(mesh, 1, posNorm);
+  }
+
+  RelatedGL(manifolds, input);
+}
+
 /**
  * These tests check the various manifold constructors.
  */
--- a/test/manifold_test.cpp
+++ b/test/manifold_test.cpp
@@ -161,25 +161,19 @@ TEST(Manifold, Decompose) {
 }
 
 TEST(Manifold, DecomposeProps) {
-  auto posNorm = [](float* newProp, glm::vec3 pos, const float* oldProp) {
-    newProp[0] = hypotf(hypotf(pos.x, pos.y), pos.z);
-  };
   std::vector<Manifold> manifoldList;
-  manifoldList.emplace_back(Manifold::Tetrahedron().SetProperties(1, posNorm));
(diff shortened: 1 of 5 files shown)


pca006132:
LGTM, @elalish what do you think about the tests?

test/manifold_test.cpp
```diff
@@ -160,6 +160,31 @@ TEST(Manifold, Decompose) {
   RelatedGL(manifolds, input);
 }
 
+TEST(Manifold, DecomposeProps) {
+  auto posNorm = [](float* newProp, glm::vec3 pos, const float* oldProp) {
+    newProp[0] = hypotf(hypotf(pos.x, pos.y), pos.z);
```
elalish, line 165: `hypotf` is a new one for me, but I think it's simpler to go with `glm::length(pos)`.

test/test_main.cpp
```diff
@@ -404,6 +404,29 @@ void RelatedGL(const Manifold& out, const std::vector<MeshGL>& originals,
   }
 }
 
+void ExpectProperties(const MeshGL& mesh, int numProp,
```
elalish, line 407: I don't think this function is necessary, as I believe `RelatedGL` already checks all of this in a more generic way.

test/manifold_test.cpp
```diff
@@ -160,6 +160,31 @@ TEST(Manifold, Decompose) {
...
+  ExpectMeshes(manifolds, {{8, 12, 1}, {6, 8, 1}, {4, 4, 1}});
+  ExpectProperties(manifolds.GetMeshGL(), 1, posNorm);
+
+  std::vector<MeshGL> input;
+
+  for (const Manifold& manifold : manifoldList) {
+    EXPECT_GE(manifold.OriginalID(), 0);
+    auto mesh = manifold.GetMeshGL();
+    input.emplace_back(mesh);
+    ExpectProperties(mesh, 1, posNorm);
```
elalish, line 182: Looking at this, I think you're just testing your inputs, rather than the result. I believe you want to run something like `newManifoldList = manifolds.Decompose()`, then call `RelatedGL(manifold, input);` - you'll need separate loops for input and output though.

elalish (changes requested):
Thanks! Let's just tighten up the test.

test/manifold_test.cpp
```diff
@@ -160,6 +160,31 @@ TEST(Manifold, Decompose) {
   RelatedGL(manifolds, input);
 }
 
+TEST(Manifold, DecomposeProps) {
+  auto posNorm = [](float* newProp, glm::vec3 pos, const float* oldProp) {
```
elalish, line 164: Actually, maybe just replace this with our test function `WithPositionColors`. That's also convenient for debug, since you can export the meshes and look at the coloration to figure out what's going on.

test/manifold_test.cpp
harmanpa: OK, I see what I did, the Decompose is actually called in the Expect method

test/manifold_test.cpp
harmanpa: Done

test/test_main.cpp
harmanpa: OK, removed

test/manifold_test.cpp
harmanpa: Probably calls the same method in the end!

test/manifold_test.cpp
```diff
@@ -160,6 +160,25 @@ TEST(Manifold, Decompose) {
...
+  manifoldList.emplace_back(WithPositionColors(Manifold::Sphere(1, 4)));
+  Manifold manifolds = Manifold::Compose(manifoldList);
+
+  ExpectMeshes(manifolds, {{8, 12, 3}, {6, 8, 3}, {4, 4, 3}});
+
+  std::vector<MeshGL> input;
+
+  for (const Manifold& manifold : manifoldList) {
+    auto mesh = manifold.GetMeshGL();
+    input.emplace_back(mesh);
```
elalish, line 176: `WithPositionColors` *is* a `MeshGL` - why not create `input` at the same time you create `manifoldList`?

test/manifold_test.cpp
```diff
@@ -160,6 +160,25 @@ TEST(Manifold, Decompose) {
...
 }
 
+TEST(Manifold, DecomposeProps) {
+  std::vector<Manifold> manifoldList;
+  manifoldList.emplace_back(WithPositionColors(Manifold::Tetrahedron()));
+  manifoldList.emplace_back(WithPositionColors(Manifold::Cube()));
+  manifoldList.emplace_back(WithPositionColors(Manifold::Sphere(1, 4)));
+  Manifold manifolds = Manifold::Compose(manifoldList);
+
+  ExpectMeshes(manifolds, {{8, 12, 3}, {6, 8, 3}, {4, 4, 3}});
```
elalish, line 170: This is only testing that `Decompose` has the right number of properties. Are they all zeros? Let's replace this `ExpectMeshes` with an explicit `Decompose` and run `RelatedGL(manifold, input)` on each returned item.

elalish (changes requested):

test/manifold_test.cpp
harmanpa: OK

test/manifold_test.cpp
harmanpa: How should I handle the order of the output from Decompose(), which I don't think is necessarily the same as the input list?

elalish: It doesn't matter - you use the whole input list for each one. It'll find the correct one to test against via OriginalID.

elalish (approved):
Thanks, this looks great!

PR #795, opened 2024-04-25
C bindings: pass context to set_properties and warp operation callbacks

harmanpa:
Similar to #427 , this PR creates and tests functions in c bindings for passing data to the callback function. We use Manifold from Java and this allows us to pass callbacks from Java that carry references to the associated Java objects, however it is also useful for passing parameters, as done in the tests in manifoldc_test.cpp.

New functions:
* manifold_set_properties_context
* manifold_warp_context
* manifold_cross_section_warp_context

New test:
* CBIND properties

Changed tests:
* CBIND warp_translation
* CBIND level_set

### DIFF
--- a/bindings/c/manifoldc.cpp
+++ b/bindings/c/manifoldc.cpp
@@ -35,9 +35,10 @@
 using namespace manifold;
 
 namespace {
-ManifoldMeshGL *level_set(
-    void *mem, float (*sdf_context)(float, float, float, void *),
-    ManifoldBox *bounds, float edge_length, float level, bool seq, void *ctx) {
+ManifoldMeshGL *level_set(void *mem,
+                          float (*sdf_context)(float, float, float, void *),
+                          ManifoldBox *bounds, float edge_length, float level,
+                          bool seq, void *ctx) {
   // Bind function with context argument to one without
   using namespace std::placeholders;
   std::function<float(float, float, float)> sdf =
@@ -258,9 +259,9 @@ ManifoldManifold *manifold_mirror(void *mem, ManifoldManifold *m, float nx,
 }
 
 ManifoldManifold *manifold_warp(void *mem, ManifoldManifold *m,
-                                        ManifoldVec3 (*fun)(float, float, float,
-                                                            void *),
-                                        void *ctx) {
+                                ManifoldVec3 (*fun)(float, float, float,
+                                                    void *),
+                                void *ctx) {
   // Bind function with context argument to one without
   using namespace std::placeholders;
   std::function<ManifoldVec3(float, float, float)> f3 =
@@ -272,9 +273,10 @@ ManifoldManifold *manifold_warp(void *mem, ManifoldManifold *m,
   return to_c(new (mem) Manifold(warped));
 }
 
-ManifoldMeshGL *manifold_level_set(
-    void *mem, float (*sdf)(float, float, float, void *), ManifoldBox *bounds,
-    float edge_length, float level, void *ctx) {
+ManifoldMeshGL *manifold_level_set(void *mem,
+                                   float (*sdf)(float, float, float, void *),
+                                   ManifoldBox *bounds, float edge_length,
+                                   float level, void *ctx) {
   return level_set(mem, sdf, bounds, edge_length, level, false, ctx);
 }
 
--- a/bindings/c/manifoldc.cpp
+++ b/bindings/c/manifoldc.cpp
@@ -35,16 +35,7 @@
 using namespace manifold;
 
 namespace {
-ManifoldMeshGL *level_set(void *mem, float (*sdf)(float, float, float),
-                          ManifoldBox *bounds, float edge_length, float level,
-                          bool seq) {
-  std::function<float(glm::vec3)> fun = [sdf](glm::vec3 v) {
-    return (sdf(v.x, v.y, v.z));
-  };
-  auto mesh = LevelSet(fun, *from_c(bounds), edge_length, level, !seq);
-  return to_c(new (mem) MeshGL(mesh));
-}
-ManifoldMeshGL *level_set_context(
+ManifoldMeshGL *level_set(
     void *mem, float (*sdf_context)(float, float, float, void *),
     ManifoldBox *bounds, float edge_length, float level, bool seq, void *ctx) {
   // Bind function with context argument to one without
@@ -267,15 +258,6 @@ ManifoldManifold *manifold_mirror(void *mem, ManifoldManifold *m, float nx,
 }
 
 ManifoldManifold *manifold_warp(void *mem, ManifoldManifold *m,
-                                ManifoldVec3 (*fun)(float, float, float)) {
-  std::function<void(glm::vec3 & v)> warp = [fun](glm::vec3 &v) {
-    v = from_c(fun(v.x, v.y, v.z));
-  };
-  auto warped = from_c(m)->Warp(warp);
-  return to_c(new (mem) Manifold(warped));
-}
-
-ManifoldManifold *manifold_warp_context(void *mem, ManifoldManifold *m,
                                         ManifoldVec3 (*fun)(float, float, float,
                                                             void *),
                                         void *ctx) {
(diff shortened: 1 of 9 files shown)


pca006132:
Do we have more functions that accept a callback without a ctx parameter? I think this will be the nice place to fix everything (if there is any). I think we don't have to introduce a `context` variant for these APIs, to avoid the API from being too complicated.

The problem here is *will it cause existing problems to break badly*? I think it should either cause compilation failure or not affecting existing program. I think adding parameters at the end of the list of arguments is fine. For x86-64, we are looking at `cdecl` calling convention, which is fine with additional parameters (it will just ignore the additional one at the end) or fewer parameters (it will just be junk). For AArch64, it should also be fine as the parameters are passed in register.

TL;DR: It is fine to just add parameters at the end. Standard C API is fine with this.

harmanpa:
I couldn't find other methods in the C bindings with a callback.

Your suggestion is to have the context parameter always but it be optional?

src/manifold/src/sort.cpp
```diff
@@ -464,6 +464,7 @@ void Manifold::Impl::GatherFaces(const Impl& old, const Vec<int>& faceNew2Old) {
     gather(policy, faceNew2Old.begin(), faceNew2Old.end(),
            old.meshRelation_.triProperties.begin(),
            meshRelation_.triProperties.begin());
+    meshRelation_.numProp = old.meshRelation_.numProp;
```
elalish, line 467: It looks like you may have started with your other PR's branch. Should probably rebase onto master to avoid merge conflicts - and to make it easier to review.

bindings/c/manifoldc.cpp
```diff
@@ -275,6 +275,21 @@ ManifoldManifold *manifold_warp(void *mem, ManifoldManifold *m,
   return to_c(new (mem) Manifold(warped));
 }
 
+ManifoldManifold *manifold_warp_context(void *mem, ManifoldManifold *m,
```
elalish, line 278: I agree that if adding an optional ctx parameter to a single function works, that does seem cleaner. It would mean a small breaking change for the existing level set function and such, but since I'm guessing you're the only user of this for now, it's probably fine.

test/manifoldc_test.cpp
```diff
@@ -23,19 +23,35 @@ TEST(CBIND, warp_translation) {
...
+  ManifoldManifold *warpedcontext =
+      manifold_warp_context(malloc(sz), sphere, warpcontext, context);
+  ManifoldManifold *diffcontext =
+      manifold_difference(malloc(sz), trans, warped);
 
   ManifoldProperties props = manifold_get_properties(diff);
+  ManifoldProperties propscontext = manifold_get_properties(diffcontext);
 
   EXPECT_NEAR(props.volume, 0, 0.0001);
+  EXPECT_NEAR(propscontext.volume, 0, 0.0001);
```
elalish, line 46: If what you're testing is translation, perhaps we should add a bounding box check to verify it moved.

test/manifoldc_test.cpp
```diff
@@ -69,10 +106,41 @@ TEST(CBIND, level_set) {
 #endif
 
   EXPECT_EQ(manifold_status(sdf_man), MANIFOLD_NO_ERROR);
+  EXPECT_EQ(manifold_status(sdf_man_context), MANIFOLD_NO_ERROR);
```
elalish, line 109: Can we add some volume and surface area checks to tighten these tests?

elalish (changes requested):
Thanks! Good idea.

pca006132:
> Your suggestion is to have the context parameter always but it be optional?

Kind of, users should put a null if they don't want to use it (and are aware of this API update).

My explanation above is just me thinking about if this will break applications if they assume the old API. Because linker does not check the number of arguments, they may not notice the API change. And my conclusion is it is probably fine.

harmanpa:
Sorry for all the commits, I'm more used to C than C++ so not knowing what libraries to use.

I've made the "context" variants of functions into the defaults. All builds and tests but don't know possible external impacts.

test/manifoldc_test.cpp
```diff
@@ -19,29 +19,76 @@ TEST(CBIND, sphere) {
...
+  context[0] = 15.0f;
+  ManifoldVec3 (*warpcontext)(
+      float, float, float, void *) = [](float x, float y, float z, void *ctx) {
+    ManifoldVec3 v = {x + ((float *)ctx)[0], y, z};
+    return v;
+  };
   ManifoldManifold *sphere = manifold_sphere(malloc(sz), 1.0f, 100);
   ManifoldManifold *trans = manifold_translate(malloc(sz), sphere, 15., 0., 0.);
-  ManifoldManifold *warped = manifold_warp(malloc(sz), sphere, warp);
+  ManifoldManifold *warped = manifold_warp(malloc(sz), sphere, warp, NULL);
```
elalish, line 36: Can we test @pca006132's idea by simply omitting `NULL` here (and in the other tests)? If that works, then this is a pretty clean API. The only breaking change is removing `manifold_level_set_context`, which I'm comfortable with, especially since you're the main user of it.

test/manifoldc_test.cpp
```diff
@@ -69,10 +137,70 @@ TEST(CBIND, level_set) {
...
+        5.0f;
+  };
+  float *context = (float *)malloc(1 * sizeof(float));
+  context[0] = 5.0f;
+  void (*propscontext)(float *, ManifoldVec3, const float *,
+                       void *) = [](float *new_prop, ManifoldVec3 position,
+                                    const float *old_prop, void *ctx) {
+    new_prop[0] =
+        glm::sqrt(glm::sqrt(position.x * position.x + position.y * position.y) +
+                  position.z * position.z) *
```
elalish, line 189: How about `glm::length(position)`?

test/manifoldc_test.cpp
```diff
@@ -69,10 +137,70 @@ TEST(CBIND, level_set) {
...
+
+  // Analytic calculations for volume and surface area
+  float a = context[0] * context[1];
+  float b = context[0] * context[2];
+  float c = context[0] * context[3];
+  float s = 4.0f * glm::pi<float>() *
+            std::pow(((std::pow(a * b, 1.6f) + std::pow(a * c, 1.6f) +
+                       std::pow(b * c, 1.6f)) /
+                      3.0f),
+                     1.0f / 1.6f);
```
elalish, line 150: Thanks for the geometry lesson! 😄

elalish (commented):
I assume you've tested this is all working with your Java bindings?

test/manifoldc_test.cpp
harmanpa: Unfortunately I can't get it to compile without the NULL in the tests.

I know there's tricks we can play with varargs or preprocessor but not sure how you would feel about that.

elalish: No, that's fine, let's leave it. I assume C doesn't have nice things like default arguments? @pca006132 Did I misunderstand what you were saying?

pca006132: Yeah I mean something else. Will give an example later today.

pca006132:
This is what I mean:

```shell
$ cat foo.c  
#include <stdio.h>

int foo(int a, int (*cb)(int, int), int b) {
  printf("%d %d %d\n", a, b, cb(a, b));
  return 0;
}

$ cat bar.c
#include <stdio.h>

int foo(int a, int (*cb)(int));

int cb(int a) {
  printf("a = %d\n", a);
  return a * 10;
}

int main() {
  printf("%d\n", foo(1, cb));
  return 0;
}

$ gcc foo.c bar.c
$ ./a.out
a = 1
1 -1245249192 10
0
```

i.e. when the users are not using the header file from C but declares the function in their language, it should not affect the original functionality.

test/manifoldc_test.cpp
elalish: Okay, i guess we can skip this one. Is this ready to merge then?

harmanpa: I'm happy if you're happy

PR #800, opened 2024-04-29
Split manifold tests

dasoya:
Split manifold test into 
manifold_test, 
manifold_advanced_test,
manifold_smooth_test,
manifold_mingap_test.


test/manifold_advanced_test.cpp
elalish, line 1: This name is pretty vague. I'm trying to think of a clearer way to categorize these tests. If you look at impl.h, you'll see the functions are grouped by which file they pertain to, e.g. smoothing.cpp, edge_op.cpp, etc. Our public API functions map pretty directly to these - perhaps we can categorize them by which file they are primarily testing? Please take a look at let me know if that makes sense or if it's difficult/confusing in practice.

test/manifold_smooth_test.cpp
elalish, line 1: Let's just call this `smooth_test.cpp` and apply that convention to the rest of the naming as well.

test/manifold_smooth_test.cpp
```diff
@@ -0,0 +1,221 @@
...
+
+#include "cross_section.h"
+#include "manifold.h"
+#include "samples.h"
+#include "test.h"
+#include "tri_dist.h"
+
+using namespace manifold;
+
+TEST(Manifold, Smooth) {
```
elalish, line 25: Let's go with the convention of `TEST(Smooth, Tetrahedron)` - where the first name refers to the testing file (smooth_test.cpp) and the second is more specific. These tests have grown over time, so some of the older ones have pretty vague names we should update.

test/manifold_mingap_test.cpp
elalish, line 1: I just noticed MinGap() is in impl.cpp - it should probably in properties.cpp instead. Then we can make this file about all the properties-related tests. Of course nearly every test checks properties, but I mean just the ones that verify the property functions are working properly.

elalish (changes requested):
Thanks! Good start.

test/manifold_advanced_test.cpp
```diff
@@ -0,0 +1,417 @@
+// Copyright 2021 The Manifold Authors.
```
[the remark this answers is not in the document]
dasoya: okay

elalish:
Thanks @dasoya and sorry that you didn't get into GSoC for this project - it would have been great to have you! Are you planning to come back to this PR, or should one of us take it over?

dasoya:
> Thanks @dasoya and sorry that you didn't get into GSoC for this project - it would have been great to have you! Are you planning to come back to this PR, or should one of us take it over?


Thank you for considering my proposal! I still want to contribute to this pr. I've been busy with exams the past few days, so I haven't been able to.. sorry!

test/manifold_smooth_test.cpp
dasoya: Okay Done

test/manifold_smooth_test.cpp
dasoya: Yes.

test/manifold_mingap_test.cpp
dasoya: I moved MinGap() to properties.cpp. changed this file to "properties_test.cpp" and added related tests.

elalish (approved):
This looks great, thank you! You're always welcome here as a contributor if you're interested - and I'm happy to give you some direction if you want it.

PR #805, opened 2024-05-10
Runs example

elalish:
I've updated our Three.js example significantly - it's now TS, with Vite building source maps for easy inspection in dev tools. It's now using our own `Merge` instead of Three's utility function. Most importantly it now demonstrates material mapping as a round-trip from Three.js constructs, through Manifold, and back to Three for rendering, with lots of comments. This is meant to address the desire for a tutorial on using `runOriginalID`, etc: https://github.com/elalish/manifold/issues/91#issuecomment-2102050474

This will be live at https://manifoldcad.org/three once this gets merged.


hrgdavor:
This is excellent, I have things to learn in that demo both about threejs and manifold :) .... my use of threejs is pretty basic ATM.

PR #808, opened 2024-05-10
Vitest update problem repro

elalish:
I'm having trouble updating vitest, so made this PR to demonstrate the regression I've found.


elalish:
Fixed by #822.

Issue #803, opened 2024-05-09
Modularize Manifold


elalish:
We've been getting more requests to have just core Manifold functionality (usually Booleans) with minimal dependencies (like Clipper, convex hulls, SDFs, etc). Our code is largely already partitioned this way, but I think we need to make some build flags to allow it to build modularly. Context: https://github.com/godotengine/godot/pull/91748#discussion_r1595679227

I'm curious if anyone knows any best practices for how to set this up? ASSIMP does a lot of this. @kintel perhaps this would also help for OpenSCAD? Curious if you have any thoughts.

pca006132:
I feel that getting rid of Clipper can be quite painful. If we want to do something about clipper dependency, it will be a large breaking change and should be in the 3.0 plan.

elalish:
Isn't clipper pretty much walled-off within `CrossSection`? I think we should be able to build Manifold without CrossSection if we want.

elalish:
Also relevant to Blender integration: https://projects.blender.org/blender/blender/issues/120182#issuecomment-1170472

pca006132:
ah, I was conflating it with glm, not something optional.

probably indicating I should sleep now :P

kintel:
Some thou^H^H^H^Hramblings:
* Splitting up into feature sets is a bit of a slippery slope, as the number of possible permutations may explode, making testing very annoying. Can be mitigated by creating some automated test "matrixes". If there is a way of splitting into distinct sub-libraries, that might be preferable. Not sure how feasible that is.
* Some larger libraries do this pretty well (e.g. Qt, boost). They pretty much split into completely distinct libraries, possibly with some interdependencies.
* I kind of assume that people asking for this are mostly thinking in terms of dependencies, not features, but it would be nice if this is surfaced as feature selectors where the dependencies follow.
* In general, be prepared to spend some quality time with CMake :)
* Once you get sufficiently fed up with CMake, don't fall for the temptation to create a build system generator-generator script, as you'll eventually reinvent one of the 86 different build systems already created to solve this problem.
* Don't skimp on automated testing

[...]

PR #810, opened 2024-05-12
make quickhull optional

elalish:
Addresses part of #803 

This seems like a pretty simple change to make our Quickhull dependency optional (or whatever other hull library we might switch to). I just added a build flag `MANIFOLD_HULL` that defaults to `ON`. When turned `OFF`, `quickhull` is no longer a dependency and our `Manifold::Hull()` methods ~~just return an empty manifold~~ just fail to link. Does this seem like a reasonable approach @fire @kintel?


pca006132:
I think we can also just avoid compiling those functions, so if users attempt to use it they will get a linker error.

src/manifold/src/manifold.cpp
```diff
@@ -904,6 +906,7 @@ CrossSection Manifold::Project() const {
 
 ExecutionParams& ManifoldParams() { return manifoldParams; }
 
+#ifdef MANIFOLD_HULL
```
elalish, line 909: Like this?

pca006132: Yes. Do note that some of the tests will not compile now due to linker error, so we need to ifdef them as well.

elalish: Already done.

elalish: Hmm, or not. I got the tests, but not the bindings. I kind of feel like Hull shouldn't be optional for the bindings. Any thoughts for how I should set up the CI to test this?

pca006132: The simplest way is to put ifdef in `manifoldc.h`, like https://github.com/elalish/manifold/blob/6b4a4ed841357f627573181402d3bb979092b5bb/bindings/c/include/manifoldc.h#L414

kintel:
This can be sorted out later, but it would be cool if consumers of Manifold could do smth. like this in CMake:
```
find_package(manifold REQUIRED COMPONENTS manifold hull polygon)
```

..and thus get an error at configure-time if Manifold doesn't offer the requested components.
I assume this organization would be easier if each optional component technically is a separate library object.

src/manifold/src/manifold.cpp
elalish: Wouldn't we then need to do the same for the Python and WASM bindings? This is why I was thinking of not letting these dependencies be optional for the bindings, since we package them all up and distribute them anyway.

pca006132: I think it may make sense for c binding, as there is no package manager for it and people need to compile on their own?

For python and js, I agree we can just not let them be optional.

elalish:
@kintel I have some odd holes in my knowledge regarding how libraries actually work together (probably stemming from my odd path to software engineering). `cross_section` (what I think you meant by `polygon`) I think is independent enough it should be usable independently of anything else. 

On the other hand, `hull` is currently just a few methods on the manifold class, rather than its own library. Is it possible to make that a separate library? Can an independent target just implement a method on a different library's class? Because that would be pretty cool if that's possible.

src/manifold/CMakeLists.txt
```diff
@@ -32,4 +32,17 @@ target_compile_features(${PROJECT_NAME}
 )
 
 install(TARGETS ${PROJECT_NAME} EXPORT manifoldTargets)
+
+if(MANIFOLD_HULL)
+add_library(hull module/hull.cpp)
+target_link_libraries(hull
+    PRIVATE manifold quickhull
```
elalish, line 39: @pca006132 here's an attempt at making this even more modular, I believe going toward what @kintel was suggesting. This appears to work, though I'd love some feedback on best practices and such. I still need to fix up the bindings.

kintel:
> On the other hand, `hull` is currently just a few methods on the manifold class, rather than its own library. Is it possible to make that a separate library? Can an independent target just implement a method on a different library's class? Because that would be pretty cool if that's possible.

Humm, that's probably not ideal. I assume this needs to be a method because it accesses class internals?

src/manifold/src/manifold.cpp
kintel: How would Python and JS customers know whether the hull feature is available?

elalish: >For python and js, I agree we can just not let them be optional.

I think the optionality will only be for C++ consumers.

elalish:
> Humm, that's probably not ideal. I assume this needs to be a method because it accesses class internals?

It makes a more pleasant API shape and it avoids a copy. But apparently a library can in fact implement methods of another class. Seems kind of cool now that I know about it. Can you explain why it might not be ideal?

kintel:
> But apparently a library can in fact implement methods of another class. Seems kind of cool now that I know about it. Can you explain why it might not be ideal?

it's possible that you may find a technical solution allowing you to do this, despite limitations imposed by shared libraries on various platforms. Static libraries are naturally fine as they're essentially just a bunch of compilation unit object files concatenated together.

It feels weird though. You'd basically offer a library which declares an API but doesn't fully implement it. ..then offer another library implementing said API, but doesn't declare it.
Would you have to offer a third library implementing a dummy implementation of the missing method too?

There surely exist commonly used patterns for designing something like this, but it might be cleaner to (re-)design for this particular type of modularization. Apart from offering the extra functionality as free helper functions, I don't have a great gut feeling for this without reading the code. In general I'd say that designing a more decoupled system tends to come with somewhat greater complexity.

elalish:
Thanks @kintel, I think I pretty much agree. Unless someone has strong opinions, I think I'll put a pin in this one for now. @Kushal-Shah-03 is working on optimizing the convex hull in #781 this summer for GSoC, so we may end up with our own implementation, in which case I'll ditch the dependency and leave the API as is. If we still need a dependency, then I'll consider your free-function approach as part of v3.0 breaking changes.

kintel:
Another option to free functions could be some sort of registration API where libraries ("plugins") can register functions and the Manifold API can offer lookup and invocation methods. This works if function signatures are standardized. Benefit is that consumers only have to deal with one API. Drawback is that registration needs to be done somewhere. Dynamic library init time is good, but we need to make sure static linkers don't kill unused functions..

fire:
A common way that c++ handles this is by api interface inheritance.

1. you provide an interface that either has no implementation or return the inputs
2. for each implementation follow the interface contract

fire:
Not that familiar with manifold but where is the c++ method prototype that is common?

src/manifold/module/hull.cpp
```diff
@@ -0,0 +1,65 @@
...
+using namespace manifold;
+
+/**
+ * Compute the convex hull of a set of points. If the given points are fewer
+ * than 4, or they are all coplanar, an empty Manifold will be returned.
+ *
+ * @param pts A vector of 3-dimensional points over which to compute a convex
+ * hull.
+ */
+Manifold Manifold::Hull(const std::vector<glm::vec3>& pts) {
```
elalish, line 28: @fire here is the implementation of these hull functions, which lives in the `hull` target. Without that linked in, these methods (defined in `manifold.h`) will fail to link if called.

fire: I expected this api to be in a .h header file. Strange.

elalish: The API is defined in `manifold.h`, this is the implementation.

elalish:
@pca006132 Do you have any idea why only the windows TBB build is failing? It's complaining about a bunch of `size_t` to `int` conversion, but I haven't changed anything here, and it's weird that it works without TBB.

pca006132:
Probably because master is merged, and I enabled warnings about implicit signedness conversion. Will have a look at that.

pca006132:
Maybe try to add `/wd4267` to msvc arguments?

```
set(MANIFOLD_FLAGS ${MANIFOLD_FLAGS} /DNOMINMAX /bigobj /wd4267)
```

We will eventually properly fix this, but that day is not today :P

elalish:
Hmm, no luck with `/wd4267` - but I still don't understand how it can be working on master and not here?

pca006132:
Oh, nvm, the warnings are fine, they never cause errors in msvc (we did not set warnings as errors).

The issue is
```
error LNK2019: unresolved external symbol "public: class manifold::Manifold __cdecl manifold::Manifold::Hull(void)const "
```

This makes more sense.

.github/workflows/manifold.yml
```diff
@@ -166,10 +143,10 @@ jobs:
         submodules: recursive
     - uses: jwlawson/actions-setup-cmake@v2
     - uses: ilammy/msvc-dev-cmd@v1
-    - name: Build ${{matrix.backend}}
+    - name: Build ${{matrix.parallel_backend}}
       shell: powershell
       run: |
-        cmake . -DCMAKE_BUILD_TYPE=Release -DBUILD_SHARED_LIBS=OFF -DMANIFOLD_DEBUG=ON -DMANIFOLD_PAR=${{matrix.parallel_backend}} -A x64 -B build
+        cmake . -DCMAKE_BUILD_TYPE=Release -DBUILD_SHARED_LIBS=OFF -DMANIFOLD_DEBUG=ON -DMANIFOLD_HULL=${{matrix.parallel_backend == 'NONE' && 'ON' || 'OFF'}} -DMANIFOLD_PAR=${{matrix.parallel_backend}} -A x64 -B build
```
elalish, line 149: @kintel You can see here that our Windows CI is now building without Hull for the TBB version - everything compiles and links fine because the hull test file is excluded. You just get a linker error if you try to actually use the `Hull` functions with `MANIFOLD_HULL` disabled. Does that look like a reasonable way to use the library and minimize dependencies?

elalish:
Sure, we can reopen if he hits a blocker.

PR #811, opened 2024-05-13
Fix `Manifold.prototype.{split,splitByPlane,rotate}` JavaScript bindings functions

errendir:
The `Manifold.prototype.split` was incorrectly calling `this._split` instead of `this._Split`. Emscripten bindings export `_Split` with capital "S": https://github.com/elalish/manifold/blob/master/bindings/wasm/bindings.cpp#L138

Similar problem was present for the `Manifold.prototype.splitByPlane`. This PR fixes that as well

`Manifold.prototype.rotate` was broken in a different way: it only accepted 3d vectors in a form of an array of three numbers. However the TypeScript types suggest you can also call this method with three parameters.

### DIFF
--- a/bindings/wasm/bindings.js
+++ b/bindings/wasm/bindings.js
@@ -248,8 +248,12 @@ Module.setup = function() {
     return this._Translate(vararg2vec3(vec));
   };
 
-  Module.Manifold.prototype.rotate = function(vec) {
-    return this._Rotate(...vec);
+  Module.Manifold.prototype.rotate = function(xOrVec, y, z) {
+    if (Array.isArray(xOrVec)) {
+      return this._Rotate(...xOrVec);
+    } else {
+      return this._Rotate(xOrVec, y || 0, z || 0);
+    }
   };
 
   Module.Manifold.prototype.scale = function(vec) {
@@ -269,14 +273,14 @@ Module.setup = function() {
   };
 
   Module.Manifold.prototype.split = function(manifold) {
-    const vec = this._split(manifold);
+    const vec = this._Split(manifold);
     const result = fromVec(vec);
     vec.delete();
     return result;
   };
 
   Module.Manifold.prototype.splitByPlane = function(normal, offset = 0.) {
-    const vec = this._splitByPlane(vararg2vec3([normal]), offset);
+    const vec = this._SplitByPlane(vararg2vec3([normal]), offset);
     const result = fromVec(vec);
     vec.delete();
     return result;
--- a/bindings/wasm/examples/public/examples.js
+++ b/bindings/wasm/examples/public/examples.js
@@ -12,6 +12,9 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+/** @type {import("../../manifold-encapsulated-types")["Manifold"]} */
+const Manifold = globalThis.Manifold;
+
 export const examples = {
   functions: {
     Intro: function() {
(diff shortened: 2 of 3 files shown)


bindings/wasm/bindings.js
```diff
@@ -248,8 +248,12 @@ Module.setup = function() {
     return this._Translate(vararg2vec3(vec));
   };
 
-  Module.Manifold.prototype.rotate = function(vec) {
-    return this._Rotate(...vec);
+  Module.Manifold.prototype.rotate = function(xOrVec, y, z) {
+    if (Array.isArray(xOrVec)) {
+      return this._Rotate(...xOrVec);
```
elalish, line 253: I believe we already have a helper function you can use here: `vararg2vec3`.

bindings/wasm/examples/public/examples.js
```diff
@@ -12,6 +12,9 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+/** @type {import("../../manifold-encapsulated-types")["Manifold"]} */
+const Manifold = globalThis.Manifold;
```
elalish, line 16: What does this do?

bindings/wasm/examples/public/examples.js
```diff
@@ -536,7 +539,37 @@ export const examples = {
         0, 0, size / Math.sqrt(2)
       ]);
       return result;
-    }
+    },
+
+    Split: function() {
```
elalish, line 544: For better or worse, these serve double duty as JS tests, but primarily as public examples in ManifoldCAD.org. I'd like to keep that list highly curated. Would you mind adding the split functionality to one of the existing examples instead? I think `RoundedFrame` would be good - you can just split by a centered cube to make the inside a different color from the outside.

elalish (changes requested):
Good find, thank you!

bindings/wasm/bindings.js
errendir: I was considering using `vararg2vec3` however `this._Rotate` seems to be a bit different from `_Scale`, `_Mirror` or `_Translate`—it doesn't take `glm::vec3` but just three floats. So we'd have to do:
```ts
Module.Manifold.prototype.rotate = function(...v) {
  const vec = vararg2vec3(v)
  return this._Rotate(vec.x, vec.y, vec.z);
};
```
which seems wasteful—we are creating an object just to destructure it immediately. Perhaps it's a negligible performance hit, so if you like this version more, let me know and I will change it in the PR

bindings/wasm/examples/public/examples.js
errendir: I added those two lines to get the TypeScript typechecking working in this .js file. TypeScript allows you to specify types in .js files via the JSDoc syntax: https://www.typescriptlang.org/docs/handbook/jsdoc-supported-types.html#type

This is purely for dev convenience, it doesn't impact the ManifoldCAD page at all, since the types are injected directly to the editor there.

bindings/wasm/examples/public/examples.js
errendir: OK, I split the frame into the inside and outside part. The inside part is slightly different color and appears/disappears as part of an animation. This makes it clear that it's a separate mesh and not just a material group on the surface of the frame.

The animation could be made a bit nice if moving from scale 1 to scale 0 wasn't animated. Is there a way to achieve this with your animation system? I could also do the following:
```ts
insideNode.translation = (t) => {
  const isInsideVisible = t < 0.5;
  return isInsideVisible ? [0,0,0] : [10000,10000,10000];
}
```
which looks better but it's a pretty nasty hack that probably should not be in the public examples

bindings/wasm/examples/public/examples.js
elalish: Nice, I didn't know that!

bindings/wasm/bindings.js
elalish: Okay, that's fine, thanks.

elalish (approved):
Thanks!

bindings/wasm/examples/public/examples.js
elalish: No, I think what you have is good. Once animation pointer support is broadly adopted, we can probably do fancier things.

PR #815, opened 2024-05-21
Set import library for Clipper2 target when using system version

imciner2:
On Windows, the import library is also needed to link against the Clipper2 DLL, and that must be specified using a different target property.

We noticed this when building manifold in Julia on Windows: https://github.com/JuliaPackaging/Yggdrasil/pull/8712#issuecomment-2122763024


elalish (approved):
Thank you!

PR #816, opened 2024-05-21
CI: use GCC 14 in Linux build

cjmayo:
Ubuntu 24.04 packages GCC 13. GitHub add GCC 14 to their runner.

--- 

There are now some CMake issues to sort out with the Clipper2 snapshot!

```
CMake Error: install(EXPORT "manifoldTargets" ...) includes target "cross_section" which requires target "Clipper2" that is not in this export set, but in multiple other export sets: lib/cmake/clipper2/Clipper2Targets.cmake, share/clipper2/clipper2Targets.cmake.
An exported target cannot depend upon another target which is exported multiple times. Consider consolidating the exports of the "Clipper2" target to a single export.
-- Generating done
CMake Generate step failed.  Build files cannot be regenerated correctly.
```

### DIFF
--- a/.github/workflows/manifold.yml
+++ b/.github/workflows/manifold.yml
@@ -59,11 +59,11 @@ jobs:
         mkdir build
         cd build
         cmake -DCMAKE_BUILD_TYPE=Release -DBUILD_SHARED_LIBS=ON -DMANIFOLD_PAR=${{matrix.parallel_backend}} -DCODE_COVERAGE=ON .. && make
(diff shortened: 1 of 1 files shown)


elalish:
Interesting, I guess file an issue on Clipper2? Or is this a problem in our own CMake?

pca006132:
Interesting, I think the clipper2 update is separate from the GCC update. I forgot a lot of cmake trickeries now, need to check if the best solution is to fix things here or ask upstream to have a single export.

pca006132:
I think it is fine to disable lcov for GCC 14 for now?

elalish (approved):
Thanks!

PR #817, opened 2024-05-22
update clipper2

pca006132:
clipper2 update of #816 for running CIs

Fixes #813, verified by #818.

PR #818, opened 2024-05-22
added project test

### LINKED ISSUE #813: Crash in Project()
`Manifold::Project()` sometimes causes a crash inside Clipper2.
The crash is deterministic and reproducible by passing specific data to Manifold.

I couldn't find any existing test cases for `Project()`, but the mesh in question can be reconstructed from this example:
https://github.com/openscad/openscad/pull/5139

See also stack trace in https://github.com/openscad/openscad/issues/5137

elalish:
This test comes from #813 - I verified that this is in fact a Clipper2 bug, which has already been fixed, so #817 will fix #813 when it lands.


elalish:
The fact that Clipper2 seg faulted on such a simple set of polygons is concerning. I had thought Clipper was better tested than that. This makes me want to decouple our libraries more, making `Manifold` not depend on `CrossSection`, since `CrossSection` depends on Clipper. This may be a good v3.0 breaking change: making Manifold's 2D inputs and outputs use `Polygons` exclusively. It means slightly more work on the part of a user, to manually feed those `Polygons` into `CrossSection` constructors, or call `ToPolygons()`, but I think it's worth it for the separation.

kintel:
My experience with Clipper says that Clipper V1 is pretty well battle tested, but Clipper V2 is relatively new and hasn't seen much action. I haven't managed to get a stable enough Clipper V2 build to switch OpenSCAD yet..

pca006132:
@kintel It would be great if you can open some issues upstream or give examples for the failed test cases. It is better to help upstream to make the library more reliable.

kintel:
I'll make another attempt at Clipper2, probably in the fall. Will create tickets as I encounter new issues.

PR #819, opened 2024-05-22
Update glm to 1.0.1

cjmayo:
Not a massive step forward but takes us past the errors for GTX extensions, unless `GLM_ENABLE_EXPERIMENTAL` is defined, as we fixed in 0b7ee87489a7.

Hopefully this will demonstrate a newer instance of that.


elalish:
That seems reasonable to me. @pca006132 do you see any problem here?

pca006132:
Set `GLM_ENABLE_EXPERIMENTAL` to enable some experimental features?

elalish:
Oh, I see what the problem is now - the glm include is before the `impl.h` include (due to clang formatting), which is what pulls in the existing `#define GLM_ENABLE_EXPERIMENTAL`. Yeah, I think adding it again is fine.

elalish (approved):
Thanks!

PR #820, opened 2024-05-23
Add manifold_meshgl_merge to c bindings

harmanpa:
Add manifold_meshgl_merge to call MeshGL::Merge() on a duplicate mesh via the c bindings


pca006132:
Why do we need to duplicate the mesh here?

harmanpa:
Only because ```from_c``` returns a const so the compiler wouldn't allow it otherwise

pca006132:
I see. LGTM.

PR #821, opened 2024-05-23
Improve smoothing of flat faces

elalish:
This includes a bunch of smoothing improvements. I'll add some screenshots below - first I need to take a look at the diff and clean things up before merging.


elalish:
Here's my original weird shape, complete with a bunch of large flat faces, as well as some quads and triangles. Many of the edges have dihedral angles of 60 or 45 degrees. I'll show various automatic smoothing of this mesh below.
image

Below, using `shape.CalculateNormals(0, 50).SmoothByNormals(0).Refine(8);`, which is the same as `shape.SmoothOut(50).Refine(8);`
image

Below using `shape.SmoothOut(180, 1).Refine(8);` - smoothing everything as much as possible.
image

Below using `shape.SmoothOut(50, 0.5).Refine(8);` - making the sharp corners a bit rounded.
image

PR #824, opened 2024-05-26
Add polygon test showing bad tessellation result + fix

kintel:
I believe this shows that a relatively complex "branchy" polygon with a single triangular hole causes `Triangulate()` to emit some pretty funky triangles.

This polygons is a very slimmed down version of the polygon emitted by https://github.com/openscad/openscad/blob/master/examples/Advanced/module_recursion.scad

The polygon itself is a result or a union of a large number of 2D rectangles using Clipper V1.
It's very possible that there is some precision-related issue causing this, but I believe that a robust triangulator should be able to handle this kind of abuse.

PS! validated by emitting the triangles as an OBJ file and rendering that externally:
```
std::cout << "# polygon-test.cc\n";
  for (auto poly : polys) {
    for (auto v : poly) {
      std::cout << "v " << v[0] << " " << v[1] << " 0\n";
    }
  }
  for (auto tri : triangles) {
    std::cout << "f " << tri[0] + 1 << " " << tri[1] + 1 << " " << tri[2] + 1
              << "\n";
  }
```


elalish:
Indeed, this is a valid polygon and a good test case. Turns out a couple of valid ears were being rejected (positive cost) due to verts near their tip. I've improved their neighbor checking to get correct costs, so now everything proceeds properly.

kintel:
Awesome - tested and works perfectly also with extra torture:

out

kintel:
@elalish I've got another one. It's a bit time consuming to manually trim polygons to minimal examples. Do you have a good/preferred way of providing data-driven polygon tests?

elalish:
If you compile with `MANIFOLD_DEBUG` and set `ManifoldParams().processOverlaps = false`, then it will error out when a bad triangulation is found and print out a test case to the command line that you can copy directly into our tests, like [this](https://github.com/elalish/manifold/blob/master/test/polygon_corpus.cpp#L576).

kintel:
After a while I figured out it's `PolygonParams()`, not `ManifoldParams()`. Will open a new ticket.

elalish:
Oh, oops - you're right.

PR #825, opened 2024-05-28
Create a release file containing submodules

### LINKED ISSUE #783: Manifold 2.4.5 release tar.gz is incomplete
For blender we use the released tgz's of projects rather than cloning from git (for compliance reasons no code can be downloaded during a build, so git, git sub-modules and cmake's `FetchContent` just aren't an options for us), while building manifold's 2.4.5 release from https://github.com/elalish/manifold/releases/tag/v2.4.5 the `src/third_party/quickhull` folder isn't populated in the release tarball, causing the build to fail, we worked around this by copying in the required code during a pre-build step, but it be nice if this is fixed for future releases. 

this is a common problem in the automatically generated tar.gz's on github, other projects that have this issue generally manually upload a tar.gz for a release that has the all the sub modules populated

cjmayo:
Fixes #783

This does assume the release tag starts with v. If this workflow checked and then exited without creating a file that could be a problem.

### DIFF
--- a/.github/workflows/release_file.yml
+++ b/.github/workflows/release_file.yml
@@ -1,0 +1,31 @@
+name: Release File
+
+on:
+  release:
+    types: [published]
+  workflow_dispatch:
+
+jobs:
+  upload_archive:
+    name: Upload archive
+    runs-on: ubuntu-latest
+    steps:
+      - run: |
+          ref=${{ github.ref_name }}
+          echo "release_name=manifold-${ref#v}" >> $GITHUB_ENV
+      - uses: actions/checkout@v4
+        with:
+          path: ${{ env.release_name }}
+          submodules: recursive
+      - name: Build archive
+        run: >
+          tar --exclude=".git*" -cz ${{ env.release_name }} -f ${{ env.release_name }}.tar.gz
+      - name: Log checksum
+        run: >
+          sha256sum ${{ env.release_name }}.tar.gz
+      - name: Add file to release
+        run: |
+          cd ${{ env.release_name }} || exit 1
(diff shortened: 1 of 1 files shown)


.github/workflows/release_file.yml
```diff
@@ -0,0 +1,28 @@
...
+        with:
+          path: ${{ env.release_name }}
+          submodules: recursive
+      - name: Build archive
+        run: >
+          tar  --exclude=".git*" -cz ${{ env.release_name }} -f ${{ env.release_name }}.tar.gz
+      - name: Add file to release
+        run: |
+            cd ${{ env.release_name }} || die
+            gh release upload ${{ github.ref_name }} ../${{ env.release_name }}.tar.gz
```
elalish, line 26: Does this override whatever Github's default release packaging is? Looks great, though I don't have much ability to review or test this. Should we just see what happens on the next release?

elalish: Sounds good, I'll trust you. Looks like the only problem is our CI - it started misbehaving yesterday for no apparent reason. I'm hoping M$ fixes it soon, though if you have any ideas what could be randomly canceling our jobs or how to work around it, I'd love to know! I already tried a few things to no avail: #826.

elalish: Looks like the CI started working again - can you prod this PR with a minor commit to trigger the CI? My manual triggering doesn't rerun the randomly cancelled jobs.

elalish: No problem - in general though I prefer no rebasing and no force-pushing. We do squash and merge anyway, so there's no history pollution. Merging makes things much easier to review because it keeps commits separate. Plus it's safer.

PR #826, opened 2024-05-28
Clean up test files

elalish:


.github/workflows/manifold.yml
```diff
@@ -17,11 +17,12 @@ jobs:
       matrix:
         parallel_backend: [NONE, TBB]
         gcc: [13, 14]
+      max-parallel: 1
+      fail-fast: false
```
elalish, line 21: I didn't change anything in the CI, but it started randomly skipping and canceling jobs for no apparent reason. I've tried these tweaks, but to no avail. @pca006132 any thoughts as to what is going wrong here? I'm tempted to think the Actions CI shipped a bug this afternoon sometime.

pca006132: Not too surprising, GitHub action can be unstable from time to time. I think we can just wait for a day or so and retry.

elalish: Indeed, today it works.

PR #829, opened 2024-06-04
added convex polygon triangulation fast path

elalish:
Addresses part of #723, an alternative to #751. This mostly improves the speed of `Cylinder` constructors and other extrusions of convex polygons.


elalish:
So, my 10,000 point `Manifold.Cylinder` constructor test (dominated by triangulating the top and bottom) was 80ms before #830, 900ms after #830, and 25ms with this PR. I'm a little concerned with the 10x slowdown from #830 - we may want to check the perf of some more polygons.

pca006132:
It seems that this is because the collider is returning almost all points for circles. Perhaps it is related to the angle?

pca006132:
This helps a bit and does not slow down the zebra case, but this is merely some heuristic which will not work in general...

```diff
@@ -430,6 +430,9 @@ class EarClip {
       const float radius = glm::length(openSide) / 2;
       openSide = glm::normalize(openSide);
 
+      if (radius >= collider.collider.BBox().Size().length() / 10.0)
+        return EarCost(precision);
+
       float totalCost = glm::dot(left->rightDir, rightDir) - 1 - precision;
       if (CCW(pos, left->pos, right->pos, precision) == 0) {
         // Clip folded ears first
```

pca006132:
Thinking about it, it may be helpful to have a segment tree to tell you the number of results, and see if you want to do a BVH query or not depending on the number of results (high -> probably no). Comparing with BVH, a segment tree runtime does not depend on the number of points hit, but it cannot tell you which points are hit.

elalish:
I was thinking something similar - compare the bounding box of the query to the bounding box of the entire collider. If it's more than some fraction of the volume, then just return the entire list instead of running through the BVH. Basically I feel like this logic ought to live in the collider, since it may help with difficult 3D cases as well.

pca006132:
My major concern is that this is merely a heuristic, and will not work well when the point distribution is not uniform across the bounding box. Segment tree is precise, but will need some experiments to see if O(log^3 n) is fast enough.

For putting this inside the collider, we will need to copy everything everytime we check for collision, which may not be very efficient, at least for the triangulator case.

pca006132:
We can probably open another ticket about using segment tree for counting collisions, that may allow us to speed up the collider memory allocation phase as well. Having this collision count before actually recording collisions may allow us to avoid storing all collision pairs when there are roughly O(n^2) collisions, which can improve memory consumption and hopefully give us some speed improvement as well (by avoiding those memory lookup).

elalish:
After a cursory look at segment trees I'm not quite sure how they're different than the BVH we have. I think they're more balanced but at the cost of being less parallelizable in construction? I'm not worried about heuristics in performance - I feel like all of performance is heuristic, especially things like cache coherence and the fact that our inputs are not random.

pca006132:
Oh forgot to say this, there are two kinds of segment trees, with the same name. One is the wikipedia entry, the *real* one, is just like BVH. One is the one used in competitive programming, see https://cp-algorithms.com/data_structures/segment_tree.html

elalish:
Thanks, that was interesting reading. Fodder for further optimization, but I think we'll get more bang for our buck out of your heuristic for now. I'll give a shot at making it a little more general and you can review.

elalish:
After poking around some more, triangulating these circles I can only get down by 2x by returning sequence in the sparse return of Collider (based on a comparison of query volume and total volume). I can get it down by 5x if I skip the sparse entirely and do the polygon while loop (a combination of not having to build useless sequence vectors and the fact the polygon skips over processed points without having to check them). However, this convex check is closer to a 50x improvement, since it skips all the collider construction and other triangulator overhead. I'm going to go with this for now, since it's simple and fixes a definite pain point. We can look into the other optimizations sometime, but I'm concerned they're not worth the code complexity since I'm not sure they'll get that much benefit on a wide range of common polygons.

PR #830, opened 2024-06-04
Optimize triangulator with collider

### LINKED ISSUE #723: triangulator optimization

elalish:
Fixes #723, or so's the hope. ~~Does anyone have a good polygon to contribute to a `TEST` that would make clear the performance improvement?~~ 35x improvement on #831 without degrading our other test times, so I think we're good! And this is just the simple version: using our 3D collider for 2D boxes. Could probably be sped up a bit more with a templatized Collider, though I'm not sure it's worth the trouble. 

@pca006132 ~~I seem to be hitting a linker error - I told CMake to link to `collider`, but apparently that's not good enough? I'm guessing it's just a line I forgot somewhere.~~

### DIFF
--- a/src/collider/include/collider.h
+++ b/src/collider/include/collider.h
@@ -30,6 +30,7 @@ class Collider {
   template <const bool selfCollision = false, const bool inverted = false,
             typename T>
   SparseIndices Collisions(const VecView<const T>& queriesIn) const;
+  static uint32_t MortonCode(glm::vec3 position, Box bBox);
 
  private:
   Vec<Box> nodeBBox_;
--- a/src/collider/src/collider.cpp
+++ b/src/collider/src/collider.cpp
@@ -265,6 +265,14 @@ struct TransformBox {
   const glm::mat4x3 transform;
   void operator()(Box& box) { box = box.Transform(transform); }
 };
+
+uint32_t SpreadBits3(uint32_t v) {
+  v = 0xFF0000FFu & (v * 0x00010001u);
+  v = 0x0F00F00Fu & (v * 0x00000101u);
+  v = 0xC30C30C3u & (v * 0x00000011u);
+  v = 0x49249249u & (v * 0x00000005u);
+  return v;
+}
 }  // namespace
 
 namespace manifold {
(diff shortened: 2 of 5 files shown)


pca006132:
Remove the `OBJECT` in

```
add_library(${PROJECT_NAME} OBJECT ${SOURCE_FILES})
```

For polygon, I think things like the polygon in https://github.com/openscad/openscad/issues/4679 can be a good test.

pca006132:
and also #831

elalish:
>Remove the `OBJECT`

You're right! What voodoo is this? Thanks - not sure what I'd do without you.

pca006132:
You will figure this out when you see the linker command used by cmake, which somehow skips the dependencies of polygon when linking with executable (minimizeTests or something) that depends on polygon lib only. I have no idea why cmake works like this, but this is what we get...

elalish:
@kintel Okay, this PR reduces the time for #831 from 11.8 seconds to 0.6 seconds. It doesn't make much difference on our overall test set (not surprising since we don't have much in the way of large polygons) but at least it's not a detriment, so that's good. 

Now I just need to figure out what's wrong with the Zebra.

elalish:
I forgot a simple optimization - with that #831 is now down to 0.33 seconds for a total of 35x improvement!

Issue #802, opened 2024-05-07
Warning comparison of integer expressions of different signedness


fire:
```
./thirdparty/manifold/src/manifold/include/manifold.h: In member function 'uint32_t manifold::MeshGL::NumVert() const':
Error: ./thirdparty/manifold/src/manifold/include/manifold.h:50:41: error: comparison of integer expressions of different signedness: 'std::vector<float>::size_type' {aka 'long unsigned int'} and 'int' [-Werror=sign-compare]
   50 |     if (vertProperties.size() / numProp >= std::numeric_limits<int>::max())
      |         ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
./thirdparty/manifold/src/manifold/include/manifold.h: In member function 'uint32_t manifold::MeshGL::NumTri() const':
Error: ./thirdparty/manifold/src/manifold/include/manifold.h:56:41: error: comparison of integer expressions of different signedness: 'std::vector<float>::size_type' {aka 'long unsigned int'} and 'int' [-Werror=sign-compare]
   56 |     if (vertProperties.size() / numProp >= std::numeric_limits<int>::max())
      |         ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

PR #832, opened 2024-06-05
Fix the zebra polygon

### LINKED ISSUE #831: Triangulation issue: Zebra
The following OpenSCAD script yields a bad Manifold triangulation:
```
$fn=24;
text(text="R",font="naturePro",size=5);
```

Note: This single font glyph consists of ~120 separate outlines with >62K vertices.

out

Saving as SVG and reading that back yields the same issue (SVG attached for convenience).
zebra

The output after enabling `MANIFOLD_DEBUG` and `PolygonParams().processOverlaps = false`:
[zebra.cc.zip](https://github.com/user-attachments/files/15572598/zebra.cc.zip)

elalish:
Fixes #831, or at least it had better before we merge this. I removed a few holes and got it to triangulate properly, which gives us a few clues. The polygon appears to be valid. I haven't managed to make a minimized version yet though.


elalish:
Nice work @kintel - this one horribly excellent polygon test revealed 4 different triangulator bugs. I still need to make the upside-down version work, but I'll save that for tomorrow. For now I want to bask in making this ridiculous zebra finally work.

elalish:
@pca006132 It seems our Windows CI got updated or something - suddenly it's very picky about possible data loss: `C:\Program Files (x86)\Microsoft Visual Studio\2019\Enterprise\VC\Tools\MSVC\14.29.30133\include\xutility(5532,1): warning C4267: 'argument': conversion from 'size_t' to 'int', possible loss of data [D:\a\manifold\manifold\build\src\manifold\manifold.vcxproj]`
Do you know an easy way to get it to shut up?

pca006132:
No idea, but should be easy to fix. Will do with #802.

pca006132:
but the actual error seems to be

```
fatal error C1128: number of sections exceeded object file format limit: compile with /bigobj
```

https://stackoverflow.com/questions/69903253/cmake-bigobj-no-such-file-or-directory

this may work:
```cmake
if (MSVC)
  add_compile_options(/bigobj)
endif ()
```

pca006132:
btw I feel like the `zebra.cpp` file should be placed within another folder in the test directory, as I can foresee us adding more and more complicated test cases in the future.

elalish:
Perfect, that's exactly what I needed! Agreed about the separate test folder, but let's wait until we have another file in that category to create it.

Issue #804, opened 2024-05-09
Build without exceptions


elalish:
We did a fair bit of work to ensure Manifold worked with exceptions disabled, but I'm concerned that may have been broken in the time since. First step is to make sure at least one of our CI builds is without exceptions. If that doesn't work, let's see how we can fix it. WDYT @pca006132?

elalish:
For reference: https://github.com/godotengine/godot/pull/91748

fire:
I am testing this in https://github.com/godotengine/godot/pull/94321

elalish:
Thank you! We should have marked this as fixed by #833. Please re-open this if you find a problem.

Issue #744, opened 2024-02-16
better error handling


pca006132:
e.g. poison the object that thrown an error, so we will not observe things in inconsistent states after throwing exceptions. This can happen when we have shared csg nodes, and got an error when evaluated for the first time, and inconsistent when trying to evaluate for the second time.

pca006132:
Thinking about it, not sure if we need more than

```diff
diff --git a/src/manifold/src/boolean_result.cpp b/src/manifold/src/boolean_result.cpp
index ca5f4612..93631e2c 100644
--- a/src/manifold/src/boolean_result.cpp
+++ b/src/manifold/src/boolean_result.cpp
@@ -671,11 +671,22 @@ Manifold::Impl Boolean3::Result(OpType op) const {
   const int c3 = op == OpType::Intersect ? 1 : -1;
 
   if (inP_.IsEmpty()) {
+    if (inP_.status_ != Manifold::Error::NoError ||
+        inQ_.status_ != Manifold::Error::NoError) {
+      auto impl = Manifold::Impl();
+      impl.status_ = Manifold::Error::InvalidConstruction;
+      return impl;
+    }
     if (!inQ_.IsEmpty() && op == OpType::Add) {
       return inQ_;
     }
     return Manifold::Impl();
   } else if (inQ_.IsEmpty()) {
+    if (inQ_.status_ != Manifold::Error::NoError) {
+      auto impl = Manifold::Impl();
+      impl.status_ = Manifold::Error::InvalidConstruction;
+      return impl;
+    }
     if (op == OpType::Intersect) {
       return Manifold::Impl();
     }
```

This will propagate errors in the boolean operands.

PR #833, opened 2024-06-08
chores

### LINKED ISSUE #802: Warning comparison of integer expressions of different signedness

pca006132:
Fix #802, #804 and https://github.com/elalish/manifold/pull/832#issuecomment-2155831898

@elalish When doing this change, I found that we are using 32-bit integers extensively. I feel like the proper way would be to use 64-bit integers. There may be some performance hit, but I doubt if it will be huge. And it just feels better if we can make things work with larger meshes, considering we don't have any reasons not to do so? Some public APIs will have to be changed, but they will be changed when we update to double precision floating point.

For #804, it should now build fine, but it is not the best way of handling exceptions. We need more careful checks everywhere, probably together with #744.


pca006132:
probably messed up the python binding somewhere, will fix tmr.

src/manifold/include/manifold.h
```diff
@@ -47,16 +47,20 @@ class CsgLeafNode;
...
+#endif
     return vertProperties.size() / numProp;
   };
   /// Number of triangles
   uint32_t NumTri() const {
+#if MANIFOLD_EXCEPTIONS 
     if (vertProperties.size() / numProp >=
         std::numeric_limits<unsigned int>::max())
       throw std::out_of_range("mesh too large");
+#endif
```
elalish, line 63: I think I don't understand something about C++ exceptions. I thought any `throw` just automatically became `exit` when exceptions are disabled in the compiler. Or does disabling exceptions just turn uses of `throw` into compile errors?

src/utilities/include/vec_view.h
```diff
@@ -46,16 +48,16 @@ class VecView {
...
     return ptr_[i];
   }
 
   inline T &operator[](size_t i) {
-    if (i >= size_) {
-      throw std::out_of_range("Vec out of range");
-    }
+#if MANIFOLD_EXCEPTIONS
+    if (i >= size_) throw std::out_of_range("Vec out of range");
+#endif
```
elalish, line 60: Should we make another wrapper for these a bit like our `ASSERT`?

elalish (commented):
Thanks, this looks great. Given the breaking changes, what do you say we hold this for ~week while we release v2.5, then put this in for v3.0 with the switch to doubles and all?

src/utilities/include/vec_view.h
pca006132: probably, my goal for now is to get it to compile
I think we need better error handling

src/manifold/include/manifold.h
pca006132: I vaguely recall that emscripten requires no `throw` statement to compile with exception disabled. Not really sure about that.

elalish: Okay, but how about in normal C++ workflows? I still don't really understand #804. Is there something we did that made it impossible to build with exceptions disabled? @fire - can you help me understand https://github.com/godotengine/godot-proposals/issues/9711#issuecomment-2146063161? Our exceptions at this point are only a convenience for debug and such. Is there something that breaks when you compile with exceptions disabled?

fire: When we compile with exceptions enabled, for example web, and a few of the platforms break. We couldn't disable exceptions.

elalish: Ah, so this is specific to web. In that case this PR may well address the issue. We could really use this kind of detail here or in #804. Which platforms break? What were the errors? Why couldn't you disable exceptions? Do you want to disable exceptions? 

I really want to help you out, but I do need at least *some* feedback.

fire: Here's what I can do. I'll push the branch with exceptions on and watch godot engine cid crash. Will post the branch.

fire: https://github.com/V-Sekai/godot/tree/vsk-csg-exceptions-4.3

pca006132:
Converting everything to size_t turns out to be quite complicated, mainly due to `SparseIndices` and `Collider`. Probably need some more time to do it, and evaluate the performance differences. This can be merged even though the changes are incomplete.

pca006132:
@elalish note that the old `ASSERT` is now `DEBUG_ASSERT`, and `ASSERT` is for things that should still be checked when `MANIFOLD_DEBUG` is disabled.

src/manifold/src/boolean3.cpp
```diff
@@ -137,7 +139,7 @@ inline thrust::pair<int, glm::vec2> Shadow01(
 
 // https://github.com/scandum/binary_search/blob/master/README.md
 // much faster than standard binary search on large arrays
-size_t monobound_quaternary_search(VecView<const int64_t> array, int64_t key) {
+ssize_t monobound_quaternary_search(VecView<const int64_t> array, int64_t key) {
```
elalish, line 142: what is `ssize_t`? Okay, I found [this](https://stackoverflow.com/questions/19224655/using-ssize-t-vs-int) and am slightly more confused.

src/manifold/src/properties.cpp
```diff
@@ -201,7 +201,7 @@ struct CheckHalfedges {
 
     const Halfedge paired = halfedges[halfedge.pairedHalfedge];
     bool good = true;
-    good &= paired.pairedHalfedge == edge;
+    good &= paired.pairedHalfedge == static_cast<int>(edge);
```
elalish, line 204: Will the update of `Halfedge` happen with `Collider`?

src/manifold/src/smoothing.cpp
```diff
@@ -434,7 +434,7 @@ std::vector<Smoothness> Manifold::Impl::SharpenEdges(
     float minSharpAngle, float minSmoothness) const {
   std::vector<Smoothness> sharpenedEdges;
   const float minRadians = glm::radians(minSharpAngle);
-  for (int e = 0; e < halfedge_.size(); ++e) {
+  for (int e = 0; e < static_cast<int>(halfedge_.size()); ++e) {
```
elalish, line 437: Why not the same as the change just below?

src/manifold/src/subdivision.cpp
```diff
@@ -125,9 +125,9 @@ class Partition {
       }
     }
     const int offset = interiorOffset - newVerts.size();
-    for (int i = newVerts.size(); i < vertBary.size(); ++i) {
-      newVerts.push_back(i + offset);
-    }
+    size_t old = newVerts.size();
+    newVerts.resize(vertBary.size());
+    std::iota(newVerts.begin() + old, newVerts.end(), old + offset);
```
elalish, line 130: 👍

src/polygon/src/polygon.cpp
```diff
@@ -949,15 +950,17 @@ std::vector<glm::ivec3> TriangulateIdx(const PolygonsIdx &polys,
                                        float precision) {
   std::vector<glm::ivec3> triangles;
   float updatedPrecision = precision;
+#if MANIFOLD_EXCEPTION
```
elalish, line 953: 👍

src/utilities/include/utils.h
```diff
@@ -18,6 +18,11 @@
 #include <mutex>
 #include <unordered_map>
 
+#if defined(_MSC_VER)
+#include <BaseTsd.h>
+typedef SSIZE_T ssize_t;
```
elalish, line 23: This makes me slightly worried about `ssize_t`...

src/utilities/include/vec.h
```diff
@@ -13,7 +13,9 @@
 // limitations under the License.
 
 #pragma once
+#if MANIFOLD_EXCEPTIONS
 #include <exception>
+#endif
```
elalish, line 18: Do we need to include this per file, or can it just be wherever we define `ASSERT`?

src/utilities/include/vec_view.h
```diff
@@ -14,7 +14,11 @@
 
 #pragma once
 
+#if MANIFOLD_EXCEPTIONS
 #include <stdexcept>
+#endif
```
elalish, line 19: same as above.

.github/workflows/manifold.yml
```diff
@@ -107,6 +107,9 @@ jobs:
     timeout-minutes: 30
     runs-on: ubuntu-22.04
     if: github.event.pull_request.draft == false
+    strategy:
+      matrix:
+        exception: [ON, OFF]
```
elalish, line 112: Is there any reason to build WASM with exceptions off? Maybe better to test one of the nix builds since they're fast?

fire: I have a use for wasm with exceptions off.

elalish: @fire I think your earlier comment mentioned you do build WASM with exceptions off - that requires special CMake instructions which I was considering removing since JS is generally exceptions-based. But if WASM sans exceptions is actually useful to someone, then @pca006132 has it set up properly now. Do you build for WASM without exceptions?

fire: I would expect to build Godot Engine WASM without exceptions since the Godot Engine editor and Godot Engine game templates build without exceptions in the default configuration on all platforms.

elalish: Thanks! Okay, let's leave it as is in this PR then.

bindings/python/manifold3d.cpp
```diff
@@ -68,7 +68,7 @@ struct nb::detail::type_caster<glm::vec<N, T, Q>> {
     int size = PyObject_Size(src.ptr());  // negative on failure
     if (size != N) return false;
     make_caster<T> t_cast;
-    for (size_t i = 0; i < size; i++) {
+    for (int i = 0; i < size; i++) {
```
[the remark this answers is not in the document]
pca006132: Ah, should fix this. Thanks.

src/cross_section/src/cross_section.cpp
```diff
@@ -177,33 +177,28 @@ void HullBacktrack(const glm::vec2& pt, std::vector<glm::vec2>& stack) {
...
-    lower.push_back(pts[i]);
+  for (auto& pt : pts) {
+    HullBacktrack(pt, lower);
+    lower.push_back(pt);
   }
   auto upper = std::vector<glm::vec2>{};
-  for (int i = len - 1; i >= 0; i--) {
-    HullBacktrack(pts[i], upper);
-    upper.push_back(pts[i]);
+  for (auto pt_iter = pts.rbegin(); pt_iter != pts.rend(); pt_iter++) {
```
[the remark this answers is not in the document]
pca006132: No, this is also backward, `rbegin` and `rend` are reverse iterators.

src/manifold/include/manifold.h
```diff
@@ -47,14 +47,16 @@ class CsgLeafNode;
 struct MeshGL {
   /// Number of property vertices
   uint32_t NumVert() const {
-    if (vertProperties.size() / numProp >= std::numeric_limits<int>::max())
-      throw std::out_of_range("mesh too large");
+    ASSERT(vertProperties.size() / numProp <
+               static_cast<size_t>(std::numeric_limits<unsigned int>::max()),
```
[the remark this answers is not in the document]
pca006132: I forgot why, need to check. Probably due to `MeshGL` indices are/should be unsigned integers?

src/manifold/src/boolean3.cpp
pca006132: `ssize_t` is signed version of `size_t`. But thinking about it, we only need that -1 to signal failure, and I can do it with `size_t` max as well.

src/manifold/src/edge_op.cpp
```diff
@@ -156,14 +156,14 @@ void Manifold::Impl::SimplifyTopology() {
     ZoneScopedN("DedupeEdge");
     Vec<SortEntry> entries;
     entries.reserve(nbEdges / 2);
-    for (int i = 0; i < nbEdges; ++i) {
+    for (unsigned int i = 0; i < nbEdges; ++i) {
```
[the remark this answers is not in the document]
pca006132: Will fix.

src/manifold/src/properties.cpp
pca006132: Yes.

src/manifold/src/smoothing.cpp
pca006132: Probably some error when I was too sleepy writing this, will fix.

src/utilities/include/utils.h
pca006132: yeah, thinking about it `ssize_t` is not really needed.

src/utilities/include/vec.h
pca006132: yeah probably just need it when we define `ASSERT`

src/cross_section/src/cross_section.cpp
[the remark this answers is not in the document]
elalish: ah, reading too quickly, thanks.

pca006132:
sorry, was having a headache today. Will update this tmr

pca006132:
I also updated nanobind version.

pca006132:
there are still a lot of uses of integer that should be fixed, but we should probably leave to another PR and unblock your iterator changes.

elalish (approved):
Thanks, this looks great!

PR #835, opened 2024-06-11
Flair normals

elalish:
This updates `SetNormals` to flair out the normals along sharp creases to better match the curvature they will connect to.


elalish:
Before:
image

After:
image

PR #848, opened 2024-06-16
[test] Smooth.SineSurface fails - the difference between prop3.volume and 8.43 exceeds 0.01

cjmayo:
~~Not for merging~~

```
 [ RUN      ] Smooth.SineSurface
/home/runner/work/manifold/manifold/test/smooth_test.cpp:361: Failure
The difference between prop3.volume and 8.43 is 0.012265510559082315, which exceeds 0.01, where
prop3.volume evaluates to 8.442265510559082,
8.43 evaluates to 8.4299999999999997, and
0.01 evaluates to 0.01.
[  FAILED  ] Smooth.SineSurface (17 ms)
```

Appears to happen only when both `-march` and `-O` are used.

Something is setting `-O3` here, but I can reproduce with `-march=x86-64-v3 -O2` and `-march=znver2 -O2`, on GCC 13 and 14 respectively.

manifold-2.5.0 passes all tests with `-march` and `-O`.

### DIFF
--- a/test/smooth_test.cpp
+++ b/test/smooth_test.cpp
@@ -358,7 +358,7 @@ TEST(Smooth, SineSurface) {
 
   Manifold smoothed3 = Manifold(surface).SmoothOut(50, 0.5).Refine(8);
   auto prop3 = smoothed3.GetProperties();
-  EXPECT_NEAR(prop3.volume, 8.43, 0.01);
+  EXPECT_NEAR(prop3.volume, 8.44, 0.01);
   EXPECT_NEAR(prop3.surfaceArea, 31.72, 0.01);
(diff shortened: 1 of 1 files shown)


elalish:
Looks like I just didn't update the number accurately - I'm getting 8.438 locally, so the check ought to be 8.44. Want to update it and check it's okay? This looks like some rounding error from doing ops in a different order probably. Let me know if you actually notice something ugly in the resulting shape: `./manifold_test gtest_filter=Smooth.SineSurface -e`

elalish:
Care to simplify the PR to that one-liner?

elalish (approved):
Thanks!

PR #850, opened 2024-06-16
removed cross_section dependency from manifold

elalish:
This is one of several steps toward #803, and specifically addressing https://github.com/elalish/manifold/discussions/664#discussioncomment-9402889. Now Manifold has no direct dependence on Clipper2 - Clipper2 is only a dependency of CrossSection (2D), which is independent of Manifold (3D). We rely only the very simple `Polygons` class as an interface between these libraries, akin to `Mesh`. 

Since we build our Python and JS bindings with Manifold and CrossSection bundled together, I'm keeping their APIs from changing, so they still accept and return `CrossSection` instead of `Polygon`, since that feels a little nicer.

### DIFF
--- a/src/manifold/src/constructors.cpp
+++ b/src/manifold/src/constructors.cpp
@@ -152,8 +152,8 @@ Manifold Manifold::Cube(glm::vec3 size, bool center) {
       glm::length(size) == 0.) {
     return Invalid();
   }
-  glm::mat4x3 m =
-      glm::translate(center ? (-size / 2.0f) : glm::vec3(0)) * glm::scale(size);
+  glm::mat4x3 m(glm::translate(center ? (-size / 2.0f) : glm::vec3(0)) *
+                glm::scale(size));
   return Manifold(std::make_shared<Impl>(Manifold::Impl::Shape::Cube, m));
 }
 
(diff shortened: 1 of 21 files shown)


src/manifold/src/constructors.cpp
```diff
@@ -322,28 +325,40 @@ Manifold Manifold::Extrude(const CrossSection& crossSection, float height,
...
+Manifold Manifold::Revolve(const Polygons& crossSection, int circularSegments,
+                           float revolveDegrees) {
   ZoneScoped;
-  Polygons polygons = crossSection.ToPolygons();
 
-  if (polygons.size() == 0) {
-    return Invalid();
+  Polygons polygons;
+  float radius = 0;
+  for (const SimplePolygon& poly : crossSection) {
```
pca006132, line 334: It seems that our only non-trivial change is here. Do we have test cases that can cover it? While the logic is simple, it is easy to have bugs here.

pca006132 (commented):
LGTM. I think we can also add a Markdown document for the list of breaking changes implemented, so it is easier for users to follow.

src/manifold/src/constructors.cpp
elalish: Indeed - yes this originally came in from an external contributor who put in test cases for it, so I think we're covered.

PR #852, opened 2024-06-25
Remove fancy iterators

elalish:
This refactor will aid #823 - by removing zip iterators all we'll need is a counting iterator, which will reduce the amount of Thrust we rely on. Frankly, I also feel like this cleans up the code a little anyway - I always hated those tuples. 

I've only done a few files so far, but it's going pretty quickly. Should be no big deal - just wanted to show my progress.


elalish:
@pca006132 Before I go too much further, do you have a sense of how much conflict this will generate with #833? Perhaps we should merge that first if it's ready soon.

elalish:
Okay, I've removed zip iterators completely and done some other cleanup, like removing `reduce_by_key`. The main iterator thing that's left is a few usages of `transform_iterator` which can probably be taken care of more easily in #823.

pca006132:
Nice, will review later.

src/collider/src/collider.cpp
```diff
@@ -154,15 +154,13 @@ struct CreateRadixTree {
 
 template <typename T, const bool selfCollision, typename Recorder>
 struct FindCollisions {
+  VecView<const T> queries;
   VecView<const Box> nodeBBox_;
   VecView<const thrust::pair<int, int>> internalChildren_;
```
pca006132, line 159: Thinking if we can remove this `thrust::pair` as well, maybe with `std::pair`?
Not really related to fancy iterator, but I feel that the reason for using `thrust::pair` is mostly because of `zip`.

src/manifold/src/boolean3.cpp
```diff
@@ -475,26 +477,11 @@ Vec<int> Winding03(const Manifold::Impl &inP, Vec<int> &vertices, Vec<int> &s02,
...
-  auto endPair = reduce_by_key<
-      thrust::pair<decltype(w03val.begin()), decltype(w03val.begin())>>(
-      policy, vertices.begin(), vertices.end(), s02.begin(), w03vert.begin(),
-      w03val.begin());
-  scatter(policy, w03val.begin(), endPair.second, w03vert.begin(), w03.begin());
-
-  if (reverse)
-    transform(policy, w03.begin(), w03.end(), w03.begin(),
-              thrust::negate<int>());
+  for_each_n(policy, countAt(0), s02.size(),
```
pca006132, line 481: this is much simpler!

src/manifold/src/impl.cpp
```diff
@@ -677,19 +625,29 @@ void Manifold::Impl::CreateHalfedges(const Vec<glm::ivec3>& triVerts) {
...
-  stable_sort(policy, zip(edge.begin(), ids.begin()),
-              zip(edge.end(), ids.end()),
-              [](const thrust::tuple<uint64_t, int>& a,
-                 const thrust::tuple<uint64_t, int>& b) {
-                return thrust::get<0>(a) < thrust::get<0>(b);
-              });
+  stable_sort(
+      policy, ids.begin(), ids.end(),
+      [&edge](const int& a, const int& b) { return edge[a] < edge[b]; });
+
```
pca006132, line 650: I am curious if we can skip the `edge` tmp buffer completely, and compute the edge information on-the-fly. Maybe we can try a benchmark later, not critical.

pca006132 (commented):
Not many comments here, I just had a glimpse over it. The new code is easier to read, which is always a nice bonus when we refactor our code :)

Will work on implementing the transform iterator, as well as maybe give some simple benchmark result to see if this is having unexpected impact on performance considering we change those sort on zip iterators to sort + permute

pca006132:
And a general comment: maybe we should use `std::pair` whenever possible? I don't see the need to use `thrust::pair` now.

pca006132:
FYI: there is basically no performance different.

pca006132:
It seems that thrust is doing something non-trivial in their copy function, causing issues with our simple custom iterator implementations. I don't want to dig into the details of that as we will be getting rid of them later anyway, so I just switch to the sequential implementation for now if there is no pstl support.

src/manifold/src/impl.cpp
elalish: Yeah, I never liked that either; I couldn't find another way before, but if you have an idea, that would be great.

elalish:
Thank you, this looks excellent! Let's go ahead and merge this and remove thrust::pair and such in the follow-up.

pca006132:
actually that was done in another PR :)

src/manifold/src/impl.cpp
pca006132: FYI: I tried this, it is slightly slower.

Issue #787, opened 2024-04-11
memory leak when TBB and PSTL is enabled


rob-rb:
Great library! I love it.

The following code shows how memory is consumed. you need to pip install psutil first.

`from manifold3d import *
import random

def get_boolean_difference(n):
    sph = Manifold.sphere(1.5, n)
    tet = Manifold.tetrahedron()
    m = sph - tet
    mesh = m.to_mesh()
    return mesh.vert_properties, mesh.tri_verts

def memory():
    import psutil
    process = psutil.Process()
    return int(process.memory_info().rss*1e-6)

def test():
    for j in range(100):
        points, faces = get_boolean_difference(2500-j)
        m1 = memory()
        points = None
        faces = None
        print(memory(),m1)

if __name__ == "__main__":
    test()`

rob-rb:
I introduced some gc.collect's but no changes I provide as well the output of this script below:


from manifold3d import *
import random
import gc


def get_boolean_difference(n):
    sph = Manifold.sphere(1.5, n)
    tet = Manifold.tetrahedron()
    m = sph - tet
    mesh = m.to_mesh()
    return mesh.vert_properties, mesh.tri_verts

def memory():
    import psutil
    process = psutil.Process()
    return int(process.memory_info().rss*1e-6)

def test():
    for j in range(100):
        points, faces = get_boolean_difference(2500-j)
        m1 = memory()
        points = None
        faces = None
        m2 = memory()
        gc.collect()
        gc.collect()
        print(memory(),m2, m1)

if __name__ == "__main__":
    test()


52 52 120
53 53 122
54 54 122
54 54 122
63 63 131
59 59 128
59 59 128
62 62 131
66 66 134
68 68 136
70 70 138
69 69 137
75 75 143
75 75 143
73 73 141
76 76 144
84 84 151
85 85 152
86 86 154
81 81 148
91 91 158
92 92 159
91 91 158
91 91 158
98 98 166
99 99 166
98 98 165
102 102 169
106 106 173
106 106 173
107 107 174
113 113 180
112 112 179
117 117 184
115 115 182
114 114 181
121 121 188
122 122 188
122 122 189
120 120 187
131 131 197
130 130 196
129 129 195
129 129 195
136 136 202
138 138 204
135 135 201
137 137 203
144 144 210
146 146 212
145 145 211
142 142 208
150 150 216
151 151 217
150 150 216
152 152 217
159 159 224
159 159 225
160 160 225
160 160 226
166 166 232
[...]

PR #856, opened 2024-07-07
Bye-bye thrust

### LINKED ISSUE #809: Remove Thrust

pca006132:
Fixes #809 

Basically #823 but based on #852.


pca006132:
OK thrust is now gone in this branch. We have some performance penalty, but can probably fix that via some tuning? idk

elalish:
I always find it annoying how Git sees one PR built on top of another PR as a giant pile of merge conflicts... You'd think they could resolve this automatically when the top commits have been merged to master.

pca006132:
fixing now

src/manifold/src/impl.cpp
```diff
@@ -539,11 +539,21 @@ void Manifold::Impl::RemoveUnreferencedVerts(Vec<glm::ivec3>& triVerts) {
...
+  auto vertIdIter = TransformIterator(countAt(0_z), [&vertOld2New](size_t i) {
+    if (vertOld2New[i + 1] > 0) return i;
+    return std::numeric_limits<size_t>::max();
+  });
+
+  auto next = copy_if<decltype(tmpBuffer.begin())>(
+      autoPolicy(tmpBuffer.size()), vertIdIter, vertIdIter + tmpBuffer.size(),
+      tmpBuffer.begin(),
+      [](size_t v) { return v != std::numeric_limits<size_t>::max(); });
+  gather(autoPolicy(tmpBuffer.size()), tmpBuffer.begin(), next,
```
elalish, line 553: Why do we now need a temporary buffer and an extra gather operation? I noticed the same pattern change up in boolean_result.

src/manifold/src/properties.cpp
```diff
@@ -63,30 +61,27 @@ struct PosMax
 };
 
 struct FiniteVert {
-  bool operator()(glm::vec3 v) { return glm::all(glm::isfinite(v)); }
+  bool operator()(glm::vec3 v) const { return glm::all(glm::isfinite(v)); }
```
elalish, line 64: We can probably just remove a bunch of these tiny functors. I've been turning them into lambdas as I come to them.

src/manifold/src/properties.cpp
```diff
@@ -387,9 +382,9 @@ void Manifold::Impl::CalculateBBox() {
  */
 bool Manifold::Impl::IsFinite() const {
   auto policy = autoPolicy(NumVert());
-  return transform_reduce<bool>(policy, vertPos_.begin(), vertPos_.end(),
-                                FiniteVert(), true,
-                                thrust::logical_and<bool>());
+  return reduce<bool>(policy, TransformIterator(vertPos_.begin(), FiniteVert()),
```
elalish, line 385: Why do you prefer `reduce` with a transform iterator to `transform_reduce`?

src/utilities/include/par.h
elalish, line 1: 💯

elalish (changes requested):
Looking great! Now I just need to make it compile in parallel on my M1 - I'll give that a try now.

elalish:
And just so we know where to look - where have you noticed a performance penalty?

src/manifold/src/properties.cpp
pca006132: Oh, I forgot there is such a thing. Will fix.

src/manifold/src/impl.cpp
pca006132: I think we need this for `copy_if` with stencil? Not sure how to do it otherwise.

src/manifold/src/properties.cpp
pca006132: Agreed. Will fix.

src/manifold/src/properties.cpp
elalish: I think I have this change in a commit of #823 if you want to grab it from there.

pca006132:
I need some more time to do rigorous profiling. I think I just noticed the unit test time increased by a bit, not much though.

pca006132:
After running the tests again, it seems that we are roughly 4% faster, not slower, regardless of whether `MANIFOLD_DEBUG` is on or off. Maybe my feeling that this is slower is due to some random thermo thing for my computer.

src/manifold/src/impl.cpp
elalish: Interesting, yeah I missed that. Fair enough, I can't think of a better way either.

pca006132:
and we probably want to add the cmake config needed to check pstl version? for #787.

elalish:
Okay, I've got OneDPL linked in now and I'm getting the same compiler problem on the CI as my local Mac, so that should aid with debug a bit. I have no idea what this means:
```
/opt/homebrew/include/oneapi/dpl/internal/../pstl/glue_numeric_impl.h:43:12: error: call to function 'transform_reduce' that is neither visible in the template definition nor found by argument-dependent lookup
    return transform_reduce(::std::forward<_ExecutionPolicy>(__exec), __first, __last, __init, __binary_op,
           ^
/Users/runner/work/manifold/manifold/src/utilities/include/par.h:132:21: note: in instantiation of function template specialization 'oneapi::dpl::reduce<const oneapi::dpl::execution::parallel_unsequenced_policy &, glm::vec<3, float> *, glm::vec<3, float>, (lambda at /Users/runner/work/manifold/manifold/src/manifold/src/properties.cpp:329:58)>' requested here
STL_DYNAMIC_BACKEND(reduce, void)
                    ^
/Users/runner/work/manifold/manifold/src/manifold/src/properties.cpp:327:15: note: in instantiation of function template specialization 'manifold::reduce<glm::vec<3, float>, glm::vec<3, float> *, glm::vec<3, float> *, glm::vec<3, float>, (lambda at /Users/runner/work/manifold/manifold/src/manifold/src/properties.cpp:329:58)>' requested here
  bBox_.min = reduce<glm::vec3>(
              ^
/opt/homebrew/include/oneapi/dpl/pstl/glue_numeric_defs.h:60:1: note: 'transform_reduce' should be declared prior to the call site or in an associated namespace of one of its arguments
transform_reduce(_ExecutionPolicy&& __exec, _ForwardIterator __first, _ForwardIterator __last, _Tp __init,
^
```
One the plus side, it is actually trying to use the PSTL from DPL, I just don't know what it doesn't like.

elalish:
I also don't understand all of the known limitations, or if we might be hitting any of them: https://oneapi-src.github.io/oneDPL/introduction.html#difference-with-standard-c-parallel-algorithms

pca006132:
are they just reduce and transform_reduce? can you get it working by changing them to sequential for the time being? just to see if oneDPL works.

elalish:
Good call, it *was* only `reduce` and `transform_reduce` - it's now running faster on my M1 than before this PR (~15% faster, not bad!). I suppose parallel reduce isn't too critical, but I'd love to understand the problem.

elalish:
Okay, now the Mac CI is just failing the cmake consumer test. What's the proper way to handle the OneDPL dependency? Do we link it in `test-cmake.sh`, or change something about the CMake install?

elalish:
[This](https://stackoverflow.com/questions/60242409/why-does-clang-require-a-function-to-be-declared-before-its-called-in-a-templat) is the closest I've come to useful info about the `reduce` error. It almost seems like it's a OneDPL bug? Unless there's something about our lambdas that are confusing it.

pca006132:
> [This](https://stackoverflow.com/questions/60242409/why-does-clang-require-a-function-to-be-declared-before-its-called-in-a-templat) is the closest I've come to useful info about the `reduce` error. It almost seems like it's a OneDPL bug? Unless there's something about our lambdas that are confusing it.

Feels like so. Maybe you can try to install oneDPL from source with the latest version and see if it works? I think we should probably file an upsteam issue if this persists.

Also, I wonder if we need PSTL at all. Most functions here are rather simple to implement with tbb `parallel_for` and `parallel_reduce`, the only exception is probably `stable_sort`?

pca006132:
and doing that will fix #787 as well.

pca006132:
@elalish things should work now.

src/utilities/CMakeLists.txt
```diff
@@ -39,10 +39,11 @@ if(MANIFOLD_PAR STREQUAL "TBB")
...
     endif()
     if(APPLE)
-      find_package(oneDPL)
-      if(oneDPL_FOUND)
-        target_link_libraries(${PROJECT_NAME} INTERFACE oneDPL)
-      endif()
+        if(oneDPL_FOUND)
+            target_link_libraries(${PROJECT_NAME} INTERFACE oneDPL)
+        else()
+            message(WARNING "oneDPL not found, sequential implementation is used instead")
```
elalish, line 45: 👍

elalish (approved):
Thank you! This is excellent; let's discuss the PSTL / parallel reduce stuff in a follow-up issue.

elalish:
>Also, I wonder if we need PSTL at all. Most functions here are rather simple to implement with tbb parallel_for and parallel_reduce, the only exception is probably stable_sort?

I see the appeal, but in the name of maintainability, PSTL feels safer long term. Maybe something better than TBB comes along, like TBB overtook OpenMP? Maybe new architectures come along that need a new parallel driver (e.g. Google TPUs)? I feel like PSTL will have the best chance of supporting all these unknown futures with minimal effort on our part. But that said, I'm open to being convinced otherwise.

PR #857, opened 2024-07-08
custom parallel algorithms based on tbb

### LINKED ISSUE #787: memory leak when TBB and PSTL is enabled

pca006132:
Most of the algorithms done. My major reasoning for this is that:

1. oneDPL seems like another huge dependency for mac users, while the algorithms we use can mostly be implemented with tbb.
2. I want to find a way to fix #787 without asking users to disable tbb.
3. It seems unlikely that (in the foreseeable) future accelerators will allow seamless interop with PSTL. If you look at how oneDPL and thrust works, they all require some specific device container/queues, which I think is unavoidable. So we are probably not losing much by ditching PSTL for now.
4. The implementations are very simple, with good performance. So far I haven't yet implemented `unique`, `uninitialized_fill`, `uninitialized_copy`, `stable_sort`. I don't think these are hard, just I probably need some sleep now. And the performance with my simple implementation for all the other algorithm is on-par with PSTL.
5. I always wanted to get rid of those explicit template specialization.

I think in the end `remove_if`, `unique`, `stable_sort` will require temporary storage buffer because they do in-place update, and I think all other parallel algorithm libraries does that anyway. Everything else can be parallelized.

I also took the time to remove some of the unused or (apparently) not useful algorithms here. E.g. `find`. Actually I am not sure if they are in a sorted array and maybe we can use avoid this generic find function? Probably does not matter much anyway...

### DIFF
--- a/src/manifold/src/boolean3.cpp
+++ b/src/manifold/src/boolean3.cpp
@@ -93,9 +93,9 @@ SparseIndices Filter11(const Manifold::Impl &inP, const Manifold::Impl &inQ,
                        const SparseIndices &p1q2, const SparseIndices &p2q1) {
   ZoneScoped;
   SparseIndices p1q1(3 * p1q2.size() + 3 * p2q1.size());
-  for_each_n(autoPolicy(p1q2.size(), 100'000), countAt(0_z), p1q2.size(),
+  for_each_n(autoPolicy(p1q2.size(), 1e5), countAt(0_z), p1q2.size(),
              CopyFaceEdges<false>({p1q2, p1q1, inQ.halfedge_, 0_z}));
-  for_each_n(autoPolicy(p2q1.size(), 100'000), countAt(0_z), p2q1.size(),
+  for_each_n(autoPolicy(p2q1.size(), 1e5), countAt(0_z), p2q1.size(),
              CopyFaceEdges<true>({p2q1, p1q1, inP.halfedge_, p1q2.size()}));
   p1q1.Unique();
   return p1q1;
@@ -263,7 +263,7 @@ std::tuple<Vec<int>, Vec<glm::vec4>> Shadow11(SparseIndices &p1q1,
   Vec<int> s11(p1q1.size());
   Vec<glm::vec4> xyzz11(p1q1.size());
 
-  for_each_n(autoPolicy(p1q1.size(), 10'000), countAt(0_z), p1q1.size(),
+  for_each_n(autoPolicy(p1q1.size(), 1e4), countAt(0_z), p1q1.size(),
              Kernel11({xyzz11, s11, inP.vertPos_, inQ.vertPos_, inP.halfedge_,
                        inQ.halfedge_, expandP, inP.vertNormal_, p1q1}));
 
@@ -355,7 +355,7 @@ std::tuple<Vec<int>, Vec<float>> Shadow02(const Manifold::Impl &inP,
   Vec<float> z02(p0q2.size());
 
   auto vertNormalP = forward ? inP.vertNormal_ : inQ.vertNormal_;
-  for_each_n(autoPolicy(p0q2.size(), 10'000), countAt(0_z), p0q2.size(),
+  for_each_n(autoPolicy(p0q2.size(), 1e4), countAt(0_z), p0q2.size(),
              Kernel02({s02, z02, inP.vertPos_, inQ.halfedge_, inQ.vertPos_,
                        expandP, vertNormalP, p0q2, forward}));
 
@@ -463,7 +463,7 @@ std::tuple<Vec<int>, Vec<glm::vec3>> Intersect12(
   Vec<glm::vec3> v12(p1q2.size());
 
   for_each_n(
-      autoPolicy(p1q2.size(), 10'000), countAt(0_z), p1q2.size(),
+      autoPolicy(p1q2.size(), 1e4), countAt(0_z), p1q2.size(),
       Kernel12({x12, v12, p0q2.AsVec64(), s02, z02, p1q1.AsVec64(), s11, xyzz11,
                 inP.halfedge_, inQ.halfedge_, inP.vertPos_, forward, p1q2}));
 
@@ -477,7 +477,7 @@ Vec<int> Winding03(const Manifold::Impl &inP, Vec<int> &vertices, Vec<int> &s02,
   ZoneScoped;
   // verts that are not shadowed (not in p0q2) have winding number zero.
   Vec<int> w03(inP.NumVert(), 0);
-  if (vertices.size() <= 100'000) {
+  if (vertices.size() <= 1e5) {
     for_each_n(ExecutionPolicy::Seq, countAt(0), s02.size(),
                [&w03, &vertices, &s02, reverse](const int i) {
                  w03[vertices[i]] += s02[i] * (reverse ? -1 : 1);
--- a/src/manifold/src/boolean3.cpp
+++ b/src/manifold/src/boolean3.cpp
@@ -93,9 +93,9 @@ SparseIndices Filter11(const Manifold::Impl &inP, const Manifold::Impl &inQ,
                        const SparseIndices &p1q2, const SparseIndices &p2q1) {
   ZoneScoped;
   SparseIndices p1q1(3 * p1q2.size() + 3 * p2q1.size());
-  for_each_n(autoPolicy(p1q2.size()), countAt(0_z), p1q2.size(),
+  for_each_n(autoPolicy(p1q2.size(), 100'000), countAt(0_z), p1q2.size(),
              CopyFaceEdges<false>({p1q2, p1q1, inQ.halfedge_, 0_z}));
-  for_each_n(autoPolicy(p2q1.size()), countAt(0_z), p2q1.size(),
+  for_each_n(autoPolicy(p2q1.size(), 100'000), countAt(0_z), p2q1.size(),
              CopyFaceEdges<true>({p2q1, p1q1, inP.halfedge_, p1q2.size()}));
   p1q1.Unique();
   return p1q1;
@@ -263,7 +263,7 @@ std::tuple<Vec<int>, Vec<glm::vec4>> Shadow11(SparseIndices &p1q1,
   Vec<int> s11(p1q1.size());
   Vec<glm::vec4> xyzz11(p1q1.size());
 
-  for_each_n(autoPolicy(p1q1.size()), countAt(0_z), p1q1.size(),
+  for_each_n(autoPolicy(p1q1.size(), 10'000), countAt(0_z), p1q1.size(),
              Kernel11({xyzz11, s11, inP.vertPos_, inQ.vertPos_, inP.halfedge_,
                        inQ.halfedge_, expandP, inP.vertNormal_, p1q1}));
 
@@ -355,7 +355,7 @@ std::tuple<Vec<int>, Vec<float>> Shadow02(const Manifold::Impl &inP,
   Vec<float> z02(p0q2.size());
 
   auto vertNormalP = forward ? inP.vertNormal_ : inQ.vertNormal_;
-  for_each_n(autoPolicy(p0q2.size()), countAt(0_z), p0q2.size(),
+  for_each_n(autoPolicy(p0q2.size(), 10'000), countAt(0_z), p0q2.size(),
              Kernel02({s02, z02, inP.vertPos_, inQ.halfedge_, inQ.vertPos_,
                        expandP, vertNormalP, p0q2, forward}));
 
@@ -463,7 +463,7 @@ std::tuple<Vec<int>, Vec<glm::vec3>> Intersect12(
   Vec<glm::vec3> v12(p1q2.size());
 
   for_each_n(
-      autoPolicy(p1q2.size()), countAt(0_z), p1q2.size(),
+      autoPolicy(p1q2.size(), 10'000), countAt(0_z), p1q2.size(),
       Kernel12({x12, v12, p0q2.AsVec64(), s02, z02, p1q1.AsVec64(), s11, xyzz11,
                 inP.halfedge_, inQ.halfedge_, inP.vertPos_, forward, p1q2}));
 
@@ -477,11 +477,17 @@ Vec<int> Winding03(const Manifold::Impl &inP, Vec<int> &vertices, Vec<int> &s02,
   ZoneScoped;
   // verts that are not shadowed (not in p0q2) have winding number zero.
   Vec<int> w03(inP.NumVert(), 0);
-  auto policy = autoPolicy(vertices.size());
-  for_each_n(policy, countAt(0), s02.size(),
-             [&w03, &vertices, &s02, reverse](const int i) {
-               AtomicAdd(w03[vertices[i]], s02[i] * (reverse ? -1 : 1));
-             });
+  if (vertices.size() <= 100'000) {
+    for_each_n(ExecutionPolicy::Seq, countAt(0), s02.size(),
+               [&w03, &vertices, &s02, reverse](const int i) {
+                 w03[vertices[i]] += s02[i] * (reverse ? -1 : 1);
+               });
+  } else {
+    for_each_n(ExecutionPolicy::Par, countAt(0), s02.size(),
+               [&w03, &vertices, &s02, reverse](const int i) {
+                 AtomicAdd(w03[vertices[i]], s02[i] * (reverse ? -1 : 1));
+               });
+  }
   return w03;
 };
 }  // namespace
--- a/src/utilities/include/par.h
+++ b/src/utilities/include/par.h
@@ -14,6 +14,7 @@
 
 #pragma once
 #if MANIFOLD_PAR == 'T'
+#include <tbb/combinable.h>
 #include <tbb/parallel_for.h>
 #include <tbb/parallel_invoke.h>
 #include <tbb/parallel_reduce.h>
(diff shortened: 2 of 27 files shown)


pca006132:
My `copy_if` and friends seems to be slower than the PSTL version. Will look at why later... About 4% performance impact for benchmark.

pca006132:
performance issue seems to be gone now, will check again later when I wake up.

src/utilities/include/par.h
```diff
@@ -49,47 +53,251 @@ inline constexpr ExecutionPolicy autoPolicy(size_t size) {
...
+  for_each(policy, first, first + n, f);
+}
+
+template <typename InputIter, typename BinaryOp,
+          typename T = typename std::iterator_traits<InputIter>::value_type>
+T reduce(ExecutionPolicy policy, InputIter first, InputIter last, T init,
+         BinaryOp f) {
+#if MANIFOLD_PAR == 'T'
+  if (policy == ExecutionPolicy::Par) {
+    // should we use deterministic reduce here?
```
elalish, line 82: I would tend to say no, but I suppose it depends how much performance penalty there really is. It still won't be identical between different architectures for floating-point answers.

src/utilities/include/par.h
```diff
@@ -49,47 +53,251 @@ inline constexpr ExecutionPolicy autoPolicy(size_t size) {
...
-  Ret NAME(ExecutionPolicy policy, Args... args) { \
-    return std::NAME(args...);                     \
+#endif
+  return std::reduce(first, last, init, f);
+}
+
+template <typename InputIter, typename BinaryOp, typename UnaryOp,
+          typename T = std::invoke_result_t<
+              UnaryOp, typename std::iterator_traits<InputIter>::value_type>>
+T transform_reduce(ExecutionPolicy policy, InputIter first, InputIter last,
```
elalish, line 99: Should we just build our auto execution policy into these wrappers so we don't need to define an `autoPolicy` all the time? We can just use e.g. `std::reduce` if we want to force a serial operation.

src/utilities/include/par.h
```diff
@@ -49,47 +53,251 @@ inline constexpr ExecutionPolicy autoPolicy(size_t size) {
...
+                                                   d_first);
+    tbb::parallel_scan(
+        tbb::blocked_range<size_t>(0, std::distance(first, last)), body);
+    return;
+  }
+#endif
+  std::exclusive_scan(first, last, d_first, init, f);
+}
+
+// TODO: use STL variant when parallelization is not enabled?
```
elalish, line 183: Looks like you already took care of this.

src/utilities/include/par.h
```diff
@@ -49,47 +53,251 @@ inline constexpr ExecutionPolicy autoPolicy(size_t size) {
...
+  size_t get_sum() const { return sum; }
+  void reverse_join(CopyIfScanBody &a) { sum = a.sum + sum; }
+  void assign(CopyIfScanBody &b) { sum = b.sum; }
+};
+#endif
+
+// note that you should not have alias between input and output...
+// in general it is impossible to check if there is any alias, as the input
+// iterator can be computed on-the-fly depending on the output position and may
+// not have a pointer
```
elalish, line 258: Might be a good idea to copy the Thrust requirements documentation into each of these wrappers - I know I find myself looking those up constantly when I use these, and it would be more convenient to have them here anyway.

src/utilities/include/par.h
elalish, line 1: Should we move your other `tbb::` usages in here as wrappers too for consistency? Kind of nice to have all references to TBB in one file.

elalish (commented):
Thank you for the clear argument - you've convinced me. This isn't as daunting as I had feared.

src/utilities/include/par.h
pca006132: Not completely. For example, for `transform_reduce`, I am calling `reduce` with transform iterator, which can be different from calling `std::transform_reduce`. Not sure about the performance impact of that, though.

src/utilities/include/par.h
pca006132: Sure.

src/utilities/include/par.h
pca006132: Yes, we can do that. This is usually just `std::distance`.

src/utilities/include/par.h
pca006132: This is a bit tricky, because our explicit usages of `tbb` doesn't quite fit into the typical PSTL paradigm, and moving them here may not be very helpful. Anyway, I can try to come up with useful abstraction for that.

pca006132:
Thinking about it, we can allow `copy_if` with stencil (the thrust extension) in our implementation here to avoid the need for allocation. And there are other things we can try to optimize for which I have not do yet, the implementations are about as simple as they can be for now.

pca006132:
Also I'm thinking if we should just update the `clang-format` version. The distro I use stopped supporting llvm11, switching to something like 15 should be good. It is annoying that `clang-format` does not provide backward compatibility options...

pca006132:
Note that `sort` is causing some performance issue here, but it is non-trivial to optimize.

src/utilities/include/par.h
pca006132: Thinking about it, we may still want execution policy for things like `for_each`, where each loop may be doing something non-trivial and will benefit from parallelization even when the number of iteration is small. Will think about how to find out all of those...

pca006132: but we can have an overload that automatically figure out the policy.

src/utilities/include/par.h
elalish: Agreed, `deterministic` feels more like a compile option to me. Honestly I'd prefer to get rid of it entirely. Might be good to talk about use cases and what we're really trying to accomplish with it.

src/utilities/include/par.h
elalish: If we've abandoned the PSTL, I think it's fine for us to make up some of our own functions too. This is more of a code organization thing; I don't think we need to be super picky about the shape of our internal functions.

src/utilities/include/par.h
pca006132: We probably want to ask the openscad folks. I think I added that option when someone complained about non-deterministic results and I thought we can have deterministic results even with parallelization.

Making it a compile-time config will probably be a bit too limiting, e.g. openscad cannot have that as optional feature setting.

elalish: @kintel Do you make use of this `deterministic` option? I don't love it, and I wonder if it's really necessary.

pca006132:
@elalish I think the performance related stuff are mostly finished. Maybe you can check if there is significant performance degradation on mac?

src/utilities/include/par.h
pca006132: Though for the record I think I *told* someone it can be used, I don't think openscad really added that option.

fire:
Can you summarize the options for parallelization? I have concerns that to add this into godot engine we'll require a tbb vendored dependency or go single threaded.

src/utilities/include/par.h
```diff
@@ -49,47 +43,430 @@ inline constexpr ExecutionPolicy autoPolicy(size_t size) {
...
+    size_t q3 = p3 + (q1 - p1) + (q2 - p2);
+    dest[q3] = src[q1];
+    tbb::parallel_invoke(
+        [=] { mergeRec(src, dest, p1, q1, p2, q2, p3, comp); },
+        [=] { mergeRec(src, dest, q1 + 1, r1, q2, r2, q3 + 1, comp); });
+  }
+}
+
+template <typename T, typename Comp>
+void mergeSortRec(T *src, T *dest, size_t begin, size_t end, Comp comp) {
```
pca006132, line 80: This parallel merge sort is really needed to get performance to comparable level. Sequential sort just cause a blowup in running time for `GenericTwinBooleanTest7081` from 7s to ~30s iirc. And tbb parallel sort is not stable, so it cannot be used.

Also, I know that this is far from optimal, we can probably do much better than this because the parallel `stable_sort` from pstl is faster than ours. But that is a problem for the other day.

pca006132:
We don't really need tbb for parallelization. Our parallelized algorithms here just rely on parallel for, invoke, reduce and scan. I think it is possible to do it in openmp, or even c++ standard thread (some custom thread pool implementation). What does godot engine use for thread pool?

fire:
We wrote a custom thread worker pool https://docs.godotengine.org/en/stable/classes/class_workerthreadpool.html

pca006132:
I think it should be possible to use custom thread pool and implement those parallel for/invoke/reduce/scan primitives. The performance will probably be worse than tbb, but should be better than nothing. I need to think a bit about how to wrap that in an API though.

elalish:
>We wrote a custom thread worker pool https://docs.godotengine.org/en/stable/classes/class_workerthreadpool.html

@fire So what would you like to see? My impression is that TBB is nice because it both uses multiple cores, but also the vector units within each CPU. Is there anything we can do to make it easier for you to hook into your custom threading system? Also keep in mind that going single threaded is currently only about a 2-3x performance hit, so depending on your use that may be sufficient (for instance, our WASM for ManifoldCAD.org is all single-threaded).

elalish:
@pca006132 I don't detect any significant difference in speed with this PR - it's still 15% faster than it was thanks to removing thrust.

fire:
The ideal case is we can use parallel c++17 instead of c++17 + tbb. I have doubts this is possible.

The fallback would be to use the single-threaded processing.

It's a research project to check if godot [class_workerthreadpool.html](https://docs.godotengine.org/en/stable/classes/class_workerthreadpool.html) will work for manifold.

pca006132:
Parallel C++ requires a backend. If you look at libstdc++ or libc++, they either use tbb or openmp, or libdispatch on mac.

elalish:
Even if you use parallel c++17, you still have to link in TBB to get the compiler to actually run `par_unseq`. This is why we consider the TBB dependency to be better than Thrust - it's what the STL maintainers depend on. Does that ease your concern? We already dropped our Thrust dependency - this PR also removes our oneDPL dependency (which was only needed for Mac), leaving us with just c++17 and TBB.

elalish:
@fire Okay, then I think we've finished your requested changes. Thrust and OpenMP are gone and we can build with exceptions disabled again now. I think TBB is intended to be used as a system library (like the STL), and it appears to be available on every system, so there should be no need to vendor it. If you ever decide that's acceptable, you should be able to get the parallel speed up too.

pca006132:
I made `GenericTwinBooleanTest7081` ~10% faster by adding my previous radix sort with parallel merge.

pca006132:
@elalish I added a benchmark comparing our parallel algorithm against stl sequential implementation and look at the speedup. From this we can probably determine a good threshold for when to switch to the parallel implementation. You can run `./extras/stlTest` to perform the tests, and show me the results to see if Mac is special.

Here are my results:
```
                               1K   10K  100K    1M   10M 
all_of                       0.02  0.29  0.84  2.41  2.95 
for_each                     0.03  0.19  0.68  2.14  1.95 
count                        1.03  0.89  0.95  2.03  2.56 
copy                         1.75  0.91  0.81  1.83  1.15 
copy_if_28                   0.64  1.76  1.07  2.08  4.23 
copy_if_55                   1.16  1.76  1.05  2.72  6.41 
copy_if_82                   0.65  2.38  1.08  2.38  3.79 
copy_if_1001                 0.44  0.91  0.98  1.94  1.57 
fill                         0.04  0.16  0.61  2.18  2.97 
sort_Rnd                     1.03  0.93  4.09 11.19 10.80 
sort_Rnd_size                2.59  2.23  7.05 17.24 13.59 
sort_Rnd_size_merge          1.01  1.02  4.12 11.64  9.76 
sort_Eq                      1.03  0.97  4.50 13.49  9.13 
sort_Eq_size_merge           0.98  0.96  3.59 11.60  8.05 
sort_Asc                     1.13  1.03  4.39 12.02  8.96 
sort_Asc_size_merge          1.03  0.99  3.50 11.00  7.88 
sort_roughly_Asc             1.02  1.00  4.46 11.91 11.04 
sort_roughly_Asc_size        3.06  2.01  5.43 16.48 10.70 
sort_roughly_Asc_size_merge  1.14  0.99  4.54 11.88  9.90 
sort_Des                     0.98  1.01  4.53 12.83  9.04 
sort_Des_size                2.09  2.28  7.86 14.63  8.45 
sort_Des_size_merge          0.99  1.00  4.50 11.45  8.51 
reduce                       0.83  1.02  1.40  2.57  3.04 
transform_reduce             0.75  0.91  1.27  2.64  2.99 
exclusive_scan               0.14  0.76  1.69  1.36  1.83 
```

They are compared across different workload sizes, different input patterns for sort and random input for copy (to avoid our smart hardware/glibc to optimize things into fills). The numbers represent speedup/slowdown with respect to the sequential version when `ExecutionPolicy::Seq` is passed into the function.

src/utilities/include/par.h
pca006132: The benchmark I added may help this as well. I think the threshold may be quite different for each algorithm.

src/utilities/include/par.h
```diff
@@ -529,11 +526,13 @@ template <typename Iter, typename P,
...
-    auto back = copy_if(policy, first, last, tmp.begin(),
-                        [&](T v) { return !pred(v); });
-    copy(policy, tmp.begin(), back, first);
-    return first + std::distance(tmp.begin(), back);
+    T *tmp = new T[std::distance(first, last)];
+    auto back =
+        copy_if(policy, first, last, tmp, [&](T v) { return !pred(v); });
+    copy(policy, tmp, back, first);
+    auto d = std::distance(tmp, back);
+    delete[] tmp;
```
elalish, line 534: Do we have any other usages of `new` or `delete` in this code base? Why is this better than a vector?

elalish (commented):
My Mac M1 results:
```
                               1K   10K  100K    1M   10M 
all_of                       0.07  0.24  1.62  4.19  4.94 
for_each                     0.03  0.19  0.51  1.36  1.65 
count                        0.63  1.05  1.40  1.34  2.15 
copy                         0.56  0.88  0.66  0.92  1.68 
copy_if_28                   1.27  0.98  1.04  2.73  5.01 
copy_if_55                   1.74  0.97  1.39  3.23  5.15 
copy_if_82                   1.16  0.91  1.15  2.47  4.37 
copy_if_1001                 0.55  0.54  0.76  1.50  2.56 
fill                         0.02  0.13  0.87  1.21  1.91 
sort_Rnd                     1.03  0.88  4.69  7.87  8.15 
sort_Rnd_size                2.67  2.56  7.62 14.14 17.29 
sort_Rnd_size_merge          1.00  0.92  4.99  7.62  7.92 
sort_Eq                      0.80  0.50  1.94  3.41  3.78 
sort_Eq_size_merge           0.76  0.41  1.50  3.78  3.94 
sort_Asc                     0.80  0.38  1.09  3.88  3.84 
sort_Asc_size_merge          0.80  0.68  2.36  3.70  3.83 
sort_roughly_Asc             0.83  0.48  1.64  4.66  4.76 
sort_roughly_Asc_size        0.75  0.75  1.85  3.86  4.73 
sort_roughly_Asc_size_merge  0.89  0.71  1.76  4.65  5.30 
sort_Des                     0.79  0.51  1.33  3.79  4.27 
sort_Des_size                2.42  1.95  6.74 10.31  6.53 
sort_Des_size_merge          1.03  0.68  4.31  6.15  5.71 
reduce                       0.98  1.00  6.48  6.38  7.90 
transform_reduce             0.91  0.99  3.80  4.95  7.66 
exclusive_scan               0.74  0.57  1.14  2.76  3.12 
```

src/utilities/include/par.h
pca006132: The idea is to avoid initializing the memory. But it is probably premature optimization...

pca006132:
I have to say that the results from M1 looks beautiful: they make more sense than what I got from my poor x86-64 system, which are sometimes quite counterintuitive (to my intuition anyway).

src/utilities/include/par.h
pca006132: nvm, I did some benchmark, this is not premature. It does matter for some of the benchmarks.

elalish: I sort of thought this was a feature of `Vec` - that `Vec<int> a(10)` was uninitialized and `Vec<int> a(10, 0)` was initialized to 0. Should it be?

pca006132: Yes, but I would rather avoid introducing dependency to `Vec` here because `Vec` also use some parallel algorithms (copy and fill).

pca006132:
Did some little tuning, the threshold I wrote for `for_each` are just gut feeling... the performance is slightly better on my system, but not by much.

src/manifold/src/boolean3.cpp
```diff
@@ -93,9 +93,9 @@ SparseIndices Filter11(const Manifold::Impl &inP, const Manifold::Impl &inQ,
                        const SparseIndices &p1q2, const SparseIndices &p2q1) {
   ZoneScoped;
   SparseIndices p1q1(3 * p1q2.size() + 3 * p2q1.size());
-  for_each_n(autoPolicy(p1q2.size()), countAt(0_z), p1q2.size(),
+  for_each_n(autoPolicy(p1q2.size(), 100'000), countAt(0_z), p1q2.size(),
```
elalish, line 96: Can we make the policy not require e.g. `p1p2.size()`? I feel like the algorithm already has enough info to determine that. This is such a common pattern, it would be great to reduce it as much as possible. Could we make it work with e.g. `for_each_n({1e5}, countAt(0), ...)`?

src/manifold/src/boolean3.cpp
```diff
@@ -93,9 +93,9 @@ SparseIndices Filter11(const Manifold::Impl &inP, const Manifold::Impl &inQ,
...
              CopyFaceEdges<false>({p1q2, p1q1, inQ.halfedge_, 0_z}));
-  for_each_n(autoPolicy(p2q1.size()), countAt(0_z), p2q1.size(),
+  for_each_n(autoPolicy(p2q1.size(), 100'000), countAt(0_z), p2q1.size(),
```
elalish, line 98: I never knew you could write `100'000`; it's cool, but I think I still prefer `1e5` if you don't mind.

elalish (approved):
Let's not go any deeper on optimization in this PR - we need to merge this. What do you think is left before you're ready?

src/manifold/src/boolean3.cpp
pca006132: yes, I already updated things in par.h, should be easy to modify these as well.

src/manifold/src/boolean3.cpp
pca006132: Yeah this is fine. I just saw this method in the pstld repo and thought it was cool :P

pca006132:
I think it should be good after addressing your two comments above.

pca006132:
@elalish maybe I will leave the execution policy change (just input a size) for later. It can be done but requires a lot of changes.

pca006132:
@elalish should I merge this now?

src/utilities/include/par.h
```diff
@@ -592,15 +803,41 @@ OutputIter copy_if(ExecutionPolicy policy, InputIter first, InputIter last,
...
+// starting from `d_first` that satisfies the predicate `pred`, i.e. `pred(x) ==
+// true`, and returns `d_first + n` where `n` is the number of times the
+// predicate is evaluated to true.
+//
+// This function is stable, meaning that the relative order of elements in the
+// output range remains unchanged.
+//
+// The input range `[first, last)` and
+// the output range `[d_first, d_first + last - first)`
+// must be equal or non-overlapping.
```
elalish, line 816: Does this really work in-place (equal)? And if so, does `remove_if` really need that `tmp` vector?

elalish (approved):
Yeah, this looks good, thanks. So I think future work is simplifying the `autoPolicy` and moving the other `tbb::` functions into `par.h`, right?

src/utilities/include/par.h
pca006132: good cach, I copied the wrong lines.

pca006132:
Yup

elalish:
Go ahead and merge when you're ready.

PR #861, opened 2024-07-10
Avoid Wundef trigger

starseeker:
When the Manifold headers are included in BRL-CAD, one of our compile flags doesn't like MANIFOLD_EXCEPTIONS being undefined - if I'm not mistaken, we can just check for the definition here, without worrying about its specific value.


elalish (approved):
Indeed, that looks like a typo to me.

Issue #868, opened 2024-07-18
operator""_z returns unsigned and is in the global scope


BenFrantzDale:
At https://github.com/elalish/manifold/blob/0617d344bcccb5b603d93e94111d6e6a4ea5e6b5/src/utilities/include/public.h#L34 it defines `constexpr std::size_t operator""_z(unsigned long long n) { return n; }`. 

1. This is inconsistent with C++23's (P0330R8)[https://wg21.link/P0330R8] operator""z, which returns `std::make_signed_t<std::size_t>`. 
2. This is presumptuous to put this in the global scope; it should go in `namespace manifold` (and possibly also in something like `namespace manifold::literals`).

Proposed solution:
Rename to `operator""_uz` and move into `namespace manifold`.

PR #869, opened 2024-07-18
Rename operator""_z to _uz per P0330R8, and move into namespace manifold

BenFrantzDale:
https://github.com/elalish/manifold/issues/868

I don't see a `public_test.cpp` anywhere. Is there somewhere I should put the test for this?

### DIFF
--- a/src/manifold/src/boolean_result.cpp
+++ b/src/manifold/src/boolean_result.cpp
@@ -155,7 +155,7 @@ std::tuple<Vec<int>, Vec<int>> SizeOutput(
   outR.faceNormal_.resize(numFaceR);
 
   Vec<size_t> tmpBuffer(outR.faceNormal_.size());
-  auto faceIds = TransformIterator(countAt(0_z), [&sidesPerFacePQ](size_t i) {
+  auto faceIds = TransformIterator(countAt(0_uz), [&sidesPerFacePQ](size_t i) {
     if (sidesPerFacePQ[i] > 0) return i;
     return std::numeric_limits<size_t>::max();
   });
@@ -168,7 +168,7 @@ std::tuple<Vec<int>, Vec<int>> SizeOutput(
          outR.faceNormal_.begin());
 
   auto faceIdsQ =
-      TransformIterator(countAt(0_z), [&sidesPerFacePQ, &inP](size_t i) {
+      TransformIterator(countAt(0_uz), [&sidesPerFacePQ, &inP](size_t i) {
         if (sidesPerFacePQ[i + inP.faceNormal_.size()] > 0) return i;
         return std::numeric_limits<size_t>::max();
       });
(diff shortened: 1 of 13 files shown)


src/utilities/include/public.h
```diff
@@ -31,10 +31,13 @@
...
 #endif
 
-constexpr std::size_t operator""_z(unsigned long long n) { return n; }
-
 namespace manifold {
 
+/**
+ * Stand-in for C++23's operator""uz (P0330R8)[https://wg21.link/P0330R8].
+ */
+[[nodiscard]] constexpr std::size_t operator""_uz(unsigned long long n) noexcept { return n; }
```
elalish, line 39: This caused me to look up `[[nodiscard]]` - that's a nice language feature, and one we should probably use more. Thanks!

elalish (approved):
Thank you, both for noticing and quickly fixing! Signing the individual CLA should be a one-click automatic thing. Once those last two checks turn green we'll merge this.

src/utilities/include/public.h
BenFrantzDale: Yeah, I'm in the habit of being very pedantic about that and using it nearly everywhere.

src/manifold/src/boolean_result.cpp
```diff
@@ -255,7 +255,7 @@ void AddNewEdgeVerts(
         [&](size_t hash) { mutexes[hash % mutexes.size()].unlock(); },
         std::placeholders::_1);
     tbb::parallel_for(
-        tbb::blocked_range<size_t>(0_z, p1q2.size(), 32),
+        tbb::blocked_range<size_t>(0_uz, p1q2.size(), 32),
```
BenFrantzDale, line 258: FWIW, `tbb::blocked_range` will inver its type, so `tbb::blocked_range(0_uz, p1q2.size(), 32),` should work just as well.

src/utilities/include/public.h
elalish: Improvements like that are always welcome, if you're in the mood to improve our API.

src/utilities/include/par.h
```diff
@@ -598,13 +598,13 @@ void transform(ExecutionPolicy policy, InputIter first, InputIter last,
...
-                                       first + range.end(),
-                                       d_first + range.begin(), f);
-                      });
+    tbb::parallel_for(
+        tbb::blocked_range<size_t>(
+            0_uz, static_cast<size_t>(std::distance(first, last))),
+        [&](const tbb::blocked_range<size_t> &range) {
+          std::transform(first + range.begin(), first + range.end(),
+                         d_first + range.begin(), f);
+        });
```
BenFrantzDale, line 607: I believe this can be
```cpp
    tbb::parallel_for(
        tbb::blocked_range(first, last),
        [&](const auto& range) {
          std::transform(range.begin(), range.end(),
                         d_first + std::distance(first, range.begin()), f);
        });
```

src/manifold/src/boolean_result.cpp
elalish: 👍 good fodder for a follow-on

src/utilities/include/public.h
pca006132: Nice to know that this is available with c++17. Thought these attributes are for c++20.

PR #871, opened 2024-07-21
Compile time optimization

pca006132:
Not very sure about this one, I guess this depends on how far we want to go for optimizing compile time.

I wanted to optimize some collider related performance issue, which requires passing functions inside the collider loop. However, I recall that the reason we put collider into `.cpp` file instead of using templates in the header function is due to compile time concern, which was pretty painful when we compiled using nvcc. We don't use nvcc now, and I wanted to optimize the build time a bit more so we can have more headroom to make it slower :)

Changes:
1. Use PCH for expensive headers. Requires minimal changes, reduces compile time slightly:
    - Total: 221s to 190s (enabled virtually everything and does LTO).
    - Minimal total: 93s to 76s (disabled virtually everything)
    - Wall clock time: 17s to 14.6s.
    - Minimal wall clock time: 7.4s to 7.2s.
2. Enable cmake unity build. This is a pretty invasive change that due to our use of anonymous namespaces. This reduces the total compile time, but wall clock time is slower than typical builds due to less parallelization (clang should try to do parallel codegen...).
    - Total: 115s (cannot directly compare as we did not have unity build before)
    - Minimal total: 38.5s (disabled virtually everything).
    - Wall clock time: 22s.
    - Minimal wall clock time: 13.3s.


elalish:
Huh, so the total time is the time for each CPU added together? Seems like wall clock time is what we care about, right? So if this is making total time 15% better, but wall clock time 30-50% worse, is that good? You also mentioned making collider into a header - are you still planning to do that? What difference would it make?

I've never heard of a Unity build before, but it looks like it's just concatenating several cpp files for compilation (I sort of thought that would take more, rather than less time, but goes to show how much I know about the compilation process). Still, another reason I like a bunch of separate files is that generally re-compiling is quicker, since I'm often only touching one of those files. How does this change affect re-compilation?

pca006132:
Yes, total time is the time for each CPU added together. I think it depends on what we are optimizing for, either the developer time or CI time, or both (some devs may have less beefy machines with less cores, idk).

Unity build is not really meant for development, it is for optimizing one-off compilation. And yes, it makes recompilation pretty slow. I did it just because I want to see how much a different it makes, and I am happy to drop that change.

I still plan to make collider into header, will see how it affects compile time. I don't expect there to be a huge increase in compile time because it is still small comparing to things like stl algorithms. We can usually optimize compiler frontend time (parsing) by using PCH. I don't think these functions will be instantiated many times (< 10 across the whole library I think), so backend time should be fine.

elalish:
Okay, that sounds reasonable. Let's see how it looks without the Unity part. I think I'd rather optimize for dev experience than for the CI.

pca006132:
OK I checked, no measurable compile time difference: 20.2s for header-only collider, 19.5s for the current version.

pca006132:
And I found that we should not expose collider to users: it uses some of our internal APIs like `vec`, which are not exposed to users.

pca006132:
the cmake error can probably be fixed by the build interface thing, will check later tmr

pca006132:
@elalish should we merge this now?

elalish (approved):
Thanks, looks good!

PR #875, opened 2024-07-22
Testing Framework for Hull

Kushal-Shah-03:
@elalish  and @pca006132, I've cleaned the code for the testing framework, keeping the tests of a High Quality Sphere, Menger Sponge and the script to run the code on the Thingi10K dataset, and then get the statistics for the implementations. I've attached instructions as to how to run the code in the comments. I would appreciate it if you could review the code once as well.

As for instructions: 
To run the executable  (run this in ./build directory)
Usage: ./test_hull_performance <Test (Sphere/Menger/Input)> <Implementation (Hull/Hull_CGAL)> <Print Header (1/0)> <Input Mesh (filename)> 
Example: ./extras/testHullPerformace Menger Hull 1
The Input test allows the user to enter a custom file to run the implementation on. You can test our current implementation as well as CGAL implementation. The Print Header refers to printing the corresponding header for the csv file (it's mainly used by the script for the dataset, for single testing purposes just use 1 there)

To run the script for the dataset (run this is ./extras directory)
Usage : ./run.sh ../build/extras/testHullPerformance {path_to_dataset_folder} {name_of_csv} {implementation(Hull,Hull_CGAL)}
example : ./run.sh ../build/extras/testHullPerformance  ./Thingi10K/raw_meshes/ Hull.csv Hull


Kushal-Shah-03:
I've tried my best to include only the relevant things from our experiments, if there's something I missed or something I should remove. I would gladly make the change.

bindings/python/manifold3d.cpp
```diff
@@ -107,7 +107,7 @@ struct nb::detail::type_caster<glm::mat<C, R, T, Q>> {
   static handle from_cpp(glm_type mat, rv_policy policy,
                          cleanup_list *cleanup) noexcept {
     T *buffer = new T[R * C];
-    nb::capsule mem_mgr(buffer, [](void *p) noexcept { delete[] (T *)p; });
+    nb::capsule mem_mgr(buffer, [](void *p) noexcept { delete[](T *) p; });
```
elalish, line 110: This is due to using the wrong version of clang-format. Make sure you've pulled master.

extras/test_hull_performace.cpp
```diff
@@ -0,0 +1,175 @@
...
+  std::cout << CGAL::Polygon_mesh_processing::area(cgalMesh) << ",";
+  std::cout << CGAL::Polygon_mesh_processing::area(cgalHull) << ",";
+  std::cout << std::distance(cgalMesh.faces_begin(), cgalMesh.faces_end())
+            << ",";
+  std::cout << std::distance(cgalHull.faces_begin(), cgalHull.faces_end())
+            << ",";
+  return;
+}
+
+void PrintManifold(Manifold input, Manifold (Manifold::*hull_func)() const) {
```
pca006132, line 93: Why should this take a manifold method? I think you can just make it take a generic function that returns a manifold given a set of points.

extras/test_hull_performace.cpp
```diff
@@ -0,0 +1,175 @@
...
+    ExportMesh("testHullManifold.glb", hullManifold.GetMesh(), {});
+  }
+#endif
+  auto end = std::chrono::high_resolution_clock::now();
+  PrintVolArea(input, hullManifold);
+  std::chrono::duration<double> elapsed = end - start;
+  std::cout << elapsed.count() << " sec";
+}
+
+void PrintCGAL(Manifold input, Manifold (Manifold::*hull_func)() const) {
```
pca006132, line 107: We probably don't need a separate CGAL one as well.

extras/run.sh
```diff
@@ -0,0 +1,41 @@
+#!/bin/bash
+
+# Go to manifold/extras directory and run the command as `./run.sh ../build/extras/testHullPerformance {path_to_dataset_folder} {name_of_csv} {implementation(Hull,Hull_CGAL)}`
```
pca006132, line 3: Ideally, let the code take one file, and use the shell script to iterate over everything in the folder and write to a csv. This way we can quickly run the tool with a single file.

pca006132:
I think we also talked about checking if the convex hull result is a valid convex polyhedron? I think that is a pretty important thing to test for.

extras/test_hull_performace.cpp
```diff
@@ -0,0 +1,175 @@
...
+    SphereTestHull(&Manifold::Hull, argv[2]);
+  else if (!strcmp(argv[1], "Menger"))
+    MengerTestHull(&Manifold::Hull, 1, 2, 3, argv[2]);
+  else if (!strcmp(argv[1], "Input")) {
+    auto inputMesh = ImportMesh(argv[4], 1);
+    Manifold inputManifold = Manifold(inputMesh);
+    if (!strcmp(argv[2], "Hull"))
+      PrintManifold(inputManifold, &Manifold::Hull);
+    else if (!strcmp(argv[2], "Hull_CGAL"))
+      PrintCGAL(inputManifold, &Manifold::Hull);
```
elalish, line 173: It looks like you're always using `Manifold::Hull()` - do we still need this input?

extras/test_hull_performace.cpp
```diff
@@ -0,0 +1,175 @@
...
+  if (!strcmp(implementation, "Hull"))
+    PrintManifold(sponge, hull_func);
+  else if (!strcmp(implementation, "CGAL"))
+    PrintCGAL(sponge, hull_func);
+}
+
+// Constructs a high quality sphere, and tests the convex hull implementation on
+// it (you can pass the specific hull implementation to be tested). Comparing
+// the volume and surface area with CGAL implementation
+void SphereTestHull(Manifold (Manifold::*hull_func)() const,
```
elalish, line 141: We talked about how this one can test our precision by checking if the number of verts is the same between input and output.

CMakeLists.txt
```diff
@@ -46,6 +46,7 @@ endif()
 option(TRACY_ENABLE "Use tracy profiling" OFF) 
 option(TRACY_MEMORY_USAGE "Track memory allocation with tracy (expensive)" OFF)
 option(BUILD_TEST_CGAL "Build CGAL performance comparisons" OFF)
+option(TEST_HULL_PERFORMANCE "Test Hulling Algorithm" OFF)
```
elalish, line 49: I think I'd rather build this under the `TEST_CGAL` flag instead of adding another.

extras/test_hull_performace.cpp
```diff
@@ -0,0 +1,175 @@
...
+  PrintVolAreaCGAL(cgalInput, cgalHull);
+  std::chrono::duration<double> elapsed = end - start;
+  std::cout << elapsed.count() << " sec";
+}
+
+// Constructs a Menger Sponge, and tests the convex hull implementation on it
+// (you can pass the specific hull implementation to be tested). Comparing the
+// volume and surface area with CGAL implementation, for various values of
+// rotation
+void MengerTestHull(Manifold (Manifold::*hull_func)() const, float rx, float ry,
```
elalish, line 128: This is fine for comparison to CGAL, but I'm actually more interested in adding to our own testing framework. Can you build tests like this directly into `hull_test.cpp` for just Manifold's implementation?

elalish (changes requested):
We don't need to bother with Draft PRs - our CI will only run on regular ones.

extras/run.sh
Kushal-Shah-03: The code takes the folder containing the files as an input and iterates over all the files in that folder. This is even simpler than specifying a file right? I can remove name of the csv parameter, if needed. The reason I kept implementation as an option is so that the user doesn't have to go and change the code to compare different implementations.

extras/test_hull_performace.cpp
Kushal-Shah-03: Oh right, I forgot that we didn't need input of the hull function there, as we're using CGAL, I'll make that change, thanks for pointing it out. And the reason I am taking this input, is that we can run a comparison between various implementations (or for the bare minimum at least with CGAL).

CMakeLists.txt
Kushal-Shah-03: Alright, I will change that.

extras/test_hull_performace.cpp
Kushal-Shah-03: So we just want to keep this testing script for the tests on datasets, and these single tests, I can shift to `hull_test.cpp`?

bindings/python/manifold3d.cpp
Kushal-Shah-03: I will try it again, thanks for pointing it out.

extras/test_hull_performace.cpp
Kushal-Shah-03: The reason, I kept it, is since we had discussed that we don't want to keep the other implementations in our code. CGAL, seemed like the simplest way to compare the outputs relatively.

extras/run.sh
pca006132: You can do the iteration in the bash script.

```bash
for f in dataset/*; do
  ../build/extras/testHullPerformance $f Hull >> output.csv
done
```

Something like this.

extras/test_hull_performace.cpp
pca006132: I mean, we can just use the function for `PrintManifold`? Just run CGAL hull in the `hull_func` function

Kushal-Shah-03: Right, since we don't need the Manfiold Method function, we can just use that. I will change that too. Now, that I think of it, that was a very redundant piece of code thanks for pointing it out.

extras/run.sh
Kushal-Shah-03: So we just assume that the input files that we wanna test will always be placed in the same dataset folder, and thus no need to take the folder for dataset as input right?

extras/test_hull_performace.cpp
Kushal-Shah-03: Ok so, since we aren't storing the output of CGAL as a manifold, I am just computing the CGAL output, and then printing the outputs using the CGAL functions. I didn't want to convert the CGAL output back to manifold as that would add time overhead, so what should I do in this case?

bindings/python/manifold3d.cpp
Kushal-Shah-03: Could you guide me on how to fix this, I had used the latest commit from master on the main repository.

extras/test_hull_performace.cpp
pca006132: The simplest way would be to do the time measurement in the callback, or you can split the function into two, the actual hull operation as well as a postprocessing operation which returns a manifold, and you only measure the time in the actual hull operation.

```cpp
class HullImpl {
  public:
    // actual hull operation, we will measure the time needed to evaluate this function.
    virtual void hull(const std::vector<glm::vec3>& points);
    // return the result of the hull operation, should be called after `hull`.
    virtual manifold::Manifold result();
};
```

And just let the implementations inherit this class. (or if you use template, you don't need inheritance).

extras/run.sh
pca006132: No, the shell script expects the same dataset folder, but the actual binary just takes a single file as input.

bindings/python/manifold3d.cpp
pca006132: What clang-format version are you using? Use a newer one, e.g. 16.

Kushal-Shah-03:
I've made the changes as suggested, except adding the check for the valid convex hull, since I was debating whether we wanted that in the code with the implementation that we're moving to our library or in this testing framework, what's your opinion on it?

pca006132:
Just ping me know when I should review it.

Kushal-Shah-03:
> Just ping me know when I should review it.

I'm currently working on the reduced points cases that Emmett had suggested, but I'm having difficulties with it, I've added the check for valid convex hull. If you plan to start the refactoring of the code soon, I can commit this much for now. And we can add these failing cases later?

pca006132:
No hurry, I just asked because I saw the conflicts are now resolved and tests seem to be passing. You can do it later.

elalish:
I think I'd like to land this PR before #881, since it'd good to have regression tests in place for a big change like that. What troubles are you having with the other test cases? I would expect some to fail for now (you can mark them as e.g. `TEST(Hull, DISABLED_HardTest)`. You can reenable them with your fixes in #881.

Kushal-Shah-03:
> I think I'd like to land this PR before #881, since it's good to have regression tests in place for a big change like that. What troubles are you having with the other test cases? I would expect some to fail for now (you can mark them as e.g. `TEST(Hull, DISABLED_HardTest)`. You can reenable them with your fixes in #881.

I was having issues with the recreating the error, with the points values (not storing them in binary). I am working on that now, I hope to resolve it soon and I will add the tests as you suggested. I will focus on this PR then, considering what we had planned for #881 is pretty much finished

Kushal-Shah-03:
I've added the test for the valid convex hull in `test_hull_performance.cpp`, and  the two cases for the reduced points in the file `hull_test.cpp`. What do we want to test for the currently failing tests, whether they are giving correct volume and area, or whether the output is a valid convex hull? Depending on that I will add the check for convex hull in `hull_test.cpp` if needed. 

Also, for checking if for any input file the output hull is valid or not, I'm using the MANIFOLD_DEBUG flag for now, should I change that in any way, and do you want that function to compile only when that flag is enabled?

test/hull_test.cpp
```diff
@@ -94,3 +94,54 @@ TEST(Hull, Sphere) {
   EXPECT_FLOAT_EQ(sphereHull.GetProperties().volume,
                   sphere.GetProperties().volume);
 }
+
+// TEST(Hull, FailingTest1) {
```
elalish, line 98: Instead of commenting these tests, you can change them to `TEST(Hull, DISABLED_FailingTest1)` - this way they get compiled, but just not run, and we get a warning to remind us.

test/hull_test.cpp
```diff
@@ -75,4 +75,73 @@ TEST(Hull, Empty) {
...
+//       {13.26748085f, -19.979951859f, 28.117856979f},
+//       {-18.286884308f, 31.673814774f, 2.1749999523f},
+//       {18.419618607f, -18.215343475f, 52.450099945f},
+//       {-24.983196259f, 43.272167206f, 52.710712433f},
+//       {-1.6232370138f, -29.794223785f, 48.394889832f},
+//       {49.865573883f, -0.0f, 55.507141113f},
+//       {-18.627283096f, -39.544368744f, 55.507141113f},
+//       {-20.442623138f, -35.407661438f, 8.2749996185f},
+//       {10.229375839f, -14.717799187f, 10.508025169f}};
+//   auto hull = Manifold::Hull(hullPts);
```
elalish, line 120: Yes, for these odd shapes I think testing `IsConvex()` is the way to go.

elalish (changes requested):
This is looking good - let's just do these small updates and get it merged.

Kushal-Shah-03:
I've made the suggested changes (`DISABLED` in tests and check for convex Mesh in `hull_test.cpp`), I also added the definition of isMeshConvex in `hull_test_performance.cpp` under the `MANFIOLD_DEBUG` flag.

elalish (approved):
Thanks!

PR #880, opened 2024-07-30
Use system nanobind if installed

cjmayo:
I think I saw mention of supporting system nanobind in one of the discussions, it is becoming a thing now.

This uses the recommended approach - with additional version checking. 

This still falls back to the third_party directory for now, but obviously that is easily replaced. For pyproject.toml users  automatically fetching nanobind is supported.

### DIFF
--- a/bindings/python/CMakeLists.txt
+++ b/bindings/python/CMakeLists.txt
@@ -14,7 +14,18 @@
 
 project(python)
 
-add_subdirectory(third_party)
+execute_process(
+  COMMAND "${Python_EXECUTABLE}" -m nanobind --version
+  OUTPUT_STRIP_TRAILING_WHITESPACE OUTPUT_VARIABLE NB_VERSION)
+if(NB_VERSION VERSION_GREATER_EQUAL 1.8.0)
+  message("Found nanobind, version ${NB_VERSION}")
+  execute_process(
+    COMMAND "${Python_EXECUTABLE}" -m nanobind --cmake_dir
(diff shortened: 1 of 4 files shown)


elalish:
I think in general that fewer submodules would be great. Any reason not to remove it from `third_party`? I'll defer to @pca006132 on approving this one.

pca006132:
It seems that when building using python `build`, it automatically checks nanobind version and fail when it is not installed, even when we have a submodule.

Re. submodule, I think we can switch to `FetchContent` as well.

pca006132:
> It seems that when building using python `build`, it automatically checks nanobind version and fail when it is not installed, even when we have a submodule.

Or maybe this is just the case when the build environment is not connected to the internet? We can add the package to nix though, it is already packaged.

```diff

diff --git a/flake.nix b/flake.nix
index 05677219..3ea2dc1c 100644
--- a/flake.nix
+++ b/flake.nix
@@ -123,6 +123,7 @@
                 tbb
                 glm
                 clipper2
+                nanobind
               ];
               nativeBuildInputs = with pkgs; [
                 cmake
```

bindings/python/CMakeLists.txt
```diff
@@ -14,7 +14,25 @@
 
 project(python)
 
-add_subdirectory(third_party)
+execute_process(
```
pca006132, line 17: I think for nix, there is no python nanobind package. I think we can do `find_package(nanobind 1.8.0 CONFIG REQUIRED)`, and if `nanobind_FOUND` is not defined we can do the fetch content?

pca006132 (approved):

PR #881, opened 2024-07-31
Removing QuickHull dependency

Kushal-Shah-03:
I have removed the dependency of quickhull, I was not sure whether we wanted to delete the Makefile for third-party so, I have kept that for now. I also removed the Vector3 struct of the original implementation,replacing it with glm::vec3, and some functions that acted on Vector3. I wanted to know how to use squaredLength() functions and length() functions that show up in editor.d.ts, but are not a part of glm, currently I just added those functions in the namespace of mathutils, but was wondering whether there was a way to access them?


Kushal-Shah-03:
No they aren't, I didn't realize that changes to nanobind were made, I am not sure how that happened. Also, I am unsure as to how to revert those, any ideas?

starseeker:
Not completely sure - maybe start here?  https://stackoverflow.com/questions/5669358/can-i-do-a-partial-revert-in-git

Kushal-Shah-03:
Thanks, for the help, seems like I've fixed that issue.

src/third_party/CMakeLists.txt
```diff
@@ -1,8 +1 @@
-add_library(quickhull OBJECT quickhull/QuickHull.cpp)
```
pca006132, line 1: Just remove the whole `third_party` directory.

src/manifold/src/manifold.cpp
```diff
@@ -919,12 +919,12 @@ Manifold Manifold::Hull(const std::vector<glm::vec3>& pts) {
...
   if (numVert < 4) return Manifold();
 
-  std::vector<quickhull::Vector3<double>> vertices(numVert);
+  std::vector<glm::vec3> vertices(numVert);
   for (int i = 0; i < numVert; i++) {
     vertices[i] = {pts[i].x, pts[i].y, pts[i].z};
   }
 
-  quickhull::QuickHull<double> qh;
+  quickhull::QuickHull<float> qh;
```
pca006132, line 927: why float?

pca006132:
also please give us a summary on what you have changed comparing to the upstream version

src/manifold/src/manifold.cpp
Kushal-Shah-03: So, glm::vec3 has floats right?, so since some of the operations that were originally a part of Vecto3.hpp would give float outputs, and others double and then typecasting would happen, I thought instead let's just use floats, especially since the output we store is in floats too, if you want I can change that though.

Kushal-Shah-03:
Ok, so basically I compiled all the header files into `quickhull.hpp`, and then modified some things to make the code compile. After, that I just worked on replacing Vector3 struct and some of it's functions with glm::vec3. I also removed the dependency from the Makefile, now will remove the directory as suggested.

src/manifold/src/manifold.cpp
pca006132: just use `glm::dvec3`, we will soon switch to double.

Kushal-Shah-03: Alright, I've shifted it to glm::dvec3 instead.

src/manifold/src/quickhull.hpp
```diff
@@ -0,0 +1,852 @@
...
+		VertexDataSource& getVertexBuffer() {
+			return m_vertices;
+		}
+		
+		const VertexDataSource& getVertexBuffer() const {
+			return m_vertices;
+		}
+		
+		// Export the mesh to a Waveform OBJ file
+		void writeWaveformOBJ(const std::string& filename, const std::string& objectName = "quickhull") const
```
elalish, line 523: Don't think we need this.

src/manifold/src/quickhull.hpp
```diff
@@ -0,0 +1,852 @@
...
+			objFile.close();
+		}
+
+	};
+
+
+
+// HalfEdgeMesh.hpp
+	
+	class HalfEdgeMesh {
```
elalish, line 545: Considering we have our own halfedge mesh data structure in `Impl` already, is there a chance we can use that instead of this one?

elalish (changes requested):
Any troubler resolving conflicts?

Kushal-Shah-03:
I've made the requested changes, and resolved them for clarity. As discussed I'll tackle the Half edge mesh data structure in another PR.

Kushal-Shah-03:
Ok, so the tests are failing because
```
CMake Error at bindings/python/CMakeLists.txt:17 (nanobind_add_module):
Unknown CMake command "nanobind_add_module".


-- Configuring incomplete, errors occurred!
```
Any ideas on how I can fix it?

bindings/python/CMakeLists.txt
```diff
@@ -14,7 +14,6 @@
 
 project(python)
 
-add_subdirectory(third_party)
```
pca006132, line 17: remove this line.

pca006132:
you removed both the third-party directory inside `src` and `bindings/python`, but they are two different things.

Kushal-Shah-03:
So, I think I'm encountering this error now, although I am not sure
```
lcov: ERROR: 'exclude' pattern '*/third_party/*' is unused.
(use "lcov --ignore-errors unused ..." to bypass this error)
```
Based on a search, I think it could be caused by 
https://github.com/Kushal-Shah-03/manifold/blob/00dd951c03d0b1117c1efe3f86d27f6f6df5dd22/.github/workflows/manifold.yml#L68
or 
https://github.com/Kushal-Shah-03/manifold/blob/00dd951c03d0b1117c1efe3f86d27f6f6df5dd22/Doxyfile#L941

But, since I'm not sure could you suggest how to solve this issue?

pca006132:
yeah just the manifold.yml, remove the third_party pattern.

src/manifold/src/quickhull.h
```diff
@@ -0,0 +1,826 @@
+#ifndef QUICKHULL_HPP_
+#define QUICKHULL_HPP_
```
pca006132, line 2: I think you should add the copyright header from the upstream library (or make one for them), as well as change the define to `pragma once` that we typically use.

src/manifold/src/quickhull.h
```diff
@@ -0,0 +1,826 @@
...
+
+inline double getSquaredDistanceBetweenPointAndRay(const glm::dvec3& p,
+                                                   const Ray& r) {
+  const glm::dvec3 s = p - r.S;
+  double t = glm::dot(s, r.V);
+  return glm::dot(s, s) - t * t * r.VInvLengthSquared;
+}
+
+inline double getSquaredDistance(const glm::dvec3& p1, const glm::dvec3& p2) {
+  return (glm::dot(p1 - p2, p1 - p2));
```
pca006132, line 565: why the additional bracket?

src/manifold/src/quickhull.h
Kushal-Shah-03: Should I add both Manfiold copyright header and the upstream header? (also they did mention that we don't need to add anything so what should I add)

src/manifold/src/quickhull.h
Kushal-Shah-03: I missed it, I'll change that.

src/manifold/src/quickhull.h
pca006132: I think we can include both.

src/manifold/src/quickhull.cpp
```diff
@@ -1,3 +1,22 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+/*
+ * Implementation of the 3D QuickHull algorithm by Antti Kuukka
+ *
+ * No copyrights. What follows is 100% Public Domain.
```
pca006132, line 18: Not sure if this is the best way to put it, I think we typically just acknowledge the source, put an URL as well as state its license. From my understanding, the code is under a certain license doesn't mean the author has no copyright over it.

Kushal-Shah-03: This was what he had included in his code. I can include the link, but I don't know which license to acknowledge, could you just tell me what exactly do I need to include I will add that below our license.

elalish: Public domain *does* mean there is no longer a copyright holder. We should modify this line, since our version is no longer public domain (see the apache license above). I would change this to "Derived from the public domain work of Antti Kuukka at github/..."

Kushal-Shah-03: I've made the change to the license, I hope it's alright now.

Kushal-Shah-03:
Thank you for merging #875, now for the next steps, I was wondering what changes should I make in this pull request?

elalish:
Start by merging master and re-enabling your tests - they should work in this branch, right?

Kushal-Shah-03:
So, yeah ideally it should pass the tests, but I realized a flaw, since while calculating in the algorithm we use a variable epsilon which changes with scale, we won't be able to find the exact epsilon to check whether the point is positive or not, and since when we store it in a mesh or a manifold, the pts are converted back to floats rather than the double used to calculate the hull. 

I am thinking we can either shift the check for valid hull inside the algorithm implementation itself, so we can check relative to the variable epsilon, or we would need to check using a floating epsilon if we check outside the implementation. 

What would you suggest? 
Also, I checked despite the fact that the tests don't "pass", the outputs of both the "failing tests" are infact convex, so we would need to modify how we check for convexity

elalish:
We'll be switching to double soon, so as a stopgap, can you just put epsilon as a parameter to `IsConvex()` and then pick values for each test that makes sense?

Kushal-Shah-03:
I've made the change you suggested, currently I'm using the `m_epsilon` value that the algorithm generates as the epsilon for the two tests. I am looking into how I can revert the nanobind change I made by mistake

elalish (approved):
Good work - when the tests pass, let's merge it! I look forward to seeing this merge deeper into our own data structures next.

PR #888, opened 2024-08-08
Add a cmake option for CrossSection

cjmayo:
On by default.


CMakeLists.txt
```diff
@@ -17,18 +17,27 @@ project(manifold LANGUAGES CXX)
...
 # fuzztest is a rather large dependency
 option(MANIFOLD_FUZZ "Enable fuzzing tests" OFF)
 option(MANIFOLD_DEBUG "Enable debug tracing/timing" OFF)
-option(MANIFOLD_PYBIND "Build python bindings" OFF)
-option(MANIFOLD_CBIND "Build C (FFI) bindings" ON)
-option(MANIFOLD_JSBIND "Build js binding" ${EMSCRIPTEN})
 option(MANIFOLD_EXCEPTIONS "Build manifold with exception enabled" ON)
 option(BUILD_SHARED_LIBS "Build shared library" ON)
+include(CMakeDependentOption)
+cmake_dependent_option(MANIFOLD_TEST "Enable testing suite" ON "MANIFOLD_CROSS_SECTION" OFF)
```
elalish, line 36: This is new cmake syntax to me. What does it mean? It looks like cross section defaults to off, but I assume I'm wrong?

elalish (approved):

CMakeLists.txt
pca006132: This means `MANIFOLD_TEST` is only enabled when `MANIFOLD_CROSS_SECTION` is `ON`. I think this is a bit too strong.

pca006132:
I think we can run most of our tests when `CrossSection` is disabled.

CMakeLists.txt
elalish: Okay, I can see why that was chosen though. I wonder how many `#ifdef`s we'll have to add to get it to run? May be a good idea to rearrange the tests a bit to put anything requiring `CrossSection` near each other.

test/CMakeLists.txt
```diff
@@ -17,7 +17,11 @@ project(manifold_test)
 set(CMAKE_CXX_STANDARD 17)
 enable_testing()
 
-set(SOURCE_FILES test_main.cpp polygon_test.cpp cross_section_test.cpp properties_test.cpp manifold_test.cpp hull_test.cpp smooth_test.cpp sdf_test.cpp boolean_test.cpp boolean_complex_test.cpp samples_test.cpp)
+set(SOURCE_FILES test_main.cpp polygon_test.cpp sdf_test.cpp boolean_test.cpp)
+
+if(MANIFOLD_CROSS_SECTION)
+  list(APPEND SOURCE_FILES boolean_complex_test.cpp cross_section_test.cpp hull_test.cpp manifold_test.cpp properties_test.cpp samples_test.cpp smooth_test.cpp)
```
elalish, line 23: How many of the tests in these files actually require cross section? I can see skipping `cross_section_test.cpp`, but it feels a bit like throwing the baby out with the bathwater to remove all of these. Can we just `#ifdef` around a few tests in each file?

elalish (approved):
Thanks for updating the CI!

elalish:
Oops, @cjmayo it looks like we missed something. We were running 261 tests before, but from this PR on we're only running 250, even with `MANIFOLD_CROSS_SECTION` enabled. It appears CMake has successfully discovered the flag, but the corresponding `#ifdef`s are still false in the tests. Any idea what's wrong?

pca006132:
Ah, we need to pass the definition to c++. https://cmake.org/cmake/help/latest/command/target_compile_definitions.html

PR #893, opened 2024-08-10
Using Inbuilt Datastructures

Kushal-Shah-03:
Removed VertexDatapool class and used Vec and VecView instead.

### DIFF
--- a/src/manifold/src/impl.cpp
+++ b/src/manifold/src/impl.cpp
@@ -490,6 +490,7 @@ void Manifold::Impl::Hull(const std::vector<glm::vec3>& vertPos) {
                       [](const glm::vec3& v) { return glm::dvec3(v); });
   QuickHull qh(pointCloudVec);
   ConvexHull hull = qh.getConvexHullAsMesh(pointCloudVec, false);
+  // TODO : Once double PR lands, replace this with move
   vertPos_.resize(hull.vertices.size());
   manifold::transform(hull.vertices.begin(), hull.vertices.end(),
                       vertPos_.begin(),
--- a/src/manifold/src/impl.cpp
+++ b/src/manifold/src/impl.cpp
@@ -478,9 +478,9 @@ Manifold::Impl::Impl(const Mesh& mesh, const MeshRelationD& relation,
   Finish();
 }
 
-void Manifold::Impl::Hull(const std::vector<glm::vec3>& vertPos){
+void Manifold::Impl::Hull(const std::vector<glm::vec3>& vertPos) {
   int numVert = vertPos.size();
-  if (numVert < 4) return; // What to return here? 
+  if (numVert < 4) return;  // What to return here?
 
   Vec<glm::dvec3> pointCloudVec(numVert);
   for (int i = 0; i < numVert; i++) {
@@ -488,16 +488,17 @@ void Manifold::Impl::Hull(const std::vector<glm::vec3>& vertPos){
   }
   QuickHull qh(pointCloudVec);
   std::pair<Vec<Halfedge>, Vec<glm::dvec3>> retVal;
-  retVal= qh.getConvexHullAsMesh(pointCloudVec, false, true);
-  for (size_t i=0;i<retVal.second.size();i++){
-    vertPos_.push_back(glm::vec3(retVal.second[i].x, retVal.second[i].y, retVal.second[i].z));
+  retVal = qh.getConvexHullAsMesh(pointCloudVec, false, true);
+  for (size_t i = 0; i < retVal.second.size(); i++) {
+    vertPos_.push_back(
+        glm::vec3(retVal.second[i].x, retVal.second[i].y, retVal.second[i].z));
   }
-  numVert=vertPos_.size();
-  halfedge_=retVal.first;
+  numVert = vertPos_.size();
+  halfedge_ = retVal.first;
   meshRelation_.originalID = ReserveIDs(1);
   InitializeOriginal();
   Finish();
-  return; 
+  return;
 }
 
 /**
--- a/src/manifold/src/impl.cpp
+++ b/src/manifold/src/impl.cpp
@@ -478,6 +478,28 @@ Manifold::Impl::Impl(const Mesh& mesh, const MeshRelationD& relation,
   Finish();
 }
 
+void Manifold::Impl::Hull(const std::vector<glm::vec3>& vertPos){
+  int numVert = vertPos.size();
+  if (numVert < 4) return; // What to return here? 
+
+  Vec<glm::dvec3> pointCloudVec(numVert);
+  for (int i = 0; i < numVert; i++) {
+    pointCloudVec[i] = {vertPos[i].x, vertPos[i].y, vertPos[i].z};
+  }
+  QuickHull qh(pointCloudVec);
+  std::pair<Vec<Halfedge>, Vec<glm::dvec3>> retVal;
+  retVal= qh.getConvexHullAsMesh(pointCloudVec, false, true);
+  for (size_t i=0;i<retVal.second.size();i++){
+    vertPos_.push_back(glm::vec3(retVal.second[i].x, retVal.second[i].y, retVal.second[i].z));
+  }
+  numVert=vertPos_.size();
+  halfedge_=retVal.first;
+  meshRelation_.originalID = ReserveIDs(1);
+  InitializeOriginal();
+  Finish();
+  return; 
+}
+
 /**
  * Create either a unit tetrahedron, cube or octahedron. The cube is in the
  * first octant, while the others are symmetric about the origin.
--- a/src/manifold/src/impl.cpp
+++ b/src/manifold/src/impl.cpp
@@ -487,14 +487,14 @@ void Manifold::Impl::Hull(const std::vector<glm::vec3>& vertPos) {
     pointCloudVec[i] = {vertPos[i].x, vertPos[i].y, vertPos[i].z};
   }
   QuickHull qh(pointCloudVec);
-  std::pair<Vec<Halfedge>, Vec<glm::dvec3>> retVal;
-  retVal = qh.getConvexHullAsMesh(pointCloudVec, false, true);
-  for (size_t i = 0; i < retVal.second.size(); i++) {
+  ConvexHull hull;
+  hull = qh.getConvexHullAsMesh(pointCloudVec, false);
+  for (size_t i = 0; i < hull.vertices.size(); i++) {
     vertPos_.push_back(
-        glm::vec3(retVal.second[i].x, retVal.second[i].y, retVal.second[i].z));
+        glm::vec3(hull.vertices[i].x, hull.vertices[i].y, hull.vertices[i].z));
   }
   numVert = vertPos_.size();
-  halfedge_ = retVal.first;
+  halfedge_ = hull.halfEdges;
   meshRelation_.originalID = ReserveIDs(1);
   InitializeOriginal();
   Finish();
--- a/src/manifold/src/impl.cpp
+++ b/src/manifold/src/impl.cpp
@@ -495,10 +495,7 @@ void Manifold::Impl::Hull(const std::vector<glm::vec3>& vertPos) {
   manifold::transform(hull.vertices.begin(), hull.vertices.end(),
                       vertPos_.begin(),
                       [](const glm::dvec3& v) { return glm::vec3(v); });
-  halfedge_ = std::move(hull.halfEdges);
-  for_each(autoPolicy(halfedge_.size(), 1e5), countAt(0_uz),
-           countAt(halfedge_.size()),
-           [this](size_t i) { halfedge_[i].face = i / 3; });
+  halfedge_ = std::move(hull.halfedges);
   meshRelation_.originalID = ReserveIDs(1);
   CalculateBBox();
   SetPrecision(bBox_.Scale() * kTolerance);
--- a/src/manifold/src/impl.cpp
+++ b/src/manifold/src/impl.cpp
@@ -479,29 +479,29 @@ Manifold::Impl::Impl(const Mesh& mesh, const MeshRelationD& relation,
 }
 
 void Manifold::Impl::Hull(const std::vector<glm::vec3>& vertPos) {
-  int numVert = vertPos.size();
+  size_t numVert = vertPos.size();
   if (numVert < 4) {
     status_ = Error::InvalidConstruction;
     return;
   }
 
   Vec<glm::dvec3> pointCloudVec(numVert);
-  for (int i = 0; i < numVert; i++) {
-    pointCloudVec[i] = {vertPos[i].x, vertPos[i].y, vertPos[i].z};
-  }
+  manifold::transform(vertPos.begin(), vertPos.end(), pointCloudVec.begin(),[](const glm::vec3& v) { return glm::dvec3(v); });
   QuickHull qh(pointCloudVec);
-  ConvexHull hull;
-  hull = qh.getConvexHullAsMesh(pointCloudVec, false);
-  for (size_t i = 0; i < hull.vertices.size(); i++) {
-    vertPos_.push_back(
-        glm::vec3(hull.vertices[i].x, hull.vertices[i].y, hull.vertices[i].z));
-  }
-  numVert = vertPos_.size();
-  halfedge_ = hull.halfEdges;
+  ConvexHull hull = qh.getConvexHullAsMesh(pointCloudVec, false);
+  vertPos_.resize(hull.vertices.size());
+  manifold::transform(hull.vertices.begin(), hull.vertices.end(),vertPos_.begin(),[](const glm::dvec3& v) { return glm::vec3(v); });
+  halfedge_ = std::move(hull.halfEdges);
+  for_each(autoPolicy(halfedge_.size(), 1e5), countAt(0_uz),countAt(halfedge_.size()),[this](size_t i) { halfedge_[i].face = i / 3; });
   meshRelation_.originalID = ReserveIDs(1);
+  CalculateBBox();
+  SetPrecision(bBox_.Scale() * kTolerance);
+  SplitPinchedVerts();
+  CalculateNormals();
   InitializeOriginal();
+  CreateFaces({});
+  SimplifyTopology();
   Finish();
-  return;
 }
 
 /**
--- a/src/manifold/src/impl.cpp
+++ b/src/manifold/src/impl.cpp
@@ -480,7 +480,10 @@ Manifold::Impl::Impl(const Mesh& mesh, const MeshRelationD& relation,
 
 void Manifold::Impl::Hull(const std::vector<glm::vec3>& vertPos) {
   int numVert = vertPos.size();
-  if (numVert < 4) return;  // What to return here?
+  if (numVert < 4) {
+    status_ = Error::InvalidConstruction;
+    return;
+  }
 
   Vec<glm::dvec3> pointCloudVec(numVert);
   for (int i = 0; i < numVert; i++) {
--- a/src/manifold/src/impl.cpp
+++ b/src/manifold/src/impl.cpp
@@ -486,13 +486,18 @@ void Manifold::Impl::Hull(const std::vector<glm::vec3>& vertPos) {
   }
 
   Vec<glm::dvec3> pointCloudVec(numVert);
-  manifold::transform(vertPos.begin(), vertPos.end(), pointCloudVec.begin(),[](const glm::vec3& v) { return glm::dvec3(v); });
+  manifold::transform(vertPos.begin(), vertPos.end(), pointCloudVec.begin(),
+                      [](const glm::vec3& v) { return glm::dvec3(v); });
   QuickHull qh(pointCloudVec);
   ConvexHull hull = qh.getConvexHullAsMesh(pointCloudVec, false);
   vertPos_.resize(hull.vertices.size());
-  manifold::transform(hull.vertices.begin(), hull.vertices.end(),vertPos_.begin(),[](const glm::dvec3& v) { return glm::vec3(v); });
+  manifold::transform(hull.vertices.begin(), hull.vertices.end(),
+                      vertPos_.begin(),
+                      [](const glm::dvec3& v) { return glm::vec3(v); });
   halfedge_ = std::move(hull.halfEdges);
-  for_each(autoPolicy(halfedge_.size(), 1e5), countAt(0_uz),countAt(halfedge_.size()),[this](size_t i) { halfedge_[i].face = i / 3; });
+  for_each(autoPolicy(halfedge_.size(), 1e5), countAt(0_uz),
+           countAt(halfedge_.size()),
+           [this](size_t i) { halfedge_[i].face = i / 3; });
   meshRelation_.originalID = ReserveIDs(1);
   CalculateBBox();
   SetPrecision(bBox_.Scale() * kTolerance);
--- a/src/manifold/src/quickhull.cpp
+++ b/src/manifold/src/quickhull.cpp
@@ -1,0 +1,575 @@
+// Copyright 2024 The Manifold Authors.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+// Derived from the public domain work of Antti Kuukka at
+// https://github.com/akuukka/quickhull
+
+#include "quickhull.h"
+
+#include <algorithm>
+#include <cassert>
+#include <cmath>
+#include <iostream>
+#include <limits>
+
+double defaultEps() { return 0.0000001; }
+
+/*
+ * Implementation of the algorithm
+ */
+
+ConvexHull QuickHull::getConvexHull(const std::vector<glm::dvec3>& pointCloud,
+                                    bool CCW, bool useOriginalIndices,
+                                    double epsilon) {
+  VertexDataSource vertexDataSource(pointCloud);
+  return getConvexHull(vertexDataSource, CCW, useOriginalIndices, epsilon);
+}
+
+ConvexHull QuickHull::getConvexHull(const glm::dvec3* vertexData,
+                                    size_t vertexCount, bool CCW,
+                                    bool useOriginalIndices, double epsilon) {
+  VertexDataSource vertexDataSource(vertexData, vertexCount);
+  return getConvexHull(vertexDataSource, CCW, useOriginalIndices, epsilon);
+}
+
+ConvexHull QuickHull::getConvexHull(const double* vertexData,
+                                    size_t vertexCount, bool CCW,
+                                    bool useOriginalIndices, double epsilon) {
+  VertexDataSource vertexDataSource((const vec3*)vertexData, vertexCount);
+  return getConvexHull(vertexDataSource, CCW, useOriginalIndices, epsilon);
+}
+
+HalfEdgeMesh QuickHull::getConvexHullAsMesh(const double* vertexData,
+                                            size_t vertexCount, bool CCW,
+                                            double epsilon) {
+  VertexDataSource vertexDataSource((const vec3*)vertexData, vertexCount);
+  buildMesh(vertexDataSource, CCW, false, epsilon);
+  return HalfEdgeMesh(mesh, originalVertexData);
+}
+
+void QuickHull::buildMesh(const VertexDataSource& pointCloud, bool CCW,
+                          bool useOriginalIndices, double epsilon) {
+  // CCW is unused for now
+  (void)CCW;
+  // useOriginalIndices is unused for now
+  (void)useOriginalIndices;
+
+  if (pointCloud.size() == 0) {
+    mesh = MeshBuilder();
+    return;
+  }
+  originalVertexData = pointCloud;
+
+  // Very first: find extreme values and use them to compute the scale of the
+  // point cloud.
+  extremeValues = getExtremeValues();
+  scale = getScale(extremeValues);
+
+  // Epsilon we use depends on the scale
+  m_epsilon = epsilon * scale;
+  epsilonSquared = m_epsilon * m_epsilon;
+
+  // Reset diagnostics
+  diagnostics = DiagnosticsData();
+  // The planar case happens when all the points appear to lie on a two
+  // dimensional subspace of R^3.
+  planar = false;
+  createConvexHalfEdgeMesh();
+  if (planar) {
+    const size_t extraPointIndex = planarPointCloudTemp.size() - 1;
+    for (auto& he : mesh.halfEdges) {
+      if (he.endVertex == extraPointIndex) {
+        he.endVertex = 0;
+      }
+    }
+    originalVertexData = pointCloud;
+    planarPointCloudTemp.clear();
+  }
+}
+
+ConvexHull QuickHull::getConvexHull(const VertexDataSource& pointCloud,
+                                    bool CCW, bool useOriginalIndices,
+                                    double epsilon) {
+  buildMesh(pointCloud, CCW, useOriginalIndices, epsilon);
+  return ConvexHull(mesh, originalVertexData, CCW, useOriginalIndices);
+}
+
+void QuickHull::createConvexHalfEdgeMesh() {
+  visibleFaces.clear();
+  horizonEdgesData.clear();
+  possiblyVisibleFaces.clear();
+
+  // Compute base tetrahedron
+  setupInitialTetrahedron();
+  assert(mesh.faces.size() == 4);
+
+  // Init face stack with those faces that have points assigned to them
+  faceList.clear();
+  for (size_t i = 0; i < 4; i++) {
+    auto& f = mesh.faces[i];
+    if (f.pointsOnPositiveSide && f.pointsOnPositiveSide->size() > 0) {
+      faceList.push_back(i);
+      f.inFaceStack = 1;
+    }
+  }
+
+  // Process faces until the face list is empty.
+  size_t iter = 0;
+  while (!faceList.empty()) {
+    iter++;
+    if (iter == std::numeric_limits<size_t>::max()) {
+      // Visible face traversal marks visited faces with iteration counter (to
+      // mark that the face has been visited on this iteration) and the max
+      // value represents unvisited faces. At this point we have to reset
+      // iteration counter. This shouldn't be an issue on 64 bit machines.
+      iter = 0;
+    }
+
+    const size_t topFaceIndex = faceList.front();
+    faceList.pop_front();
+
+    auto& tf = mesh.faces[topFaceIndex];
+    tf.inFaceStack = 0;
+
+    assert(!tf.pointsOnPositiveSide || tf.pointsOnPositiveSide->size() > 0);
+    if (!tf.pointsOnPositiveSide || tf.isDisabled()) {
+      continue;
+    }
+
+    // Pick the most distant point to this triangle plane as the point to which
+    // we extrude
+    const vec3& activePoint = originalVertexData[tf.mostDistantPoint];
+    const size_t activePointIndex = tf.mostDistantPoint;
+
+    // Find out the faces that have our active point on their positive side
+    // (these are the "visible faces"). The face on top of the stack of course
+    // is one of them. At the same time, we create a list of horizon edges.
+    horizonEdgesData.clear();
+    possiblyVisibleFaces.clear();
+    visibleFaces.clear();
+    possiblyVisibleFaces.emplace_back(topFaceIndex,
+                                      std::numeric_limits<size_t>::max());
+    while (possiblyVisibleFaces.size()) {
+      const auto faceData = possiblyVisibleFaces.back();
+      possiblyVisibleFaces.pop_back();
+      auto& pvf = mesh.faces[faceData.faceIndex];
+      assert(!pvf.isDisabled());
+
+      if (pvf.visibilityCheckedOnIteration == iter) {
+        if (pvf.isVisibleFaceOnCurrentIteration) {
+          continue;
+        }
+      } else {
+        const Plane& P = pvf.P;
+        pvf.visibilityCheckedOnIteration = iter;
+        const double d = glm::dot(P.N, activePoint) + P.D;
+        if (d > 0) {
+          pvf.isVisibleFaceOnCurrentIteration = 1;
+          pvf.horizonEdgesOnCurrentIteration = 0;
+          visibleFaces.push_back(faceData.faceIndex);
+          for (auto heIndex : mesh.getHalfEdgeIndicesOfFace(pvf)) {
+            if (mesh.halfEdges[heIndex].opp != faceData.enteredFromHalfEdge) {
+              possiblyVisibleFaces.emplace_back(
+                  mesh.halfEdges[mesh.halfEdges[heIndex].opp].face, heIndex);
+            }
+          }
+          continue;
+        }
+        assert(faceData.faceIndex != topFaceIndex);
+      }
+
+      // The face is not visible. Therefore, the halfedge we came from is part
+      // of the horizon edge.
+      pvf.isVisibleFaceOnCurrentIteration = 0;
+      horizonEdgesData.push_back(faceData.enteredFromHalfEdge);
+      // Store which half edge is the horizon edge. The other half edges of the
+      // face will not be part of the final mesh so their data slots can by
+      // recycled.
+      const auto halfEdgesMesh = mesh.getHalfEdgeIndicesOfFace(
+          mesh.faces[mesh.halfEdges[faceData.enteredFromHalfEdge].face]);
+      const std::int8_t ind =
+          (halfEdgesMesh[0] == faceData.enteredFromHalfEdge)
+              ? 0
+              : (halfEdgesMesh[1] == faceData.enteredFromHalfEdge ? 1 : 2);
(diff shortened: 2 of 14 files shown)


Kushal-Shah-03:
Sorry, for the delay, it took me some time to figure out where to use Vec and VecView. There is another class called Pool, I think since it's using unique_ptrs, we can't replace it with Vec, but I'm not sure, could you @pca006132, help me out with this?

src/manifold/src/quickhull.cpp
```diff
@@ -32,33 +32,11 @@ double defaultEps() { return 0.0000001; }
 ConvexHull QuickHull::getConvexHull(const std::vector<glm::dvec3>& pointCloud,
                                     bool CCW, bool useOriginalIndices,
                                     double epsilon) {
-  VertexDataSource vertexDataSource(pointCloud);
+  manifold::Vec<glm::dvec3> vertexDataSource(pointCloud);
```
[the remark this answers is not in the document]
pca006132: Also, will this change the timing? You are copying the points into a `Vec`. Should we modify the input `pointCloud` to `VecView`?

pca006132:
For the pool thing, just change the inner `vector` into `Vec`. And I am curious how much reuse does this pool enable.

src/manifold/src/quickhull.cpp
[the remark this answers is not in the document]
Kushal-Shah-03: > @elalish I vaguely recall you said we should move the quickhull thing into the `manifold` namespace right?
> 
> If we move the namespace, we don't need this `manifold::` namespace specifier.

That was regarding removing the namespace for quickhull right? So, should I also wrap the code in namespace manifold?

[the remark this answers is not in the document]
Kushal-Shah-03: > Also, will this change the timing? You are copying the points into a `Vec`. Should we modify the input `pointCloud` to `VecView`?

I tried modifying the input 'pointCloud' to VecView, but it was showing the error that VecView had a protected constructor. I can try to surpass that though, I'll give it one more try.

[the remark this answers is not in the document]
Kushal-Shah-03: Also, `optimizedVertexBuffer` is a `unique_ptr` so, it won't be compatible with `VecView` right?, so when I'm calling

`vertices = manifold::VecView<glm::dvec3>(*optimizedVertexBuffer);` it won't work right?

Kushal-Shah-03:
> For the pool thing, just change the inner `vector` into `Vec`. And I am curious how much reuse does this pool enable.

So, at just one place it actually uses the erase function call of vector, should I just work around that by manually erasing it by iterating the Vec (since the inbuilt call was O(n) anyway), or should I modify our implementation of Vec to add a erase implementation? [Also, note this won't add a massive overhead, this is only called when the hull algorithm fails to solve a horizon edge]

Kushal-Shah-03:
Also, regarding the HalfEdge datastructure, since most of the parameters are the same except one, I decided to add the inbuilt structure in the new structure as a variable, but since our HalfEdge structure stores `startVert`,`endVert`,`pairedHalfEdge` and `face` as `int` while the implementation uses `size_t`, what should I be doing, should I just change the implementation to now use `int` instead of `size_t`, because otherwise there are lot's of warnings and errors.

Also, I feel that despite it using almost the same variables, we cannot afford to remove the implementations structure, because it uses functions to disable the edges when not in use [and I think this disabling might be the reason this implementation gives such a good performance in terms of speed].

elalish:
Manual erasing and switching to `int` sound fine to me for now. When in doubt, go with fewer lines of added/changed code. 

How does the disabling of edges work? Is it just a marker per edge, or a more complex/efficient structure? If the former, you can probably just add a parallel vector.

Kushal-Shah-03:
> Manual erasing and switching to `int` sound fine to me for now. When in doubt, go with fewer lines of added/changed code.
> 
> How does the disabling of edges work? Is it just a marker per edge, or a more complex/efficient structure? If the former, you can probably just add a parallel vector.

Alright I'll switch to int.
It is sort of a marker (it sets the `endVertex` to `MAX` to mark it as unused), Instead of a parallel vector I was just planning to add our HalfEdge structure as a variable inside each HalfEdge that also works right? [since this HalfEdge has a next halfedge corresponding variable (so it actually isn't stored chronologically), so an independent vector of our halfedge structure would not be very useful]

src/manifold/src/quickhull.cpp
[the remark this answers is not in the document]
pca006132: 1. Just change the namespace for quickhull to manifold.
2. Can you show me the modification and error message about protected constructor?
3. `optimizedVertexBuffer` is an internal buffer, you want `Vec` for that.

pca006132:
> Instead of a parallel vector I was just planning to add our HalfEdge structure as a variable inside each HalfEdge that also works right? [since this HalfEdge has a next halfedge corresponding variable (so it actually isn't stored chronologically), so an independent vector of our halfedge structure would not be very useful]

I don't understand this part, can you elaborate a bit or maybe just show some pseudocode?

> So, at just one place it actually uses the erase function call of vector, should I just work around that by manually erasing it by iterating the Vec (since the inbuilt call was O(n) anyway), or should I modify our implementation of Vec to add a erase implementation? [Also, note this won't add a massive overhead, this is only called when the hull algorithm fails to solve a horizon edge]

Adding erase method to our `Vec` implementation is fine. It should be very simple.

elalish:
We actually use a similar system - we set a halfedge to `{-1, -1, -1, -1}` to indicate it has been removed, for instance. You can feel free to use any negative flags you need for your implementation.

src/manifold/src/quickhull.cpp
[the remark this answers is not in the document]
Kushal-Shah-03: In terms of modification all I'm doing is since we want to make pointCloud VecView, I make some Vec's VecView's for the code to compile (without incorrect function definitions) but this error persists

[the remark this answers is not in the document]
Kushal-Shah-03: Alright, I'll commit the changes I've made to the Mesh structure first and then I'll work on this, to separate the commits

Kushal-Shah-03:
> I don't understand this part, can you elaborate a bit or maybe just show some pseudocode?

I'm just using our HalfEdge implementation as a variable inside the HalfEdge implementation of the algortihm:
So, the code looks like this
```
struct HalfEdge {
    manifold::Halfedge halfEdgeManifold;
    int next;
    // functions
}
```

pca006132:
I don't understand why you still need the old halfedge data structure there. The two fields left are a next index (which manifold's halfedge data structure already has) and manifold's halfedge data structure.

Kushal-Shah-03:
> I don't understand why you still need the old halfedge data structure there. The two fields left are a next index (which manifold's halfedge data structure already has) and manifold's halfedge data structure.

https://github.com/elalish/manifold/blob/e01d31611ed5894094cc89005db89b1ca2f417bf/src/manifold/src/shared.h#L123C1-L132C3 
I might have missed something, but this does not have the next index right?

pca006132:
oh sorry, I forgot that we assume that the halfedge data structure is for triangular mesh, and the next id is always computable from the current id (using the next function)

it is very likely that the halfedge here also has this property.

Kushal-Shah-03:
> oh sorry, I forgot that we assume that the halfedge data structure is for triangular mesh, and the next id is always computable from the current id (using the next function)
> 
> it is very likely that the halfedge here also has this property.

Do you mean the `NextHalfEdge` function? I just thought, that since we are disabling, we might reuse it later with other values, leading to it not being in the order of indexes (which is what I think the NextHalfEdge function assumes)

pca006132:
They disable all three halfedges around the face, so this should not be an issue.

pca006132:
I see. I think this is fine for now. In the future if we have time we should look at how to avoid those halfedge reordering, so we can directly use as our halfedge data structure.

Kushal-Shah-03:
Also, I've made the change for the VecView (the error I was having) and Vec for Pool, could you go through them once as well when you have time

src/manifold/src/quickhull.h
```diff
@@ -67,25 +67,27 @@
...
-  void reclaim(std::unique_ptr<std::vector<size_t>>& ptr) {
+  void reclaim(std::unique_ptr<manifold::Vec<size_t>>& ptr) {
     data.push_back(std::move(ptr));
   }
 
-  std::unique_ptr<std::vector<size_t>> get() {
+  std::unique_ptr<manifold::Vec<size_t>> get() {
     if (data.size() == 0) {
-      return std::unique_ptr<std::vector<size_t>>(new std::vector<size_t>());
+      return std::unique_ptr<manifold::Vec<size_t>>(
```
pca006132, line 86: `std::make_unique<manifold::Vec<size_t>>()`

src/manifold/src/quickhull.h
```diff
@@ -400,8 +381,8 @@ class ConvexHull {
     indices = o.indices;
     if (o.optimizedVertexBuffer) {
       optimizedVertexBuffer.reset(
-          new std::vector<glm::dvec3>(*o.optimizedVertexBuffer));
-      vertices = VertexDataSource(*optimizedVertexBuffer);
+          new manifold::Vec<glm::dvec3>(*o.optimizedVertexBuffer));
+      vertices = manifold::Vec<glm::dvec3>(*optimizedVertexBuffer);
```
pca006132, line 385: can you just use `vertices = *optimizedVertexBuffer`?

Kushal-Shah-03:
I've made the changes @elalish and @pca006132 , could you review this once. I will start working on profiling, so that we can plan further.

src/manifold/src/quickhull.h
```diff
@@ -264,86 +242,86 @@ class MeshBuilder {
 
     // Create halfedges
     HalfEdge AB;
-    AB.endVertex = b;
-    AB.opp = 6;
-    AB.face = 0;
+    AB.halfEdgeManifold.endVert = b;
```
elalish, line 245: Didn't we talk about condensing this huge list?

src/manifold/src/quickhull.h
```diff
@@ -129,51 +130,28 @@ struct Ray {
...
-    size_t endVertex;
-    size_t opp;
-    size_t face;
-    size_t next;
-
-    void disable() { endVertex = std::numeric_limits<size_t>::max(); }
+    // size_t endVertex;
+    // size_t opp;
+    // size_t face;
+    manifold::Halfedge halfEdgeManifold;
```
elalish, line 141: If we can't actually use our `halfedge_` vector, you can go ahead and put this back the way it was so the members are easier to access.

src/manifold/src/quickhull.h
Kushal-Shah-03: Yes, sorry I forgot about that will change this.

src/manifold/src/quickhull.h
Kushal-Shah-03: So, I had an idea regarding the same, I understand that we can't just use the `halfedge_` vector during each iteration, but at the end, I can reorder the halfedges and return our `halfedge_` vector. That would be a O(n) operation, so wouldn't add much overhead as well, and we will have the halfedges in our format if we want to do some further operation. What do you think about this? (Also another option is we can modify our halfedge implementation and add next in it, which I will only use in the quickhull file for now, it would still add a unused variable at every other place halfedge is being used, which does not seem ideal)

If you think we should implement any of these ideas, I'll go ahead with it otherwise I'll revert the changes

pca006132: If you want to do reordering, you should probably just have a vector of `manifold::Halfedge` as well as a vector of `int` as the `next` index.

Kushal-Shah-03: > If you want to do reordering, you should probably just have a vector of `manifold::Halfedge` as well as a vector of `int` as the `next` index.

Yes, I will remove the class and move the functions outside as well. I just wanted to get the idea approved first. Let's see what Emmett says, till then I'll see if I can start profiling

elalish: Oh, yes, thanks for bringing this up - I just reread how your function in `manifold.cpp` works. Yes, let's make quickhull into a member function of `Impl` and directly construct the `halfedge_` vector. The public Manifold function can just wrap that. You can use [this](https://github.com/elalish/manifold/blob/master/src/manifold/src/impl.cpp#L485-L532) as an example. The halfedge reordering part can remain as an internal detail of the quickhull function.

Kushal-Shah-03: So, just to verify if I understood it correct, I want to create a function like this in the file

> Manifold Manifold::Impl::Quickhull(const std::vector<glm::dvec3>& points) {

inside this function I wall call the quickhull function, which would return a `halfedge_` vector and also return the convexhull right, and if that's the case how do we plan to use the `halfedge_` vector here? [Also, I will be reordering the edges inside the algorithm after it computes the hull]

elalish: The Impl is the internal mesh data structure of Manifold. You'll make a new Impl constructor that creates `vertPos_` and `halfedge_` because that is what represents a Manifold mesh - the convex hull in this case. If you look at how the Manifold constructors work you'll see they just mostly wrap `Impl` methods. Then you won't need to use `Mesh` at all, which is good because we're deprecating it.

Kushal-Shah-03: I'm sorry but I'm still slightly confused. I can directly get vertPos and halfedge from the quickhull function, so why would I need to create a new Impl constructor for that? If you don't mind could write a pseudo code or function heading to explain it a bit.

elalish: My understanding of this code right now is that `QuickHull` operates internally on a `MeshBuilder` object called `mesh` - this is where the algorithmic work actually happens. Once complete, it goes through a series of transforms: `MeshBuilder -> ConvexHull -> Mesh -> Manifold`, where `Manifold` is really just a wrapper of `Impl` which is our actual internal halfedge data structure. I propose simplifying this into `MeshBuilder -> Impl` - delete the `ConvexHull` class entirely and don't use our `Mesh` class at all. Simply translate the `MeshBuilder` halfedge data structure into our `Impl` halfedge data structure.

This can either happen in your existing `Manifold Manifold::Hull(const std::vector<vec3>& pts)` function, or in a new `void Manifold::Impl::Hull(const std::vector<vec3>& pts)` function - the latter just lets you avoid writing a lot of `impl_->...` but we have plenty of examples both ways, so I don't have a strong opinion.

Kushal-Shah-03: Thank you for the clarification, I understand it now. The last thing I wanted to ask is once, I translate the `MeshBuilder` halfedge data structure into `Impl` halfedge, then we plan to call functions like this https://github.com/elalish/manifold/blob/952546b761b343a21a951ff03b22a8d0a8a570f8/src/manifold/src/impl.cpp#L528C1-L531C17, so we can then directly use this 'Impl` as a wrapper function right?

elalish: exactly - you can probably skip CreateFaces though - it's a bit slow and not so necessary on a hull.

Kushal-Shah-03: Alright, thank you for clearing it up, I was quite confused about what I had to do, but I think I've got a very good idea of it now.

Kushal-Shah-03:
I was trying to profile using tracy, and I was encountering this error, while setting up the tracy server

> 
> [0/2] Re-checking globbed directories...
> ninja: error: '/usr/share/wayland-protocols/staging/cursor-shape/cursor-shape-v1.xml', needed by 'wayland-cursor-shape-client-protocol.h', missing and no known rule to make it
> 

Any idea, on how to fix this I tried doing 

>  sudo apt-get install libwayland-dev wayland-protocols

Also, I noticed there was a interactive demo on the github https://tracy.nereid.pl/ , is there a way to directly use this rather than setting it up locally?

pca006132:
just commit it and let us have a look. I think maybe the halfedge vector is having some invalid indices.

elalish:
Yeah, we use `face` and `tri` a bit interchangeably in the code base. It means exactly what it says - you set the `pairedHalfedge` to point to a halfedge index that doesn't exist. So it's just a bug in how you constructed `halfedge_`.

Kushal-Shah-03:
So, I got it to work, but I am still having an issue, in the case where I'm supposed to return the empty hull (numVert<4), could you help me with what would I need to return there?

Kushal-Shah-03:
So, I realized that the test that was failing was actually the coplanar test, and that was because it isn't being marked as empty, the output has 0 volume, 2 surface area, both of which are intended as far as I understand. So, what should I do here?

And I also tested the entire dataset, we are able to mimic the accuracy of the earlier results. So, I think it works now.

Also, I remember @pca006132 saying something about saving the original manifold. I am not sure what that meant? Could you guide me on what I should do regarding the same?

PS: I got tracy to work (had to install some yaml files manually and tweak a couple of lines of code)

pca006132:
1. For empty manifold, I guess you can return this: https://github.com/elalish/manifold/blob/master/src/manifold/src/manifold.cpp#L81
2. https://github.com/elalish/manifold/blob/7787cc04d8cdce0994b83fb08c00cd7d0402ce65/src/manifold/src/manifold.cpp#L926 you should modify this one as well, which returns a new manifold. You should probably not use `GetMesh` directly, but only call `GetCsgLeafNode().GetImpl()` and get the vertex position vector for hull.

Kushal-Shah-03:
> 1. For empty manifold, I guess you can return this: https://github.com/elalish/manifold/blob/master/src/manifold/src/manifold.cpp#L81
> 
>     2. https://github.com/elalish/manifold/blob/7787cc04d8cdce0994b83fb08c00cd7d0402ce65/src/manifold/src/manifold.cpp#L926
>         you should modify this one as well, which returns a new manifold. You should probably not use `GetMesh` directly, but only call `GetCsgLeafNode().GetImpl()` and get the vertex position vector for hull.

I've made these changes, and I've also added manifold namespace, and removed the excess lines needed to add HalfEdges at the start. Now, the tests aren't passing because of the `IsEmpty()` check for coplanar test, as far as I understand `isEmpty()` checks if there are any triangles in the Manifold, and ideally triangles should be returned in this case right, so what shoud I do?

Also, I think I've refactored the code, and it's ready for review @elalish and @pca006132.

Kushal-Shah-03:
I was trying out tracy on the main branch to get some profiling done
image
This, result is on the tictac hull test

I think the changes we've done here have resulted in quite a bit of an improvement in the time. I say, this because functions like CreateNormals, CreateHalfEdges that were called inside the impl function are now not used. And, I also tried testing this implementation on the entire Thingi10K dataset again, and I got around **0.7-0.75 times** of the original time (before resolving dependency, now I understand that these times could vary from run to run, but I think the profiling results, suggest that we did make an imporvement). And, based on these results, I think the next function, we can look to parallelize the calls to is `addPointToFace`, and I'll also try to see by adding `Zonescoped` in other functions inside the `CreateConvexHalfEdgeMesh` to see which other function or part of the function we should look into. I think, once we get this PR merged, I'll open another PR, where we can start working on parallelization, to look to improve the speed further.

pca006132:
Fix:

```diff
diff --git a/src/manifold/src/impl.cpp b/src/manifold/src/impl.cpp
index d4462ffe..f3cb09a0 100644
--- a/src/manifold/src/impl.cpp
+++ b/src/manifold/src/impl.cpp
@@ -479,29 +479,34 @@ Manifold::Impl::Impl(const Mesh& mesh, const MeshRelationD& relation,
 }
 
 void Manifold::Impl::Hull(const std::vector<glm::vec3>& vertPos) {
-  int numVert = vertPos.size();
+  size_t numVert = vertPos.size();
   if (numVert < 4) {
     status_ = Error::InvalidConstruction;
     return;
   }
 
   Vec<glm::dvec3> pointCloudVec(numVert);
-  for (int i = 0; i < numVert; i++) {
-    pointCloudVec[i] = {vertPos[i].x, vertPos[i].y, vertPos[i].z};
-  }
+  manifold::transform(vertPos.begin(), vertPos.end(), pointCloudVec.begin(),
+                      [](const glm::vec3& v) { return glm::dvec3(v); });
   QuickHull qh(pointCloudVec);
-  ConvexHull hull;
-  hull = qh.getConvexHullAsMesh(pointCloudVec, false);
-  for (size_t i = 0; i < hull.vertices.size(); i++) {
-    vertPos_.push_back(
-        glm::vec3(hull.vertices[i].x, hull.vertices[i].y, hull.vertices[i].z));
-  }
-  numVert = vertPos_.size();
-  halfedge_ = hull.halfEdges;
+  ConvexHull hull = qh.getConvexHullAsMesh(pointCloudVec, false);
+  vertPos_.resize(hull.vertices.size());
+  manifold::transform(hull.vertices.begin(), hull.vertices.end(),
+                      vertPos_.begin(),
+                      [](const glm::dvec3& v) { return glm::vec3(v); });
+  halfedge_ = std::move(hull.halfEdges);
+  for_each(autoPolicy(halfedge_.size(), 1e5), countAt(0_uz),
+           countAt(halfedge_.size()),
+           [this](size_t i) { halfedge_[i].face = i / 3; });
   meshRelation_.originalID = ReserveIDs(1);
+  CalculateBBox();
+  SetPrecision(bBox_.Scale() * kTolerance);
+  SplitPinchedVerts();
+  CalculateNormals();
   InitializeOriginal();
+  CreateFaces({});
+  SimplifyTopology();
   Finish();
-  return;
 }
 
 /**
```

this is because:

1. You need `SimplifyTopology` to simplify the coplanar mesh.
2. You need to calculate the bounding box, precision to setup the data structure.
3. Your halfedge data structure face id is wrong, we expect the face id to be the index divided by 3. We should probably document this better somewhere.

Also, I think there is still a `HalfEdge` data structure in `MeshBuilder`?

Kushal-Shah-03:
So, I've made the changes that you suggested, and now the test passes, but it does make me wonder whether we need `SimplifyTopology` here, since technically the output will still have the same volume and area [based on the tests as well, since they gave the same accuracy] (which was what we were mainly concerned with to begin with), and removing these calls, was what improved the speed in the first place right? [if you look at the profiler image I attached above, you can see that `CalculateNormals` was one of the functions that took a lot of time]

And yes, the `halfEdge` structure is still needed, since I don't want to rearrange them everytime right, that would add overhead, so I would need the next variable to identify the order, and keeping them separate would just add more lines of code when copying the `halfEdge` and performing other such operations (like trying to reference the same indices), having this data structure won't cause any other problems right? (If you want I can change the name to avoid confusion with our current structure)

src/manifold/src/quickhull.cpp
```diff
@@ -70,104 +66,127 @@ std::pair<manifold::Vec<manifold::Halfedge>,std::vector<glm::dvec3>> QuickHull::
...
-        continue;
-      } else {
-        while(halfEdgeProcessed[top]!=true) {
-          if (mesh.halfEdges[top].opp == std::numeric_limits<size_t>::max()) {
-            std::cout << "ERROR: Halfedge " << top << " has no paired halfedge" << std::endl;
+    size_t top = halfEdgeStack[index];
+
+    if (mesh.halfEdges[top].isDisabled() ||
+        mesh.faces[mesh.halfEdges[top].face].isDisabled()) {
+      std::cout << "ERROR: Halfedge " << top << " is not disabled" << std::endl;
```
elalish, line 92: Hmm, ideally this should fail to compile. We only include `iostream` if we're in `MANIFOLD_DEBUG` mode (it adds a lot of binary size to our WASM, etc. Let's just use our internal `ASSERT` and `DEBUG_ASSERT` functions instead.

src/manifold/src/quickhull.h
```diff
@@ -241,89 +254,30 @@ class MeshBuilder {
...
-    halfEdges.push_back(BD);
-
-    HalfEdge DC;
-    DC.endVertex = c;
-    DC.opp = 4;
-    DC.face = 3;
-    DC.next = 9;
-    halfEdges.push_back(DC);
+    // AB
+    halfEdges.push_back(HalfEdge(b, 6, 0, 1));
```
elalish, line 258: 💯

src/manifold/src/impl.cpp
```diff
@@ -478,26 +478,35 @@ Manifold::Impl::Impl(const Mesh& mesh, const MeshRelationD& relation,
...
-  for (size_t i=0;i<retVal.second.size();i++){
-    vertPos_.push_back(glm::vec3(retVal.second[i].x, retVal.second[i].y, retVal.second[i].z));
-  }
-  numVert=vertPos_.size();
-  halfedge_=retVal.first;
+  ConvexHull hull = qh.getConvexHullAsMesh(pointCloudVec, false);
+  vertPos_.resize(hull.vertices.size());
+  manifold::transform(hull.vertices.begin(), hull.vertices.end(),
+                      vertPos_.begin(),
+                      [](const glm::dvec3& v) { return glm::vec3(v); });
```
elalish, line 496: Let's make a note that we can replace with this `move` once our `double`s PR lands.

src/manifold/src/impl.cpp
```diff
@@ -478,6 +478,37 @@ Manifold::Impl::Impl(const Mesh& mesh, const MeshRelationD& relation,
...
+           countAt(halfedge_.size()),
+           [this](size_t i) { halfedge_[i].face = i / 3; });
+  meshRelation_.originalID = ReserveIDs(1);
+  CalculateBBox();
+  SetPrecision(bBox_.Scale() * kTolerance);
+  SplitPinchedVerts();
+  CalculateNormals();
+  InitializeOriginal();
+  CreateFaces({});
+  SimplifyTopology();
```
elalish, line 508: I agree with your analysis, but let's leave it to the next PR to optimize exactly what's needed here.

src/manifold/src/quickhull.h
```diff
@@ -385,155 +317,6 @@ class MeshBuilder {
   }
 };
 
-// ConvexHull.hpp
-
-class ConvexHull {
```
elalish, line 390: 💯 - removing big piles of code is the best!

elalish (approved):
This looks fantastic! I'd appreciate clearing out any `std::cout` usages, but I think this is pretty much ready to merge.

pca006132:
I am not sure if we can skip `CalculateNormals`, as we still need the normal vector in `Impl` for further operations. For `SimplifyTopology`, apart from those degenerate cases, I think it can be helpful when there are points closer than our precision value?

But other than that, I think this is ready for merging. Nice job!

Kushal-Shah-03:
I've removed the `std::cout` and replaced it with `DEBUG_ASSERT`, and I also added the comment as TODO where we wanted to change the code when `double` PR get's accepted.

pca006132:
I moved things around to reduce the size of the header file, also changed some code to make them simpler.

pca006132:
I just went ahead and removed the custom `HalfEdge` data structure, and modified the halfedge reordering code. It should now be simpler.

I should give these as comments but I figured that it would be faster to just make the changes...

Kushal-Shah-03:
Thanks for the changes. I agree that it simplifies the code a lot. Also looking at these changes also helped me learn how to simplify the code in the future, thanks.

pca006132:
The compilation error is probably caused by missing include, instead of conditional compilation.

src/utilities/include/public.h
```diff
@@ -598,7 +599,6 @@ class Quality {
  *  @brief Custom Exceptions
  * @{
  */
-#ifdef MANIFOLD_DEBUG
```
elalish, line 601: @pca006132 Is this okay to remove? I thought this was required to make us able to compile with exceptions disabled?

elalish (approved):

src/utilities/include/public.h
pca006132: It is fine, the header files are still there even when we use `-fno-exceptions`. We just cannot throw them. I tried compiling it and it is fine.

src/manifold/src/quickhull.h
```diff
@@ -55,37 +55,37 @@
...
+#include "shared.h"
+#include "vec.h"
 
-// Pool.hpp
+namespace manifold {
 
+class ConvexHull {
+ public:
+  Vec<Halfedge> halfedges;
+  std::vector<glm::dvec3> vertices;
```
pca006132, line 69: I found that I forgot to change this to `Vec`.

src/manifold/src/quickhull.h
```diff
@@ -658,27 +246,28 @@ class QuickHull {
...
 
   // Temporary variables used during iteration process
-  std::vector<size_t> newFaceIndices;
-  std::vector<size_t> newHalfEdgeIndices;
-  std::vector<std::unique_ptr<std::vector<size_t>>> disabledFacePointVectors;
-  std::vector<size_t> visibleFaces;
-  std::vector<size_t> horizonEdgesData;
+  Vec<size_t> newFaceIndices;
+  Vec<size_t> newHalfedgeIndices;
+  std::vector<std::unique_ptr<Vec<size_t>>> disabledFacePointVectors;
```
pca006132, line 258: Maybe reorder this to make `std::vector` at the start/end? Just make it a bit more consistent...

src/manifold/src/quickhull.h
```diff
@@ -658,27 +246,28 @@ class QuickHull {
...
+  Vec<size_t> visibleFaces;
+  Vec<size_t> horizonEdgesData;
   struct FaceData {
```
pca006132, line 261: and maybe move this struct definition to the start of the outer class.

src/manifold/src/quickhull.h
```diff
@@ -780,51 +321,15 @@ class QuickHull {
   //   distance to a plane to consider a point being on positive side of it (for
   //   a point cloud with scale 1)
   // Returns:
-  //   Convex hull of the point cloud as a mesh object with half edge structure.
-  HalfEdgeMesh getConvexHullAsMesh(const double* vertexData, size_t vertexCount,
-                                   bool CCW, double eps = defaultEps());
+  //   Convex hull of the point cloud as halfEdge vector and vertex vector
+  ConvexHull getConvexHullAsMesh(const Vec<glm::dvec3>& pointCloud, bool CCW,
```
pca006132, line 325: Questions: Is it useful to keep CCW here? Can we pass the mesh epsilon from our mesh function?

No need to fix it in this PR, just food for thought.

src/manifold/src/quickhull.cpp
```diff
@@ -18,56 +18,199 @@
...
 
 #include <algorithm>
-#include <cassert>
-#include <cmath>
-#include <iostream>
 #include <limits>
 
+namespace manifold {
+
 double defaultEps() { return 0.0000001; }
```
pca006132, line 25: Should we change to `kTolerance`? After the double precision PR the epsilon will be much smaller though, we should see how this default epsilon affects precision and runtime.

We can do it in a later PR.

src/manifold/src/quickhull.cpp
```diff
@@ -573,3 +778,40 @@ void QuickHull::setupInitialTetrahedron() {
     }
   }
 }
+
+std::unique_ptr<Vec<size_t>> QuickHull::getIndexVectorFromPool() {
+  auto r = indexVectorPool.get();
+  r->resize(0);
```
pca006132, line 784: This resize actually causes reallocation, so currently the pool is pretty much useless. We should try and see if this pool actually helps (by implementing an alternative resize function in `vec.h` to avoid it from shrinking) or we can remove it all together.

pca006132: Anyway, it seems that the pool magically helps despite this resize will cause reallocation, and changing this resize to something that does not perform reallocation will not help. I guess it is about the size of `Face` and allocating empty unique pointers...

Should further investigate later.

pca006132:
One potential direction for optimization is to look at splitting `std::vector<Face>` to multiple vectors (vec of structs to struct of vecs), as `Face` is now a huge struct, and some of the tight loops are not using all fields of `Face` and may benefit from better spatial locality from struct of vecs.

Kushal-Shah-03:
That's an excellent idea, I think I'll just push a final commit making the modifications that you suggested, then we should move to a new PR, where we can start discussing these points related to optimization, because otherwise these two PR's will get mixed up, and we'll loose track of our ideas. So, once this PR get's merged I'll open another PR where I'll add the tracy headers, and then we can look into various optimizations there, does that sound alright?

Kushal-Shah-03:
I've made the changes that were suggested. I'm also testing this code on the Thingi10K dataset again, just in case we modified something incorrectly leading to bugs. I'll update once it's done.

Kushal-Shah-03:
Ok, so this comes as a surprise to me, but while testing, this test case file `101213.stl`, seems to go in an infinite loop (since it never caused a timeout before and since the time was 0.0006 seconds before, I doubt it's due to a timeout). I am looking to investigate as to what change could have caused this issue. @pca006132 it seemed to work after my [last commit](https://github.com/elalish/manifold/pull/893/commits/8b3963e8ad8ba7762ba3d3be719d190f2cb0bad2) , any places you would suggest looking at? (I am thinking I'll start by going back commits one by one to see which commit caused the issue and then we can try to find what caused it?)

Kushal-Shah-03:
> Ok, so this comes as a surprise to me, but while testing, this test case file `101213.stl`, seems to go in an infinite loop (since it never caused a timeout before and since the time was 0.0006 seconds before, I doubt it's due to a timeout). I am looking to investigate as to what change could have caused this issue. @pca006132 it seemed to work after my [last commit](https://github.com/elalish/manifold/pull/893/commits/8b3963e8ad8ba7762ba3d3be719d190f2cb0bad2) , any places you would suggest looking at? (I am thinking I'll start by going back commits one by one to see which commit caused the issue and then we can try to find what caused it?)

So, to clarify it wasn't a timeout, it was a segfault, I apologize for that miscommunication, I've identified that it's occurring in `SplitPinchedVerts()` (there were no segfaults in the algorithm call) https://github.com/elalish/manifold/blob/74e15b1574ebe6ae01d1fd2cffbe75aeeb7b8fab/src/manifold/src/edge_op.cpp#L671C1-L675C8 (Somewhere here).
I understand that it's probably not an error there, it's probably related to some other bug in the algorithm, and I will try to check out older commits and see if I can find the issue

Kushal-Shah-03:
Alright, so it's one of d9edae68f74edf64ff442adb5b986ab80774b2bd or 76d4126987b862c33daed2a3141e1c14e851db12 that caused the issue. The former wasn't compiling on my system, but the latter is the first time the file `101213.stl` leads to a segfault. @pca006132 there are lots of changes in these two commits, since you made them do you have any ideas as to what could have caused it? Because I am slightly lost as to where I should look for it, since I don't quite understand all the simplifications you made to the code.

pca006132:
Interesting. To start you can call the ismanifold method to check if the halfedge vector is wrong. I am busy these two days so I cannot debug this for now, sorry about that.

Kushal-Shah-03:
> Interesting. To start you can call the ismanifold method to check if the halfedge vector is wrong. I am busy these two days so I cannot debug this for now, sorry about that.

No problem, I had some spare time and tried to understand the code as much as I could, and I noticed that when you're reordering the halfedges, you've assumed that all faces are valid (and there will be no disabled faces, which is not the case). I've added return when the face is disabled or the halfedge is disabled and since the all the faces are not being considered now, I used j from 0 incrementing it by 3 instead of the face_index*3. Good news is that it is working on all cases now, with the same accuracy.
Could you go through it once, and see if the code follows how you wanted to simplify it?

src/manifold/src/quickhull.cpp
```diff
@@ -85,34 +228,92 @@ void QuickHull::buildMesh(const VertexDataSource& pointCloud, bool CCW,
...
+  Vec<int> counts(mesh.halfedges.size(), 0);
+  Vec<int> mapping(mesh.halfedges.size());
+
+  // reorder halfedges
+  size_t j = 0;  // Since all faces are not used now (so we should start from 0
+                 // and just increment, we can later set the face id as index/3
+                 // for each halfedge)
+  for_each(autoPolicy(mesh.halfedges.size()), countAt(0_uz),
+           countAt(mesh.halfedges.size()), [&](size_t i) {
+             if (mesh.halfedges[i].pairedHalfedge < 0) return;
```
pca006132, line 254: This part will not work when run in parallel. You can change the autoPolicy to ExecutionPolicy::Seq for now and optimize that later.

Kushal-Shah-03:
I've checked it again on the dataset, it's working correctly. I think the PR is ready to merge now.

src/manifold/src/quickhull.cpp
```diff
@@ -85,34 +228,94 @@ void QuickHull::buildMesh(const VertexDataSource& pointCloud, bool CCW,
...
-  buildMesh(pointCloud, CCW, useOriginalIndices, epsilon);
-  return ConvexHull(mesh, originalVertexData, CCW, useOriginalIndices);
+  // reorder halfedges
+  Vec<Halfedge> halfedges(mesh.halfedges.size());
+  Vec<int> counts(mesh.halfedges.size(), 0);
+  Vec<int> mapping(mesh.halfedges.size());
+  Vec<int> faceMap(mesh.faces.size());
+
+  // reorder halfedges
+  // Since all faces are not used now (so we should start from 0 and just
```
pca006132, line 250: This is a bit confusing to read. I think this means 'some faces are disabled and should not go into the halfedge vector'.

src/manifold/src/quickhull.cpp
```diff
@@ -85,34 +228,94 @@ void QuickHull::buildMesh(const VertexDataSource& pointCloud, bool CCW,
...
+  // increment, we can later set the face id as index/3 for each halfedge
+  int j = 0;
+  for_each(
+      autoPolicy(mesh.halfedges.size()), countAt(0_uz),
+      countAt(mesh.halfedges.size()), [&](size_t i) {
+        if (mesh.halfedges[i].pairedHalfedge < 0) return;
+        if (mesh.faces[mesh.halfedges[i].face].isDisabled()) return;
+        if (AtomicAdd(counts[mesh.halfedges[i].face], 1) > 0) return;
+        int currIndex = AtomicAdd(j, 3);
```
pca006132, line 259: Atomic works here, but typically I would avoid this kind of frequent atomic updates to the same variable as it is quite expensive. If you think about it, the CPU must synchronize all access to this variable, and it is quite frequent. Anyway, just food for thought and I should have actual benchmark data supporting my claim (or the null hypothesis, i.e. atomic add is good enough).

And it is nice that you can figure this out! This is the interesting thing about parallel programming!

src/manifold/src/quickhull.cpp
```diff
@@ -85,34 +228,94 @@ void QuickHull::buildMesh(const VertexDataSource& pointCloud, bool CCW,
...
+  for_each(
+      autoPolicy(halfedges.size()), halfedges.begin(), halfedges.end(),
+      [&](Halfedge& he) { he.pairedHalfedge = mapping[he.pairedHalfedge]; });
+  counts.resize(pointCloud.size() + 1);
+  fill(counts.begin(), counts.end(), 0);
+
+  // remove unused vertices
+  for_each(autoPolicy(halfedges.size() / 3), countAt(0_uz),
+           countAt(halfedges.size() / 3), [&](size_t i) {
+             AtomicAdd(counts[halfedges[3 * i].startVert], 1);
```
pca006132, line 285: Food for thought: maybe atomic set will be better for performance, and we don't need the transform iterator later.

src/manifold/src/quickhull.cpp
```diff
@@ -85,34 +228,94 @@ void QuickHull::buildMesh(const VertexDataSource& pointCloud, bool CCW,
...
+               vertices[counts[i]] = pointCloud[i];
+             }
+           });
+  for_each(autoPolicy(halfedges.size()), halfedges.begin(), halfedges.end(),
+           [&](Halfedge& he) {
+             he.startVert = counts[he.startVert];
+             he.endVert = counts[he.endVert];
+           });
+  // setting face id
+  for (size_t index = 0; index < halfedges.size(); index++) {
```
pca006132, line 305: parallel? and combine with the loop before?

test/hull_test.cpp
```diff
@@ -52,6 +52,7 @@ bool isMeshConvex(manifold::Manifold hullManifold, double epsilon = 0.0000001) {
 
       // If any vertex lies on the opposite side of the normal direction
       if (distance > epsilon) {
+        std::cout << distance << std::endl;
```
pca006132, line 55: remove this?

src/manifold/src/quickhull.cpp
```diff
@@ -212,10 +418,19 @@ void QuickHull::createConvexHalfEdgeMesh() {
...
+        if ((*tf.pointsOnPositiveSide)[index] == activePointIndex) {
+          change_flag = 1;
+        } else if (change_flag == 1) {
+          change_flag = 2;
+          (*tf.pointsOnPositiveSide)[index - 1] =
+              (*tf.pointsOnPositiveSide)[index];
+        }
+      }
+      if (change_flag == 1)
+        tf.pointsOnPositiveSide->resize(tf.pointsOnPositiveSide->size() - 1);
```
pca006132, line 432: we can use `pop_back()`, I forgot to fix this.

src/manifold/src/impl.cpp
```diff
@@ -478,6 +478,35 @@ Manifold::Impl::Impl(const Mesh& mesh, const MeshRelationD& relation,
   Finish();
 }
 
+void Manifold::Impl::Hull(const std::vector<glm::vec3>& vertPos) {
```
pca006132, line 481: btw why are we taking `std::vector` here? shouldn't we take `VecView`?

pca006132:
oops, we can address the comments in the next PR I guess

elalish:
oops, sorry @pca006132! Yeah, let's address in the next PR - this discussion is getting too long to keep track of.

Kushal-Shah-03:
Yeah I agree with @elalish, it's gotten a bit hard to keep track of the conversation. I'll open another PR, where I will add the tracy headers and do some of the changes you suggested, then we can look into it further.

Kushal-Shah-03:
> Great job working together on this! If it's not too onerous, it would be great to include the test object that caught this error in our `TEST` set, but let's do that as a follow-on.

That's a great idea, it's a testcase where the algorithm disables some faces, so adding that case will ensure we've handled disabling the faces correctly.

PR #894, opened 2024-08-11
SDF collapse

elalish:
This makes the SDF faster and higher quality by collapsing narrow triangles before they are even built by moving grid points near the surface to the surface. I've also moved `LevelSet` from `MeshGL` to `Manifold` as a static constructor. This allows me to build a manifold faster by avoiding some of our default construction slowness - in particular `CreateFaces()`. If the faces are needed, `AsOriginal()` can be used, but we should probably make a similar method with a clearer name. 

I'm still having some trouble with the Python bindings and one test.


bindings/python/manifold3d.cpp
```diff
@@ -433,6 +433,24 @@ NB_MODULE(manifold3d, m) {
...
+
+            std::function<float(glm::vec3)> cppToPython = [&f](glm::vec3 v) {
+              return f(v.x, v.y, v.z);
+            };
+            return Manifold::LevelSet(cppToPython, bound, edgeLength, level,
+                                      precision, false);
+          },
+          nb::arg("f"), nb::arg("bounds"), nb::arg("edgeLength"),
+          nb::arg("level") = 0.0, nb::arg("precision") = -1,
+          manifold__level_set__sdf__bounds__edge_length__level__precision)
```
elalish, line 453: @wrongbad I can't seem to get this doc string to be found - any idea what I'm doing wrong?

bindings/python/manifold3d.cpp
elalish, line 1: @pca006132 I'm now getting a bunch of compile errors about lambdas in this file on parts I haven't touched. Can you repro? I'm wondering if I got some update that's being more picky now or something.

elalish:
The only test I'm having trouble with locally is `Smooth.SineSurface`, but it turns out that's been flaky since #828, so I'm going to fix that now in a separate PR.

bindings/python/manifold3d.cpp
pca006132: I think you need `manifold__level_set__sdf__bounds__edge_length__level__precision__can_parallel`.

bindings/python/manifold3d.cpp
pca006132: No, I can't reproduce. Maybe you should try to clean build it or check if the nanobind submodule is up-to-date?

src/manifold/src/impl.h
```diff
@@ -108,7 +108,7 @@ struct Manifold::Impl {
   SparseIndices VertexCollisionsZ(VecView<const glm::vec3> vertsIn,
                                   bool inverted = false) const;
 
-  bool IsEmpty() const { return NumVert() == 0; }
+  bool IsEmpty() const { return NumTri() == 0; }
```
pca006132, line 111: Just curious, does this change affect anything?

src/manifold/src/impl.h
```diff
@@ -150,13 +150,14 @@ struct Manifold::Impl {
...
   // edge_op.cu
+  void CleanupTopology();
   void SimplifyTopology();
   void DedupeEdge(int edge);
   void CollapseEdge(int edge, std::vector<int>& edges);
   void RecursiveEdgeSwap(int edge, int& tag, std::vector<int>& visited,
                          std::vector<int>& edgeSwapStack,
                          std::vector<int>& edges);
-  void RemoveIfFolded(int edge);
+  bool RemoveIfFolded(int edge);
```
pca006132, line 160: Do we need this boolean return value?

bindings/python/manifold3d.cpp
wrongbad: I haven't looked in detail yet, but in general you can find the auto-generated header in the cmake build folder, and look for the exact var-name. If it's missing then the source file might not be white-listed by the script yet, or the comment formatting is different somehow.

src/manifold/src/impl.h
elalish: I was getting manifolds from `Decompose` that had one vertex and no triangles, but they weren't marked empty. I fixed the underlying problem too, but this still feels like a better definition.

src/manifold/src/impl.h
elalish: Not anymore - will revert.

bindings/python/manifold3d.cpp
elalish: nevermind, seems to be working now.

elalish:
@pca006132 I think this is PR should be ready, but I have a strange problem I don't understand. As you can see, `BooleanComplex.MeshRelation` is throwing a Vec out of Range on some of these tests. I can repro this locally, but only when I run all the tests. When I run it by itself (or even all of BooleanComplex) it passes just fine. Stranger yet, I compiled with address sanitizer enabled and ran the whole test suite - took a few minutes, but everything passed with no spew! Any idea what might be going on?

pca006132:
you can use this patch to catch the error with gdb/lldb, and do the debugging:

```
diff --git a/src/utilities/include/optional_assert.h b/src/utilities/include/optional_assert.h
index 9a8f9ba8..ae2e5ac2 100644
--- a/src/utilities/include/optional_assert.h
+++ b/src/utilities/include/optional_assert.h
@@ -42,7 +42,7 @@ void Assert(bool condition, const char* file, int line, const std::string& cond,
 
 #ifdef MANIFOLD_EXCEPTIONS
 #define ASSERT(condition, EX) \
-  if (!(condition)) throw(EX);
+  if (!(condition)) __builtin_debugtrap();
 #else
 #define ASSERT(condition, EX)
 #endif
```

and it feels like some sort of error caused by parallelization? just a guess.

pca006132:
It seems that it is not related to the partition cache, and not related to parallelization (in the subdivide part). Maybe it is something caused by other tests modifying some global variables?

pca006132:
OK, I think I have some idea about why. The issue is due to reading from uninitialized memory in `Vec`, which address sanitizer cannot catch (requires MSan, but it is hard to use as we need to instrument everything including libc++), and this bug is hard to reproduce because it depends on the uninitialized memory value.

I will try to look at where is assuming the memory to be initialized when it shouldn't.

pca006132:
And while making `Vec`'s constructor always return something initialized can fix the symptom, it is not a proper fix because the bug is caused by reading something undefined in the algorithm...

elalish:
Interesting! Any any what line is reading from uninitialized data? Sounds like I made an error somewhere.

pca006132:
```diff
diff --git a/src/manifold/src/subdivision.cpp b/src/manifold/src/subdivision.cpp
index 8f052d47..fe1e7a01 100644
--- a/src/manifold/src/subdivision.cpp
+++ b/src/manifold/src/subdivision.cpp
@@ -515,6 +515,8 @@ Vec<Barycentric> Manifold::Impl::Subdivide(
                  numVert);
 
   Vec<Barycentric> vertBary(edgeOffset.back() + edgeAdded.back());
+  for (size_t i = 0; i < vertBary.size(); i++)
+    vertBary[i].tri = -1;
   const int totalEdgeAdded = vertBary.size() - numVert;
   FillRetainedVerts(vertBary);
   for_each_n(policy, countAt(0), numEdge,
@@ -621,6 +623,7 @@ Vec<Barycentric> Manifold::Impl::Subdivide(
   for_each_n(policy, countAt(0), vertBary.size(),
              [&newVertPos, &vertBary, &faceHalfedges, this](const int vert) {
                const Barycentric bary = vertBary[vert];
+               ASSERT(bary.tri != -1, logicErr(""));
                const glm::ivec4 halfedges = faceHalfedges[bary.tri];
                if (halfedges[3] < 0) {
                  glm::mat3 triPos;
```

Found it, some `vertBary` are uninitialized and is still being used.

elalish:
I made another vicious test object:
image
This shows the power of the little heuristic I added: I only collapse verts if at most 3 of the 7 edge directions have intersections on both sides. This means we preserve the extra verts for thin surfaces and narrow details, but still remove effectively all narrow triangles on smoothly curved surfaces. It cuts the LevelSet time down quite a bit too, simply because it outputs so many fewer verts and tris.

hrgdavor:
Wow

PR #904, opened 2024-08-22
Cleaning Hull Algorithm Code

Kushal-Shah-03:
I've removed some redundant functions and structs, I also removed some comments. I had made some changes in the previous PR, which also removed some redundancy in the code. Should I make the same changes here? (like modifying the `for_each` loop to merge two for loops)

I was thinking I can pass the `vertPos_` and `halfedge_` to the function as arguments and then modify them inside the quickhull implementation to reduce the size of the `impl` Hull function and also reduce the `ConvexHull` class. What do you think?

### DIFF
--- a/src/manifold/src/impl.cpp
+++ b/src/manifold/src/impl.cpp
@@ -488,13 +488,8 @@ void Manifold::Impl::Hull(const std::vector<glm::vec3>& vertPos) {
   Vec<glm::dvec3> pointCloudVec(numVert);
   manifold::transform(vertPos.begin(), vertPos.end(), pointCloudVec.begin(),
                       [](const glm::vec3& v) { return glm::dvec3(v); });
-  ConvexHull hull = getConvexHullAsMesh(pointCloudVec, false);
-  // TODO : Once double PR lands, replace this with move
-  vertPos_.resize(hull.vertices.size());
-  manifold::transform(hull.vertices.begin(), hull.vertices.end(),
-                      vertPos_.begin(),
-                      [](const glm::dvec3& v) { return glm::vec3(v); });
-  halfedge_ = std::move(hull.halfedges);
+  QuickHull qh(pointCloudVec);
+  std::tie(halfedge_, vertPos_) = qh.buildMesh();
   meshRelation_.originalID = ReserveIDs(1);
   CalculateBBox();
   SetPrecision(bBox_.Scale() * kTolerance);
--- a/src/manifold/src/impl.cpp
+++ b/src/manifold/src/impl.cpp
@@ -478,29 +478,6 @@ Manifold::Impl::Impl(const Mesh& mesh, const MeshRelationD& relation,
   Finish();
 }
 
-void Manifold::Impl::Hull(const std::vector<glm::vec3>& vertPos) {
-  size_t numVert = vertPos.size();
-  if (numVert < 4) {
-    status_ = Error::InvalidConstruction;
-    return;
-  }
-
-  Vec<glm::dvec3> pointCloudVec(numVert);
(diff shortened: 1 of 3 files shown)


src/manifold/src/quickhull.cpp
```diff
@@ -826,4 +818,11 @@ bool QuickHull::addPointToFace(typename MeshBuilder::Face& f,
   }
   return false;
 }
+
+ConvexHull getConvexHullAsMesh(const Vec<glm::dvec3>& pointCloud, bool CCW,
+                               double epsilon) {
+  Vec<glm::dvec3> pointCloudVec(pointCloud);
+  QuickHull qh(pointCloudVec);
+  return qh.buildMesh(CCW, epsilon);
```
elalish, line 826: Let's get rid of CCW too - CCW is always positive in Manifold.

src/manifold/src/impl.cpp
```diff
@@ -488,8 +488,7 @@ void Manifold::Impl::Hull(const std::vector<glm::vec3>& vertPos) {
   Vec<glm::dvec3> pointCloudVec(numVert);
   manifold::transform(vertPos.begin(), vertPos.end(), pointCloudVec.begin(),
                       [](const glm::vec3& v) { return glm::dvec3(v); });
-  QuickHull qh(pointCloudVec);
-  ConvexHull hull = qh.getConvexHullAsMesh(pointCloudVec, false);
+  ConvexHull hull = getConvexHullAsMesh(pointCloudVec, false);
```
elalish, line 491: You might just merge `getConvexHullAsMesh` into `Impl::Hull` and put this combined implementation in `quickhull.cpp`. Then you can remove `ConvexHull` and directly construct `halfedge_` and `vertPos_`.

elalish (commented):
You should make your changes in one PR or the other - you don't want to generate a bunch of merge conflicts for yourself. Do whatever feels easiest to keep track of.

Kushal-Shah-03:
I was going through the `Face` struct to see how we could break it into struct of vectors, I noticed that 
`isVisibleFaceOnCurrentIteration`, `visibilityCheckedOnIteration` and `horizonEdgesOnCurrentIteration` are accessed close to each other.
and 
`inFaceStack`, `mostDistantPoint`, `mostDistantPointDist` and `pointsOnPositiveSide` are also accessed close to each other.
`he` and `P` are accessed close to all the variables.

So maybe we can break the `Face` struct into these three parts?

src/manifold/src/impl.cpp
Kushal-Shah-03: I have merged `getConvexHullAsMesh` into `Impl::Hull` and removed `ConvexHull`. I think it's better to keep the `Hull` function inside `Impl`, since it follows a similar flow to the other functions inside `Impl`. Although if you think it's better to put it into `quickhull.cpp`, I can make that change.

elalish: I mean both - notice how our `Impl` functions are implemented across many files? You can implement `Impl::Hull` in `quickhull.cpp` if you want, so long as it imports `impl.h`.

pca006132:
> I was going through the `Face` struct to see how we could break it into struct of vectors, I noticed that `isVisibleFaceOnCurrentIteration`, `visibilityCheckedOnIteration` and `horizonEdgesOnCurrentIteration` are accessed close to each other. and `inFaceStack`, `mostDistantPoint`, `mostDistantPointDist` and `pointsOnPositiveSide` are also accessed close to each other. `he` and `P` are accessed close to all the variables.
> 
> So maybe we can break the `Face` struct into these three parts?

I think you probably want to do this in the optimization PR.

elalish (approved):
This is looking like great progress to me - @pca006132 any changes you want to see before we merge?

pca006132:
LGTM

PR #907, opened 2024-09-01
implemented typedoc documentation

msantic:
Addresses #906


bindings/wasm/package.json
```diff
@@ -3,6 +3,9 @@
   "version": "2.5.1",
   "description": "Geometry library for topological robustness",
   "main": "manifold.js",
+  "scripts": {
+    "docs": "npx typedoc"
```
elalish, line 7: Hmm, are all the extra files here created automatically by running this command? If so, I think instead of checking them in to the repo, we should just tweak our Github Actions CI to run this script as part of `deploy-documentation` and copy them straight into our served Pages. Would that be reasonable?

elalish (commented):
Any trouble with signing the CLA? It should be automatic as an individual.

bindings/wasm/package.json
msantic: @elalish I was not sure what do you prefer but sure, instead of checking them in to the rep we could move them to GitHub Action CI. Would you like me to do it, or would you prefer to handle it? I'm not yet familiar with your file organization.

Btw I've resolved CLA and removed bindings/wasm/docs folder from the repo.

elalish:
It'd be great if you can take a stab at it - I think you'll probably just have to modify [deploy.yml](https://github.com/elalish/manifold/blob/master/.github/workflows/deploy.yml) to build that site and put it in another directory under `./public`. I'll take care of trouble-shooting it when we merge, since there will inevitably be some path error with the website the first time it deploys. 

I realized we didn't copy our docs for `Mesh` into our TS [declarations](https://github.com/elalish/manifold/blob/master/bindings/wasm/manifold-encapsulated-types.d.ts#L943-L964), which is why they're missing in your website. Would you mind copying them over and checking the page gets generated correctly? While you're at it, it would be great to have a second set of eyes on the rest of those doc comments in case they've gotten out of sync with the C++.

elalish:
Thanks! The TS `Mesh` is `MeshGL` - the old C++ `Mesh` is deprecated.

msantic:
@elalish I went through the entire doc comments and already fixed a few. Here’s what is still missing: 

- resetToCircularDefaults - ResetToDefaults missing doc in C++, public.h:579
- Mesh methods:
- merge(): boolean;
- verts(tri: number): SealedUint32Array<3>;
- position(vert: number): SealedFloat32Array<3>;
- extras(vert: number): Float32Array;
- tangent(halfedge: number): SealedFloat32Array<4>;
- transform(run: number): Mat4;

msantic:
@elalish Yes, the definitions are entirely missing. I could try to add doc comments, but I think it would be better for someone more familiar with the internal logic to do it.

PR #908, opened 2024-09-01
Added DisabledFaceTest

Kushal-Shah-03:
I added the narrowed down version (12 points) of the test case `101213.stl`. It is a case where one of the `halfEdge` in the list of halfEdges has a disabled face. This case caused a segfault before, and it's a good check to identify if disabled faces are handled correctly. Thanks, @elalish for suggesting this. Let me know if I should change anything.

### DIFF
--- a/test/hull_test.cpp
+++ b/test/hull_test.cpp
@@ -198,3 +198,28 @@ TEST(Hull, FailingTest2) {
 #endif
   EXPECT_TRUE(isMeshConvex(hull, 2.13966e-05));
 }
+
+TEST(Hull, DisabledFaceTest) {
+  // 101213.stl
+  // Replace glm::vec3 with vec3 later once Double PR merges
+  const std::vector<glm::vec3> hullPts = {
+      {65.398902893, 58.303115845, 58.765388489},
+      {42.147319794, 44.512584686, 75.703102112},
+      {89.208251953, 97.092460632, 41.632453918},
+      {69.860748291, 69.860748291, 56.492958069},
+      {45.375354767, 39.067985535, 64.844772339},
+      {26.555616379, 18.671405792, 81.067504883},
+      {88.179382324, 81.083595276, 43.981628418},
+      {51.823883057, 50.247039795, 70.359062195},
+      {58.489616394, 72.681190491, 51.274829865},
+      {110, 10, 65},
+      {29.590316772, 20.917686462, 73.143547058},
+      {101.61526489, 98.461585999, 30.909877777}};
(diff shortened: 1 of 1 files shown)


test/hull_test.cpp
```diff
@@ -197,4 +197,29 @@ TEST(Hull, FailingTest2) {
...
+      {110, 10, 65},
+      {29.590316772, 20.917686462, 73.143547058},
+      {101.61526489, 98.461585999, 30.909877777}};
+  auto hull = Manifold::Hull(hullPts);
+#ifdef MANIFOLD_EXPORT
+  if (options.exportModels) {
+    ExportMesh("disabledFaceTest.glb", hull.GetMesh(), {});
+  }
+#endif
+  EXPECT_TRUE(isMeshConvex(hull));
```
elalish, line 224: Does `isMeshConvex` return true for an empty mesh? If so, we should probably also expect `!hull.isEmpty()` too.

test/hull_test.cpp
```diff
@@ -197,4 +197,29 @@ TEST(Hull, FailingTest2) {
   }
 #endif
   EXPECT_TRUE(isMeshConvex(hull, 2.13966e-05));
+}
+
+TEST(Hull, DisabledFaceTest) {
```
elalish, line 202: When you say it has a disabled face, do you mean during the coarse of the algorithm it disables a face? Did you verify this test fails if you revert whatever the fix was?

elalish (approved):
Thanks!

test/hull_test.cpp
Kushal-Shah-03: Yes, and that disabled face is linked to one of the `halfEdge` returned after the `CreateConvexHalfedgeMesh` function call. The algorithm doesn't remove them during computation to reduce time. But, we need to filter those at the end.

Yes, I verified that the test fails without the fix.

test/hull_test.cpp
Kushal-Shah-03: Yeah it does. I have added that as well. Thanks for pointing that out.

elalish:
Thanks! Have you had a chance to look at `Hull.Tictac` yet? Any idea where it's losing that vert?

Kushal-Shah-03:
To be honest I still don't understand why is that considered an error. Isn't it possible that it drops some verts that lie outside or on the surface of the hull?

elalish:
It's because it is two axis-aligned spheres. If you drop the n down to, say 12, and look at the result, I think it'll become clear. To be fair though, getting a repro of that windows problem is a higher priority.

PR #911, opened 2024-09-03
Fix Double Error On Windows

### LINKED ISSUE #910: Fix Windows CI
For some reason #895 caused our Windows CI to start inexplicably failing on two tests: `Polygon.Zebra` and `Polygon.Zebra3` with `Process exited with code 127`. It's a pretty opaque error, but 100% repro on Windows, 0% on all other platforms. For now we have a workaround - details in this [thread](https://github.com/elalish/manifold/pull/895#issuecomment-2323857439), but we should fix the root cause before release.

Kushal-Shah-03:
Fixes #910, hopefully.

I had an idea through my experimentation on Windows and just wanted to check if it's correct.


Kushal-Shah-03:
@elalish @pca006132 I might have misinterpreted the outputs, but it seems to me like the error is merely caused by overflow of memory in the stack? and we would just need to allocate it in the heap? I was trying to fix it, but I wasn't sure about my interpretation of these errors. So wanted to get your approval first.

elalish:
Ah, a Stack Overflow - I suppose that could make sense since it's only happening on our largest input files and only after switching to double precision. But what is being allocated on the stack? The vector should just be a pointer to its heap allocation. I wonder if this is some idiosyncrasy of the way initializer lists are handled in our tests? By all means, if you have an idea, feel free to try it out here.

Kushal-Shah-03:
So, I think it's possible that when we allocate the polygons which are bigger in input. The entire vector declared is initially in the stack, since we're declaring it in one go?(I'm not sure if that's the case). So, we can maybe try to `push_back` each point in each `SimplePolygon`?

 Also, since this input is so large, why don't we write it all in a file and read from it by adding the points 1 by 1 in the process?

elalish:
>Also, since this input is so large, why don't we write it all in a file and read from it by adding the points 1 by 1 in the process?

Yeah, that's probably the right approach. Our current system was designed only to be very easy to copy-paste from our std out spew to a new test, not so much for extremely large polygons.

Kushal-Shah-03:
Alright, I had started some work on it, but I see that @pca006132 is already much further along than I am. If there's something I can help out with, ping me. Till then, I'll go back to my prior tasks.

elalish:
Closing in favor of #913.

PR #913, opened 2024-09-04
use custom file format for polygon tests

### LINKED ISSUE #910: Fix Windows CI

pca006132:
Fixes #910

see https://github.com/elalish/manifold/pull/911#issuecomment-2327446700


pca006132:
Probably need to use https://emscripten.org/docs/porting/files/packaging_files.html#packaging-files for wasm. Will try it later tonight.

test/polygons/conversion.py
```diff
@@ -0,0 +1,66 @@
+# Usage:
+# python conversion.py file.cpp output.txt
```
elalish, line 2: Do we need this script anymore if all the old test files have already been converted?

src/polygon/src/polygon.cpp
```diff
@@ -123,7 +123,7 @@ void CheckGeometry(const std::vector<ivec3> &triangles,
 
 void Dump(const PolygonsIdx &polys) {
   for (auto poly : polys) {
-    std::cout << "polys.push_back({" << std::setprecision(9) << std::endl;
+    std::cout << "polys.push_back({" << std::setprecision(15) << std::endl;
```
elalish, line 126: Seems like there's not much point dumping this format now - can we update it to the format you now use in the `txt` files?

test/polygon_fuzz.cpp
```diff
@@ -103,21 +104,39 @@ void TestPoly(Polygons polys, int _unused, float precision = -1.0) {
...
 std::vector<TestCase> SeedProvider() {
+  std::string file = __FILE__;
+  std::string dir = file.substr(0, file.rfind('/'));
+  auto f = std::ifstream(dir + "/polygons/" + "polygon_corpus.txt");
+  // for each test:
+  //   test name, expectedNumTri, precision, num polygons
+  //   for each polygon:
+  //     num points
+  //     for each vertex:
+  //       x coord, y coord
```
elalish, line 116: 👍

elalish (changes requested):
Thanks, that's a significant improvement! Remember to undo my change to `manifold.yml` that's skipping the zebra tests on windows.

test/polygons/conversion.py
pca006132: We don't need this if we dump the new format directly.

src/polygon/src/polygon.cpp
pca006132: Sure. My major concern is that the new format can be a bit hard to read (as I want to keep the parser simple), and the old c++ format is more self-explaining.

pca006132:
I have no idea why the formatting thing is now failing...

src/polygon/src/polygon.cpp
```diff
@@ -144,12 +144,12 @@ void PrintFailure(const std::exception &e, const PolygonsIdx &polys,
   std::cout << "-----------------------------------" << std::endl;
   std::cout << "Triangulation failed! Precision = " << precision << std::endl;
   std::cout << e.what() << std::endl;
-  if (triangles.size() > 1000) {
+  if (triangles.size() > 1000 && !PolygonParams().verbose) {
```
elalish, line 147: 👍

elalish (approved):
Thanks!

PR #914, opened 2024-09-04
Fix missing 'this' for property access in wasm bindings

dsafa:
The js bindings throw `numProp is not defined` when  trying to use `Mesh.position()` or `Mesh.extras()`. They were just missing a `this`

### DIFF
--- a/bindings/wasm/bindings.js
+++ b/bindings/wasm/bindings.js
@@ -368,12 +368,12 @@ Module.setup = function() {
     }
 
     position(vert) {
-      return this.vertProperties.subarray(numProp * vert, numProp * vert + 3);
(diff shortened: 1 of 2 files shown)


dsafa:
hmm running format.sh is not doing anything :/

elalish (approved):
Thanks! I generally just run clang-format inside of VSCode. If all else fails you can just copy the corrected lines from the CI's format checker and paste them in place.

dsafa:
Thanks, that works. fixed.

I took a look and bash needs `globstar` option set for `**` to work it seems. So I enabled it but then `node_modules` starts matching the pattern and I'm not familiar enough to know the best way to ignore that so I'll just leave this for anyone else if they encounter the same thing and want to fix it

elalish (approved):
Thanks!

PR #918, opened 2024-09-08
change directory structure

### LINKED ISSUE #701: header file naming/directory structure

### LINKED ISSUE #917: Expose Parallel Algorithms
Expose our parallel algorithm API, so other projects such as openscad can use it.

pca006132:
Closes #701. Closes #917.

1. Moved every (public) headers into the `manifold` directory.
2. Changed `public.h` to `common.h` and `par.h` to `parallel.h`.
3. Install `parallel.h` and `iters.h`


hzeller:
Nice! LGTM.

elalish (approved):
Sounds good. My only thoughts, and I'm not at all sure these are good, is would it be nice if all the public manifold headers were in the same directory? Or is that confusing? And likewise, would it be helpful to separate the public headers like `common.h` from the internal headers like `vec.h` into different directories?

pca006132:
Separating the headers sounds good, but I am not sure how to do this with our current module organization. Having 2 header directories in the same module seems a bit too cluttered to me. Anyway, that will not be public facing changes, so it should be fine to do it later I guess.

hzeller:
maybe there could be an `internal/` subdirectory. But also agree, that can be refined later.

PR #924, opened 2024-09-12
use mxe to cross compile to windows

pca006132:
I just noticed we were not actually using mxe properly to cross compile to windows, we were merely compiling in a docker environment with mxe installed.


manifoldDeps.cmake
```diff
@@ -3,6 +3,7 @@ include(GNUInstallDirs)
 find_package(PkgConfig QUIET)
 
 if(MANIFOLD_PAR STREQUAL "TBB")
+    find_package(Threads REQUIRED)
```
elalish, line 6: What's this? Why do we need this dependency now and didn't before?

pca006132: This is pthreads. Tbb always require pthreads, but for some reason with mxe it somehow does not detect pthread and we need to add this find package call. OpenSCAD does not have this call but it somehow works, I don't have much idea why we were getting errors regarding this...

elalish (approved):
Thanks!

PR #927, opened 2024-09-13
add MeshGL64

pca006132:
Added `MeshGL64` and removed usages of `Mesh` inside our library. `Mesh` is not yet removed because we use it extensively in tests, we should soon remove it soon though.

@elalish we have two failures related to refine and smooth though, they are `Smooth.ToLength` and `Samples.Scallop`. The stats are different significantly, so perhaps I did something wrong in my refactoring.

Note that we haven't yet exposed `MeshGL64` to C and python bindings, I think that can be done later. I did remove `Mesh` from the C binding.


pca006132:
And it seems that we miss a vertex normal calculation function that write normals to property channels? If I understand it correctly, `CalculateNormals` calculate face normals.

pca006132:
and I am thinking maybe we should add some simple iterators to MeshGL so we can easily iterate over vertex positions and tri verts as triples, without changing the memory representation.

elalish:
> and I am thinking maybe we should add some simple iterators to MeshGL so we can easily iterate over vertex positions and tri verts as triples, without changing the memory representation.

Yeah, good idea. That fits with the helper functions we put in the WASM bindings too.

samples/src/scallop.cpp
```diff
@@ -27,29 +27,35 @@ Manifold Scallop() {
...
     int j = i + 1;
     if (j == 2 * wiggles) j = 0;
 
     double smoothness = 1 - sharpness * glm::cos((theta + delta / 2) / 2);
     size_t halfedge = 3 * scallop.triVerts.size() + 1;
     sharpenedEdges.push_back({halfedge, smoothness});
-    scallop.triVerts.push_back({0, 2 + i, 2 + j});
+    scallop.triVerts.insert(
+        scallop.triVerts.end(),
+        {0, static_cast<uint32_t>(2 + i), static_cast<uint32_t>(2 + j)});
```
elalish, line 52: Can we skip the casting if we just define `i` and `j` as `uint32_t`?

src/manifold/include/manifold/manifold.h
```diff
@@ -106,18 +100,26 @@ struct MeshGL {
...
 };
+
+/**
+ * An alternative to Mesh for output suitable for pushing into graphics
+ * libraries directly. This may not be manifold since the verts are duplicated
+ * along property boundaries that do not match. The additional merge vectors
+ * store this missing information, allowing the manifold to be reconstructed.
+ */
+using MeshGL = MeshGLP<float>;
+using MeshGL64 = MeshGLP<double>;
```
elalish, line 122: 👍

src/manifold/src/impl.h
```diff
@@ -65,7 +65,183 @@ struct Manifold::Impl {
...
+
+    halfedgeTangent_.resize(meshGL.halfedgeTangent.size() / 4);
+    for (size_t i = 0; i < halfedgeTangent_.size(); ++i) {
+      for (const int j : {0, 1, 2, 3})
+        halfedgeTangent_[i][j] = meshGL.halfedgeTangent[4 * i + j];
+    }
+
+    if (meshGL.runOriginalID.empty()) {
+      // FIXME: should we do this?
+      meshRelation_.originalID = Impl::ReserveIDs(1);
```
elalish, line 149: Huh, it does seem to conflict with the statement at the end of the function. Does anything break if you remove it? I should probably think through the cases here...

src/manifold/src/impl.h
```diff
@@ -65,7 +65,183 @@ struct Manifold::Impl {
...
+      ivec3 tri;
+      for (const size_t j : {0, 1, 2}) {
+        const int vert = meshGL.triVerts[3 * i + j];
+        if (vert < 0 || vert >= static_cast<int>(numVert)) {
+          MarkFailure(Error::VertexOutOfBounds);
+          return;
+        }
+        tri[j] = prop2vert[vert];
+        // FIXME: do we need this check?
+        if (tri[j] < 0 || tri[j] >= static_cast<int>(numVert)) {
```
elalish, line 194: Agreed, looks like that's already checked above.

src/manifold/src/impl.h
```diff
@@ -65,7 +65,183 @@ struct Manifold::Impl {
...
+      }
+    }
+
+    Vec<ivec3> triVerts;
+    triVerts.reserve(numTri);
+    for (size_t i = 0; i < numTri; ++i) {
+      ivec3 tri;
+      for (const size_t j : {0, 1, 2}) {
+        const int vert = meshGL.triVerts[3 * i + j];
+        if (vert < 0 || vert >= static_cast<int>(numVert)) {
```
elalish, line 188: Can we avoid the cast by declaring `vert` as `size_t`?

elalish (commented):
This is looking great!

samples/src/scallop.cpp
pca006132: It depends if we need signed results for other things that depend on i and j, iirc there are results that can be negative before this line?

src/manifold/src/impl.h
pca006132: probably yes

pca006132:
Btw, MeshGLP is just a random naming, feel free to suggest another if you think of something better.

pca006132:
> And it seems that we miss a vertex normal calculation function that write normals to property channels? If I understand it correctly, `CalculateNormals` calculate face normals.

@elalish btw what about this? Am I overlooking some API or we should add this?

pca006132:
For `Smooth.ToLength`, if we switch to using `Smooth` with a `Mesh`, it works:

```diff
diff --git a/test/smooth_test.cpp b/test/smooth_test.cpp
index 8d11dd89..88efcba1 100644
--- a/test/smooth_test.cpp
+++ b/test/smooth_test.cpp
@@ -109,7 +109,7 @@ TEST(Smooth, ToLength) {
       CrossSection::Circle(10, 10).Translate({10, 0}).ToPolygons(), 2, 0, 0,
       {0, 0});
   cone += cone.Scale({1, 1, -5});
-  Manifold smooth = Manifold::Smooth(cone.GetMeshGL64());
+  Manifold smooth = Manifold::Smooth(cone.GetMesh());
   smooth = smooth.RefineToLength(0.1);
   ExpectMeshes(smooth, {{85250, 170496}});
   auto prop = smooth.GetProperties();
```

So maybe the mesh is not being simplified due to not being original, and it is somehow giving a different volume?
The weird thing though is that the volume difference is pretty large:

```
The difference between prop.volume and 4604 is 77.845313791107401, which exceeds 1, where
prop.volume evaluates to 4526.1546862088926,
4604 evaluates to 4604, and
1 evaluates to 1.
```

pca006132:
OK looking at the outputs it seems that they are indeed very different:

The expected one:
image

The current one with MeshGL:
image

It is like getting punched in the face XD

src/manifold/src/impl.h
```diff
@@ -144,8 +132,10 @@ struct Manifold::Impl {
         halfedgeTangent_[i][j] = meshGL.halfedgeTangent[4 * i + j];
     }
 
+    Vec<TriRef> triRef;
     if (meshGL.runOriginalID.empty()) {
-      // FIXME: should we do this?
+      // FIXME: This affects Impl::InitializeOriginal, and removing this
+      // apparently make things fail. Not sure if this is expected.
```
elalish, line 138: Yeah, I think I remember what's going on there actually. I think it's benign, but I should probably refactor it at some point to make the code clearer.

src/manifold/src/manifold.cpp
```diff
@@ -653,11 +655,20 @@ Manifold Manifold::SetProperties(
...
+                                   pImpl->vertPos_.data(), triProperties.data(),
+                                   pImpl->halfedge_.data(), propFunc}));
+    else
+      for_each_n(autoPolicy(NumTri()), countAt(0), NumTri(),
+                 UpdateProperties(
+                     {pImpl->meshRelation_.properties.data(), numProp,
+                      oldProperties.data(), oldNumProp, pImpl->vertPos_.data(),
+                      triProperties.data(), pImpl->halfedge_.data(),
+                      [](double* newProp, vec3 position,
+                         const double* oldProp) { *newProp = 0; }}));
```
elalish, line 671: Could you just replace `propFunc` with this function conditionally instead of having to duplicate the whole loop?

test/manifold_test.cpp
```diff
@@ -86,43 +67,39 @@ TEST(Manifold, ValidInput) {
...
-      if (tri[i] == 2) tri[i] = 4;
-    }
+  MeshGL in = TetGL();
+  for (uint32_t& triVert : in.triVerts) {
+    if (triVert == 2) triVert = 4;
   }
   Manifold tet(in);
   EXPECT_TRUE(tet.IsEmpty());
-  EXPECT_EQ(tet.Status(), Manifold::Error::VertexOutOfBounds);
+  EXPECT_EQ(tet.Status(), Manifold::Error::NotManifold);
```
elalish, line 102: This is surprising - the tetrahedron only has 4 verts, so this is clearly out of bounds. Are we checking for errors in a different order now?

test/manifold_test.cpp
```diff
@@ -364,47 +341,86 @@ TEST(Manifold, WarpBatch) {
...
-      {5, 3, 6},    {6, 4, 21},   {26, 22, 25}, {21, 25, 22}, {25, 15, 26},
-      {26, 6, 22},  {21, 4, 25},  {21, 22, 6},  {16, 26, 15}, {16, 6, 26},
-      {4, 15, 25},  {15, 1, 16},  {16, 2, 6},   {4, 1, 15},   {1, 2, 16},
-      {12, 14, 23}, {12, 13, 14}, {12, 11, 13}, {18, 9, 12},  {11, 7, 17},
-      {7, 9, 18},   {17, 7, 18},  {13, 11, 19}, {17, 18, 20}, {19, 11, 17},
-      {19, 17, 20}, {14, 13, 20}, {18, 12, 24}, {20, 13, 19}, {20, 18, 27},
-      {12, 10, 11}, {24, 12, 23}, {9, 10, 12},  {9, 8, 10},   {8, 11, 10},
-      {8, 7, 11},   {8, 9, 7},    {14, 20, 27}, {24, 28, 18}, {27, 18, 28},
-      {23, 14, 27}, {24, 23, 28}, {28, 23, 27}};
+  // TODO
```
elalish, line 344: What's the TODO?

elalish (commented):
As for `CalculateNormals`, I realize now it's a bit confusing since we have both a `Manifold` and an `Impl` method with that name that don't do the same thing. We should rename the Impl one - it calculates both face and vertex normals, but just the simple internal versions. The public method does much more complicated vertex normal calculations that are put into the public property channels.

So, I don't *think* we're missing anything. Certainly some tests should fail if we are.

elalish:
I'm pretty certain the issue with `Smooth.ToLength` is that certain triangle pairs used to be found to be coplanar and treated as quads for smoothing, but now they are not. Would be good to know if it's some kind of precision error or something else. `AsOriginal` may fix it, but I'd like to understand why it's behaving differently in the first place. Take a look at the `faceID`s right after `CreateFaces` runs to see the difference.

test/manifold_test.cpp
pca006132: `TetGL` has duplicated vertices. Or should we change this test to test vertex out of bounds?

pca006132:
> So, I don't _think_ we're missing anything. Certainly some tests should fail if we are.

Oh, I don't mean we miss some calculation, I just mean there is no simple way to get vertex normal with `MeshGL` and it makes it hard to get rid of `Mesh` in our tests.

test/manifold_test.cpp
pca006132: Ah, apparently we already have this check in `InvalidInput7`. And I should probably rename that to `InvalidInput6`.

test/manifold_test.cpp
pca006132: I just forgot to remove that.

src/manifold/src/impl.h
```diff
@@ -17,14 +17,33 @@
...
-#include "manifold/utils.h"
 #include "manifold/vec.h"
-#include "quickhull.h"
 #include "shared.h"
 
+#if defined(__GNUC__)
+
+#define IGNORE_DEPRECATED_BEGIN  \
+  _Pragma("GCC diagnostic push") \
+      _Pragma("GCC diagnostic ignored \"-Wdeprecated-declarations\"")
```
elalish, line 29: This is cool, but should we just go ahead and remove `Mesh` for v3.0? It would be a lot less maintenance burden.

test/manifold_test.cpp
elalish: Oh, I see now - let's just update the test to change that index to 7 instead of 4, since this one has more verts.

elalish (commented):
>I just mean there is no simple way to get vertex normal with MeshGL and it makes it hard to get rid of Mesh in our tests.

Interesting - which tests is this making difficult? I can take a look at refactoring them.

src/manifold/src/impl.h
pca006132: true

test/manifold_test.cpp
pca006132: we already have InvalidInput6

pca006132:
> Interesting - which tests is this making difficult? I can take a look at refactoring them.

In general, `WithNormals`, `ExpectMeshes`, `CheckNormals` in `test_main.cpp`, as well as `Manifold.Normals`.

And I also reverted the deprecated commit, that is a bit hard to get right on all platforms (e.g. clang 18 has a bug that breaks `pragma diagnostic push`.

pca006132:
@elalish so should I just disable the failed test xase and merge this?

test/smooth_test.cpp
```diff
@@ -111,8 +109,7 @@ TEST(Smooth, ToLength) {
       CrossSection::Circle(10, 10).Translate({10, 0}).ToPolygons(), 2, 0, 0,
       {0, 0});
   cone += cone.Scale({1, 1, -5});
-  Manifold smooth = Manifold::Smooth(cone.GetMesh());
-  smooth = smooth.RefineToLength(0.1);
+  Manifold smooth = cone.AsOriginal().SmoothOut(180).RefineToLength(0.1);
```
elalish, line 112: Okay, I fixed it with `AsOriginal` and I think I understand why - our old Mesh didn't keep track of IDs, which allowed those faces to be collapsed, but MeshGL keeps more info. I think this is fine.

elalish (approved):
Looks great, thanks!

elalish:
Hmm, I should have asked - did you add any tests of `MeshGL64`? Not that it's likely to behave much differently, but at least one that tests, say, the difference of round-trip precision between `MeshGL` and `MeshGL64`? Fodder for a follow-on.

pca006132:
Yeah, we should definitely add some tests for `MeshGL64`. I think I used it sparingly for some tests, but it is kind of random (depends on my mood) and only a few of them uses `MeshGL64`.

PR #930, opened 2024-09-15
wasm: fix CrossSection slice and project bindings

### LINKED ISSUE #929: Error when `slice`-ing the bottom face of a manifold
It appears the [`slice` function doesn't work as advertised](https://github.com/elalish/manifold/blob/4760ba3a33edded2b619557a789b70b86a97cfc2/bindings/wasm/manifold-encapsulated-types.d.ts#L830-L838) for extracting the bottom faces of a manifold. I'm hitting an undefined error that I can reproduce with a trivial example on [manifoldcad.org](https://manifoldcad.org)

```js
const box = Manifold.cube(1)
console.log(box.boundingBox().min);
console.log(box.slice(0));
```

Produces

```
[
    0,
    0,
    0
]
TypeError: A[0] is undefined
```

Related, is there a way to use `slice` to extract the top faces?

> using a height equal to the top of the bounding box will return empty.

Assuming this bug gets resolved, I can work around it by mirroring and slicing the bottom, but that feels like a kludge.

neilpa:
Resolves #929 by calling the `CrossSectionCtor` directly. These functions already have the vector returned by the `_Slice`/`_Project` so the `polygons2vec` isn't needed. I just copied what the `CrossSection` wrapper did w/out it.

I'm not sure if the memory management with `disposePolygons` is correct, but my tests in with a local ManifoldCAD work after this change.

I also couldn't figure out how to add new tests. AFAICT, it's just running the examples? It might be worth adding a `slice` example anyway. Happy to add as part of this PR if you point me in the right direction.

### DIFF
--- a/.github/workflows/manifold.yml
+++ b/.github/workflows/manifold.yml
@@ -153,7 +153,7 @@ jobs:
         cp ../manifold.* ./dist/
     - name: Upload WASM files
       uses: actions/upload-artifact@v4
-      if: matrix.exception == 'ON'
+      if: matrix.exception == 'ON' && github.event_name == 'push'
       with:
         name: wasm
         path: bindings/wasm/examples/dist/
--- a/bindings/wasm/bindings.js
+++ b/bindings/wasm/bindings.js
@@ -284,11 +284,17 @@ Module.setup = function() {
   };
 
   Module.Manifold.prototype.slice = function(height = 0.) {
-    return Module.CrossSection(this._Slice(height));
+    const polygonsVec = this._Slice(height);
+    const result = new CrossSectionCtor(polygonsVec, fillRuleToInt('Positive'));
+    disposePolygons(polygonsVec);
+    return result;
   };
 
   Module.Manifold.prototype.project = function() {
-    return Module.CrossSection(this._Project()).simplify(this.precision());
+    const polygonsVec = this._Project();
+    const result = new CrossSectionCtor(polygonsVec, fillRuleToInt('Positive'));
+    disposePolygons(polygonsVec);
+    return result.simplify(this.precision);
   };
 
   Module.Manifold.prototype.split = function(manifold) {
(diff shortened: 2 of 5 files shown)


bindings/wasm/bindings.js
```diff
@@ -284,11 +284,17 @@ Module.setup = function() {
   };
 
   Module.Manifold.prototype.slice = function(height = 0.) {
-    return Module.CrossSection(this._Slice(height));
+    const polygonsVec = this._Slice(height);
+    const result = new CrossSectionCtor(polygonsVec, 2 /* positive */);
```
elalish, line 288: How about `fillRule2Int('Positive')` as elsewhere, for maintainability?

elalish (commented):
Yes, another example / test would be great to show people how to use slice/project. The difficulty is I like our examples to be high quality - simple, readable, cool-looking. If you're up to the challenge that would be great.

neilpa:
I leaned into the default yellow rendering and made a cheesy example. Let me know if that works for you.

Also, how do I run the auto code formatter to fix the one broken CI check? I installed `clang-format` via homebrew and ran the `format.sh` script locally but that didn't do anything.

elalish:
Would you mind adding a screenshot?

As to formatting, it looks like our `format.sh` is restricting the paths a bit too much - can you update accordingly? Thanks!

neilpa:
Screen Shot 2024-09-16 at 5 54 08 PM

bindings/wasm/examples/public/examples.js
```diff
@@ -51,6 +51,34 @@ export const examples = {
       return result;
     },
 
+    Auger: function() {
```
elalish, line 54: Thanks for the example! But I'm also really picky about public examples, so I went ahead and made my own instead. Please don't feel slighted; it's just my own aesthetics.

.github/workflows/manifold.yml
```diff
@@ -153,7 +153,7 @@ jobs:
         cp ../manifold.* ./dist/
     - name: Upload WASM files
       uses: actions/upload-artifact@v4
-      if: matrix.exception == 'ON'
+      if: matrix.exception == 'ON' && github.event_name == 'push'
```
elalish, line 156: Hopefully this will fix the accidental update problem.

elalish (approved):
image

PR #931, opened 2024-09-16
optimizations

pca006132:
1. Make `AssignNormals` a template, so we can optimize away one branch.
2. Add "refinement" for collider to check for collision more precisely, so we get less collision pairs, and less work has to be done in `Boolean3`.

### DIFF
--- a/test/hull_test.cpp
+++ b/test/hull_test.cpp
@@ -81,15 +81,6 @@ TEST(Hull, Tictac) {
   EXPECT_NEAR(sphere.NumVert() + tictacSeg, tictac.NumVert(), 1);
 }
 
-#ifdef MANIFOLD_EXPORT
-TEST(Hull, Fail) {
-  Manifold body = ReadMesh("hull-body.glb");
-  Manifold mask = ReadMesh("hull-mask.glb");
-  Manifold ret = body - mask;
-  MeshGL mesh = ret.GetMeshGL();
-}
-#endif
-
 TEST(Hull, Hollow) {
   auto sphere = Manifold::Sphere(100, 360);
   auto hollow = sphere - sphere.Scale({0.8, 0.8, 0.8});
--- a/src/manifold/src/impl.cpp
+++ b/src/manifold/src/impl.cpp
@@ -45,13 +45,13 @@ struct Transform4x3 {
   vec3 operator()(vec3 position) { return transform * vec4(position, 1.0); }
 };
 
+template <bool calculateTriNormal>
 struct AssignNormals {
   VecView<vec3> faceNormal;
   VecView<vec3> vertNormal;
   VecView<const vec3> vertPos;
   VecView<const Halfedge> halfedges;
   const double precision;
-  const bool calculateTriNormal;
 
   void operator()(const int face) {
     vec3& triNormal = faceNormal[face];
@@ -657,9 +657,14 @@ void Manifold::Impl::CalculateNormals() {
     faceNormal_.resize(NumTri());
     calculateTriNormal = true;
   }
-  for_each_n(policy, countAt(0), NumTri(),
-             AssignNormals({faceNormal_, vertNormal_, vertPos_, halfedge_,
-                            precision_, calculateTriNormal}));
+  if (calculateTriNormal)
+    for_each_n(policy, countAt(0), NumTri(),
+               AssignNormals<true>({faceNormal_, vertNormal_, vertPos_,
+                                    halfedge_, precision_}));
+  else
+    for_each_n(policy, countAt(0), NumTri(),
+               AssignNormals<false>({faceNormal_, vertNormal_, vertPos_,
+                                     halfedge_, precision_}));
   for_each(policy, vertNormal_.begin(), vertNormal_.end(),
            [](vec3& v) { v = SafeNormalize(v); });
 }
--- a/test/boolean_complex_test.cpp
+++ b/test/boolean_complex_test.cpp
@@ -996,4 +996,11 @@ TEST(BooleanComplex, CraycloudBool) {
   EXPECT_TRUE(res.IsEmpty());
 }
 
+TEST(BooleanComplex, HullMask) {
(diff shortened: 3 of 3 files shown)


pca006132:
OK, the edge-triangle distance calculation is actually pretty trivial, and seems to fix the issue we had previously. I guess the previous issue is just some precision issue. However, I'm thinking if we should also have some error margin to avoid numerical errors from affecting manifoldness guarantee.

pca006132:
Testing this with openscad shows that it is still not robust... need a bit more investigation.

pca006132:
In my testing, it seems that the computed squared distance can be about 1e7 times larger than precision squared and still has to be added to the collision pairs. I have verified that sliver triangle is not an issue here (which may throw off the normal calculation...), and it doesn't help much even if I use fma to improve the precision of calculation.

pca006132:
Ah, it works when I use long double (80bit float), so apparently it is still some precision issue, but not sure where is causing such a precision loss.

pca006132:
OK, cross products and dot products are getting catastrophic cancellation in some cases...

src/utilities/include/manifold/tri_dist.h
```diff
@@ -82,6 +82,53 @@ inline void EdgeEdgeDist(vec3& x, vec3& y,  // closest points
   y = q + b * u;
 }
 
+inline double EdgeTriangleDistanceSquared(const vec3 p, const vec3 q,
+                                          const std::array<vec3, 3>& abc) {
+  // precompute normals and edge plane perpendicular to the triangle plane
+  vec3 normal = glm::normalize(glm::cross(abc[1] - abc[0], abc[2] - abc[1]));
```
elalish, line 88: This can be NaN, or wildly inaccurate even if finite.

elalish (commented):
The Tri-edge intersection part of this is going to be very hard. This is breaking the core guarantee of how our Boolean operates, unless we can completely prove that this approximation is always conservative compared to our symbolic-perturbation shadowing answer. We can already tell that our test set is not adequate to verify this.

pca006132:
Indeed, eventually I resorted to interval arithmetic to try to make this test robust, not sure if there are other ways. Haven't yet change the edge-edge distance computation.

pca006132:
Thinking about it, though, I should probably account for floating point operation errors when doing the interval arithmetic? Maybe I should just copy a robust interval arithmetic library in.

neilpa:
As someone who's been mildly interested in robust comp-geom for a while, anything that computes distance is going to be hard and/or slow to do robustly.

If the goal is to remove more false positive triangle pairs early, there are conservative approaches that rely only on plane orientation predicates. Those you can do robustly in floating point and relatively fast with something like [Shewchuk's robust predicates](http://www.cs.cmu.edu/~quake/robust.html).

In this case, you could use one triangle to define the plane, then check the three points of the other triangle. If the orientation of all three points have the same sign, then the triangles can't intersect.

elalish:
> Thinking about it, though, I should probably account for floating point operation errors when doing the interval arithmetic? Maybe I should just copy a robust interval arithmetic library in.

The more fundamental problem is that I'm not sure how we could make the mathematical guarantee. Keep in mind, being exact isn't good enough either - it has to be conservative relative to our other imprecise (but self-consistent) method of finding intersections. 

I think what you're doing fundamentally is collapsing our `s11` and `s02` calculations into this single loop, trading reduced memory for some duplicate calculation. I actually did a similar thing some years ago when I removed the `s01` vectors, and it definitely improved speed. 

I think if we want to do this optimization robustly, we'll have to take that approach, which is a larger refactor. It would mean collapsing all the lines from `Filter11` to `Winding03` down into the collision functions.

pca006132:
I see, I was thinking if we can guarantee that all pairs where the distance is less than epsilon times some factor, we can over-approximate the set of inputs that our s11 and s02 calculation consider as overlapping, but this is probably not as simple and hard to reason about.

pca006132:
Thinking about it, is it possible to establish a bound, in which the intersection checks (s11 and s02) can only accept things closer than some epsilon, and everything further apart must be rejected?

With that, we can make the overapproximation here return true for everything closer than that epsilon. This should not be hard because we can use interval arithmetic, and account for floating point errors very conservatively by assuming the max error for each operation.

pca006132:
btw converting to draft to avoid having that green button that I always wanted to click :)

elalish:
>Thinking about it, is it possible to establish a bound, in which the intersection checks (s11 and s02) can only accept things closer than some epsilon, and everything further apart must be rejected?

I'm just not sure - keep in mind that the winding numbers of the vertices are also fundamentally related to these calculations and must also be consistent. I would need a proof, and I'm not sure how to write one for this. In any case, if the goal is optimization, I think the approach I outlined above will likely give even better results and won't break our existing proof. But something that large we should probably push till after v3.0.

pca006132:
Yeah, we can take this slow. This should not affect v3.0.

elalish:
For now, would you like to reduce this to just the `AssignNormals` change and we can merge that?

pca006132:
Sure

test/hull_test.cpp
```diff
@@ -81,15 +81,6 @@ TEST(Hull, Tictac) {
   EXPECT_NEAR(sphere.NumVert() + tictacSeg, tictac.NumVert(), 1);
 }
 
-#ifdef MANIFOLD_EXPORT
```
elalish, line 84: I think it needs this `#ifdef` since it's using Assimp for ReadMesh.

elalish (changes requested):

test/hull_test.cpp
pca006132: No, the last few `BooleanComplex` tests are wrapped inside a large `#ifdef MANIFOLD_EXPORT` block.

elalish: Oh, got it, thanks!

pca006132:
I think it is something related to draft PR. I can open a new one.

PR #933, opened 2024-09-17
Checking on a potential bug

elalish:
This test passes for me locally on Mac, so checking the CI. Found [here](https://github.com/elalish/manifold/pull/666#issuecomment-2356152331).


pca006132:
ok, we didn't really run the boolean lol

elalish:
Oh, duh, thanks!

elalish:
Okay, and just like that - I have a local repro. Just another example of why every piece of pasted code should just be PR instead. 😄

pca006132:
My original snippet doesn't have `if (options.exportModels)` :P

elalish:
This one's not a valid test - the first mesh is strongly self-intersecting, hence why we get self-intersecting polygons that make inverted triangles. 

https://github.com/user-attachments/assets/f90d5d4b-ac33-4988-bdd7-a40ff59a9b66

PR #935, opened 2024-09-17
Make create faces optional

elalish:
`CreateFace()` scales pretty poorly with model size and isn't always needed, so I'm removing it from a bunch of our functions and expecting users to call `AsOriginal()` instead when they want coplanar faces identified. I've also simplified some tests, which identified some existing bugs, particularly in `SetProperties`, so there are some fixes here as well. 

I'm still thinking through the API implications of all this and trying to make sure I have the right shape. I've also removed `propertyTolerance` from our constructors - it is only needed for `CreateFaces`, so now it's just an input to `AsOriginal`. 

Next I'm going to work on a follow-on that splits the concepts of `faceID` and coplanarity. I believe this will allow me to remove the slow part of `CreateFaces` entirely (connected components) and allow `Warp` and `Refine` to retain their mesh relations - currently those get destroyed since coplanarity gets broken.


src/manifold/src/impl.h
```diff
@@ -133,11 +132,7 @@ struct Manifold::Impl {
     }
 
     Vec<TriRef> triRef;
-    if (meshGL.runOriginalID.empty()) {
-      // FIXME: This affects Impl::InitializeOriginal, and removing this
-      // apparently make things fail. Not sure if this is expected.
-      meshRelation_.originalID = Impl::ReserveIDs(1);
```
elalish, line 139: FYI @pca006132 I fixed this as part of the rest of this cleanup.

src/manifold/src/manifold.cpp
```diff
@@ -645,13 +642,12 @@ Manifold Manifold::SetProperties(
     if (triProperties.size() == 0) {
       const int numTri = NumTri();
       triProperties.resize(numTri);
-      int idx = 0;
       for (int i = 0; i < numTri; ++i) {
         for (const int j : {0, 1, 2}) {
-          triProperties[i][j] = idx++;
+          triProperties[i][j] = pImpl->halfedge_[3 * i + j].startVert;
```
elalish, line 647: This bug was causing duplicate propVerts to be created unnecessarily.

src/manifold/src/quickhull.cpp
```diff
@@ -830,14 +830,10 @@ void Manifold::Impl::Hull(VecView<vec3> vertPos) {
...
   QuickHull qh(vertPos);
   std::tie(halfedge_, vertPos_) = qh.buildMesh();
-  meshRelation_.originalID = ReserveIDs(1);
   CalculateBBox();
   SetPrecision(bBox_.Scale() * kTolerance);
-  SplitPinchedVerts();
   CalculateNormals();
   InitializeOriginal();
-  CreateFaces({});
-  SimplifyTopology();
```
elalish, line 840: FYI @Kushal-Shah-03 I simplified the hull setup - hopefully improves the speed a bit.

PR #937, opened 2024-09-18
fix deployment

elalish:
Hopefully - before it was looking for a pull request, but we'd rather use the push to master.


elalish:
@pca006132 Would you mind merging this once the tests pass and verifying that it actually fixed the deployment? I need to go to bed. Thanks (and yay time zones)!

pca006132 (approved):

pca006132:
Deployment seems to be working, but I don't think this is a proper fix... I think the proper fix should make the deployment workflow depends on the earlier build workflow. Anyway, can check later.

PR #938, opened 2024-09-18
Add RefineToPrecision

elalish:
Now instead of uniformly segmenting edges either to a fixed number of divisions (Refine) or a fixed length (RefineToLength), we now also have a heuristic to subdivide each edge depending on its curvature to create enough facets to approximate the smooth surface within roughly the input precision value. It is very much approximate, but I think a useful tool.


pca006132:
Will have a look this Friday, busy preparing presentation slides :(

pca006132 (approved):
Can't see anything problematic. LGTM.

PR #939, opened 2024-09-18
removed face from halfedge

pca006132:
Just some cleanup, I think we talked about `face` being redundant because we can just compute it from the index, so I took the time to remove it.

### DIFF
--- a/src/manifold/src/impl.cpp
+++ b/src/manifold/src/impl.cpp
@@ -108,16 +108,18 @@ struct CoplanarEdge {
   void operator()(const int edgeIdx) {
     const Halfedge edge = halfedge[edgeIdx];
     const Halfedge pair = halfedge[edge.pairedHalfedge];
+    const int edgeFace = edgeIdx / 3;
+    const int pairFace = edge.pairedHalfedge / 3;
 
-    if (triRef[edge.face].meshID != triRef[pair.face].meshID) return;
+    if (triRef[edgeFace].meshID != triRef[pairFace].meshID) return;
 
     const vec3 base = vertPos[edge.startVert];
-    const int baseNum = edgeIdx - 3 * edge.face;
-    const int jointNum = edge.pairedHalfedge - 3 * pair.face;
+    const int baseNum = edgeIdx - 3 * edgeFace;
+    const int jointNum = edge.pairedHalfedge - 3 * pairFace;
 
     if (numProp > 0) {
-      const int prop0 = triProp[edge.face][baseNum];
-      const int prop1 = triProp[pair.face][jointNum == 2 ? 0 : jointNum + 1];
+      const int prop0 = triProp[edgeFace][baseNum];
+      const int prop1 = triProp[pairFace][jointNum == 2 ? 0 : jointNum + 1];
       bool propEqual = true;
       for (int p = 0; p < numProp; ++p) {
         if (std::abs(prop[numProp * prop0 + p] - prop[numProp * prop1 + p]) >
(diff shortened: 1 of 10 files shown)


src/manifold/src/impl.cpp
```diff
@@ -496,8 +498,9 @@ void Manifold::Impl::CreateHalfedges(const Vec<ivec3>& triVerts) {
       if (h0.startVert != h1.endVert || h0.endVert != h1.startVert) break;
       if (halfedge_[NextHalfedge(pair0)].endVert ==
           halfedge_[NextHalfedge(pair1)].endVert) {
-        h0.face = -1;
-        h1.face = -1;
+        // FIXME: this actually does nothing
```
pca006132, line 501: This is an interesting one, I was trying to use a vector instead for this and was having weird issues. It turns out we declared `h0` and `h1` as `Halfedge`, so these assignments aren't doing anything. Changing them to `Halfedge&` breaks things, so I just commented them out for now.

src/manifold/src/quickhull.cpp
```diff
@@ -78,13 +78,15 @@ size_t MeshBuilder::addHalfedge() {
     return index;
   }
   halfedges.push_back({});
+  halfedgeToFace.push_back(0);
```
pca006132, line 81: Face index is needed in `quickhull`, so I keep a `halfedgeToFace` vector for that.

src/manifold/src/sort.cpp
```diff
@@ -256,21 +231,33 @@ void Manifold::Impl::Finish() {
...
 #ifdef MANIFOLD_DEBUG
+  auto MaxOrMinus = [](int a, int b) {
+    return std::min(a, b) < 0 ? -1 : std::max(a, b);
+  };
+  int face = 0;
   Halfedge extrema = {0, 0, 0, 0};
-  extrema = reduce(halfedge_.begin(), halfedge_.end(), extrema, Extrema());
-#endif
-
+  for (size_t i = 0; i < halfedge_.size(); i++) {
```
pca006132, line 239: Simplified the code a bit as well. Was thinking about using pair iterators, but it is too much work. We don't need parallelization here anyway.

src/manifold/src/impl.cpp
elalish: Ha, I think I put this in to try to debug something and then forgot about it. Let's remove.

src/manifold/src/impl.cpp
```diff
@@ -513,13 +516,13 @@ void Manifold::Impl::CreateHalfedges(const Vec<ivec3>& triVerts) {
...
-      halfedge_[pair0] = {-1, -1, -1, -1};
-      halfedge_[pair1] = {-1, -1, -1, -1};
-    } else {
-      halfedge_[pair0].pairedHalfedge = pair1;
-      halfedge_[pair1].pairedHalfedge = pair0;
-    }
+    // if (halfedge_[pair0].face < 0) {
+    //   halfedge_[pair0] = {-1, -1, -1, -1};
+    //   halfedge_[pair1] = {-1, -1, -1, -1};
+    // } else {
```
elalish, line 522: Same as above.

src/manifold/src/quickhull.cpp
```diff
@@ -295,9 +315,9 @@ std::pair<Vec<Halfedge>, Vec<vec3>> QuickHull::buildMesh(double epsilon) {
              he.endVert = counts[he.endVert];
            });
   // setting face id
-  for (size_t index = 0; index < halfedges.size(); index++) {
-    halfedges[index].face = index / 3;
-  }
+  // for (size_t index = 0; index < halfedges.size(); index++) {
+  //   halfedges[index].face = index / 3;
```
elalish, line 319: remove?

elalish (approved):
Thanks for the cleanup! I'm guessing this didn't make any noticeable speed change?

elalish:
Yeah, I thought about that when I made `Halfedge` - I think it's worth leaving start and end in. It's also useful in other places, like the Boolean where the edges get created and operated on before they know what their pair is.

PR #940, opened 2024-09-19
Misc. changes

### LINKED ISSUE #744: better error handling

pca006132:
Fixes #744.

1. Propagate errors in mesh boolean, transform and compose. Probably there are still some other operations that we should propagate the error.
2. Use 64-bit integers for MeshGL64. Note that this is not yet tested, I just checked it can compile. (ideally nothing else can go wrong considering we don't do raw pointer operations...). We should check if the APIs make sense.

I don't plan to do 64-bit run ID or mesh ID, as I don't think users will ever want that...? It might make sense in some scientific computation situation to handle large meshes, but it doesn't make sense to do boolean on several hundred million meshes, I guess...

Checklist for error propagation, feel free to add to this list, so I can add them in this PR:
1. Warp

### DIFF
--- a/src/manifold/include/manifold/manifold.h
+++ b/src/manifold/include/manifold/manifold.h
@@ -37,22 +37,12 @@ class CsgLeafNode;
  *  @{
  */
 
-template <typename Precision>
+template <typename Precision, typename I = uint32_t>
 struct MeshGLP {
   /// Number of property vertices
-  uint32_t NumVert() const {
-    ASSERT(vertProperties.size() / numProp <
-               static_cast<size_t>(std::numeric_limits<uint32_t>::max()),
-           std::out_of_range("mesh too large for MeshGL"));
-    return vertProperties.size() / numProp;
-  };
+  I NumVert() const { return vertProperties.size() / numProp; };
   /// Number of triangles
-  uint32_t NumTri() const {
-    ASSERT(triVerts.size() / 3 <
-               static_cast<size_t>(std::numeric_limits<uint32_t>::max()),
-           std::out_of_range("mesh too large for MeshGL"));
-    return triVerts.size() / 3;
-  };
+  I NumTri() const { return triVerts.size() / 3; };
   /// Number of properties per vertex, always >= 3.
   uint32_t numProp = 3;
   /// Flat, GL-style interleaved list of all vertex properties: propVal =
@@ -61,14 +51,14 @@ struct MeshGLP {
   std::vector<Precision> vertProperties;
   /// The vertex indices of the three triangle corners in CCW (from the outside)
   /// order, for each triangle.
-  std::vector<uint32_t> triVerts;
+  std::vector<I> triVerts;
   /// Optional: A list of only the vertex indicies that need to be merged to
   /// reconstruct the manifold.
-  std::vector<uint32_t> mergeFromVert;
+  std::vector<I> mergeFromVert;
   /// Optional: The same length as mergeFromVert, and the corresponding value
   /// contains the vertex to merge with. It will have an identical position, but
   /// the other properties may differ.
-  std::vector<uint32_t> mergeToVert;
+  std::vector<I> mergeToVert;
   /// Optional: Indicates runs of triangles that correspond to a particular
   /// input mesh instance. The runs encompass all of triVerts and are sorted
   /// by runOriginalID. Run i begins at triVerts[runIndex[i]] and ends at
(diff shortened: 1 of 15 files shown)


src/manifold/include/manifold/manifold.h
```diff
@@ -37,22 +37,12 @@ class CsgLeafNode;
  *  @{
  */
 
-template <typename Precision>
+template <typename Precision, typename I = uint32_t>
```
elalish, line 40: Honestly, does it hurt anything for runIndex and runOriginalID to be 64 bit? They're very short vectors anyway. Seems simpler and more fitting with the name `MeshGL64` for everything to be 64 bit instead of just some.

src/manifold/src/csg_tree.cpp
```diff
@@ -478,9 +483,9 @@ std::shared_ptr<Manifold::Impl> CsgOpNode::BatchBoolean(
           continue;
         }
         group.run([&, a, b]() {
-          const Manifold::Impl *aImpl;
-          const Manifold::Impl *bImpl;
```
elalish, line 482: Were these lines even doing anything before? Can we just delete them now too?

src/manifold/src/impl.h
```diff
@@ -183,8 +183,9 @@ struct Manifold::Impl {
         }
         if (numProp > 0) {
           meshRelation_.triProperties.push_back(
-              ivec3(meshGL.triVerts[3 * i], meshGL.triVerts[3 * i + 1],
-                    meshGL.triVerts[3 * i + 2]));
+              ivec3((uint32_t)meshGL.triVerts[3 * i],
+                    (uint32_t)meshGL.triVerts[3 * i + 1],
+                    (uint32_t)meshGL.triVerts[3 * i + 2]));
```
elalish, line 188: which casting syntax do you prefer?

elalish (approved):
As far as a checklist, I guess it's everything non-static in `manifold.h` that returns a `Manifold`?

src/manifold/src/csg_tree.cpp
pca006132: sure

src/manifold/src/impl.h
pca006132: ah, actually `static_cast<uint32_t>`, will change.

src/manifold/include/manifold/manifold.h
pca006132: Sure

pca006132:
Thinking about it, some functions in `Manifold` should also return 64-bit integer for this.

pca006132:
Notably, `OriginalID` is now returning `std::optional<size_t>`, and is empty when the mesh is not an original.

pca006132:
I was looking into other functions that do unary operations on a manifold. Apparently, they copy the `Impl` data structure, so I think they should have the error code copied with them together. I forgot who can we have those error conditions and whether our internal data structures are still safe to operate on when an error occurs though, so not sure if I should add guard in those operations to early return.

pca006132:
maybe I will fix the wasm issue tmr

elalish:
IIRC, the manifold is set to empty when any of the error conditions occur - at least that was the intention. So I believe that yes, they should all be in a safe state.

src/manifold/include/manifold/manifold.h
```diff
@@ -222,14 +223,14 @@ class Manifold {
...
-  int NumPropVert() const;
+  size_t NumVert() const;
+  size_t NumEdge() const;
+  size_t NumTri() const;
+  size_t NumProp() const;
+  size_t NumPropVert() const;
   Box BoundingBox() const;
   double Precision() const;
-  int Genus() const;
+  size_t Genus() const;
```
elalish, line 233: Genus is signed - you get negative values for multiple disconnected objects.

src/utilities/include/manifold/iters.h
```diff
@@ -190,8 +190,7 @@ struct CountingIterator {
   }
 };
 
-template <typename T>
-CountingIterator<T> countAt(T i) {
+constexpr CountingIterator<size_t> countAt(size_t i) {
```
elalish, line 193: 👍

test/test_main.cpp
```diff
@@ -203,7 +203,7 @@ MeshGL WithIndexColors(const MeshGL& in) {
   out.runIndex.clear();
   out.runTransform.clear();
   out.faceID.clear();
-  out.runOriginalID = {Manifold::ReserveIDs(1)};
+  out.runOriginalID = {static_cast<uint32_t>(Manifold::ReserveIDs(1))};
```
elalish, line 206: Hmm, I think I see now why you recommended keeping the IDs as uint32_t across the board. The casting is annoying - and *we* know the values are sequential and so down-casting is safe, but that's not a common expectation (people might think this is a GUID, and maybe it could be someday). Sorry for the flip-flop, but I think I like it better the first way now.

elalish (changes requested):

elalish (approved):
Thanks!

PR #946, opened 2024-09-22
Install manifold.pc under CMAKE_INSTALL_LIBDIR

cjmayo:
On platforms that install libmanifold under lib64, manifold.pc should be installed in the pkgconfig directory under lib64.


pca006132 (approved):

pca006132:
Btw did you find any other hardcoded paths in our other cmake files? Ideally those should be changed as well, if any.

PR #948, opened 2024-09-23
Use CMAKE_INSTALL_LIBDIR when setting installed rpath

cjmayo:
I found one more - no sign of `usr`.

I can't say I fully understand rpath, but this does seem to make a difference to the value of RUNPATH in my installed libmanifold as reported by `objdump -x`.


elalish (approved):
This looks reasonable to me. Thanks!

PR #949, opened 2024-09-24
use nanobind builtin stubgen

### LINKED ISSUE #855: Auto generated python stubs inconsistent with cpp API
This is a stellar project, I love it! I originally came here looking for a replacement for OpenSCAD's proprietary language. And I found a little gem :)

Now, here's a technical aspect I would like to hear the authors's / maintainer's thoughts on:

Following the stub generation from [bindings/python/README.md](https://github.com/elalish/manifold/blob/master/bindings/python/README.md) via `nanobind-stubgen manifold3d` I observe that the stubs don't properly reflect the API.

I do believe it is mostly a limitation of `nanobind`, however `manifold3d` can make an informed decision on whether to improve the situation.

My main issue with the stubs is that any IDE / linter / etc. will be confused, impairing the `manifold3d` python library user experience. For example, the `Manifold` cpp class has a few `static` constructors. Those are not marked as `@staticmethod` in `manifold.pyi`. Also, the operators for boolean combinations are not included in the the stub file (i.e. `__add__` etc. are missing).

I know that there are bindings for a couple of languages, and that this is not a first class python project. Still, solid stubs could greatly improve the developer experience when using the python package.

pca006132:
Closes #855.

The stub is not perfect because it is python, but I think we are pretty good now.


pca006132:
@t-fi @wrongbad cannot request your review, so I will ping you instead.

bindings/python/CMakeLists.txt
```diff
@@ -17,7 +17,8 @@ project(python)
 execute_process(
   COMMAND "${Python_EXECUTABLE}" -m nanobind --version
   OUTPUT_STRIP_TRAILING_WHITESPACE OUTPUT_VARIABLE NB_VERSION)
-if(NB_VERSION VERSION_GREATER_EQUAL 1.8.0)
+# we are fine with 2.0.0
+if(NB_VERSION VERSION_GREATER_EQUAL 2.0.0)
```
elalish, line 21: Should we just go with 2.1.0 across the board?

bindings/python/examples/bricks.py
```diff
@@ -84,7 +84,7 @@ def floor(length, width):
             )
         )
     if length == 1 and width > 1:
-        results.append(row(width - 1).rotate(0, 0, 90))
+        results.append(row(width - 1).rotate((0, 0, 90)))
```
elalish, line 87: What are the extra parens for?

bindings/python/manifold3d.cpp
```diff
@@ -327,7 +327,8 @@ NB_MODULE(manifold3d, m) {
       .def("refine", &Manifold::Refine, nb::arg("n"), manifold__refine__n)
       .def("refine_to_length", &Manifold::RefineToLength, nb::arg("length"),
            manifold__refine_to_length__length)
-      .def("to_mesh", &Manifold::GetMeshGL, nb::arg("normal_idx") = ivec3(0),
+      .def("to_mesh", &Manifold::GetMeshGL,
+           nb::arg("normal_idx") = std::make_tuple(0, 0, 0),
```
elalish, line 331: 👍

elalish (commented):
Does this mean we can remove the python auto-gen docs now?

bindings/python/CMakeLists.txt
pca006132: Some distros may have 2.1.0? Not sure about if we should support that.

pca006132:
and no, we still need to autogen docs for getting the doc strings.

pca006132:
will figure out the wasm issue tmr, probably some library version issue.

bindings/python/CMakeLists.txt
elalish: Okay, so 2.0 is significantly more widely available than 2.1? This is fine if so - I was just thinking we could reduce CMake complexity a bit.

pca006132:
Maybe we can just disable `build_nix` with js for now until I figure out what is going on, seems to be related to emscripten version.

pca006132:
@elalish should we merge this?

elalish (approved):
Thanks!

PR #951, opened 2024-09-25
Remove CreateFaces entirely

elalish:
This is a sizable refactor and moderate breaking change. Manifold will no longer use equal `faceID`s to mean that those triangles are coplanar. This was always problematic, because it meant that using Warp or Refine would destroy the entire mesh relation, making material association quite difficult. This also means we can remove any concept of property tolerances from our API, which I never liked. Now edges can collapse if the two triangles are coplanar, have the same meshID, and reference the same propVerts on both sides of the edge. 

Hopefully this will also make things a bit faster, since `CreateFaces` was always slow due to the connected components algorithm. I'm still on the fence as to whether we should store and update the `internalEdges` list, or generate it on the fly. For the first pass it's on the fly. This PR is very much WIP.


src/manifold/src/constructors.cpp
```diff
@@ -92,7 +94,9 @@ Manifold Manifold::Smooth(const MeshGL64& meshGL64,
                "rather than Smooth().");
 
   std::shared_ptr<Impl> impl = std::make_shared<Impl>(meshGL64);
-  impl->CreateTangents(impl->UpdateSharpenedEdges(sharpenedEdges));
+  const Vec<bool> internalEdges = impl->InternalEdges();
+  impl->CreateTangents(impl->UpdateSharpenedEdges(sharpenedEdges),
+                       internalEdges);
```
elalish, line 99: @pca006132 Any ideas for reducing the duplication between these two `MeshGL` functions?

pca006132: Move this to impl?

pca006132:
Btw, do you have any particular test where `CreateFaces` is very slow? I ran both our `perfTest` and our Samples/SDF/BooleanComplex tests, `CreateFaces` is typically pretty fast:

image

elalish:
Huh, I thought it was slow, but it was awhile back - maybe your union find optimization helped more than I realized. I believe in general the slowness comes with importing a large mesh. Anyway, the primary purpose of this PR isn't speed, but making the mesh relation work better.

pca006132:
Btw, I guess this is the final breaking change (feature-wise) for 3.0? Others are mostly some API cleanup, as well as binding changes to match our updated APIs?

elalish:
> Btw, I guess this is the final breaking change (feature-wise) for 3.0? Others are mostly some API cleanup, as well as binding changes to match our updated APIs?

Close, but I think removing GLM from our public API is one more: #962.

elalish:
Hmm, definitely still some work to do on this one. Something's still a bit broken because it's not collapsing all edges it should. However, it's also *dramatically* slower, which has me a little more worried - may need to change the approach. @pca006132 if you want to have a look, maybe see if Tracy has any interesting insights (since I can't run it on Mac), that would be great.

pca006132:
Do you have any tests that are slower in particular? I tried running some tests, but they fail with vec out of range immediately, so it is hard to get profiler result for this.

And I think tracy does support mac?

elalish:
Oh, did you pull my latest? I thought I fixed the vec out of range stuff. I guess I should try Tracy again, maybe I'm remembering wrong.

pca006132:
Yeah, apparently I forgot to pull the latest. Will test later tonight.

pca006132:
Just had a look at `perf` output, nothing too unusual. I think the problem is just we are not simplifying enough, causing everything to run slower?

elalish:
Ha, looks like #975 fixed the slowness - sure enough it just wasn't running parallel. That makes me feel better - now I just need to debug.

.vscode/launch.json
```diff
@@ -13,7 +13,10 @@
       "args": [
         "--gtest_break_on_failure",
         "--gtest_catch_exceptions=0",
+<<<<<<< HEAD
```
pca006132: merge conflict :)

fire:
I noticed you mentioned tracy, samply has a nice integration with the mozilla tools for profiling.

https://github.com/mstange/samply

elalish:
This is looking like a larger can of worms than I wanted. I'm closing for now in favor of a reduced-scope version in #991. I might come back to this someday since it seems like it should have some promise for speed improvements.

PR #952, opened 2024-09-25
Adjust exported target location, remove Deps downloading

starseeker:
This allows a find_package in BRL-CAD to locate Manifold using the manifoldConfig.cmake file, and avoids performing the dependency downloads as a consequence of the find_package call.


starseeker:
To clarify, this does not eliminate manifoldDeps.cmake from the Manifold build itself - just from the exported targets.

CMakeLists.txt
```diff
@@ -162,13 +162,15 @@ write_basic_package_version_file(
   VERSION ${MANIFOLD_VERSION}
   COMPATIBILITY SameMajorVersion
 )
-install(EXPORT manifoldTargets DESTINATION ${CMAKE_INSTALL_DATADIR}/manifold)
+install(EXPORT manifoldTargets
```
pca006132, line 165: What are the files being installed with this install call?

starseeker: The DESTINATION is ${CMAKE_INSTALL_LIBDIR}/cmake/manifold, so lib/cmake/manifold.  That seems to be where CMake is looking for these files,  at least on Linux:

https://cmake.org/cmake/help/latest/command/find_package.html#search-procedure

I'm open to a better specification for where to place the files that will be better for cross-platform scenarios - the EXPORT feature is new to me, so I'm not properly conversant with best practices yet.

elalish: Do you want to try to merge this PR, or just do this as part of your larger reorganization?

starseeker: I'm game either way - if you think there's a good chance of the larger reorg getting merged there's probably no particular need for this, but if that's problematic this would be helpful.

elalish: Let's see how the other does, since this isn't passing the CI anyway.

starseeker:
Closing in favor of https://github.com/elalish/manifold/pull/961

PR #954, opened 2024-09-26
Fix comment

wtholliday:
guessing a search and replace turned floating-point into "doubleing-point"


pca006132 (approved):
Thanks for actually reading our comments :)

PR #956, opened 2024-09-26
use std::swap

wtholliday:
Just a minor change for code clarity


pca006132:
Thanks a lot, but ideally these cleanup PRs can be grouped into one to reduce the number of notification.

Maybe we can wait a bit and see if you still have other cleanup that you can find?

wtholliday:
@pca006132 I'd just merge this one. I'm not sure I'm going to find other small cleanups, this one just jumped out at me

PR #958, opened 2024-09-26
Type safety for indices

wtholliday:
Manifold uses ints for indices which shouldn't be assigned to each other, such as vertex indices and half-edge indices. This proposal adds custom index support to `VecView`.

This is a WIP, and I will attempt the rest of the changes with an OK from @elalish :)

### DIFF
--- a/src/manifold/src/boolean3.cpp
+++ b/src/manifold/src/boolean3.cpp
@@ -81,7 +81,7 @@ struct CopyFaceEdges {
       const int q1 = 3 * q2 + j;
       const Halfedge edge = halfedgesQ[q1];
       int a = pX;
-      int b = edge.IsForward() ? q1 : edge.pairedHalfedge;
+      int b = edge.IsForward() ? q1 : edge.pairedHalfedge.ix;
       if (inverted) std::swap(a, b);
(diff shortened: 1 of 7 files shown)


pca006132:
LGTM

elalish (commented):
I think this is probably a good idea, but can you show a little bit more, specifically how using these new indices looks? My main concern is code readability, so this will be okay so long as it's not adding lots of characters. 

Also, should `GetIndex` be a member function instead of a free function?

wtholliday:
I don't think it will make code readability worse.

I think perhaps some policy template could keep `GetIndex` from being a free function. The issue is that `VecView` still needs to work for `size_t` indices. I'll look into it.

elalish:
We do a lot of index addition/multiplication, so I just want to make sure that's as automatic with `HalfedgeIdx` as it is with `size_t`.

wtholliday:
Gotcha. Well, we see how it looks :)

wtholliday:
I think this is just too hard a change to make, even if the result would be nice.

elalish:
Yeah, I'm not too surprised; thanks anyway.

PR #963, opened 2024-09-29
Wasm: update CrossSection.toPolygons() return value to match type

### LINKED ISSUE #943: js bindings: CrossSection.toPolygons() return value doesn't match typescript type
The points are objects of type {x,y} instead of an array [x,y] like the types indicate.
example: run this on the web example `console.log(CrossSection.square().toPolygons());`
output:
```
[
    [
        {
            "x": 0,
            "y": 0
        },
        {
            "x": 1,
            "y": 0
        },
        {
            "x": 1,
            "y": 1
        },
        {
            "x": 0,
            "y": 1
        }
    ]
]
```

expected:
```
[
    [
        [0,0],
        [1,0],
        [1,1],
        [0,1]
    ]
]
```

I can look into fixing it later if no one wants to fix it.

dsafa:
Closes #943

I used the existing support for a map callback in `vec2polygons`.

Also started with a simple test for bindings. I guess the hard part of writing the tests is trying to test the custom bindings and not emscripten or the c++ code.


bindings/wasm/examples/bindings.test.ts
```diff
@@ -0,0 +1,21 @@
...
+let manifoldModule: ManifoldToplevel;
+
+beforeAll(async () => {
+  manifoldModule = await Module();
+  manifoldModule.setup();
+});
+
+suite('CrossSection Bindings', () => {
+  test('ToPolygons return correct shape', () => {
+    const polygons = manifoldModule.CrossSection.square().toPolygons();
```
elalish, line 13: So, is the CI running this test yet? Our current JS tests are in the form of examples, but we could certainly use better coverage.

bindings/wasm/bindings.js
```diff
@@ -190,7 +190,7 @@ Module.setup = function() {
 
   Module.CrossSection.prototype.toPolygons = function() {
     const vec = this._ToPolygons();
-    const result = vec2polygons(vec);
+    const result = vec2polygons(vec, v => [v.x, v.y]);
```
elalish, line 193: Just to double-check, which API do we like better? The one we implemented or the one we documented? Because we could equally update the type instead of this.

bindings/wasm/examples/bindings.test.ts
dsafa: Looking at the actions, seems like its running since vitest should should be running `*.test.*` files by default

bindings/wasm/bindings.js
dsafa: Personally, since there is a static factory function called `fromPolygons`, I would assume that `toPolygons` would return the same type.

elalish (approved):
Looks great, thanks! Would you mind pulling master? We just had a huge refactor that changed the names of our CI runs - this should clear it up. Also, looks like you need to run clang-format.

dsafa:
Thanks, I keep forgetting to run format

dsafa:
Also this slipped my mind, but I guess this would count as a breaking change for the npm package? although its a ts bug fix, if someone was using this in js, it would break

elalish:
Yes, good timing, as we're stacking up breaking changes right now anyway for v3.0.

PR #964, opened 2024-09-29
feat: exposes Quality::ResetToDefaults() via C bindings

NickUfer:
Yesterday I started implementing rust bindings for this project and I wanted to run tests if my bindings for changing quality settings work as expected. I stumbled across the problem that tests fail when I run them all together instead of one by one.
I then pinned it down to the changed defaults, but I also figured out that there is no C binding for the reset function.

TL;DR: this PR exposes the Quality::ResetToDefaults() function via C bindings with roughly the same function name as for WASM (`resetToCircularDefaults`).


elalish (approved):
Thanks! Great to hear about Rust bindings - when you have something to share, please add it to our README; I want to make it easy for people to find all the Manifold bindings / packages that exist.

PR #966, opened 2024-09-29
Python CMake Installation Workaround

cjmayo:
- `EXCLUDE_FROM_ALL` means doing a separate install for the `manifold3d`  target? No problem with that - except it doesn't install the Python module for me. (although if you have set `MANIFOLD_PYBIND` you want to install it?)

- Installing `py.typed` directly into `site-packages` is ambiguous. a `manifold3d` directory needed? `__init__.py`, new library name???

I don't have any answers, just illustrating with a workaround.


elalish:
I have no idea, but if you mark it ready for review at least we'll see if the CI likes it. @starseeker @pca006132 any opinions here?

starseeker:
Um.  Not offhand - is this something the new hierarchy broke?

elalish:
Github doesn't handle Draft PRs very well - I think you'll need to pull master and push to get the CI unstuck.

pca006132:
We should probably remove condition to run CI only on non-draft PRs... It seems that I added it two years ago for some unknown reason.

pca006132:
1. Which command did you use to install `manifold3d`?
2. According to https://peps.python.org/pep-0561/#packaging-type-information, it seems that we should install it alongside to `manifold3d.pyi`?

pca006132:
1. Can you give me a build log? I mean the `cmake --install` command you use to install manifold. Maybe gentoo can configure that to install some specific target?
2. It seems that this is the current result, installed with nix:

   ```
   $ ls result/lib/python3.12/site-packages/
   manifold3d-2.5.1.dist-info  manifold3d.cpython-312-x86_64-linux-gnu.so  manifold3d.pyi  py.typed
   ```
   If this is not what you got, can you show me the result you got?

pca006132:
1. I guess it is fine to remove it. I added it so it will not be installed automatically for nixos (and it cannot be installed due to path issues), but maybe we should not build the python bindings for normal packages instead.
2. I'm thinking if `py.typed` is needed at all in our case. I think it is for type checking normal python modules, but we have none here. Maybe we can just remove it.

pca006132:
Closing in flavor of #973

PR #967, opened 2024-09-30
Try to restore C bindings as a separate library build.

starseeker:
elalish asked if we could restore the C bindings being an optional component - this takes a stab at it, but needs testing.


elalish (approved):
Thanks, this looks better to me.

PR #968, opened 2024-09-30
feat: exposes allocation functions via C bindings

NickUfer:
The current C bindings require the caller to allocate memory for new structs, pass it to Manifold, and manually deallocate it using the exposed delete functions. This approach breaks the best practice of allocating and deallocating memory in the same context, especially in the context of FFI.

This PR introduces functions that leverage Manifold's internal allocator to handle memory allocation for each relevant struct. This provides an alternative to the current method of external allocation, while preserving backward compatibility with the existing approach.

### DIFF
--- a/bindings/c/include/manifold/manifoldc.h
+++ b/bindings/c/include/manifold/manifoldc.h
@@ -392,6 +392,18 @@ size_t manifold_box_size();
 size_t manifold_rect_size();
 size_t manifold_curvature_size();
 
+// allocation
+
+ManifoldManifold *manifold_alloc_manifold();
+ManifoldManifoldVec *manifold_alloc_manifold_vec();
+ManifoldCrossSection *manifold_alloc_cross_section();
+ManifoldCrossSectionVec *manifold_alloc_cross_section_vec();
+ManifoldSimplePolygon *manifold_alloc_simple_polygon();
+ManifoldPolygons *manifold_alloc_polygons();
+ManifoldMeshGL *manifold_alloc_meshgl();
+ManifoldBox *manifold_alloc_box();
+ManifoldRect *manifold_alloc_rect();
+
 // destruction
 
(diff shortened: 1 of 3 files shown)


elalish:
Thanks! The C bindings are not my forte, so I'll be looking to others for review. However, one question - could this be simplified/improved if we didn't need to preserve backward compatibility? I ask because we're making a number of breaking changes right now for v3.0, so if we'd want to eventually, now is a good time.

geoffder:
> Thanks! The C bindings are not my forte, so I'll be looking to others for review. However, one question - could this be simplified/improved if we didn't need to preserve backward compatibility? I ask because we're making a number of breaking changes right now for v3.0, so if we'd want to eventually, now is a good time.

I think that adding the alternative is better than removing the existing option that allows the user to provide their own memory. It allows buffer re-use if that is desired, and depending on the host language it might be less fuss to have it handle the allocation rather than registering something allocated elsewhere. I'm not an expert in how other languages handle FFI, but I've never had problems with the current approach.

NickUfer:
> Also, does this fix the issue you [found](https://github.com/elalish/manifold/discussions/681#discussioncomment-10802177), or only provide a working alternative?

To also clear this one up: The code I posted was my own testing and wasn't in this project 😄 This was basically the first approach of this PR, where I found out it was leaking memory. This PR contains the fixed version and does not leak anymore.

elalish (approved):
Okay, this looks pretty safe - thanks! Also, you may have noticed that the testing on our C-bindings is a bit sparse. Feel free to add some, particularly things where you'd like some assurance they keep working as you expect.

NickUfer:
For now I replaced the manual malloc + struct size function calls by the new allocaton functions. Under the hood the same happens but it now tests the existing and new functions at the same time

elalish (approved):
Good call, thanks.

Issue #969, opened 2024-10-01
Cleanup


pca006132:
A list of cleanup to do before v3.0.

- [x] https://github.com/elalish/manifold/blob/master/RELEASE_CHECKLIST.md
- [x] Cleanup `.vscode/launch.json`.
- [x] Move the public header install logic to the main cmake file?
- [x] Remove the single mesh file in `extras/Thingi10K/raw_meshes`.
- [ ] Organize `extras`, e.g. move things related to hull testing to its own folder, add readme to explain the use of each file.
- [x] `format.sh` format cmake.
- [x] Remove `.gitmodules`.
- [x] Update readme about the `clang-format` version requirement.
- [x] Workaround #944 by adding ignore flags.
- [ ] Try to make fuzzing work again.
- [x] Check the documentation to make sure things are up-to-date.

starseeker:
For the formatting of cmake, did you have a preferred solution in mind?

pca006132:
No, maybe https://github.com/cheshirekow/cmake_format? I haven't yet tried it though.

pca006132:
@elalish for fuzzing, I encountered https://github.com/google/fuzztest/issues/1022, not sure how to fix it.

starseeker:
@pca006132 It's not cleanup, but were we going to decide what to do about glm in the public headers et. al. prior to 3.0?  Didn't know if that should be added to the 3.0 milestone...

pca006132:
Yes, but not sure if I should convert the discussion into issue, the discussion is very long and has multiple threads.

elalish:
Yeah, 1) sounds like the right approach for v3.0. I'm comfortable crossing the 2) bridge when we come to it. We did a similar thing we when we built `MeshGL` parallel to `Mesh` without a breaking change - we can do that again if necessary. Unless someone has a great idea for a simple thing we should do right now, I'm going to ignore it.

pca006132:
Currently, about half of our files in extras do not work:

1. `minimize_testcase.cpp` does not parse the current polygon dump format.
2. `perf_test_cgal.cpp` does not compile (we should remove it anyway)
3. `test_hull_performance.cpp` does not compile.

I'm thinking if we should fix these before the release, or it is fine to just leave them as is for now. These are just for our internal use only anyway.

elalish:
Yeah, I wouldn't call anything in extras a priority. Might be good to remove `perf_test_cgal` at least so there's one less optional dependency to confuse people.

[...]

PR #973, opened 2024-10-01
simple cleanup

pca006132:


.github/workflows/manifold.yml
```diff
@@ -27,7 +27,6 @@ jobs:
     env:
       CC: gcc-${{ matrix.gcc }}
       CXX: g++-${{ matrix.gcc }}
-    if: github.event.pull_request.draft == false
```
elalish, line 30: 👍

src/boolean3.cpp
```diff
@@ -35,7 +35,7 @@ vec2 Interpolate(vec3 pL, vec3 pR, double x) {
   const bool useL = fabs(dxL) < fabs(dxR);
   const vec3 dLR = pR - pL;
   const double lambda = (useL ? dxL : dxR) / dLR.x;
-  if (!isfinite(lambda) || !isfinite(dLR.y) || !isfinite(dLR.z))
+  if (!std::isfinite(lambda) || !std::isfinite(dLR.y) || !std::isfinite(dLR.z))
```
elalish, line 38: Looks good, but for my own info, what is the difference between e.g. `isfinite` and `std::isfinite`?

src/boolean_result.cpp
```diff
@@ -14,9 +14,12 @@
 
 #include <algorithm>
 #include <array>
-#include <iostream>
 #include <map>
 
+#ifdef MANIFOLD_DEBUG
+#include <iostream>
```
elalish, line 20: Good catch! How long has that been there? Should we have this conditional include of string stuff in `utils.h` and then have a rule against including these headers anywhere else?

elalish (changes requested):
Thanks, this looks great! How much of #969 does this address?

src/boolean3.cpp
pca006132: One is defined in cmath without namespace, one is in C++. I think they are practically the same.

src/boolean_result.cpp
pca006132: Sounds good.

pca006132:
Fixed 4 of them. Was focused on code cleanup, and removed a few files, so there are still trivial things that are not yet addressed here.

elalish (approved):
Thanks! Is this ready to merge?

PR #975, opened 2024-10-03
format cmake files

pca006132:
Use `cmake-format` for formatting, and changed some of the old `set` calls to `list(APPEND ...)` which should be nicer.


pca006132:
This uses https://github.com/cheshirekow/cmake_format

@starseeker see if you have any comments.

The issue now is that I don't have a nice way to check for cmake format in the CI. `cmake-format` does not provide any error option to exit with error when the files are not properly formatted. I think the simplest thing we can do is to run it in the CI and use git diff to check for differences?

starseeker:
@pca006132 https://github.com/marketplace/actions/cmake-format-lint might be useful here?  I'm not 100% sure, but it looks as if they may be checking for differences as a result of formatting.  Their code is at https://github.com/neg-c/cmake-format-action

starseeker:
Did you check https://github.com/BlankSpruce/gersemi as well?  It's got a --diff option that might be handy for checking for changes.

starseeker:
Or actually, in gersemi's case, the -c option might be exactly what we want.

pca006132:
hmmm I can try that

pca006132:
yeah gersemi looks better, and easier to integrate with CI

src/CMakeLists.txt
```diff
@@ -113,31 +114,24 @@ if(MANIFOLD_EXPORT)
 endif()
 target_compile_features(manifold PUBLIC cxx_std_17)
 if(MANIFOLD_PAR)
-  target_compile_options(manifold INTERFACE -DMANIFOLD_PAR)
+  target_compile_options(manifold PUBLIC -DMANIFOLD_PAR)
```
pca006132, line 117: This is an interesting one, `INTERFACE` will make the compile option only available to external code linking with our library, so it is not actually using parallelization. Changing to public fixes that.

pca006132: and this makes me thinking if we should *always* set these to some non-zero values and assert the thing should be non-zero in our code, to avoid this sort of issue (and potential typos, which we had before)

pca006132:
@starseeker regarding the `TODO` you placed, I cannot find where tbb said we should not be doing fetch content to download it. Can you give me a reference to the issue/pr?

elalish (approved):
Thanks!

starseeker:
> @starseeker regarding the `TODO` you placed, I cannot find where tbb said we should not be doing fetch content to download it. Can you give me a reference to the issue/pr?

I think I was conflating that with some discussions I recollected about issues building it as a static library.  You can go ahead and remove that TODO.

pca006132:
> > @starseeker regarding the `TODO` you placed, I cannot find where tbb said we should not be doing fetch content to download it. Can you give me a reference to the issue/pr?
> 
> I think I was conflating that with some discussions I recollected about issues building it as a static library. You can go ahead and remove that TODO.

Sure, we can fix that when we fix our documentation in another clean up.

src/CMakeLists.txt
starseeker: Urk.  That's my mistake - I was misunderstanding the PUBLIC vs. INTERFACE settings.  I thought PRIVATE was just for the target, PUBLIC was for all other targets in the current CMake project linking that target, and INTERFACE meant everything... I see now that is incorrect, apologies.

For future reference (mostly for me) here's the link to CMake's docs on those three keywords:
https://cmake.org/cmake/help/latest/manual/cmake-buildsystem.7.html#target-command-scope

starseeker: @pca006132 Were you thinking something like this for asserting non-zero for parallel values?  https://github.com/starseeker/manifold/commit/c94e8a20b1205b0cfc725d72c9d568a011d07e62

pca006132: @starseeker maybe a bit simpler, 1 for true and -1 for false. This will work for other macros as well.

starseeker: @pca006132 Updated.  Pull request up: https://github.com/elalish/manifold/pull/980

PR #978, opened 2024-10-05
Collider optimization

pca006132:
Unrelated:

1. Fixed tracy (again the cmake interface vs public issue)
2. Moved some APIs from `Vec` to `VecView`.

Optimizations:

1. Only do collision check once, use thread local buffers to store the result and merge in parallel when everything is completed. Note that this requires making the vector copy sequential, so I added some flags to help.
2. Allow reusing collider result buffer, reduce reallocation during complex polygon triangulation.
3. Some low level optimization.

This results in about 8% time reduction for `perfTest` (2.5s to 2.3s), and 10% time reduction for running triangulation tests (~1900ms to ~1700ms)

### DIFF
--- a/src/polygon.cpp
+++ b/src/polygon.cpp
@@ -524,7 +524,7 @@ class EarClip {
           }
           return cost;
         }
-        return std::numeric_limits<double>::min();
+        return std::numeric_limits<double>::lowest();
       };
       totalCost = transform_reduce(
           countAt(0), countAt(collider.ind.size()), totalCost,
--- a/src/polygon.cpp
+++ b/src/polygon.cpp
@@ -489,7 +489,8 @@ class EarClip {
     // values < -precision so they will never affect validity. The first
     // totalCost is designed to give priority to sharper angles. Any cost < (-1
     // - precision) has satisfied the Delaunay condition.
-    double EarCost(double precision, const IdxCollider &collider) const {
+    double EarCost(double precision, const IdxCollider &collider,
+                   SparseIndices &toTest) const {
       vec2 openSide = left->pos - right->pos;
       const vec2 center = 0.5 * (left->pos + right->pos);
       const double scale = 4 / glm::dot(openSide, openSide);
@@ -506,7 +507,7 @@ class EarClip {
       earBox.push_back({vec3(center.x - radius, center.y - radius, 0),
                         vec3(center.x + radius, center.y + radius, 0)});
       earBox.back().Union(vec3(pos, 0));
-      const SparseIndices toTest = collider.collider.Collisions(earBox.cview());
+      collider.collider.Collisions(earBox.cview(), toTest);
 
       const int lid = left->mesh_idx;
       const int rid = right->mesh_idx;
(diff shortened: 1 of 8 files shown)


src/polygon.cpp
```diff
@@ -489,7 +489,8 @@ class EarClip {
     // values < -precision so they will never affect validity. The first
     // totalCost is designed to give priority to sharper angles. Any cost < (-1
     // - precision) has satisfied the Delaunay condition.
-    double EarCost(double precision, const IdxCollider &collider) const {
+    double EarCost(double precision, const IdxCollider &collider,
+                   SparseIndices &toTest) const {
```
pca006132, line 493: Thinking about it, I should probably bundle `toTest` to `IdxCollider`.

src/polygon.cpp
```diff
@@ -502,27 +504,32 @@ class EarClip {
...
-          totalCost = std::max(totalCost, cost);
+          return cost;
         }
-      }
-
+        return std::numeric_limits<double>::lowest();
+      };
+      totalCost = transform_reduce(
+          countAt(0), countAt(collider.ind.size()), totalCost,
+          [](double a, double b) { return std::max(a, b); }, f);
```
elalish, line 531: If `f` is only being used in one place, should we just put it here directly?

src/vec.h
```diff
@@ -145,25 +143,31 @@ class Vec : public VecView<T> {
...
       this->ptr_[this->size_++] = val_copy;
       return;
     }
     this->ptr_[this->size_++] = val;
   }
 
-  void reserve(size_t n) {
+  inline void extend(size_t n, bool seq = false) {
+    if (this->size_ + n >= capacity_)
+      reserve(capacity_ == 0 ? 128 : capacity_ * 2, seq);
```
elalish, line 159: Is this right? If `n` is much larger than `capacity_`, will it actually increase enough?

elalish (changes requested):
Thanks, looks great!

src/vec.h
pca006132: Nice catch, I copied the code from `push_back` and forgot that n can be large.

elalish (approved):
LGTM, thanks! Any idea what's wrong with the Black formatter?

pca006132:
Hmmm no idea, it seems that everything somehow went wrong. We can try to run the CI again...

pca006132:
The weird thing is that it starts to fail after the last commit, but the last commit should not affect anything...

pca006132:
I will investigate this later this week.

pca006132:
Just got some time, I still don't understand why the previous test failure can happen, but I guess it is caused by the cmake changes in the master branch. The weird thing is that I haven't merged the changes at [f632177](https://github.com/elalish/manifold/pull/978/commits/f6321777acb4d0c082907256b8fb16696d616d5f). Anyway, the tests pass now.

For the black formatter error, I think this is caused by the latest version (released a day ago) no longer supports python 3.8, and Ubuntu 20.04 ship with python 3.8 by default. Should be fixed by installing python 3.12. (which I did previously for the cmake formatter because I forgot Ubuntu ships with python preinstalled...)

elalish (approved):
Thanks!

PR #980, opened 2024-10-07
Always check in the code for explicit parallel/serial choice

starseeker:
In order to avoid accident serial (or parallel, for that matter) compilation settings, always check in the code for an explicit and valid MANIFOLD_PAR value.


pca006132:
Nice. Maybe we want to do this to other flags as well? e.g. MANIFOLD_DEBUG.

@elalish what do you think?

elalish (approved):
Thanks!

elalish:
> Nice. Maybe we want to do this to other flags as well? e.g. MANIFOLD_DEBUG.
> 
> @elalish what do you think?

Yeah, actually it's probably a good idea to be uniform about this. This is basically to catch our own CMake errors, right? And those could happen with any flag, so let's make this part of our coding style.

starseeker:
The thing about doing it for the debugging flag... I would expect, by default, user codes wouldn't turn debugging on - so it may feel a little weird to them to have to explicitly say it is off.  Parallel I think makes sense because of the importance - if you try to turn parallel on or off and get it wrong the consequences can be severe - but we might not want to force client codes to set a large number of defines to build.

pca006132:
Yeah, putting it in our internal files will be good enough.

starseeker:
Probably should do the debug flag in a separate PR?

elalish (approved):
Looks good, thanks!

PR #984, opened 2024-10-11
Replace GLM with linalg

### LINKED ISSUE #976: Refactor public API to hide glm as an implementation detail
Per discussions in https://github.com/elalish/manifold/discussions/962 we want to look into only exposing standard data types in the public manifold headers, translating them (preferably using casting) to glm types under the hood.

Scope is deliberately limited in this issue to the changes needed to allow applications to compile against an installed Manifold instance without also requiring glm.  (Other discussed topics such as replacement of glm as the internal math library, "zero-copy" inputs, etc. are considered separate topics from this issue.)

elalish:
Fixes #976 
Related to #962 

Here I'm removing our GLM dependency in favor of including a forked version of [linalg](https://github.com/sgorsten/linalg) directly, which is much smaller and simpler. 

A question: should we have a follow-on that converts our public API from `linalg::vec` to `std::array` so that the `linalg.h` header is internal instead of public? `linalg` already provides these implicit conversions (through a copy).


include/manifold/linalg.h
```diff
@@ -618,6 +618,12 @@ struct op_or {
 };
 
 // Function objects for applying standard library math functions
+struct std_isfinite {
+  template <class A>
+  auto operator()(A a) const -> decltype(std::isfinite(a)) {
+    return std::isfinite(a);
```
elalish, line 624: I added a couple of functions that we commonly use.

include/manifold/linalg.h
```diff
@@ -828,10 +841,10 @@ struct vec<T, 4> {
...
   constexpr explicit vec(const T &s) : vec(s, s, s, s) {}
   constexpr explicit vec(const T *p) : vec(p[0], p[1], p[2], p[3]) {}
-  template <class U>
-  constexpr explicit vec(const vec<U, 4> &v)
-      : vec(static_cast<T>(v.x), static_cast<T>(v.y), static_cast<T>(v.z),
-            static_cast<T>(v.w)) {}
+  template <class U, int N>
+  constexpr explicit vec(const vec<U, N> &v)
+      : vec(static_cast<T>(v.x), N < 2 ? 0 : static_cast<T>(v.y),
+            N < 3 ? 0 : static_cast<T>(v.z), N < 4 ? 0 : static_cast<T>(v.w)) {}
```
elalish, line 847: I changed their constructor to allow it to take vectors of different lengths like GLM does.

include/manifold/linalg.h
```diff
@@ -0,0 +1,2189 @@
...
+constexpr apply_t<detail::op_rsh, A, B> operator>>(const A &a, const B &b) {
+  return apply(detail::op_rsh{}, a, b);
+}
+
+// Binary `operator *` was originally defined component-wise for all patterns,
+// in a fashion consistent with the other operators. However, this was one of
+// the most frequent sources of confusion among new users of this library, with
+// the binary `operator *` being used accidentally by users who INTENDED the
+// semantics of the algebraic matrix product, but RECEIVED the semantics of the
+// Hadamard product. While there is precedent within the HLSL, Fortran, R, APL,
```
elalish, line 1214: This is dumb - I hate Hadamard products. I'd like to switch to the opposite syntax: make `*` an alias for `mul`. @pca006132 I'm having a little trouble following their templating syntax - do you know how to make this change?

include/manifold/linalg.h
elalish: Does anyone know a nice templatey way to allow these to take a wider variety of inputs a la GLM, e.g. `vec3(double, vec2)` or `vec4(vec3, double)`?

elalish:
I've got 80% done I think. Most of the remaining errors will be cleared by taking care of the comments above, and then I've just got a few more specialized math functions to rename and/or implement.

include/manifold/linalg.h
elalish: Okay, I think I figured this one out.

include/manifold/linalg.h
elalish: Actually, linalg already has the second type - that's probably good enough. What I really need is `mat4(mat3)` and `mat3(mat4x3)` etc.

elalish: Okay, I think we're good here now too.

pca006132:
> A question: should we have a follow-on that converts our public API from linalg::vec to std::array so that the linalg.h header is internal instead of public? linalg already provides these implicit conversions (through a copy).

Probably not needed, considering the header is included here already and it is small. Even if we return `std::array`, users still have to do conversion to their own math library data type, which is the same as exposing linalg types.

elalish:
Okay, still some work to do to get this compiling, but it's pretty close.

include/manifold/common.h
```diff
@@ -13,59 +13,66 @@
...
+using vec3 = la::vec<double, 3>;
+using vec4 = la::vec<double, 4>;
+using bvec4 = la::vec<bool, 4>;
+using mat2 = la::mat<double, 2, 2>;
+using mat3x2 = la::mat<double, 3, 2>;
+using mat4x2 = la::mat<double, 4, 2>;
+using mat2x3 = la::mat<double, 2, 3>;
+using mat3 = la::mat<double, 3, 3>;
+using mat4x3 = la::mat<double, 4, 3>;
+using mat3x4 = la::mat<double, 3, 4>;
```
elalish, line 39: I fixed our matrix naming from the stupid backwards way GLM had it.

include/manifold/common.h
```diff
@@ -13,59 +13,66 @@
...
 
+constexpr double kPi = 3.14159265358979323846264338327950288;
+constexpr double kTwoPi = 6.28318530717958647692528676655900576;
+constexpr double kHalfPi = 1.57079632679489661923132169163975144;
+
+inline double radians(double a) { return a * kPi / 180; }
+inline double degrees(double a) { return a * 180 / kPi; }
+
+inline mat3x4 Identity3x4() { return mat3x4(mat3(la::identity), vec3(0.0)); }
+inline mat2x3 Identity2x3() { return mat2x3(mat2(la::identity), vec2(0.0)); }
```
elalish, line 55: linalg doesn't do non-square identities like GLM did.

src/utils.h
```diff
@@ -255,34 +260,38 @@ inline mat4x3 RotateUp(vec3 up) {
...
+template <class T>
+std::ostream& operator<<(std::ostream& out, const la::vec<T, 4>& v) {
+  return out << '{' << v[0] << ',' << v[1] << ',' << v[2] << ',' << v[3] << '}';
 }
 
-inline std::ostream& operator<<(std::ostream& stream, const mat4x3& mat) {
-  mat3x4 tam = glm::transpose(mat);
-  return stream << tam[0] << std::endl
-                << tam[1] << std::endl
-                << tam[2] << std::endl;
```
elalish, line 285: Actually, I think I like my old ones better than what I grabbed from linalg here - I may revert or combine.

include/manifold/common.h
```diff
@@ -13,59 +13,66 @@
...
+using mat4 = la::mat<double, 4, 4>;
+using ivec2 = la::vec<int, 2>;
+using ivec3 = la::vec<int, 3>;
+using ivec4 = la::vec<int, 4>;
+using quat = la::vec<double, 4>;
 ///@}
 
+constexpr double kPi = 3.14159265358979323846264338327950288;
+constexpr double kTwoPi = 6.28318530717958647692528676655900576;
+constexpr double kHalfPi = 1.57079632679489661923132169163975144;
```
elalish, line 49: does this seem like a reasonable place to define these?

elalish:
Okay, I've got it compiling and running now (locally at least). Some tests even pass! I definitely have some math errors to track down. @pca006132 if you have a chance to look, it would be great to have another set of eyes on this. I bet the problems are mostly just typos. Also the CI is suddenly more annoyed about `size_t -> int` conversions.

src/utils.h
elalish: nevermind, I think I like theirs better after all.

include/manifold/common.h
```diff
@@ -13,59 +13,71 @@
...
-using mat3x2 = glm::dmat3x2;
-using mat3 = glm::dmat3;
-using mat3x4 = glm::dmat3x4;
-using mat4x3 = glm::dmat4x3;
-using mat4 = glm::dmat4;
-using ivec2 = glm::vec<2, int>;
-using ivec3 = glm::vec<3, int>;
-using ivec4 = glm::vec<4, int>;
-using quat = glm::dquat;
+namespace la = linalg;
```
pca006132, line 28: Not sure if we want this namespace alias in our public header. Maybe define this in our internal header?

include/manifold/common.h
pca006132: I think `cmath` should have `M_PI` and `M_PI_2` defined, but anyway this should be fine.

include/manifold/common.h
```diff
@@ -13,59 +13,71 @@
...
+using ivec3 = la::vec<int, 3>;
+using ivec4 = la::vec<int, 4>;
+using quat = la::vec<double, 4>;
 ///@}
 
+constexpr double kPi = 3.14159265358979323846264338327950288;
+constexpr double kTwoPi = 6.28318530717958647692528676655900576;
+constexpr double kHalfPi = 1.57079632679489661923132169163975144;
+
+inline double radians(double a) { return a * kPi / 180; }
```
pca006132, line 51: `constexpr`?

src/csg_tree.cpp
```diff
@@ -389,10 +388,10 @@ std::shared_ptr<CsgNode> CsgOpNode::Boolean(
   return std::make_shared<CsgOpNode>(children, op);
 }
 
-std::shared_ptr<CsgNode> CsgOpNode::Transform(const mat4x3 &m) const {
+std::shared_ptr<CsgNode> CsgOpNode::Transform(const mat3x4 &m) const {
   auto node = std::make_shared<CsgOpNode>();
   node->impl_ = impl_;
-  node->transform_ = m * mat4(transform_);
+  node->transform_ = m * Mat4(transform_);
```
pca006132, line 394: why is this `Mat` (uppercase)?

pca006132 (commented):
I think for the narrowing warnings, we can just disable them for now. We do so many narrowing operations that it is not worth it to mark them explicitly each time.

src/csg_tree.cpp
pca006132: Ah this is a function, no problem now.

src/csg_tree.cpp
```diff
@@ -111,29 +111,29 @@ std::shared_ptr<CsgNode> CsgNode::Boolean(
 }
 
 std::shared_ptr<CsgNode> CsgNode::Translate(const vec3 &t) const {
-  mat4x3 transform(1.0);
+  mat3x4 transform(1.0);
```
pca006132, line 114: This is not the identity matrix, this is a matrix with all entries = 1.

include/manifold/common.h
pca006132: It seems that this is fine, we are having `la` alias inside our manifold namespace, so this will not cause conflict elsewhere.

CMakeLists.txt
```diff
@@ -156,13 +156,18 @@ else()
       APPEND
       WARNING_FLAGS
       -Wall
+      -Wno-unknown-warning-option
       -Wno-unused
-      -Wno-array-bounds
```
pca006132, line 160: I think those are left from the thrust era.

test/test_main.cpp
```diff
@@ -341,10 +342,10 @@ void RelatedGL(const Manifold& out, const std::vector<MeshGL>& originals,
       }
       const int inTri = output.faceID.empty() ? tri : output.faceID[tri];
       ASSERT_LT(inTri, inMesh.triVerts.size() / 3);
-      ivec3 inTriangle = {inMesh.triVerts[3 * inTri],
-                          inMesh.triVerts[3 * inTri + 1],
-                          inMesh.triVerts[3 * inTri + 2]};
-      inTriangle *= inMesh.numProp;
+      ivec3 inTriangle(inMesh.triVerts[3 * inTri],
```
pca006132, line 345: Apparently, the compiler is more picky when it performs implicit conversion when putting things inside the implicit constructor. Maybe the more implicit it gets, the more likely it is an issue...

src/csg_tree.cpp
elalish: Good catch - I had fixed a bunch of those, but must have missed this one.

include/manifold/common.h
```diff
@@ -578,5 +578,3 @@ struct ExecutionParams {
 };
 
 }  // namespace manifold
-
-#undef HOST_DEVICE
```
elalish, line 572: 👍

elalish (commented):
Thanks for debugging this for me, not to mention all the cleanup!

include/manifold/common.h
elalish: Oof, well an hour later I know understand painfully well to never use `M_PI`.

elalish:
I like the idea of simplifying the determinant above - let's do that as a follow-up, once we have assurance from Windows that it's not slower.

fire:
Horray!

pca006132:
> Windows users wanted! Can someone using Windows check if this PR is faster/slower comparing with the current master, when running `extras/perfTest` and `test/manifold_test`.
> 
> In addition, try the following patch and see if it affects anything:
> 
> ```diff
> diff --git a/src/polygon.cpp b/src/polygon.cpp
> index 8f8fcdaa..6625288f 100644
> --- a/src/polygon.cpp
> +++ b/src/polygon.cpp
> @@ -30,11 +30,6 @@ static ExecutionParams params;
>  
>  constexpr double kBest = -std::numeric_limits<double>::infinity();
>  
> -// it seems that MSVC cannot optimize la::determinant(mat2(a, b))
> -constexpr double determinant2x2(vec2 a, vec2 b) {
> -  return a.x * b.y - a.y * b.x;
> -}
> -
>  #ifdef MANIFOLD_DEBUG
>  struct PolyEdge {
>    int startVert, endVert;
> @@ -180,7 +175,7 @@ bool IsConvex(const PolygonsIdx &polys, double precision) {
>      for (size_t v = 0; v < poly.size(); ++v) {
>        const vec2 edge =
>            v + 1 < poly.size() ? poly[v + 1].pos - poly[v].pos : firstEdge;
> -      const double det = determinant2x2(lastEdge, edge);
> +      const double det = la::determinant(mat2(lastEdge, edge));
>        if (det <= 0 ||
>            (std::abs(det) < precision && la::dot(lastEdge, edge) < 0))
>          return false;
> @@ -454,11 +449,11 @@ class EarClip {
>      // goes to the outside. No need to check the other side, since all verts are
>      // processed in the EarCost loop.
>      double SignedDist(VertItr v, vec2 unit, double precision) const {
> -      double d = determinant2x2(unit, v->pos - pos);
> +      double d = la::determinant(mat2(unit, v->pos - pos));
>        if (std::abs(d) < precision) {
> -        double dR = determinant2x2(unit, v->right->pos - pos);
> +        double dR = la::determinant(mat2(unit, v->right->pos - pos));
>          if (std::abs(dR) > precision) return dR;
> -        double dL = determinant2x2(unit, v->left->pos - pos);
> +        double dL = la::determinant(mat2(unit, v->left->pos - pos));
>          if (std::abs(dL) > precision) return dL;
>        }
>        return d;
> @@ -470,7 +465,7 @@ class EarClip {
>        double cost = std::min(SignedDist(v, rightDir, precision),
>                               SignedDist(v, left->rightDir, precision));
>  
> -      const double openCost = determinant2x2(openSide, v->pos - right->pos);
> +      const double openCost = la::determinant(mat2(openSide, v->pos - right->pos));
>        return std::min(cost, openCost);
>      }
>  
> @@ -672,7 +667,7 @@ class EarClip {
>      auto AddPoint = [&](VertItr v) {
>        bBox.Union(v->pos);
>        const double area1 =
> -          determinant2x2(v->pos - origin, v->right->pos - origin);
> +          la::determinant(mat2(v->pos - origin, v->right->pos - origin));
>        const double t1 = area + area1;
>        areaCompensation += (area - t1) + area1;
>        area = t1;
> ```

@starseeker do you have windows machines around for benchmark? If yes, can you try this and see if there is any difference in performance?

starseeker:
> @starseeker do you have windows machines around for benchmark? If yes, can you try this and see if there is any difference in performance?

I do, but it will take some time to get set up on on it.

starseeker:
Without patch:

```
.\perfTest.exe
nTri = 512, time = 0.002399 sec
nTri = 2048, time = 0.0057338 sec
nTri = 8192, time = 0.0191217 sec
nTri = 32768, time = 0.0676792 sec
nTri = 131072, time = 0.256196 sec
nTri = 524288, time = 1.04938 sec

.\largeSceneTest.exe
n = 20
nTri = 91814, time = 24.9764 sec
```

With patch:
```
.\perfTest.exe
nTri = 512, time = 0.0020762 sec
nTri = 2048, time = 0.0060967 sec
nTri = 8192, time = 0.0194322 sec
nTri = 32768, time = 0.0685545 sec
nTri = 131072, time = 0.265982 sec
nTri = 524288, time = 1.10732 sec
nTri = 2097152, time = 4.72829 sec

.\largeSceneTest.exe
n = 20
nTri = 91814, time = 23.7445 sec
```

elalish:
Okay, looks like we should apply the patch then, thanks! 👍

pca006132:
The weird thing though is that the performance difference is pretty large and inconsistent, maybe we should try multiple runs and account for deviations, and let the system rest before running another run because the CPU may be hotter for example. Alternatively, we may want to run these single threaded with turboboost disabled to avoid deviation due to CPU states.

E.g. for nTri = 524288, the patch made it slower by 4.5%, and pretty consistently slower for other sphere sizes. For largeSceneTest however, it is somehow 5% faster.

PR #986, opened 2024-10-13
CI: libthrust-dev dependency no longer needed

cjmayo:


fire:
libglm-dev is also not needed now.

pca006132:
nice catch

PR #987, opened 2024-10-15
fix python binding

pca006132:
Fixes https://github.com/elalish/manifold/issues/947#issuecomment-2399029206

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -26,7 +26,7 @@ dependencies = [
 
 [build-system]
 requires = [
-    "nanobind>=1.8.0",
+    "nanobind>=1.8.0,<=2.2.0",
     "scikit-build-core",
 ]
 build-backend = "scikit_build_core.build"
--- a/bindings/python/CMakeLists.txt
+++ b/bindings/python/CMakeLists.txt
@@ -96,6 +96,8 @@ set(
   ${CMAKE_SOURCE_DIR}/src/cross_section/cross_section.cpp
   ${CMAKE_SOURCE_DIR}/src/polygon.cpp
   ${CMAKE_SOURCE_DIR}/include/manifold/common.h
+  ${CMAKE_CURRENT_SOURCE_DIR}/gen_docs.py
+  ${CMAKE_CURRENT_SOURCE_DIR}/docstring_override.txt
 )
 add_custom_command(
   OUTPUT autogen_docstrings.inl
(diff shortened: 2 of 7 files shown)


pca006132:
Not entirely sure why the test is failing. May need a docker to test this.

pca006132:
I guess `warp_batch` was broken... will see how to fix that.

pca006132:
In fact, I am not sure if we should keep the `warp_batch` API. I think this kind of modifying array by reference is quite brittle with transfer of ownership. Returning a new array will always work, and I don't think that will add too much overhead.

pyproject.toml
```diff
@@ -26,7 +26,7 @@ dependencies = [
 
 [build-system]
 requires = [
-    "nanobind>=1.8.0",
+    "nanobind>=1.8.0,<=2.2.0",
```
elalish, line 29: Do we need this if we upgraded to 2.2 above?

elalish (commented):
Returning a new array sounds good to me.

pyproject.toml
pca006132: Just to be safe. I don't think they provide a stable API for returning ndarray without copy. Maybe we should just ask the upstream.

elalish: Will it matter anymore if we switch to returning a new array?

pca006132: Still matters, because we still want to return arrays without copying. Returning a new array just avoids reading back the array after passing it to the user, in case numpy did some weird things to the array.

elalish: Okay, yeah let's ask upstream. I think I'd rather eat the copy cost than rely on a non-stable API.

pca006132: Looking at the code, it seems that we are doing a copy anyway... Maybe we can find a simpler way.

pca006132:
Summary:

1. No longer depends on unstable API. I found the solution to be surprisingly simple: just use the type caster for ndarray, it automatically performs conversion from ndarray to python handle (which was why we were trying to call `ndarray_wrap`).
2. Changed the `warp_batch` API to require returning an array. I feel that we were lucky the previous code works, because it is probably relying on numpy not changing the underlying buffer for some operations.
3. Added docstring override functionality, so we can customize some docstring. I think we may also want to add replacement function later, for example to replace different function names due to differences in naming convention.
4. Download 2.2.0 of nanobind, and set the version string if we download it using `FetchContent` (nanobind is unavailable in python search path, so we cannot get the version string automatically in this case).
5. Limit nanobind dependency to 2.2.0. I don't think we relied on unstable APIs now, but to avoid potential future build failure, it is better to specify the latest version that we support. As long as we have regular releases, we can still update nanobind version.

elalish (approved):

PR #988, opened 2024-10-15
Minor cleanup

pca006132:
Update documentation and removed some todos.

Regarding files in `extras`, I found that:

1. All CGAL related code are not compiling due to our breaking APIs. I think we should build these in our CI as well.
2. `minimize_testcase.cpp` will not work now due to the different dump format.
3. `large_scene_test.cpp` is really slow. 

Will fix those in another PR.


elalish (approved):
Thanks!

PR #990, opened 2024-10-15
Make sure _MSC_VER is defined before checking

starseeker:
Avoids the following:

error: "_MSC_VER" is not defined, evaluates to 0 [-Werror=undef]
   59 | #if _MSC_VER <= 1900
      |     ^~~~~~~~


starseeker:
Starting integration testing with BRL-CAD, hit that error from linalg.h with our build flags.

pca006132 (approved):

PR #991, opened 2024-10-16
Separate coplanarity from faceID

elalish:
Simpler alternative to #951, since it doesn't change the logic we use for edge collapsing. 

Manifold will no longer use equal faceIDs to mean that those triangles are coplanar. This was always problematic, because it meant that using Warp or Refine would destroy the entire mesh relation, making material association quite difficult. This also means we can remove any concept of property tolerances from our API, which I never liked. Now edges can collapse if the two triangles are coplanar, have the same meshID, and reference the same propVerts on both sides of the edge.


elalish:
@pca006132 You're right, I can't find any noticeable performance impact from `CreateFaces` anymore, so I've gone ahead and put it back into the functions I removed it from a few PRs back. Nice job optimizing it!

include/manifold/manifold.h
```diff
@@ -78,13 +78,10 @@ struct MeshGLP {
...
-  /// comes from. When auto-generated, this ID will be a triangle index into the
-  /// original mesh. All neighboring coplanar triangles from that input mesh
-  /// will refer to a single triangle of that group as the faceID. When
-  /// supplying faceIDs, ensure that triangles with the same ID are in fact
-  /// coplanar and have consistent properties (within some tolerance) or the
-  /// output will be surprising.
+  /// Optional: Length NumTri, contains the source face ID this
+  /// triangle comes from. When auto-generated, this ID will be a triangle index
+  /// into the original mesh. This index/ID is purely for external use (e.g.
+  /// recreating polygonal faces) and will not affect Manifold's algorithms.
```
elalish, line 84: One question: should we rename our public `faceID`, since it really has nothing to do with faces now? Maybe `triID` or `triIndex`? 

The worst part is now in our internal `TriRef` struct we have `faceID` which really is a face ID and `tri` which is the internal analogue to this public `faceID`. That's pretty much bound to confuse someone eventually, probably me.

pca006132: Yeah, `triID` sounds better fit here. We should add some documentation for the internal `faceID` to clarify things a bit as well.

src/impl.cpp
```diff
@@ -339,40 +300,67 @@ void Manifold::Impl::RemoveUnreferencedVerts() {
...
   const int meshID = ReserveIDs(1);
   meshRelation_.originalID = meshID;
   auto& triRef = meshRelation_.triRef;
   triRef.resize(NumTri());
   for_each_n(autoPolicy(NumTri(), 1e5), countAt(0), NumTri(),
-             [meshID, &triRef](const int tri) {
-               triRef[tri] = {meshID, meshID, tri};
+             [meshID, keepFaceID, &triRef](const int tri) {
+               triRef[tri] = {meshID, meshID, tri,
+                              keepFaceID ? triRef[tri].faceID : tri};
```
pca006132, line 311: What will happen if we always keep face ID?

pca006132 (approved):

src/impl.cpp
elalish: I think that would be weird - for a regular initialization, that would mean every faceID was zero, which would imply the whole object was a single face. It would probably confuse something at some point.

include/manifold/manifold.h
elalish: Since that will be a big change across all of our binding surfaces and such, I'll leave that to its own PR.

Issue #993, opened 2024-10-16
Performance regression with many multiple sequential union operations


starseeker:
In BRL-CAD, we convert an implicit volume mesh to an explicit volume the "hard way" by representing each mesh element as a shape and unioning them together.  It's a brute force approach, but with Manifold's booleans it has proven able to handle simple shapes reasonably well.

With the latest code (started testing after linalg change, but delta also includes MeshGL64 and double precision changes) a problem that was taking ~3 seconds to process with v2.4.5 is now taking a couple of minutes, and considerably longer if the build isn't compiled optimized.

I'm working on making a more isolated encapsulation of the issue to try and help narrow down what's going on - discussion in #989 got to the point were we needed another issue, so setting this up.  As I get more info, will update here.

starseeker:
OK, first progress at setting up a self contained reproduction of what we're doing:

https://github.com/starseeker/manifold/commit/592b51ebcff5a18cf29973297e9bae2ee51811c8

It's using the new APIs, so it'll have to be back ported to properly get a sense of how it behaves with older codes in this form.  The triangle count does end up higher than I expected, so that could very well be what's going on - I think it peaked out at about 600k.

starseeker:
Oh, wow.  Yeah, that's gotta be it.  A quick-and-dirty back-port to the v2.4.5 code shows a peak triangle count of just shy of 11k.

Hrm.  If the simplification function is exposed, could I apply that between iterations with a tolerance closer to the old version, maybe?

elalish:
Take a look at the discussion this issue started: #994. We're looking at exactly that kind of API - feedback welcome.

starseeker:
Did you want this in some form or other as a pull request?  I wasn't sure it was suitable, since it was basically the shortest path for me to be (reasonably) sure I was reproducing my specific issue....  I doubt it can be a normal test though, since it's ultimately timing based, so if you do want it in some form please advise what would be best.

elalish:
This is perfect as a test - you can add it to `boolean_complex_test.cpp`. You'll see others in there that also import meshes - follow their example. You can simply add a surface area and volume check to serve as a regression test of correctness. GTest tells us how long each test takes to run, and we tend to pay a fair bit of attention when something gets much slower. I ask that you size it so it takes no more than a few seconds when running properly; we'll notice anything that adds a couple of seconds to the overall runtime of the tests, so that'll serve as the performance regression test.

[...]

Issue #989, opened 2024-10-15
accidental quadratic runtime in csg tree


pca006132:
This causes `large_scene_test.cpp` to be slow. Because instead of running $n^3$ operations, we are actually having $n^6$ (simple) operations.

This is introduced in https://github.com/elalish/manifold/pull/368. Basically, the problem is that due to eager flattening, repeatedly unioning things will create and drop vectors of length $i$ from 1 to $n$, making the runtime quadratic. We should instead let the csg tree grow, flatten the csg tree when we evaluate it. To avoid stack overflow, we can avoid recursion, manually maintain a stack.

elalish:
Interesting! Good catch.

starseeker:
@pca006132 is this manifesting only with the linalg conversion in place?  Doing my initial tests with that version in BRL-CAD, I'm seeing an enormous slowdown of the logic where we do booleans on shapes representing each triangle of a mesh to perform offsetting.  I've not filed an issue yet since I've got multiple things in play, but I've somehow managed to have a problem that was previously ~3 seconds to calculate jump to taking many minutes.

pca006132:
It should not be related to `linalg`. If you can reduce to simpler example or purely manifold example, just file an issue.

pca006132:
> @pca006132 is this manifesting only with the linalg conversion in place? Doing my initial tests with that version in BRL-CAD, I'm seeing an enormous slowdown of the logic where we do booleans on shapes representing each triangle of a mesh to perform offsetting. I've not filed an issue yet since I've got multiple things in play, but I've somehow managed to have a problem that was previously ~3 seconds to calculate jump to taking many minutes.

Thinking about it, are you comparing against a version before the double precision PR? It may be caused by mesh simplification. Basically, previously we were doing very aggressive mesh simplification due to the epsilon value being quite large (1e-6 times the maximum absolute value of coordinates, roughly speaking). However, after the double precision PR, the epsilon was drastically reduced, so we can only do less mesh simplification.

This is also an issue in openscad as well, but we don't think it is something manifold can fix, because we cannot arbitrarily increase the epsilon value or things will become imprecise later. Users should try to do remeshing based on their target tolerance value, which should be much larger than epsilon (epsilon is the tolerance value that we *hope to* guarantee, considering imprecision in floating point arithmetic, so this should be as small as possible). We may expose our simplification function, but that one is aiming for fast simplification, so it is not trying too hard.

starseeker:
[...]

PR #997, opened 2024-10-18
[WIP] Add tolerance value

pca006132:
Implements #994.

- [x] Hide precision from public API and change its name.
- [x] Add tolerance value API and logic.
- [x] Read the field from MeshGL (somehow forgot that)
- [ ] Fix documentation.
- [ ] Test for using tolerance.
- [ ] Test for mesh simplification with increased tolerance.


src/manifold.cpp
```diff
@@ -381,14 +378,45 @@ size_t Manifold::NumPropVert() const {
...
-double Manifold::Precision() const {
-  return GetCsgLeafNode().GetImpl()->precision_;
+double Manifold::GetUncertainty() const {
+  return GetCsgLeafNode().GetImpl()->uncertainty_;
+}
+
+Manifold Manifold::SetUncertainty(double uncertainty) const {
+  auto impl = std::make_shared<Impl>(*GetCsgLeafNode().GetImpl());
+  impl->SetUncertainty(uncertainty);
+  impl->SimplifyTopology();
```
elalish, line 393: I would gate this on the tolerance increasing, as below.

src/properties.cpp
```diff
@@ -229,7 +229,7 @@ bool Manifold::Impl::Is2Manifold() const {
 bool Manifold::Impl::MatchesTriNormals() const {
   if (halfedge_.size() == 0 || faceNormal_.size() != NumTri()) return true;
   return all_of(countAt(0_uz), countAt(NumTri()),
-                CheckCCW({halfedge_, vertPos_, faceNormal_, 2 * precision_}));
+                CheckCCW({halfedge_, vertPos_, faceNormal_, 2 * tolerance_}));
```
elalish, line 232: Should this be uncertainty instead?

src/properties.cpp
```diff
@@ -252,7 +252,7 @@ Properties Manifold::Impl::GetProperties() const {
   double volumeCompensation = 0;
   for (size_t i = 0; i < NumTri(); ++i) {
     auto [area1, volume1] =
-        FaceAreaVolume({halfedge_, vertPos_, precision_})(i);
+        FaceAreaVolume({halfedge_, vertPos_, tolerance_})(i);
```
elalish, line 255: uncertainty instead?

test/properties_test.cpp
```diff
@@ -37,27 +37,26 @@ TEST(Properties, GetProperties) {
 
 TEST(Properties, Precision) {
```
elalish, line 38: Update test names

src/boolean_result.cpp
```diff
@@ -543,7 +543,7 @@ struct Barycentric {
 
     for (const int i : {0, 1, 2}) {
       const int vert = halfedgeR[3 * tri + i].startVert;
-      uvw[3 * tri + i] = GetBarycentric(vertPosR[vert], triPos, precision);
+      uvw[3 * tri + i] = GetBarycentric(vertPosR[vert], triPos, tolerance);
```
elalish, line 546: uncertainty instead?

src/face_op.cpp
```diff
@@ -130,7 +130,7 @@ void Manifold::Impl::Face2Tri(const Vec<int>& faceEdge,
     const PolygonsIdx polys =
         Face2Polygons(halfedge_.cbegin() + faceEdge[face],
                       halfedge_.cbegin() + faceEdge[face + 1], projection);
-    return TriangulateIdx(polys, precision_);
+    return TriangulateIdx(polys, tolerance_);
```
elalish, line 133: I think `Face2Tri` should be using uncertainty instead of tolerance. Otherwise it will generate negative area triangles with a thickness of tolerance, which I don't think we want.

elalish (changes requested):
Good start; I think we need to be pretty careful about where we use tolerance vs uncertainty. I think uncertainty is the epsilon in epsilon-valid, while tolerance is our simplication parameter.

src/face_op.cpp
pca006132: Yeah, thinking about it, tolerance should only be used when we are doing simplification.

elalish: Also, maybe rename `uncertainty` to `epsilon`? It would fit with our algorithm docs better.

src/edge_op.cpp
```diff
@@ -502,14 +502,14 @@ void Manifold::Impl::CollapseEdge(const int edge, std::vector<int>& edges) {
           // Don't collapse if the edges separating the faces are not colinear
           // (can happen when the two faces are coplanar).
           if (CCW(projection * pOld, projection * pLast, projection * pNew,
-                  precision_) != 0)
+                  tolerance_) != 0)
```
pca006132, line 505: I'm thinking about these CCW checks. I think some should probably be using `epsilon` instead of `tolerance`, but not sure which is which. E.g. for the next one which checks for if collapsing will cause inverted triangle, we should probably be using `epsilon`, but not sure if that is the only one required.

src/impl.cpp
```diff
@@ -139,11 +138,11 @@ struct CoplanarEdge {
     triArea[edgeFace] = area;
     triArea[pairFace] = areaPair;
     // Don't link degenerate triangles
-    if (area < length * precision || areaPair < lengthPair * precision) return;
+    if (area < length * tolerance || areaPair < lengthPair * tolerance) return;
```
pca006132, line 141: maybe this should be epsilon as well?

pca006132 (commented):
Thinking about this, I guess I should just go for the stricter one and use epsilon whenever I am not sure.

src/impl.cpp
```diff
@@ -139,11 +138,11 @@ struct CoplanarEdge {
     triArea[edgeFace] = area;
     triArea[pairFace] = areaPair;
     // Don't link degenerate triangles
-    if (area < length * precision || areaPair < lengthPair * precision) return;
+    if (area < length * epsilon || areaPair < lengthPair * epsilon) return;
 
     const double volume = std::abs(la::dot(normal, pairVec));
     // Only operate on coplanar triangles
-    if (volume > std::max(area, areaPair) * precision) return;
+    if (volume > std::max(area, areaPair) * epsilon) return;
```
elalish, line 145: I think tolerance is probably natural for coplanarity.

src/impl.cpp
```diff
@@ -171,7 +170,7 @@ struct CheckCoplanarity {
       // If any component vertex is not coplanar with the component's reference
       // triangle, unmark the entire component so that none of its triangles are
       // marked coplanar.
-      if (std::abs(la::dot(normal, vert - origin)) > precision) {
+      if (std::abs(la::dot(normal, vert - origin)) > epsilon) {
```
elalish, line 173: tolerance again.

src/manifold.cpp
```diff
@@ -374,14 +371,46 @@ size_t Manifold::NumPropVert() const {
...
-  return GetCsgLeafNode().GetImpl()->precision_;
+double Manifold::GetEpsilon() const {
+  return GetCsgLeafNode().GetImpl()->epsilon_;
+}
+
+Manifold Manifold::SetEpsilon(double epsilon) const {
+  auto impl = std::make_shared<Impl>(*GetCsgLeafNode().GetImpl());
+  auto oldTolerance = impl->tolerance_;
+  impl->SetEpsilon(epsilon);
+  if (impl->tolerance_ > oldTolerance) impl->SimplifyTopology();
```
elalish, line 387: If we moved this logic into the Impl functions could we reduce duplication by having SetEpsilon call SetTolerance?

test/test_main.cpp
```diff
@@ -299,6 +299,11 @@ void RelatedGL(const Manifold& out, const std::vector<MeshGL>& originals,
   ASSERT_FALSE(out.IsEmpty());
   const ivec3 normalIdx = updateNormals ? ivec3(3, 4, 5) : ivec3(0);
   MeshGL output = out.GetMeshGL(normalIdx);
+
+  float epsilon = std::max(static_cast<float>(out.GetEpsilon()),
+                           std::numeric_limits<float>::epsilon() *
+                               static_cast<float>(out.BoundingBox().Scale()));
```
elalish, line 305: Why is this necessary?

src/manifold.cpp
pca006132: yes

test/test_main.cpp
pca006132: This is getting MeshGL, which uses single precision. The epsilon value we track is for double precision, which is too small here.

pca006132:
For the test naming, I think I will leave it here as we should fix a bunch of other documentations and naming as well. Should we change `RefineToPrecision` to `RefineToTolerance`, for example?

pca006132:
@starseeker can you check if this helps? if you set the tolerance value after importing the mesh.

starseeker:
@pca006132 Trying various settings in this context doesn't seem to change the timing:  https://github.com/BRL-CAD/brlcad/commit/9a82ed3df4cf37b3ab0eb8383cced8a0bab7e3e6

test/test_main.cpp
elalish: Hmm, that sounds like a bug we should fix at the MeshGL level. Also, I thought I already had?

elalish:
@starseeker How certain are you that it was faster before? Manifold can be tricky to time properly - based on the operations you're doing, I'd be a bit surprised if it was really finishing orders of magnitude faster than now. Can you check out a particular commit, add your test and show the timing difference, for instance?

starseeker:
> @starseeker How certain are you that it was faster before? Manifold can be tricky to time properly - based on the operations you're doing, I'd be a bit surprised if it was really finishing orders of magnitude faster than now. Can you check out a particular commit, add your test and show the timing difference, for instance?

It's possible I'm not doing something right, but If I go back to the stand-alone test that got boiled down into the boolean_complex case, I did have that working in both v2.4.5 and current:

```
manifold/build (perf_test_v245) $ time ./extras/offsetTest ~/Generic_Twin_91.1.t0.glb
...
Face 264: 915 vertices, 1830 triangles
Face 265: 907 vertices, 1810 triangles
Processing 266 triangles... done.

real   0m3.488s
user  0m3.436s
sys	  0m0.052s
```

Putting the same offsetTest into the tolerance branch, pulling in the RotateUp fix and setting the tolerance in the input mesh:
```
MeshGL input;
input.tolerance = std::numeric_limits<float>::min();
input = ImportMesh(filename);
```
then running the same input file:

```
manifold/build (tolerance) $ time ./extras/offsetTest ~/Generic_Twin_91.1.t0.glb
...
Face 264: 13244 vertices, 26488 triangles
Face 265: 13251 vertices, 26498 triangles
Processing 266 triangles... done.

real	1m54.895s
user	1m54.496s
sys	0m0.397s

```

starseeker:
The v2.4.5 version:  https://github.com/elalish/manifold/compare/master...starseeker:manifold:perf_test_v245
The tolerance version:  https://github.com/starseeker/manifold/commit/60ea167d8bda2dc890c73956fbf22f717ce4541b

starseeker:
Ah, sorry, that sha1 for the tolerance is one behind - this is the file that compiles:  https://github.com/starseeker/manifold/blob/tolerance/extras/offset_test.cpp

starseeker:
A quick backport of the new boolean_complex test also shows similar timing differences:

https://github.com/elalish/manifold/compare/master...starseeker:manifold:test_backport

```
manifold/build (test_backport) $ time ./test/manifold_test --gtest_filter="*SimpleOffset*"
Note: Google Test filter = *SimpleOffset*
[==========] Running 1 test from 1 test suite.
[----------] Global test environment set-up.
[----------] 1 test from BooleanComplex
[ RUN      ] BooleanComplex.SimpleOffset
[       OK ] BooleanComplex.SimpleOffset (1417 ms)
[----------] 1 test from BooleanComplex (1417 ms total)

[----------] Global test environment tear-down
[==========] 1 test from 1 test suite ran. (1417 ms total)
[  PASSED  ] 1 test.

real	0m1.428s
user	0m1.410s
sys	0m0.017s
```

tolerance branch in starseeker/manifold:
```
manifold/build (tolerance) $ time ./test/manifold_test --gtest_filter="*SimpleOffset*"
Note: Google Test filter = *SimpleOffset*
[==========] Running 1 test from 1 test suite.
[----------] Global test environment set-up.
[----------] 1 test from BooleanComplex
[ RUN      ] BooleanComplex.SimpleOffset
[       OK ] BooleanComplex.SimpleOffset (51897 ms)
[----------] 1 test from BooleanComplex (51897 ms total)

[----------] Global test environment tear-down
[==========] 1 test from 1 test suite ran. (51897 ms total)
[  PASSED  ] 1 test.

real	0m51.967s
user	0m51.938s
sys	0m0.028s
```

starseeker:
tolerance branch in starseeker/manifold with tolerance settings per https://github.com/elalish/manifold/commit/74f2fe4ff832d59db8e0f515a05e66d567891da0:

```
manifold/build (tolerance) $ time ./test/manifold_test --gtest_filter="*SimpleOffset*"
Note: Google Test filter = *SimpleOffset*
[==========] Running 1 test from 1 test suite.
[----------] Global test environment set-up.
[----------] 1 test from BooleanComplex
[ RUN      ] BooleanComplex.SimpleOffset
[       OK ] BooleanComplex.SimpleOffset (51735 ms)
[----------] 1 test from BooleanComplex (51735 ms total)

[----------] Global test environment tear-down
[==========] 1 test from 1 test suite ran. (51735 ms total)
[  PASSED  ] 1 test.

real	0m51.810s
user	0m51.779s
sys	0m0.031s
```

starseeker:
@pca006132 Let me know if I should adjust anything else or try something different - I'll work with the SimpleOffset test since that's the most compact now that I have it building with v2.4.5 as a control.

elalish:
@starseeker Okay, I just pulled your `test_backport` branch and ran `./manifold_test --gtest_filter=BooleanComplex.SimpleOffset` - I'm getting 1.2 seconds. On master, the same command is giving me 3.3 seconds. So indeed, a speed regression, but only 2x, not 40x. Are you sure you have TBB enabled on our latest? There have been a lot of changes...

starseeker:
@elalish I typically build without tbb - we don't (yet) have it integrated properly as a dependency in BRL-CAD, so it's the serial performance that I see.  (Also, I'm a little leery of enabling tbb seeing the failure that cropped up in the SimpleOffset test...) 

So checking with vs. without TBB, here's what I'm seeing:
```
              No TBB  |   TBB
v2.4.5:      1414 ms  |  803 ms
tolerance:  51947 ms  | 3559 ms
```

elalish:
Okay, thanks. Can you change your tolerance to `1e-5` times your seed object's boundingBox.Scale()? That will be a better approximation of the precision from v2.4.5. Also, I think these two lines are in the wrong order to affect tolerance:
```
origin_cyl.SetTolerance(mtol);
origin_cyl = manifold::Manifold::Cylinder(len, 1, 1, 8);
```

starseeker:
OK, that definitely made a difference:

Without tbb:
```
manifold/build (tolerance) $ time ./test/manifold_test --gtest_filter="*SimpleOffset*"
Note: Google Test filter = *SimpleOffset*
[==========] Running 1 test from 1 test suite.
[----------] Global test environment set-up.
[----------] 1 test from BooleanComplex
[ RUN      ] BooleanComplex.SimpleOffset
bbox: 4646.88,4646.88,-20170.1,1483.28,5361.18,-18787.7,5448.96
mtol: 0.201701
[       OK ] BooleanComplex.SimpleOffset (20141 ms)
[----------] 1 test from BooleanComplex (20141 ms total)

[----------] Global test environment tear-down
[==========] 1 test from 1 test suite ran. (20141 ms total)
[  PASSED  ] 1 test.
```

With tbb:
```
manifold/build (tolerance) $ time ./test/manifold_test --gtest_filter="*SimpleOffset*"
Note: Google Test filter = *SimpleOffset*
[==========] Running 1 test from 1 test suite.
[----------] Global test environment set-up.
[----------] 1 test from BooleanComplex
[ RUN      ] BooleanComplex.SimpleOffset
bbox: 4646.88,4646.88,-20170.1,1483.28,5361.18,-18787.7,5448.96
mtol: 0.201701
[       OK ] BooleanComplex.SimpleOffset (1766 ms)
[----------] 1 test from BooleanComplex (1766 ms total)

[----------] Global test environment tear-down
[==========] 1 test from 1 test suite ran. (1767 ms total)
[  PASSED  ] 1 test.
```

starseeker:
Above timings per this commit:  https://github.com/starseeker/manifold/commit/ad1fe81ff4fc2134a01e9e474e8215558b937655

elalish:
Huh, so keeping the old tolerance was good for a 2x reduction, but that still leaves a 15x difference for serial and a 2x difference for parallel. Definitely something going on here...

elalish:
With the latest tolerance are you at least getting roughly the same number of faces now as in the original version?

starseeker:
1810 tris in the old vs 4708 in the new, looks like.

starseeker:
Correction - per the test_backport branch, I'm seeing 3004 triangles exported in the final output mesh.

elalish:
Okay, thanks - that definitely doesn't seem like enough to be responsible for most the slowdown. Anyway, this is all a long tangent from the question @pca006132 was asking regarding this PR. I believe this testing shows that indeed, this PR's change is good and working as intended, even though the problem it's fixing is not the whole regression identified by the SimpleOffset test.

elalish (approved):
This looks great, thanks!

pca006132:
In combination with #989, increasing the number of faces may cause a significant slowdown. But should do a detailed profiling to see the reason.

Will finish the remaining of this PR later when I have time, hopefully this week.

pca006132:
@starseeker and you can try to use BatchBoolean instead of doing += repeatedly. It should help a bit, even though the older version should also have this performance issue regarding +=...

starseeker:
I don't know if this is helpful, but here are Linux perf recordings of the test_backport and current master running the simple offset test case" ```perf record ./test/manifold_test --gtest_filter="*SimpleOffset*"``` (only saved 0.40% and up)

https://brlcad.org/~starseeker/perf_test_backport.txt
https://brlcad.org/~starseeker/perf_master.txt

starseeker:
Interestingly, SimplifyTopology in master wound up below the 0.40% cutoff:

 0.12%  manifold_test  libmanifold.so.2.5.1  [.] manifold::Manifold::Impl::SimplifyTopology()

starseeker:
Actually, the call graph version of master might be more helpful:

```
Samples: 185K of event 'cycles:P', Event count (approx.): 207025003327
  Children      Self  Command        Shared Object         Symbol
-   99.07%     0.00%  manifold_test  libmanifold.so.2.5.1  [.] manifold::CsgOpNode::BatchBoolean(manifold::OpType, st
   - manifold::CsgOpNode::BatchBoolean(manifold::OpType, std::vector<std::shared_ptr<manifold::Manifold::Impl const>,
      - 86.85% manifold::Boolean3::Result(manifold::OpType) const
         - 40.08% manifold::Manifold::Impl::SimplifyTopology()
            - 26.49% void manifold::for_each_n<manifold::CountingIterator<unsigned long>, manifold::Manifold::Impl::S
               - void manifold::for_each<manifold::CountingIterator<unsigned long>, manifold::Manifold::Impl::Simplif
                  - 26.43% manifold::Manifold::Impl::SimplifyTopology()::{lambda(unsigned long)#3} std::for_each<mani
                     - 26.31% manifold::Manifold::Impl::SimplifyTopology()::{lambda(unsigned long)#3}::operator()(uns
                        - 25.95% (anonymous namespace)::SwappableEdge::operator()(int) const
                           + 15.56% auto linalg::operator*<linalg::mat<double, 2, 3>, linalg::vec<double, 3> >(linalg
                           + 5.06% manifold::CCW(linalg::vec<double, 2>, linalg::vec<double, 2>, linalg::vec<double,
                           + 3.20% manifold::GetAxisAlignedProjection(linalg::vec<double, 3>)
                             0.63% (anonymous namespace)::TriOf(int)
            + 7.24% manifold::Manifold::Impl::CleanupTopology()
            + 2.74% void manifold::for_each_n<manifold::CountingIterator<unsigned long>, manifold::Manifold::Impl::Si
            + 2.42% void manifold::for_each_n<manifold::CountingIterator<unsigned long>, manifold::Manifold::Impl::Si
              0.70% manifold::Manifold::Impl::CollapseEdge(int, std::vector<int, std::allocator<int> >&)
         - 25.70% manifold::Manifold::Impl::Finish()
            - 10.38% manifold::Manifold::Impl::CalculateNormals()
               - 9.96% void manifold::for_each_n<manifold::CountingIterator<unsigned long>, (anonymous namespace)::As
                  - 9.96% void manifold::for_each<manifold::CountingIterator<unsigned long>, (anonymous namespace)::A
                     - 9.94% (anonymous namespace)::AssignNormals<false> std::for_each<manifold::CountingIterator<uns
                        - 9.41% (anonymous namespace)::AssignNormals<false>::operator()(int)
                           + 2.85% (anonymous namespace)::AtomicAddVec3(linalg::vec<double, 3>&, linalg::vec<double,
                           + 2.04% linalg::vec<double, 3> linalg::normalize<double, 3>(linalg::vec<double, 3> const&)
                           + 0.83% double linalg::dot<double, 3>(linalg::vec<double, 3> const&, linalg::vec<double, 3
                           + 0.77% auto linalg::operator*<double, linalg::vec<double, 3> >(double const&, linalg::vec
                           + 0.71% linalg::detail::apply<linalg::detail::op_sub, void, linalg::vec<double, 3>, linalg
            + 5.87% manifold::Manifold::Impl::GetFaceBoxMorton(manifold::Vec<manifold::Box>&, manifold::Vec<unsigned
            + 4.36% manifold::Collider::Collider(manifold::VecView<manifold::Box const> const&, manifold::VecView<uns
            + 2.66% manifold::Manifold::Impl::SortFaces(manifold::Vec<manifold::Box>&, manifold::Vec<unsigned int>&)
            + 2.02% manifold::Manifold::Impl::SortVerts()
         - 17.16% manifold::Manifold::Impl::Face2Tri(manifold::Vec<int> const&, manifold::Vec<manifold::TriRef> const
            - 10.23% manifold::Manifold::Impl::CreateHalfedges(manifold::Vec<linalg::vec<int, 3> > const&)
               - 7.42% void manifold::stable_sort<int*, int, manifold::Manifold::Impl::CreateHalfedges(manifold::Vec<
                    void manifold::stable_sort<int*, int, manifold::Manifold::Impl::CreateHalfedges(manifold::Vec<lin
                    void std::stable_sort<int*, manifold::Manifold::Impl::CreateHalfedges(manifold::Vec<linalg::vec<i
                    void std::__stable_sort<int*, __gnu_cxx::__ops::_Iter_comp_iter<manifold::Manifold::Impl::CreateH
                  - void std::__stable_sort_adaptive<int*, int*, __gnu_cxx::__ops::_Iter_comp_iter<manifold::Manifold
                     - 7.01% void std::__merge_sort_with_buffer<int*, int*, __gnu_cxx::__ops::_Iter_comp_iter<manifol
                        - 5.49% void std::__merge_sort_loop<int*, int*, long, __gnu_cxx::__ops::_Iter_comp_iter<manif
                           + 5.11% int* std::__move_merge<int*, int*, __gnu_cxx::__ops::_Iter_comp_iter<manifold::Man
                        + 1.51% void std::__chunk_insertion_sort<int*, long, __gnu_cxx::__ops::_Iter_comp_iter<manifo
               + 1.18% void manifold::for_each_n<manifold::CountingIterator<unsigned long>, manifold::Manifold::Impl:
                 0.59% manifold::Manifold::Impl::RemoveUnreferencedVerts()
            + 6.79% void std::_Bind<manifold::Manifold::Impl::Face2Tri(manifold::Vec<int> const&, manifold::Vec<manif
         + 1.21% (anonymous namespace)::AppendWholeEdges(manifold::Manifold::Impl&, manifold::Vec<int>&, manifold::Ve
           1.17% (anonymous namespace)::SizeOutput(manifold::Manifold::Impl&, manifold::Manifold::Impl const&, manifo
      - 12.21% manifold::Boolean3::Boolean3(manifold::Manifold::Impl const&, manifold::Manifold::Impl const&, manifol
         + 3.77% manifold::Manifold::Impl::EdgeCollisions(manifold::Manifold::Impl const&, bool) const
         + 3.26% (anonymous namespace)::Shadow11(manifold::SparseIndices&, manifold::Manifold::Impl const&, manifold:
         + 2.16% (anonymous namespace)::Intersect12(manifold::Manifold::Impl const&, manifold::Manifold::Impl const&,
         + 1.67% (anonymous namespace)::Filter11(manifold::Manifold::Impl const&, manifold::Manifold::Impl const&, ma
         + 0.66% (anonymous namespace)::Shadow02(manifold::Manifold::Impl const&, manifold::Manifold::Impl const&, ma

```

starseeker:
Urk.  Wait a minute.  If I explicitly specify CMAKE_BUILD_TYPE=Release for configure, I'm getting an execution time in master of 5143 ms.  Maybe it's as simple as the v2.4.5 build getting some build flags by default without having to specify the CMake build type.

pca006132:
Hmm, interesting. But we do have default build type as configured in https://github.com/elalish/manifold/blob/e647093bbea6698a04b62da147a85102bb1f11fa/CMakeLists.txt#L115-L125

Also, I guess with this PR, the problem is essentially solved now? If you use the correct build type.

starseeker:
Bingo.  Old build has "-O3" in the compile lines even when I don't specify a build type, and the master build does not.  Looks like I somehow managed to drop those in the CMake refactor.  Good grief.  Pull request imminent.

pca006132:
To be fair, I think it is fine to drop the `-O3` flags, or at least only do it when we do release/relwithdebinfo build? It feels weird to do that with debug build.

starseeker:
Well... Hmm.  That's a point - the perf reports do actually illustrate why you might not *always* want the O3 flag on.  @elalish any thoughts on that one?

pca006132:
Ah, and another issue is that because you are calling `c.Status()` for every iteration, this will prohibit some of our CSG reordering, so things will be slower (but you also avoids #989).

starseeker:
@pca006132 Eventually I'd like to fully stomp on the gas - enable tbb, remove all the Status() and exports mid-execution, etc. - but that triangulation failure that only seems to manifest with tbb enabled in particular has me a little gun-shy.  Slower and reliable wins over faster and occasionally failing (well, within reason) in my particular use case.

pca006132:
I see. I would suggest you to try the fast way and fallback to the slower way in case something failed, because the timing difference is huge. I managed to cut the time down from 5450ms to 220ms just by changing these.

We will add some debug info in csg tree to report back the offending boolean pair when something bad happens, so it should be easier to debug later.

pca006132:
Anyway, probably too much discussion about things tangential to this PR. We should continue the discussion in #993 if needed. Wish github can just allow me to select and move some of the discussions to another issue...

elalish:
Good teamwork on the debugging - I'm glad this is solved! @pca006132 should we just go ahead and merge this and do the docs & tests in a follow-on PR?

pca006132:
> For the test naming, I think I will leave it here as we should fix a bunch of other documentations and naming as well. Should we change `RefineToPrecision` to `RefineToTolerance`, for example?

@elalish so what do you think about the naming?

elalish:
Yeah, I agree. I'm working on PRs now to update these names and to remove the need for that epsilon setting in `RelatedGL`.

starseeker:
Just to check - three of the original checkboxes in the initial PR ended up unchecked - are they still needed?

pca006132:
Yes, we will do it later

Issue #970, opened 2024-10-01
Potential boolean bug


pca006132:
See https://github.com/elalish/manifold/pull/666#issuecomment-2356564020

Just to remind us not to forget about this after finishing v3.0 release.

elalish:
Yeah, I feel like we could use an approximate `IsOverlapping` function  that we can run in debug mode after each Boolean so we can figure out where the problem first occurs, and dump out a test case of those meshes.

pca006132:
For the record: We now check for self-intersection in https://github.com/elalish/manifold/blob/3a0c251b0e785022fcef36ac008143612e238680/src/properties.cpp#L213 and has a simple fuzzer in https://github.com/elalish/manifold/blob/master/test/manifold_fuzz.cpp that can fuzz this quickly.

pca006132:
It seems that this may be related to the mesh simplification pass. For

https://github.com/elalish/manifold/blob/ac827679272a396972e92933c79602ba0aa496b9/test/boolean_complex_test.cpp#L1040-L1057

The result is not self-intersecting (doesn't even require epsilon-valid!) if we disable the mesh simplification pass, the part that flags redundant edges for collapsing. I think this suggests that the redundant edge removal may produce self-intersection.

However, disabling mesh simplification doesn't fix all the test cases. For example, the following case is still failing:

https://github.com/elalish/manifold/blob/ac827679272a396972e92933c79602ba0aa496b9/test/boolean_test.cpp#L425-L435

Probably two issues at play here.

---

Also, about the edge flagging code, why don't we do the following, which updates `ref1` to make sure `ref1` never points to the same triangle as `ref0`? The old code will not flag more edges, but it may flag less.

```c++
  bool operator()(int edge) const {
    if (halfedge[edge].pairedHalfedge < 0) return false;
    // Flag redundant edges - those where the startVert is surrounded by only
    // two original triangles.
    const TriRef ref0 = triRef[edge / 3];
    int current = NextHalfedge(halfedge[edge].pairedHalfedge);
    TriRef ref1 = triRef[current / 3];
    bool ref1Updated = !ref0.SameFace(ref1);
    while (current != edge) {
      current = NextHalfedge(halfedge[current].pairedHalfedge);
      int tri = current / 3;
      const TriRef ref = triRef[tri];
      if (!ref.SameFace(ref0) && !ref.SameFace(ref1)) {
        if (ref1Updated) {
          ref1 = ref;
          ref1Updated = true;
        } else {
[...]

PR #998, opened 2024-10-18
Define a test for simple offset booleans

starseeker:
This is intended to check for performance regressions - the BRL-CAD logic performing this operation using v2.4.5 ran considerably faster.

Note - removing in-loop assertions for triangle count > 0 with the intermediate forms greatly increases the speed, but they are placed there deliberately to simulate what BRL-CAD does to validate the output of the booleans on an incremental basis.  In the event of a boolean failure or invalid output being produced for any reason, we want to be able to capture the *exact* inputs responsible - which means we need to check after each boolean operation to validate the results and catch any bad intermediate states at the time they are produced.


starseeker:
@elalish I haven't back-ported to v2.4.5 yet to confirm it runs faster there in this form, but that should be the general idea.

test/boolean_complex_test.cpp
```diff
@@ -1030,4 +1031,160 @@ TEST(BooleanComplex, HullMask) {
...
+    int e11, e12, e21, e22, e31, e32;
+    e11 = (eind[0] < eind[1]) ? eind[0] : eind[1];
+    e12 = (eind[1] < eind[0]) ? eind[0] : eind[1];
+    e21 = (eind[1] < eind[2]) ? eind[1] : eind[2];
+    e22 = (eind[2] < eind[1]) ? eind[1] : eind[2];
+    e31 = (eind[2] < eind[0]) ? eind[2] : eind[0];
+    e32 = (eind[0] < eind[2]) ? eind[2] : eind[0];
+    edges.insert(std::make_pair(e11, e12));
+    edges.insert(std::make_pair(e21, e22));
+    edges.insert(std::make_pair(e31, e32));
```
elalish, line 1054: Just a nit, but this is way more complicated than it needs to be. Have you considered something like:
```
std::vector<std::pair<int, int>> edges;
for (size_t i = 0; i < seeds.NumTri(); i++) {
  int k[3]={1,2,0};
  for(const int j:{0,1,2}){
    int v1=seeds.triVerts[i*3+j];
    int v2=seeds.triVerts[i*3+k[j]];
    if(v2>v1) edges.push_back(std::make_pair(v1, v2));
  }
}
```

test/boolean_complex_test.cpp
```diff
@@ -1030,4 +1031,159 @@ TEST(BooleanComplex, HullMask) {
...
+    vec3 evec(-1 * edge.x, -1 * edge.y, edge.z);
+    manifold::Manifold rotated_cyl =
+        origin_cyl.Transform(manifold::RotateUp(evec));
+    manifold::Manifold right = rotated_cyl.Translate(ev1);
+    if (!right.NumTri()) continue;
+    manifold::Manifold left = c;
+    c = left.Boolean(right, manifold::OpType::Add);
+    EXPECT_TRUE(c.NumTri() > 0);
+  }
+  // Triangle Volumes
```
elalish, line 1093: Have you tried `Manifold::Extrude`?

test/boolean_complex_test.cpp
```diff
@@ -1030,4 +1031,159 @@ TEST(BooleanComplex, HullMask) {
...
+    // Extrude the points above and below the plane of the triangle
+    vec3 pnts[6];
+    pnts[0] = ev1 + n;
+    pnts[1] = ev2 + n;
+    pnts[2] = ev3 + n;
+    pnts[3] = ev1 - n;
+    pnts[4] = ev2 - n;
+    pnts[5] = ev3 - n;
+    // Construct the points and faces of the new manifold
+    double pts[3 * 6];
```
elalish, line 1118: Mind converting this to an initializer list so it's not quite so verbose? e.g.
```
double pts[3 * 6]={pnts[4].x, pnts[4].y, pnts[4].z,\\
                              ...
                             };
```

test/boolean_complex_test.cpp
```diff
@@ -1030,4 +1031,159 @@ TEST(BooleanComplex, HullMask) {
...
+    manifold::MeshGL64 tri_m;
+    for (int j = 0; j < 18; j++)
+      tri_m.vertProperties.insert(tri_m.vertProperties.end(), pts[j]);
+    for (int j = 0; j < 24; j++)
+      tri_m.triVerts.insert(tri_m.triVerts.end(), faces[j]);
+    manifold::Manifold left = c;
+    manifold::Manifold right(tri_m);
+    if (!right.NumTri()) continue;
+    c = left.Boolean(right, manifold::OpType::Add);
+    EXPECT_TRUE(c.NumTri() > 0);
```
elalish, line 1185: If this is affecting the timing, let's move it out of the loop so that our CSG tree can optimize its operations. No need for this test to pinpoint the location of the problem - it just needs to tell if anything went wrong. Also, a `Status` check is probably a more canonical way to handle this.

elalish (changes requested):
Thanks, looks like it takes about 10 seconds to run on the CI currently. We should definitely be able to get that back down.

test/boolean_complex_test.cpp
starseeker: Oh,  to create the triangle based shape?  I hadn't yet, actually - I can give that a go if you think it'll be simpler.

elalish: Up to you - let's just try to shrink the code and make it readable as much as we can.

test/boolean_complex_test.cpp
starseeker: Ah - I hadn't boiled it down that far.  Updated.

test/boolean_complex_test.cpp
starseeker: Hmm.  Without the NumTri() calls, things complete very quickly - it might not be slow enough to usefully flag the problem?  (Assuming the problem is the larger triangle counts due to the tolerance issues...)

It's more than possible that my BRL-CAD side implementation of all this is sub-optimal (i.e. doing things I shouldn't during the iterations on the boolean processing), so I suppose another thing for me to try is to "back port" how this test is doing things into our code and see if that fixes the speed issues.

starseeker: OK, looks like Status() is slow as well (at least here) so it's not NumTri() specifically.

elalish: yeah, @pca006132 has done some pretty awesome optimizations of our CSG tree that you certainly don't want to skip just for a debug check. Would you mind moving that check out of the loop so I can see the CI's timing? It'd be great if that's the whole problem, but I'd bet it's not.

elalish: Yeah, every check (num tri, status, genus, etc) requires the manifold to be evaluated -it's the forced evaluation before we can reorder them that causes the slowness.

test/boolean_complex_test.cpp
starseeker: So looking at the same example you pointed me to for the Cylinder logic, it looks like you made a 2D simple polygon triangle, extruded it, and then used Warp and a matrix of the triangle positions to make copies of the extruded triangle at the appropriate locations?

starseeker:
Uh... that failure is unexpected.  A triangulation failure?? Did I break the logic somehow?

elalish:
> Uh... that failure is unexpected. A triangulation failure?? Did I break the logic somehow?

Actually, this is why my Offset PR never got merged. I'm guessing I hit a case where the symbolic perturbation did something unexpected, created some extremely degenerate polygons that compounded enough until the triangulor gave up. However, I never found the time to properly debug it. Interesting that you only saw this error once you adopted my technique - perhaps my technique is flawed. It also means you might just have an algorithm to contribute for an `Offset` method, which would be awesome! 

Always fun when one bug leads to another - they're all important, so please follow up on any/all that interest you. But let's try to keep them in separate PRs.

elalish:
Hmm, actually I take it back, this didn't start when you switched to `Cylinder`, it started when you allowed our CSG tree to reorder. That's *very* interesting. @pca006132 could this be a bug in the CSG optimization somehow? It would be amazing if we could cut this down to find exactly the Boolean op that's failing, but for that I think we'll have to put some debug code into the CSG tree itself.

test/boolean_complex_test.cpp
starseeker: Updated.  I'm not quite following the Warp + Extrude logic yet, so my initial attempt there didn't work - I'll have to look at it next week unless this makes things readable enough.

elalish: Thanks, that's loads better - you're welcome to leave it. I guess go ahead and put your status checks back the way they were so we at least have something to test. Maybe leave these ones commented out with some info on how they caused the triangulator problem.

test/boolean_complex_test.cpp
```diff
@@ -1030,4 +1031,102 @@ TEST(BooleanComplex, HullMask) {
...
+  // Vertex Spheres
+  Manifold sph = Manifold::Sphere(1, 8);
+  for (size_t i = 0; i < seeds.NumVert(); i++) {
+    vec3 vpos(seeds.vertProperties[3 * i + 0], seeds.vertProperties[3 * i + 1],
+              seeds.vertProperties[3 * i + 2]);
+    Manifold vsph = sph.Translate(vpos);
+    manifold::Manifold left = c;
+    manifold::Manifold right(vsph);
+    if (!right.NumTri()) continue;
+    c = left.Boolean(right, manifold::OpType::Add);
```
elalish, line 1060: is there a reason you don't use `c += right;`?

starseeker: That is partially left over from my initial wiring in BRL-CAD and partially my avoiding confusion in our code.  We use + for intersection.   No harm here though,  since the context is Manifold - I'll switch it once I get back to my computer.

starseeker:
Um.  Again not what I was expecting...

```
Triangulation failed! Precision = 2.303e-05
Error in file: /home/runner/work/manifold/manifold/src/polygon.cpp (118): 'std::all_of(triangles.begin(), triangles.end(), [&vertPos, precision](const ivec3 &tri) { return CCW(vertPos[tri[0]], vertPos[tri[1]], vertPos[tri[2]], precision) >= 0; })' is false: triangulation is not entirely CCW!
Polygon 0 2.303e-05 1
8
0.966774 -23812.8
0.966774 0.0240642
-2653.02 -9970.37
-2653.02 -21874.9
0.966774 -23812.8
0.966774 0.0231964
-5230.69 -19910.1
-5230.67 -19992.8
# ... 
show(array([
  [0.966774, -23812.8],
  [0.966774, 0.0240642],
  [-2653.02, -9970.37],
  [-2653.02, -21874.9],
  [0.966774, -23812.8],
  [0.966774, 0.0231964],
  [-5230.69, -19910.1],
  [-5230.67, -19992.8],
]))
produced this triangulation:
17, 26, 12329
26, 12658, 12329
26, 12235, 12658
12235, 12234, 12658
12235, 12330, 12234
12330, Triangulation failed! Precision = 2.303e-05
Error in file: /home/runner/work/manifold/manifold/src/polygon.cpp (118): 'std::all_of(triangles.begin(), triangles.end(), [&vertPos, precision](const ivec3 &tri) { return CCW(vertPos[tri[0]], vertPos[tri[1]], vertPos[tri[2]], precision) >= 0; })' is false: triangulation is not entirely CCW!
Polygon 0 2.303e-05 1
8
-0.966774 -23812.8
5266.72 -20170.1
-0.954981 -0.0210915
-0.966774 0.0231964
-0.966774 -23812.8
-0.957262 -23812.8
-0.957262 18-0.0123562
-0.966774 0.0240642
# ... 
, 12234show(array([
  [-0.966774, -23812.8],
  [
5266.72, -20170.1],
  [-0.954981, -0.0210915],
  [-0.966774, 0.0231964],
  [-0.966774, -23812.8],
  [-0.957262, -23812.8],
  [-0.957262, -0.0123562],
  [-0.966774, 0.0240642],
]))
produced this triangulation:
17, 10818, 26
10818, 12308, 26
10818, 12307, 12308
12307, 12233, 12308
12307, 12234, 12233
12234, 18, 12233
unknown file: Failure
C++ exception with description "Error in file: /home/runner/work/manifold/manifold/src/polygon.cpp (118): 'std::all_of(triangles.begin(), triangles.end(), [&vertPos, precision](const ivec3 &tri) { return CCW(vertPos[tri[0]], vertPos[tri[1]], vertPos[tri[2]], precision) >= 0; })' is false: triangulation is not entirely CCW!" thrown in the test body.
```

starseeker:
New data point - the triangulation failure only triggers with tbb enabled.  I was wondering why I wasn't seeing it locally - it's because I normally build without tbb.  When I do enable it here, I get the failure.

elalish:
Thanks for all the debug - we're getting closer. Since this test is passing, I suppose we might as well merge it. That should make it easier to test various ideas for fixes, not to mention we can modify it to repro the other bug you found.

elalish (approved):
Thanks for the perf test!

starseeker:
I don't know if this matters or helps, but I was experimenting with doing a number of iterations on the triangle potion of the unioning before calling Status().  At least on my machine here,  if I run 15 iterations before calling Status(), I get intermittent success and failure - things aren't reproducible/deterministic from run to run.  If I do more than 15 iterations it seems to reliably fail, and less than 15 I'm not seeing a failure.

starseeker:
Sorry, I need to correct my terms - it's not iterations, it's number of faces processed.  So unioning in 14 triangle-generated solids and then calling Status() seems to work, trying to do 16 fails, and 15 sometimes works and sometimes doesn't.  I'm not iterating over all the triangles multiple times trying to union them in again and again (that would just be evil.)

elalish:
Can you give me any background on what this model is supposed to be? I'm curious why it's 3,000 meters tall and offset by 1 meter? And is this cutout in the top intentional?
image

starseeker:
It's a piece of the FAA's GenericTwin plane model.  They model thin surfaces with just a surface mesh (hence why we need the offset for boolean ops) and that just happened to be the piece I grabbed for testing - there's lots more like it.  The ones that look more like "airplane" are considerably larger, and this one was one of the smaller ones - just picked at random.

As far as I can tel, the cutout is intentional.

pca006132:
> Hmm, actually I take it back, this didn't start when you switched to `Cylinder`, it started when you allowed our CSG tree to reorder. That's _very_ interesting. @pca006132 could this be a bug in the CSG optimization somehow? It would be amazing if we could cut this down to find exactly the Boolean op that's failing, but for that I think we'll have to put some debug code into the CSG tree itself.

Not sure, csg reordering can give different results, and if things were close to triggering an issue, csg reordering can push it to actually trigger the issue I guess.

But yeah, some mechanism to dump the offending boolean operation will be nice to catch this kind of error. I actually did that for #970, forgot why I did not make a PR for it.

PR #1000, opened 2024-10-19
removed RotateUp and uint

### LINKED ISSUE #996: Odd failure creating and unioning multiple cylinders
In the process of simplifying down my test code for the performance regression (https://github.com/starseeker/manifold/blob/perf_test/extras/offset_test.cpp), I've hit an odd failure that I'm having trouble reproducing in isolation.  I'm running fine, and then I suddenly hit this error (enabling MANIFOLD_DEBUG doesn't seem to change anything):

```
Edge 333: 13254 vertices, 27144 triangles
Edges - manifold boolean op failure: Vec out of range
```

I've tried to dump the glb meshes producing this output, but something is wrong:  I can generate them, but for the Manifold generated by Cylinder all I'm getting is a mesh that contains 1 vertex:

```
./extras/convertFile cyl.glb 
1 vertices, 28 triangles
0 objects:
```

However, if I interrogate the cyl Manifold itself in the C++ code with methods like NumVert() and NumTri(), both the Manifold itself and a GetMeshGL version, there appears to be more data there:

Short Edge Cylinder counts (Manifold, pre bool): 16 vertices, 28 triangles
Short Edge Cylinder counts (GetMeshGL, pre bool) Manifold: 16 vertices, 28 triangles
Vec out of range
Short Edge Cylinder counts (GetMeshGL, post bool): 16 vertices, 28 triangles

Compared to the other Cylinders being generated this one is short (a little over 0.025) but if I try to create an isolated test with just the Cylinder call and a glb read of the other input mesh the test passes.  If I skip that one short cylinder the rest of the test runs to completion.

Am I using Cylinder incorrectly, or somehow mismanaging memory in such a way that I could be causing corruption?  I'm at a bit of a loss, particularly since things seem to complete without that individual cylinder but trying to isolate it for a test case isn't going very well...

### LINKED ISSUE #999: meshIO.cpp: uint not standard C++
Commit e067653 removed glm:: at line 58 of meshIO.cpp, left the uint type declaration.  This is not standard C++.

Edit: There are two occurrences in this file, one at line 58 and the other at 254.

elalish:
Fixes #999 
Fixes #996 

@starseeker I don't have exactly your repro, but I believe this should fix the problem you reported. Would you mind verifying that?


test/boolean_complex_test.cpp
```diff
@@ -1084,9 +1083,8 @@ TEST(BooleanComplex, SimpleOffset) {
     if (len < 0.03) continue;
     manifold::Manifold origin_cyl = manifold::Manifold::Cylinder(len, 1, 1, 8);
     vec3 evec(-1 * edge.x, -1 * edge.y, edge.z);
-    manifold::Manifold rotated_cyl =
-        origin_cyl.Transform(manifold::RotateUp(evec));
-    manifold::Manifold right = rotated_cyl.Translate(ev1);
+    quat q = rotation_quat(normalize(evec), vec3(0, 0, 1));
```
elalish, line 1086: @pca006132 I was a little suprised I didn't need `la::rotation_quat` or `manifold::la::rotation_quat`. I recall you made a comment about my namespacing - did I do it wrong?

elalish:
On second thought, considering the very non-obvious behavior from #996, perhaps we should add assertions on all our transform inputs that they are finite.

test/boolean_complex_test.cpp
pca006132: I think this is a weird thing called [Argument-dependent lookup](https://en.cppreference.com/w/cpp/language/adl) for functions, which basically adds the argument type namespaces into the lookup search path. I don't think we have anything wrong with our namespacing.

starseeker:
I'll try it out when I get a chance, but the first thing to do is remove the "if (len < 0.03) continue;" test with the TODO comment from the offset test, if that's not already removed - that was hitting the issue for me.

test/boolean_complex_test.cpp
elalish: Oh, that's cool! Thanks for the explanation.

starseeker:
Confirmed - removing the len < 0.03 filter still results in successful running of the offset test.

starseeker:
@elalish did you want to remove that workaround as part of this PR, or as a separate one?

elalish:
Yeah, I'll deal with it here.

src/boolean_result.cpp
```diff
@@ -669,23 +669,23 @@ Manifold::Impl Boolean3::Result(OpType op) const {
...
 
+  if (inP_.status_ != Manifold::Error::NoError) {
+    auto impl = Manifold::Impl();
+    impl.status_ = inP_.status_;
+    return impl;
+  }
+  if (inQ_.status_ != Manifold::Error::NoError) {
+    auto impl = Manifold::Impl();
+    impl.status_ = inQ_.status_;
+    return impl;
```
elalish, line 680: I changed this to pass through the input error instead of changing it, which I think makes it a be easier to deduce what went wrong first.

include/manifold/manifold.h
```diff
@@ -323,4 +323,44 @@ class Manifold {
...
+    case Manifold::Error::MergeIndexOutOfBounds:
+      return stream << "Merge Index Out Of Bounds";
+    case Manifold::Error::TransformWrongLength:
+      return stream << "Transform Wrong Length";
+    case Manifold::Error::RunIndexWrongLength:
+      return stream << "Run Index Wrong Length";
+    case Manifold::Error::FaceIDWrongLength:
+      return stream << "Face ID Wrong Length";
+    case Manifold::Error::InvalidConstruction:
+      return stream << "Invalid Construction";
```
elalish, line 361: I got tired of not being able to see what the error was. I really wish C++ had a way to turn enums into strings automatically, but apparently we need this lovely boiler plate instead.

src/impl.cpp
```diff
@@ -513,6 +513,10 @@ Manifold::Impl Manifold::Impl::Transform(const mat3x4& transform_) const {
     result.status_ = status_;
     return result;
   }
+  if (!all(isfinite(transform_))) {
+    result.MarkFailure(Error::NonFiniteVertex);
```
elalish, line 517: Added this to catch any kind of bad transform.

src/properties.cpp
```diff
@@ -126,16 +126,12 @@ struct UpdateProperties {
...
   VecView<const Halfedge> halfedges;
-  VecView<const vec3> vertPos;
 
   bool operator()(size_t edge) const {
     const Halfedge halfedge = halfedges[edge];
     if (halfedge.startVert == -1 || halfedge.endVert == -1) return true;
     if (halfedge.pairedHalfedge == -1) return false;
 
-    if (!std::isfinite(vertPos[halfedge.startVert][0])) return false;
-    if (!std::isfinite(vertPos[halfedge.endVert][0])) return false;
```
elalish, line 137: Not sure when this showed up, but we have a dedicated `IsFinite` method which was being elided by this throwing a NonManifold error, which was pretty unclear.

elalish:
@pca006132 I think this one is ready to merge, but I have no idea why the Windows compiler is suddenly pissed - doesn't look like the spew has anything to do with these changes. Thoughts?

pca006132:
From the error message:

```
D:\a\manifold\manifold\src\impl.cpp(516,6): error C2668: 'linalg::isfinite': ambiguous call to overloaded function [D:\a\manifold\manifold\build\src\manifold.vcxproj]
D:\a\manifold\manifold\include\manifold/linalg.h(1298,44): message : could be 'linalg::mat<bool,3,4> linalg::isfinite<manifold::mat3x4>(const A &)' [found using argument-dependent lookup] [D:\a\manifold\manifold\build\src\manifold.vcxproj]
          with
          [
              A=manifold::mat3x4
          ]
C:\Program Files (x86)\Windows Kits\10\Include\10.0.17763.0\ucrt\corecrt_math.h(401,32): message : or       'bool isfinite<manifold::mat3x4>(_Ty) noexcept' [D:\a\manifold\manifold\build\src\manifold.vcxproj]
          with
          [
              _Ty=manifold::mat3x4
          ]
```

Apparently it is some weird header in MSVC (TM) that tries to match for every `isfinite(T)` and causes lookup issue...

include/manifold/manifold.h
pca006132: I feel that a more general approach would be to implement this as a `toString` function.

src/impl.cpp
pca006132: Apparently we need to make it explicit, i.e. `all(la::isfinite(transform_))` to make msvc happy.

starseeker: fwiw, in BRL-CAD when we use C++ we generally are explicit with the namespaces to avoid lookup surprises.  It's definitely a readability tradeoff though.

PR #1002, opened 2024-10-19
Support googletest with the CMake package name GTest

cjmayo:
This check fails for me, (and I guess for the alternative Ubuntu `libgtest-dev` package:
https://packages.ubuntu.com/noble/amd64/libgtest-dev/filelist
)

According to the CMake docs the lower-case name will get checked too. Let's see...


elalish (approved):
Thanks!

PR #1007, opened 2024-10-22
update tolerance defaults

elalish:
Now input `MeshGL`s get a default tolerance of 1e-6 * scale, reflecting the uncertainty of their coordinates, and more closely aligning with the amount of default decimation we did before the update to double precision. The tolerance is likewise reduced for the floating point output of `MeshGL`.


elalish:
I believe this now also covers the TODO list from #997. However, in the process of adding the test, I found a serious problem with our simplification:
image
Basically lots of edges are being marked short and then being consecutively collapsed until far too large a change has been made. I'm going to have to think about how best to fix that.

src/edge_op.cpp
```diff
@@ -206,7 +206,7 @@ void Manifold::Impl::SimplifyTopology() {
   {
     ZoneScopedN("CollapseShortEdge");
     numFlagged = 0;
-    ShortEdge se{halfedge_, vertPos_, tolerance_};
+    ShortEdge se{halfedge_, vertPos_, epsilon_};
```
elalish, line 209: This is my quick fix - seems to be working okay. I need to come back and think about this more deeply sometime, but I think this will be okay for v3.0 - it's pretty close to what we were doing before.

test/properties_test.cpp
```diff
@@ -44,23 +44,37 @@ TEST(Properties, Precision) {
...
+  Manifold cylinder2 = cylinder.SetTolerance(0.01);
   MeshGL mesh = cylinder.GetMeshGL();
-  mesh.faceID.clear();
-  Manifold cylinder2 = Manifold(mesh).SetEpsilon(0.001);
+  mesh.tolerance = 0.01;
+  Manifold cylinder3(mesh);
+
+  EXPECT_EQ(cylinder.NumTri(), 3996);
+  EXPECT_EQ(cylinder2.NumTri(), 3996);
+  EXPECT_EQ(cylinder3.NumTri(), 3996);
```
elalish, line 64: This is kind of a stupid test right now - the high res cylinder bends too much, so `CreateFaces` gives up and doesn't allow any simplification at all. This was true before too - our `SimplifyTopology` is not a real decimator; it's considerably simpler. However, we should probably come back someday and turn it into a proper decimator.

Meanwhile, I'll update this test to something more useful before we merge this PR.

src/manifold.cpp
```diff
@@ -71,8 +71,12 @@ MeshGLP<Precision, I> GetMeshGLImpl(const manifold::Manifold::Impl& impl,
...
   MeshGLP<Precision, I> out;
   out.numProp = 3 + numProp;
-  out.tolerance = std::max(std::numeric_limits<Precision>::epsilon(),
-                           static_cast<Precision>(impl.tolerance_));
+  out.tolerance = impl.tolerance_;
+  if (std::is_same<Precision, float>::value)
+    out.tolerance =
+        std::max(out.tolerance,
+                 static_cast<Precision>(std::numeric_limits<float>::epsilon() *
+                                        impl.bBox_.Scale()));
```
pca006132, line 79: Should we perform simplification after this?

src/manifold.cpp
```diff
@@ -382,17 +386,10 @@ double Manifold::GetEpsilon() const {
...
-  impl->SetEpsilon(epsilon);
-  if (impl->tolerance_ > oldTolerance) impl->SimplifyTopology();
-  return Manifold(impl);
-}
-
 /**
- * Returns the tolerance of this Manifold's vertices.
- * Edges shorter than this tolerance value will be collapsed.
+ * Returns the tolerance value of this Manifold. Triangles that are coplanar
+ * within tolerance tend to be merged and edges shorter than tolerance tend to
```
pca006132, line 391: I'm thinking if this will be true later when we implement better simplification. Not sure how to express this properly, but the final result should be between dilation and erosion by epsilon?

src/edge_op.cpp
pca006132: Yeah, I think we just need to get the APIs correct for v3.0, detailed implementation can be done a bit later.

pca006132 (approved):
Nice catch! I am sure there should be papers on fast mesh decimation, maybe we should look into that later?

starseeker:
> Nice catch! I am sure there should be papers on fast mesh decimation, maybe we should look into that later?

I doubt the code would be a good fit for Manifold as-is, but Alexis Naveros has done quite a bit of work on a liberally licensed mesh decimation implementation that might be a useful resource:  https://github.com/BRL-CAD/mmesh/blob/main/src/meshdecimation.c  (He tends to heavily optimize his codes for high performance, so there's a lot of logic in mmesh pertaining to that sort of optimization in addition to the decimation algorithms themselves.) 

We currently using an older version of that code as one of the BRL-CAD options for decimation (still need to update to that newer mmesh version - ah, the never ending TODO list...) and it does pretty well in my experience at decimating without "ruining" the mesh shape.  I tried wiring up the OpenMesh decimation as an alternative and had problems with the output from it not being manifold, so that's at least one data point.  It might be simpler overall to grab a paper and start fresh, but I figured I might as well raise awareness just in case.

elalish:
Well, this is irritating - finally got some time to work on this again and without any changes on my side, suddenly my Mac is giving me a pile of `fatal error: 'cstdint' file not found` - apparently it suddenly can't find its own standard library? WTF? 

Since I couldn't think of anything else to do, I went ahead and upgraded OS 14 to OS 15 and did a big `brew upgrade` of all my dependencies and upped my compiler from clang 15 to clang 16. Still no luck - exactly the same errors. Has anyone seen this on a mac lately? I'm a bit at a loss as to how to tell it how to find its own STL.

pca006132:
Did you try a clean build? This sounds really weird...

elalish:
No, plenty of clean builds have had no effect. This is what it looks like - does anyone see anything weird about the command? It fails the same way on any file (fatal for whatever STL lib is included):
```
[build] [ 16%] Building CXX object bindings/python/CMakeFiles/nanobind-static.dir/__/__/_deps/nanobind-src/src/implicit.cpp.o
[build] cd /Users/elalish/Code/manifold/build/bindings/python && /usr/bin/clang++ -DNB_COMPACT_ASSERTIONS -I/Users/elalish/Code/manifold/build/_deps/nanobind-src/ext/robin_map/include -I/opt/homebrew/opt/python@3.13/Frameworks/Python.framework/Versions/3.13/include/python3.13 -I/Users/elalish/Code/manifold/build/_deps/nanobind-src/include -isystem /Library/Developer/CommandLineTools/usr/include/c++/v1 -isystem /Library/Developer/CommandLineTools/usr/lib/clang/16/include -isystem /Library/Developer/CommandLineTools/SDKs/MacOSX15.0.sdk/usr/include -isystem /Library/Developer/CommandLineTools/usr/include -O3 -DNDEBUG -std=c++17 -arch arm64 -isysroot /Library/Developer/CommandLineTools/SDKs/MacOSX15.0.sdk -fPIC -fvisibility=hidden -fno-strict-aliasing -MD -MT bindings/python/CMakeFiles/nanobind-static.dir/__/__/_deps/nanobind-src/src/implicit.cpp.o -MF CMakeFiles/nanobind-static.dir/__/__/_deps/nanobind-src/src/implicit.cpp.o.d -o CMakeFiles/nanobind-static.dir/__/__/_deps/nanobind-src/src/implicit.cpp.o -c /Users/elalish/Code/manifold/build/_deps/nanobind-src/src/implicit.cpp
[build] In file included from /Users/elalish/Code/manifold/build/_deps/nanobind-src/src/trampoline.cpp:10:
[build] In file included from /Users/elalish/Code/manifold/build/_deps/nanobind-src/include/nanobind/trampoline.h:13:
[build] /Users/elalish/Code/manifold/build/_deps/nanobind-src/include/nanobind/nanobind.h:30:10: fatal error: 'cstdint' file not found
```

pca006132:
By clean build, are you just trying ninja clean or did you remove the entire build directory?

elalish:
Removed the whole build directory. Very strange.

pca006132:
is there an option to install libc++ (dev version)? sounds very weird

elalish:
I don't know - but I can't even build a simple C++ test project now. I think I got screwed by some background update. Super weird.

pca006132:
Maybe try to use a docker environment for now?

elalish:
Yay, I managed to fix my build! My only guess is that somehow Clang's paths got messed up internally - I had to not only upgrade XCode and the command line tools, but actually build a C++ "hello world" program in XCode - which worked and somehow fixed the broken paths. Now my VSCode builds work too. So strange.

src/manifold.cpp
elalish: Huh, that's an interesting idea. It feels kind of wrong because you'd have to pay the cost of an extra simplification check on every mesh output, as well as paying the cost of extra faces in all of the boolean operations up to that point. It would be much better to set tolerance to the right level earlier instead. I feel like we should design our API to push people toward that golden path.

test/properties_test.cpp
```diff
@@ -44,23 +44,42 @@ TEST(Properties, Precision) {
...
+  const auto prop = imperfect.GetProperties();
+
+  Manifold imperfect2 = imperfect.SetTolerance(tol);
+  MeshGL mesh = imperfect.GetMeshGL();
+  mesh.tolerance = tol;
+  Manifold imperfect3(mesh);
+
+  EXPECT_EQ(imperfect.NumTri(), 28);
+  EXPECT_EQ(imperfect2.NumTri(), 16);  // TODO: should be 12
+  EXPECT_EQ(imperfect3.NumTri(), 22);  // TODO: should be 12
```
elalish, line 67: Our simplification is obviously failing here, but I don't think it's due to the tolerance work in this PR (e.g. no change with a larger value). I think I'll merge this for now and come back and work on our simplification later.

pca006132:
I have been thinking about the mesh simplification thing. I think it is probably impossible to guarantee the result of applying several consecutive simplifications to be equivalent to the result of applying simplification once, considering the process is lossy. And this makes it hard to have guarantees on boolean results, e.g. `simpilfy(simplify(A) & simplify(B))` can be very different from `simplify(A & B)`...

elalish:
Agreed that it's not exactly the same, but I think I feel okay as long as it's giving "good" results - along the lines of the volume and surface area are still close, or ideally some metric on maximum difference between the shapes (which we should probably implement someday). If you can find a situation where it's drifting way too much, please open a bug. It'll be helpful when I go through and overhaul the simplification.

PR #1008, opened 2024-10-22
Precision renaming

elalish:
We have now split our old idea of precision into epsilon and tolerance. Here we are renaming as appropriate to help minimize future confusion in the code base.


elalish:
@pca006132 I'm just going to go ahead and merge this - it's more of an FYI than a review request. I'm sure we'll find a few remaining typos here and there.

PR #1014, opened 2024-10-29
Doxygen updates

elalish:
I've updated our version of Doxygen and reorganized our C++ docs quite a bit - needed after all our refactors. Probably still needs work, but it's better than it was. I'm going to merge this quickly so I can test the deployment is actually working.


README.md
```diff
@@ -1,3 +1,5 @@
+# Readme
```
pca006132, line 1: do we really want this?

pca006132 (commented):
will provide some more detailed feedback this weekend

README.md
elalish: I put it in because it makes the doxygen table of contents make more sense. Open to other ideas though.

PR #1018, opened 2024-11-02
bindings update

pca006132:
1. Add MeshGL64 to C and python bindings.
2. Fixed C binding examples memory leak and address sanitizer complaints.

Basically, `ManifoldManifold*` returned from `manifold_alloc_manifold` **CANNOT** be used as `void*` buffer arguments to other functions, because those functions treat the buffer argument as *uninitialized* and will not attempt to destruct them, this causes memory leak for the default-constructed object. Not something serious but should be keep in mind.

Also, I think we must match alloc/free and new/delete pairs, mixing them apparently is an undefined behavior (see https://stackoverflow.com/questions/9238731/how-serious-is-the-new-delete-operator-mismatch-error).


elalish (approved):
Thanks! Have you found any other missing bindings, or are we pretty up to date?

pca006132:
probably still need to check if there is anything missing, but I think we are pretty up-to-date.

PR #1024, opened 2024-11-05
v3.0

elalish:
This is a huge release! We've packed in all the [requested breaking changes](https://github.com/elalish/manifold/discussions/664) to put us on a solid path of API stability going forward. We have removed **all** of our required dependencies (no more Thrust or GLM!) and dramatically simplified our directory structure and CMake. We have upgraded internally to double precision and added `MeshGL64` for those who need to access it. And we have simplified and improved the API in many ways and added a few powerful new features. Please check out our new and improved [C++](https://manifoldcad.org/docs/html/classmanifold_1_1_manifold.html) and [TypeScript](https://manifoldcad.org/jsdocs) docs pages!

## Breaking Changes
- Use `size_t` instead of `int` across our API, @pca006132 in #833
- Made `CrossSection` (and Clipper2 dependency) optional, @elalish in #850
- Removed Thrust dependency, use optional TBB instead, @elalish, @pca006132 in #852, #856, #857
- Moved `LevelSet()` in as a static member of `Manifold`, @elalish in #876, #894
- Upgraded from single to double precision, @pca006132 in #895
- Simplify directory structure, @pca006132, @starseeker in #918, #961
- Remove property tolerance, @elalish in #935
- Remove `Mesh`, @elalish in #950
- Overhaul our CMake, @starseeker in #961, #965, #967, #980, #990, #1005
- Replace GLM with linalg - down another dependency, @elalish in #984
- Separate coplanarity from faceID, @elalish in #991
- Split `Precision` into `Epsilon` and `Tolerance`, @pca006132, @elalish in #997, #1008, #1007
- Switched property indexing from vector to scalar, @elalish in #1016
- Replace `GetProperties()` with `Volume()` and `SurfaceArea()`, @elalish in #1022
- SDF boundary cleanup, @elalish in #1037
- Disable exceptions by default, @elalish in #1044

## New Features
- SDF interpolation, @elalish in #866
- Add `MeshGL64`, @pca006132 in #927, #1038
- Add `RefineToTolerance()`, @elalish in #938
- Add `SetTolerance()`, @pca006132 in #997

## Bindings
- Fix missing 'this' for property access in wasm bindings, @dsafa in #914
- wasm: fix `CrossSection` slice and project bindings, @neilpa in #930
- Expose `Quality::ResetToDefaults()` via C bindings, @NickUfer in #964
- Wasm: update `CrossSection.toPolygons()` return value to match type, @dsafa in #963
- Expose allocation functions via C bindings, @NickUfer in #968
[...]

PR #1025, opened 2024-11-07
Update Readme

elalish:
Addresses 1-6 of https://github.com/elalish/manifold/pull/1024#issuecomment-2458984880 (not sure what to do for 7-9) and added lots of links to users and packages of Manifold.

### DIFF
--- a/README.md
+++ b/README.md
@@ -65,11 +65,11 @@ CMake flags (usage e.g. `-DMANIFOLD_DEBUG=ON`):
 - `MANIFOLD_JSBIND=[OFF, <ON>]`: Build js binding (when using the emscripten toolchain).
 - `MANIFOLD_CBIND=[<OFF>, ON]`: Build C FFI binding.
 - `MANIFOLD_PYBIND=[OFF, <ON>]`: Build python binding.
-- `MANIFOLD_PAR=[<OFF>, ON]`: Provides multi-thread parallelization, requires `libtbb-dev` if enabled.
+- `MANIFOLD_PAR=[<OFF>, ON]`: Provides multi-thread parallelization, requires `tbb` if enabled.
 - `MANIFOLD_CROSS_SECTION=[OFF, <ON>]`: Build CrossSection for 2D support (needed by language bindings), requires `Clipper2` if enabled.
-- `MANIFOLD_EXPORT=[<OFF>, ON]`: Enables GLB export of 3D models from the tests, requires `libassimp-dev`.
+- `MANIFOLD_EXPORT=[<OFF>, ON]`: Enables `MeshIO` and GLB export of 3D models from the tests, requires `libassimp-dev`.
 - `MANIFOLD_DEBUG=[<OFF>, ON]`: Enables internal assertions and exceptions. This incurs around 20% runtime overhead.
-- `MANIFOLD_TEST=[OFF, <ON>]`: Build unittests.
+- `MANIFOLD_TEST=[OFF, <ON>]`: Build unit tests.
 - `TRACY_ENABLE=[<OFF>, ON]`: Enable integration with tracy profiler. 
   See profiling section below.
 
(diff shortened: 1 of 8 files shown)


README.md
```diff
@@ -5,9 +5,34 @@
...
+
+If C++ isn't your cup of tea, Manifold also has bindings to many other languages, some maintained in this repository, and others elsewhere.
+
+| Language | Packager | Name | Maintenance |
+| --- | --- | --- | --- |
+| C | N/A | N/A | internal |
+| TS/JS | npm | [manifold-3d](https://www.npmjs.com/package/manifold-3d) | internal |
+| Python | PyPI | [manifold3d](https://pypi.org/project/manifold3d/) | internal |
+| Java | N/A | [manifold](https://github.com/SovereignShop/manifold) | external |
+| Clojure | N/A | [clj-manifold3d](https://github.com/SovereignShop/clj-manifold3d) | external |
```
pca006132, line 32: Should we be using github url for all external bindings?

README.md
```diff
@@ -27,15 +52,17 @@ If you prefer Python to JS/TS, make your own copy of the example notebook above.
...
-- ['Emscripten'](https://github.com/emscripten-core/emscripten): WASM compiler that provides our JS bindings (only when `MANIFOLD_JSBIND=ON`)
-- [`GTest`](https://github.com/google/googletest/): Google test library (only when test is enabled, i.e. `MANIFOLD_TEST=ON`)
+Manifold no longer has **any** required dependencies! However, we do have several optional dependencies, of which the first two are strongly encouraged:
+| Name | CMake Flag | Provides |
+| --- | --- | --- |
+| [`TBB`](https://github.com/oneapi-src/oneTBB/) |`MANIFOLD_PAR=ON` | Parallel acceleration |
+| [`Clipper2`](https://github.com/AngusJohnson/Clipper2) | `MANIFOLD_CROSS_SECTION=ON` | 2D: [`CrossSection`](https://manifoldcad.org/docs/html/classmanifold_1_1_cross_section.html) |
+| [`Assimp`](https://github.com/assimp/assimp) | `MANIFOLD_EXPORT=ON` | Basic Mesh I/O |
+| [`Nanobind`](https://github.com/wjakob/nanobind) | `MANIFOLD_PYBIND=ON` | Python bindings |
+| [`Emscripten`](https://github.com/emscripten-core/emscripten) | `MANIFOLD_JSBIND=ON` | JS bindings via WASM |
```
pca006132, line 64: I feel like we can just say emscripten is the build system for js bindings, because users can't use this in a normal build, unlike the other dependencies.

README.md
```diff
@@ -65,27 +92,25 @@ CMake flags (usage e.g. `-DMANIFOLD_DEBUG=ON`):
 - `MANIFOLD_JSBIND=[OFF, <ON>]`: Build js binding (when using the emscripten toolchain).
 - `MANIFOLD_CBIND=[<OFF>, ON]`: Build C FFI binding.
 - `MANIFOLD_PYBIND=[OFF, <ON>]`: Build python binding.
-- `MANIFOLD_PAR=[<OFF>, ON]`: Provides multi-thread parallelization, requires `libtbb-dev` if enabled.
+- `MANIFOLD_PAR=[<OFF>, ON]`: Provides multi-thread parallelization, requires `tbb` if enabled.
 - `MANIFOLD_CROSS_SECTION=[OFF, <ON>]`: Build CrossSection for 2D support (needed by language bindings), requires `Clipper2` if enabled.
-- `MANIFOLD_EXPORT=[<OFF>, ON]`: Enables GLB export of 3D models from the tests, requires `libassimp-dev`.
+- `MANIFOLD_EXPORT=[<OFF>, ON]`: Enables `MeshIO` and GLB export of 3D models from the tests, requires `libassimp-dev`.
```
pca006132, line 97: requires `assimp`

README.md
elalish: No, I only use the Github link when there's nothing else. I think a project looks more legit if it has a separate documentation page or better yet if an official package manager lists it, so I use those when I can. They all link back to their githubs anyway.

README.md
elalish: I suppose - really I was thinking of Embind, which we're using a bit like Nanobind, but I think Embind is just part of emscipten like its own STL. The idea was just to list anything that might break if its version updates, which seems a lot more likely for emscripten than for clang.

PR #1026, opened 2024-11-07
Reorder README intro contents slightly

starseeker:
Just a suggestion, but for new potential users coming in to to check out Manifold it might be helpful to have the one sentence summary of what it is right at the beginning?


starseeker:
Oh, whoops - this might be OBE due to #1025

elalish:
Yup, great minds think alike 👍 - I'll take care of this in #1025.

PR #1029, opened 2024-11-08
Fix target include directory when building as a submodule

cjmayo:
Do check, but by trial and error it seems that for the project using the git submodule Manifold's src/ is the `CMAKE_SOURCE_DIR`, in any event it can't find Manifold's include directory which is the level above src.

Manifold still builds for me on its own and the installed cmake files are the same.


elalish (approved):
Thanks!

PR #1033, opened 2024-11-09
remove deterministic flag

elalish:
I want to remove the `deterministic` flag from our execution parameters - it leaves untested code paths that concern me. The question is what is more important: determinism or speed? I'm not quite ready to merge this as is because forcing everything to be deterministic all the time is making our tests run about 20% slower (at least on my M1 macbook pro). WDYT @pca006132?

### DIFF
--- a/src/boolean_result.cpp
+++ b/src/boolean_result.cpp
@@ -253,7 +253,7 @@ void AddNewEdgeVerts(
 #if (MANIFOLD_PAR == 1) && __has_include(<tbb/tbb.h>)
   // parallelize operations, requires concurrent_map so we can only enable this
   // with tbb
-  if (!ManifoldParams().deterministic && p1q2.size() > kParallelThreshold) {
+  if (p1q2.size() > kParallelThreshold) {
     // ideally we should have 1 mutex per key, but kParallelThreshold is enough
     // to avoid contention for most of the cases
     std::array<std::mutex, kParallelThreshold> mutexes;
@@ -486,9 +486,7 @@ void AppendWholeEdges(Manifold::Impl &outR, Vec<int> &facePtrR,
                       bool forward) {
   ZoneScoped;
   for_each_n(
-      ManifoldParams().deterministic ? ExecutionPolicy::Seq
-                                     : autoPolicy(inP.halfedge_.size()),
-      countAt(0), inP.halfedge_.size(),
+      autoPolicy(inP.halfedge_.size()), countAt(0), inP.halfedge_.size(),
       DuplicateHalfedges({outR.halfedge_, halfedgeRef, facePtrR, wholeHalfedgeP,
                           inP.halfedge_, i03, vP2R, faceP2R, forward}));
 }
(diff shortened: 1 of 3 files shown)


pca006132:
I see. I think the `AddNewEdgeVerts` is likely the reason for the slowdown, due to the lack of parallelization previously. We can keep that one and see if it makes the results non-deterministic. There should be ways to speed it up without sacrificing determinism, but this is probably something for later.

For the other change, can you check if removing it will cause major slowdown?

pca006132:
Some additional context:  Due to things like floating point, and changes to mesh simplification, enabling the deterministic version will not give you the same result across different versions or with different machines. I think this also makes deterministic result less important for our use case because we can't have that in general, and not sure if users demanding deterministic result are happy with only this same version guarantee (if it holds at all... we never actually tested it anyway).

I think we can just remove the flag and keep the things that may cause nondeterminism, and try to deal with them later when someone complains.

pca006132:
Another reason: Floating point operations are not really associative, and making this deterministic will require a lot more changes. The topology may be deterministic previously with the deterministic flag on, but things like normal and positions are probably not. If users want deterministic results, they should probably run with parallelization disabled (either compile with that or limit to 1 CPU core).

src/boolean_result.cpp
```diff
@@ -253,7 +253,7 @@ void AddNewEdgeVerts(
 #if (MANIFOLD_PAR == 1) && __has_include(<tbb/tbb.h>)
   // parallelize operations, requires concurrent_map so we can only enable this
   // with tbb
-  if (!ManifoldParams().deterministic && p1q2.size() > kParallelThreshold) {
+  if (p1q2.size() > kParallelThreshold) {
```
elalish, line 256: @pca006132 Actually, switching this to serial had the smallest performance impact of anything, but for now I'm going with the `determinism == false` path for everything.

src/csg_tree.cpp
```diff
@@ -468,35 +468,32 @@ std::shared_ptr<Manifold::Impl> CsgOpNode::BatchBoolean(
...
-              std::make_shared<Manifold::Impl>(boolean.Result(operation)));
-          return group.run(process);
-        });
-      }
-    };
-    group.run_and_wait(process);
-    SharedImpl r;
-    queue.try_pop(r);
-    return *std::get_if<std::shared_ptr<Manifold::Impl>>(&r);
+  tbb::task_group group;
```
elalish, line 471: This one had a noticeable perf impact when removed, but not huge.

src/csg_tree.cpp
```diff
@@ -604,10 +601,8 @@ std::vector<std::shared_ptr<CsgNode>> &CsgOpNode::GetChildren(
 
   if (forceToLeafNodes && !impl->forcedToLeafNodes_) {
     impl->forcedToLeafNodes_ = true;
-    for_each(impl->children_.size() > 1 && !ManifoldParams().deterministic
-                 ? ExecutionPolicy::Par
-                 : ExecutionPolicy::Seq,
-             impl->children_.begin(), impl->children_.end(), [](auto &child) {
+    for_each(ExecutionPolicy::Par, impl->children_.begin(),
```
elalish, line 604: Changing this to `Seq` caused about 75% of the performance regression I was seeing. Even just the `impl->children_.size() > 1` check was causing noticeable perf regression.

elalish (commented):
Okay, reversing course and taking the `determinism == false` branches all the time now. I agree it's likely impossible to get deterministic results across architectures. I'd rather users don't start assuming determinism, particularly across our point releases.

pca006132 (approved):

pca006132:
maybe we can document this in https://github.com/elalish/manifold/wiki/Performance-Considerations

PR #1034, opened 2024-11-10
Add C bindings for Triangulate()

james-bern:
Here's my attempt at adding C bindings for `Triangulate()`. The code seems to work; but there are still a couple things I need help with.

* `Triangulate()` returns a `std::vector<ivec3>`, which I tried to respect. Since `ivec3` is 3 `int`'s, I have `manifold_triangulation_tri_verts(void *mem, ...)` `copy_data` assume `mem` has size `(num_tri * 3 * sizeof(int))`.

* **NOTE:** manifoldc.cpp now includes `"manifold/polygon.h"`

* **NOTE:** I couldn't figure out `manifold_destruct_triangulation(...)`; **need some help on this.**
  * Also, I don't understand the use case(s) of the destruct functions (literally don't understand; I am sure they exist). **Help on this too would be great.**

* **NOTE:** I have not attempted to add bindings for `TriangulateIdx()`

* Usage Code

  ```c
  ManifoldTriangulation *triangulation = manifold_triangulate(manifold_alloc_triangulation(), polygons, epsilon);
  size_t num_tri = manifold_triangulation_num_tri(triangulation);
  int *tri_verts = (int *) manifold_triangulation_tri_verts(malloc(num_tri * 3 * sizeof(int)), triangulation);
  manifold_delete_triangulation(triangulation);
  ```

* Test
  * **NOTE:** I hacked this into my current codebase. I took the polygons I was already getting from `ManifoldPolygons *polygons = manifold_cross_section_to_polygons(manifold_alloc_polygons(), cross_section);`, triangulated them as in the usage code above, and drew them as outlined triangles.

  Screenshot 2024-11-10 at 8 13 30 AM

### DIFF
--- a/bindings/c/manifoldc.cpp
+++ b/bindings/c/manifoldc.cpp
@@ -21,6 +21,7 @@
 #include "manifold/cross_section.h"
 #include "manifold/manifold.h"
 #include "manifold/types.h"
+#include "manifold/polygon.h"
 
 using namespace manifold;
 
@@ -702,6 +703,19 @@ int manifold_get_circular_segments(double radius) {
 
 void manifold_reset_to_circular_defaults() { Quality::ResetToDefaults(); }
 
+ManifoldTriangulation *manifold_triangulate(void *mem, ManifoldPolygons *ps, double epsilon) {
+  auto triangulation = manifold::Triangulate(*from_c(ps), epsilon);
+  return to_c(new (mem) std::vector<ivec3>(triangulation));
+}
+
+size_t manifold_triangulation_num_tri(ManifoldTriangulation *m) {
+  return from_c(m)->size();
+}
+
+int *manifold_triangulation_tri_verts(void *mem, ManifoldTriangulation *m) {
+  return reinterpret_cast<int *>(copy_data<ivec3>(mem, *from_c(m)));
+}
+
 // memory size
 size_t manifold_cross_section_size() { return sizeof(CrossSection); }
 size_t manifold_cross_section_vec_size() {
@@ -716,6 +730,7 @@ size_t manifold_meshgl_size() { return sizeof(MeshGL); }
 size_t manifold_meshgl64_size() { return sizeof(MeshGL64); }
 size_t manifold_box_size() { return sizeof(Box); }
 size_t manifold_rect_size() { return sizeof(Rect); }
+size_t manifold_triangulation_size() { return sizeof(std::vector<ivec3>); }
 
 // allocation
 ManifoldManifold *manifold_alloc_manifold() {
@@ -740,6 +755,9 @@ ManifoldMeshGL *manifold_alloc_meshgl() { return to_c(new MeshGL); }
 ManifoldMeshGL64 *manifold_alloc_meshgl64() { return to_c(new MeshGL64); }
 ManifoldBox *manifold_alloc_box() { return to_c(new Box); }
 ManifoldRect *manifold_alloc_rect() { return to_c(new Rect); }
+ManifoldTriangulation *manifold_alloc_triangulation() {
+    return to_c(new std::vector<ivec3>);
+}
 
 // pointer free + destruction
 void manifold_delete_cross_section(ManifoldCrossSection *c) {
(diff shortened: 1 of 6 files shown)


bindings/c/manifoldc.cpp
```diff
@@ -702,6 +703,20 @@ int manifold_get_circular_segments(double radius) {
 
 void manifold_reset_to_circular_defaults() { Quality::ResetToDefaults(); }
 
+ManifoldTriangulation *manifold_triangulate(void *mem, ManifoldPolygons *ps,
+                                            double epsilon) {
+  auto triangulation = manifold::Triangulate(*from_c(ps), epsilon);
+  return to_c(new (mem) std::vector<ivec3>(triangulation));
```
elalish, line 709: Isn't `auto` already just `std::vector<ivec3>`?

elalish (commented):
This C memory management stuff is not my cup of tea. @pca006132 can you review, or perhaps @geoffder if you happen to have a spare moment while your daughter is asleep?

bindings/c/manifoldc.cpp
pca006132: For in-place new I think you do need this to specify the type, it is not a type cast.

bindings/c/manifoldc.cpp
```diff
@@ -780,6 +802,10 @@ void manifold_destruct_meshgl(ManifoldMeshGL *m) { from_c(m)->~MeshGL(); }
 void manifold_destruct_meshgl64(ManifoldMeshGL64 *m) { from_c(m)->~MeshGL64(); }
 void manifold_destruct_box(ManifoldBox *b) { from_c(b)->~Box(); }
 void manifold_destruct_rect(ManifoldRect *r) { from_c(r)->~Rect(); }
+// error: '~' in destructor name should be after nested name specifier
```
pca006132, line 805:
```diff
diff --git a/bindings/c/manifoldc.cpp b/bindings/c/manifoldc.cpp
index 2072f0da..c82ade1f 100644
--- a/bindings/c/manifoldc.cpp
+++ b/bindings/c/manifoldc.cpp
@@ -802,10 +802,9 @@ void manifold_destruct_meshgl(ManifoldMeshGL *m) { from_c(m)->~MeshGL(); }
 void manifold_destruct_meshgl64(ManifoldMeshGL64 *m) { from_c(m)->~MeshGL64(); }
 void manifold_destruct_box(ManifoldBox *b) { from_c(b)->~Box(); }
 void manifold_destruct_rect(ManifoldRect *r) { from_c(r)->~Rect(); }
-// error: '~' in destructor name should be after nested name specifier
-// void manifold_destruct_triangulation(ManifoldTriangulation *m) {
-//     from_c(m)->~std::vector<ivec3>();
-// }
+void manifold_destruct_triangulation(ManifoldTriangulation *m) {
+    from_c(m)->~vector<ivec3>();
+}
 
 #ifdef __cplusplus
 }
```

pca006132 (changes requested):
I think you should also add a simple test case in `test/manifoldc_test.cpp` to use the API.

pca006132:
Regarding destruct, the thing to keep in mind is that you can have two separate regions of memory: One managed by C/C++ and one managed by the alternative host language, e.g. Java heap. The in-place new syntax (`new (mem) ...`) initializes the data structure in the provided buffer `mem`, but the buffer is allocated separately. Users can either use our `manifold_alloc_*` API, which allocates from the heap managed by C/C++, or use a buffer provided by their host language runtime.

When you destroy the object, it should be done in two steps: 1) call the destructor to drop the internal pointers; 2) free the buffer. When you use the `manifold_delete_*` API, we call C++ `delete`, which does the two steps for us. However, this is invalid when the buffer is user provided instead of managed by C/C++ runtime, and in that case the user should just call the `manifold_destruct_*` API, and manually free the buffer. Note that the internal pointers are pointing to regions managed by the C/C++ runtime, so we don't have to worry about those. As to why the outer object resides in the user buffer... I am not the designer, and we should probably ask @geoffder :)

pca006132 (approved):

pca006132:
Yeah, this is fine, I worked on it because we want to push for v3.0, normally we will wait a bit more :)

PR #1037, opened 2024-11-12
SDF boundary cleanup

elalish:
A final minor breaking change: I've updated our SDF so that it does not exceed its given bounds. Before we would get a funny egg-crate shape due to the body-centered cubic grid, but that seems like an internal detail that users should not need to be aware of. It doesn't seem to affect performance noticeably, so I think it's worthwhile for clarity.

### DIFF
--- a/src/sdf.cpp
+++ b/src/sdf.cpp
@@ -117,6 +117,10 @@ vec3 Position(ivec4 gridIndex, vec3 origin, vec3 spacing) {
   return origin + spacing * (vec3(gridIndex) + (gridIndex.w == 1 ? 0.0 : -0.5));
 }
 
+vec3 Bound(vec3 pos, vec3 origin, vec3 spacing, ivec3 gridSize) {
+  return min(max(pos, origin), origin + spacing * (vec3(gridSize) - 1));
+}
+
 double BoundedSDF(ivec4 gridIndex, vec3 origin, vec3 spacing, ivec3 gridSize,
                   double level, std::function<double(vec3)> sdf) {
   const ivec3 xyz(gridIndex);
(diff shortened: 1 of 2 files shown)


elalish:
Before:
image

After:
image

src/sdf.cpp
```diff
@@ -119,11 +119,9 @@ vec3 Position(ivec4 gridIndex, vec3 origin, vec3 spacing) {
 
 double BoundedSDF(ivec4 gridIndex, vec3 origin, vec3 spacing, ivec3 gridSize,
                   double level, std::function<double(vec3)> sdf) {
-  auto Min = [](ivec3 p) { return std::min(p.x, std::min(p.y, p.z)); };
-
   const ivec3 xyz(gridIndex);
-  const int lowerBoundDist = Min(xyz);
-  const int upperBoundDist = Min(gridSize - xyz);
+  const int lowerBoundDist = minelem(xyz);
```
pca006132, line 123: The naming of this function is funny, I was thinking about "lem" as lemma and thinking about what "mine lemma" means :P

Anyway, it is nice that this is already provided by linalg.

src/sdf.cpp
```diff
@@ -258,12 +256,14 @@ struct NearSurface {
...
-      const vec3 pos = FindSurface(gridPos, gridVert.distance,
-                                   Position(neighborIndex, origin, spacing),
-                                   vMax, tol, level, sdf);
+      vec3 pos = FindSurface(gridPos, gridVert.distance,
+                             Position(neighborIndex, origin, spacing), vMax,
+                             tol, level, sdf);
       // Bound the delta of each vert to ensure the tetrahedron cannot invert.
       if (la::all(la::less(la::abs(pos - gridPos), kS * spacing))) {
         const int idx = AtomicAdd(vertIndex[0], 1);
+        pos = max(pos, origin);
```
pca006132, line 265: Maybe we can make a function out of this, this pattern is used several times.

PR #1040, opened 2024-11-14
Better faces

elalish:
I'm replacing our `CreateFaces` algorithm. Our previous was based on connected components, but in the case of e.g. a high-res sphere, it would connect everything into one big face. So then we added an algorithm to check global tolerance of each face after the fact, and if that failed, it would just throw out all the faces and say each triangle was individual. 

This new algorithm skips all that and instead finds faces by starting with the largest triangle and searching its neighbors until none are within tolerance. Then it gets the next largest unprocessed triangle and does it again, serially, until all triangles have been assigned to faces. It's a bit arbitrary, but works decently. 

Ideally we can eventually follow this on with some improved edge collapsing code.


fire:
This could be a way to attach an edge-length refiner with collapse or subdivide done per polygon.

src/impl.cpp
```diff
@@ -318,75 +215,105 @@ void Manifold::Impl::InitializeOriginal(bool keepFaceID) {
...
+      const vec3 v = vertPos_[halfedge_[h].endVert];
+      if (abs(dot(v - base, normal)) < tolerance_) {
+        meshRelation_.triRef[h / 3].faceID = tp.tri;
+        if (!interiorHalfedges.empty() &&
+            h == halfedge_[interiorHalfedges.back()].pairedHalfedge) {
+          interiorHalfedges.pop_back();
+        } else {
+          interiorHalfedges.push_back(h);
+        }
+        const int hNext = NextHalfedge(h);
```
pca006132, line 260: Why are we also working with the front of the queue?

src/impl.cpp
```diff
@@ -318,75 +215,105 @@ void Manifold::Impl::InitializeOriginal(bool keepFaceID) {
...
+      const int h =
+          NextHalfedge(halfedge_[interiorHalfedges.back()].pairedHalfedge);
+      interiorHalfedges.pop_back();
+      if (meshRelation_.triRef[h / 3].faceID >= 0) continue;
+
+      const vec3 v = vertPos_[halfedge_[h].endVert];
+      if (abs(dot(v - base, normal)) < tolerance_) {
+        meshRelation_.triRef[h / 3].faceID = tp.tri;
+        if (!interiorHalfedges.empty() &&
+            h == halfedge_[interiorHalfedges.back()].pairedHalfedge) {
```
pca006132, line 255: Considering `h = NextHalfedge(halfedge_[interiorHalfedges.back()].pairedHalfedge)`, will this condition ever hold?

src/impl.cpp
```diff
@@ -318,75 +215,105 @@ void Manifold::Impl::InitializeOriginal(bool keepFaceID) {
...
-  }
+  const int numTri = NumTri();
+  struct TriPriority {
+    double area2;
+    int tri;
+  };
+  Vec<TriPriority> triPriority(numTri);
+  for_each_n(autoPolicy(numTri), countAt(0), numTri,
+             [&triPriority, this](int tri) {
+               meshRelation_.triRef[tri].faceID = -1;
```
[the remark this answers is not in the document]
elalish: No, because they get reassigned on lines 240 and 253.

src/impl.cpp
elalish: No, because the `back` got popped off in line 248.

fire:
can the random selection be more deterministic

elalish:
> can the random selection be more deterministic for like ai.. and import recreation / reproducibility

Since this is based on sorting, it should be as deterministic as we can hope for.

fire:
Can you see what happens when you try to merge https://github.com/user-attachments/files/17969298/node_3d.zip ?

elalish:
Okay, I got bogged down on this one because I tried to solve too many things at once. The actual point of this PR, which is to create faces better and faster is working just fine. Then I tried to make it into a general purpose decimator, but got stuck on some Boolean regressions. Now I've undone that to get back to basics here. I'll do further work on the decimator in separate PRs.

elalish:
@pca006132 This is odd - the CI is passing almost everywhere, but on three jobs it's failing, and not just a few tests, but a ton of them and even seg-faulting. On my mac everything is fine - even the address sanitizer isn't finding any undefined behavior. Do you have any idea what I did wrong?

pca006132:
It seems that I can reproduce them locally with parallelization disabled. Will check.

pca006132:
OK, for some reason Linux `stdlib.h` defines an `int abs(int);`, changing to `std::abs` fixes the issue.

elalish:
Wow, thanks for the fix! I wonder what their `abs` was doing... One of these days I'll remember to always put `std::` in front of everything.

PR #1044, opened 2024-11-14
Disable exceptions by default

### LINKED ISSUE #1021: C API exception handling
Users can build manifold with exception enabled, and I expect distros to package manifold with exception enabled. Typical C bindings try to catch the exception before the exception crossing FFI boundary. Should we do the same here? And what should be the return value of various functions when we catch an exception?

@geoffder @NickUfer @cartesian-theatrics

### LINKED ISSUE #1041: Merge MANIFOLD_EXCEPTIONS and MANIFOLD_DEBUG
Currently our readme says our `MANIFOLD_DEBUG` flag controls both assertions and exceptions, but in fact we have a second CMake flag: `MANIFOLD_EXCEPTIONS`, which defaults to ON. However, nearly all of our exceptions use `DEBUG_ASSERT`, for which *both* debug and exceptions must be enabled to fire. Only `Vec` uses `ASSERT` (and hull, but I think that's a mistake) which is enabled even when debug is off. I can't imagine a situation where an app is going to be able to do much if we throw a `Vec` assertion - it's basically just a seg-fault, and it certainly shouldn't happen if we've written our code properly. 

I propose we merge `MANIFOLD_EXCEPTIONS` into `MANIFOLD_DEBUG` and use only a single kind of `ASSERT`. This will have the effect of turning exceptions off by default, which seems right? Feedback welcome - and WDYT @pca006132?

elalish:
Fixes #1021 
Fixes #1041 

I've removed our `MANIFOLD_EXCEPTIONS` CMake flag, opting to control it with `MANIFOLD_DEBUG` instead. Very strangely, I'm now getting a 50% slowdown on running my tests (even though I had exceptions and debug already enabled in my build). I'm a bit at a loss as to how this can change the performance. Any ideas @pca006132?


pca006132:
It turns out this is due to the function accepting `const std::string&`, causing C++ to *always* allocate a temporary `std::string` from `const char*` and immediately deallocate it afterwards. This is the major bottleneck. And then there is also function call overhead, but that is minor comparing to the allocation and deallocation overhead.

pca006132:
To be honest I have no idea how the copying causes the g++ error, but it is indeed a bug in our code.

PR #1045, opened 2024-11-15
Enable emscripten build with parallelization

pca006132:
1. Enabled parallelization when building with emscripten. It turns out we have to warm up the tbb workers in our tests, according to https://github.com/oneapi-src/oneTBB/blob/master/WASM_Support.md#limitations.
2. Fix nix build. The nixpkg upstream PR is still not updated, so I wrote an overlay to build binaryen instead. Because it can take a while in the CI, I enabled a workflow cache.

Note that the parallel build is still experimental, it seems that things can be a bit finicky and I had to increase the stack size. The build is also a bit finicky, I got errors running "Gyroid module" with npm/browser, will debug that a bit later... But overall the thing works in npm and browser.


pca006132:
Looking at the emscripten bug reports, I feel like the issue we got with gyroid module may be something related to the older version of emscripten that nix currently uses. Maybe building with the latest version will make things work.

pca006132:
@elalish indeed the gyroid module issue is just due to parallelization being enabled for level set.

elalish:
> `test_manifold` is still a bit finicky, disabling test for now so to not block the CI.

Mind putting this comment on the line of that change that disabled the test? I'm having trouble finding it, and seems like a good one to keep an eye on when we're getting ready to merge this.

How much performance improvement do you see on our npm tests?

flake.nix
```diff
@@ -9,141 +9,203 @@
...
+            cp bindings/wasm/manifold.* $out/
+          '';
+        };
+      in
+      {
+        packages = {
+          manifold-tbb = manifold { };
+          manifold-none = manifold { parallel = false; };
+          manifold-js = manifold-emscripten { parallel = false; };
+          # disable check for now
```
pca006132, line 153: This one with `doCheck = false` disables the test for nix.

pca006132:
For timing improvement, overall it is not huge, 11805ms to 9484ms. But for individual tests, I think I saw 2x speedup or something, so this really depends on the specific test. Do note that I limited the number of pthread worker to 4.

pca006132:
@elalish I rewrote the nix CI to pack everything into one single action with multiple variants.

elalish:
This is looking great - okay if we delay merging this to after v3.0?

pca006132:
sure

pca006132:
@elalish It seems to work reliably now, but IDK why. Maybe some linker flags were not actually applied previously, but are now applied correctly?

bindings/wasm/fixup.py
```diff
@@ -0,0 +1,19 @@
...
+
+filename = sys.argv[1]
+with open(filename, "r") as file:
+    data = file.read()
+
+data = data.replace(
+    'var workerOptions={type:"module",workerData:"em-pthread",name:"em-pthread"};', ""
+)
+data = data.replace(
+    "workerOptions", '{type:"module",workerData:"em-pthread",name:"em-pthread"}'
```
elalish, line 15: What is this?

elalish (approved):
How much speedup do you see coming out of this now?

bindings/wasm/fixup.py
pca006132: emscripten generating some non-static worker options that Vite cannot handle, and it can be trivially modified to become static.

pca006132:
> How much speedup do you see coming out of this now?

Overall for the c++ test, 4x, probably more for some individual benchmark. For npm test, roughly 11s to 8s, 2x faster in some benchmark.

PR #1049, opened 2024-11-16
manifold alias

pca006132:
For openscad/openscad#5308, basically let them link against `manifold::manifold` regardless if our library is installed or put in a subdirectory.


pca006132:
@elalish this is the change I want to get into v3.0.

elalish (approved):
Great, v3.0 here we come!

PR #1056, opened 2024-11-18
Fix publishing

elalish:
Looks like our [npm](https://github.com/elalish/manifold/actions/runs/11886415327/job/33117548674) and [PyPI](https://github.com/elalish/manifold/actions/runs/11886415289/job/33117549096) publishing workflows failed again. The npm one failed because it didn't pick up our artifacts - I'm taking a shot in the dark that bumping our actions versions might help. 

There seem to be two different problems for Python, which are more baffling. Ubuntu works, and MacOS works for almost everything, but its x86_64 wheel is failing saying it needs 'arm64e' or 'arm64', which seems odd. And the the SDist failed because apparently CMake can't manage to add the `test`, `samples`, or `extras` directories (??). Seems to work fine everywhere else...

@pca006132 any thoughts?

### DIFF
--- a/.github/workflows/build_wheels.yml
+++ b/.github/workflows/build_wheels.yml
@@ -19,9 +19,9 @@ jobs:
           git config --global submodule.fetchJobs 8
           git config --global core.longpaths true
       - if: matrix.os == 'macos-13'
-        run: echo "CIBW_ARCHS=\"x86_64\"" >> "$GITHUB_ENV"
+        run: echo "CIBW_ARCHS=x86_64" >> "$GITHUB_ENV"
       - if: matrix.os == 'macos-14'
-        run: echo "CIBW_ARCHS=\"arm64 universal2\"" >> "$GITHUB_ENV"
+        run: echo "CIBW_ARCHS=arm64 universal2" >> "$GITHUB_ENV"
       - uses: actions/checkout@v4
       - name: Build wheels
         uses: pypa/cibuildwheel@v2.19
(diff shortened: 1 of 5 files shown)


pca006132:
The SDist problem is due to we excluding some of the directories, but forgot to make cmake skip those directories when building the SDist.

pca006132:
Manually triggering https://github.com/elalish/manifold/actions/runs/11886656652

pca006132:
I feel like publish NPM failure is caused by improper CI dependency, which caused deployment issues previously. I triggered the job manually and it works now :)

pca006132:
I wonder if the python CI problem is caused by us trying to run some python script during cross-compilation, which uses a version of python incompatible with the architecture. I can look at it later today.

elalish:
Damn, our WASM is twice as fast and half the size for this release! Nice to see that up on npm. I'm going to bed - I'll check in on the PyPI situation in the morning. Thanks for looking into it.

hrgdavor:
> Damn, our WASM is twice as fast and half the size for this release! Nice to see that up on npm. I'm going to bed - I'll check in on the PyPI situation in the morning. Thanks for looking into it.

woooooooooooow that is a massive improvement. Thank you for all the hard work.

pca006132:
see if this works... https://github.com/elalish/manifold/actions/runs/11888407884

elalish:
> The action will automatically deploy to pypi, I guess we should make a version that doesn't do that later to verify before release.

That sounds reasonable if you have a plan for how to verify it. Thoughts?

pca006132:
Just see if it builds?

elalish:
Oh, I thought it wouldn't successfully deploy unless it built anyway. Wasn't that was happened here before this PR?

pca006132:
Yeah, but we can check if our PR can be built successfully before merging.

PR #1061, opened 2024-11-18
parallel.h is private again, so we can make iters.h private as well.

starseeker:


Doxyfile
```diff
@@ -972,8 +972,8 @@ INPUT                  = include/manifold/manifold.h \
                          include/manifold/meshIO.h \
                          include/manifold/linalg.h \
                          include/manifold/vec_view.h \
-                         include/manifold/iters.h \
                          include/manifold/optional_assert.h \
+                         src/iters.h \
```
elalish, line 976: Actually, you can remove this and parallel.h from Doxygen - we don't document non-public features.

elalish (approved):
Thanks!

PR #1062, opened 2024-11-18
fix download-artifact

elalish:
I think this'll fix our WASM deployment. The error was `The following inputs cannot be used together: pr, commit, branch, run_id`

### DIFF
--- a/.github/workflows/publish_npm.yml
+++ b/.github/workflows/publish_npm.yml
@@ -24,7 +24,7 @@ jobs:
         workflow: manifold.yml
         workflow_conclusion: completed
         branch: master
-        check_artifact: true
+        check_artifacts: true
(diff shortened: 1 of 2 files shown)


.github/workflows/publish_npm.yml
```diff
@@ -24,7 +24,7 @@ jobs:
         workflow: manifold.yml
         workflow_conclusion: completed
         branch: master
-        check_artifact: true
+        check_artifacts: true
```
elalish, line 27: @pca006132 any reason we should use `branch: master` here instead of `run_id` like above?

pca006132: `publish_npm` is not triggered by another workflow, so we can't do that.

PR #1065, opened 2024-11-21
cancel in-progress runs upon new push

pca006132:
Thought it is a bit wasteful to keep running the old job when we push some new commits. This will now cancel the old job.

And this removes the parallel limit for some builds, I forgot why they were there in the first place, probably copied from openscad?


elalish:
Good idea, thanks!

PR #1079, opened 2024-11-28
wasm-tbb: increase stack size and add CI

pca006132:
Now run in "normal" CI as well with npm test.

Also updates dependency versions.


pca006132:
Emscripten may soon fix the vite issue: https://github.com/emscripten-core/emscripten/issues/22394

pca006132:
hmmm, the vite thing is annoying, I guess it is due to emscripten version issue...

pca006132:
And it seems that emscripten pthread support is not that stable for now...

elalish (approved):
So, is this one ready? Should I update our branch rules so we can merge it?

elalish:
Is this PR intended to fix this problem? https://github.com/elalish/manifold/actions/runs/12076790849/job/33678681856

pca006132:
> So, is this one ready? Should I update our branch rules so we can merge it?

Branch rule: Yes, the name should be stable.

I am not entirely sure if I want to experiment with newer versions of emscripten or not. iirc newer version runs fine in the CI (with only a few runs though, not sure if it is a lot more stable than what we have now), but because nixos haven't yet support the newer versions of emscripten, I will need to use it inside a container to figure out how to patch the js file to make vite happy.

Currently, I simply avoid running the tests on the CI, and just treat this as a compilation test, because we know that it is unstable for the emscripten version we have tested.

pca006132:
Hmmm, for some reason vite did not complain about workerOptions issue with emscripten 3.1.73 when run inside a docker, idk what happened. I think we can merge this for now.

pca006132:
But apparently 3.1.73 works fine without issue, at least on my machine. 3.1.61/64 will sometimes produce the issue.

pca006132:
> But apparently 3.1.73 works fine without issue, at least on my machine. 3.1.61/64 will sometimes produce the issue.

May break when I add some assertions, very non-deterministic. And sadly emscripten does not support address sanitizer on wasm workers for now...

pca006132:
Added warnings to the readme for enabling parallelization with emscripten, as well as changed the emscripten `CMAKE_BUILD_TYPE` to MinSizeRel. This shrinks the size of the wasm from 559K to 470K for emscripten 3.1.64.

PR #1080, opened 2024-11-28
make format run in parallel

pca006132:
Originally, the format script will block for about 2s because things are run in sequence, and this will cause git commit to lag when this is added to git commit hook. This is basically instant now (takes ~0.2s).


elalish (approved):
So simple, nice!

PR #1085, opened 2024-12-01
fix: change manifold_calculate_normals min_sharp_angle type to double

### LINKED ISSUE #1083: C binding function manifold_calculate_normals contains parameter type mistmatch
The [manifold_calculate_normals](https://github.com/elalish/manifold/blob/11387fb871077953cedd3e08c0c70afe5969b3f0/bindings/c/include/manifold/manifoldc.h#L216-L218) function has a parameter called `min_sharp_angle` of type `int`. This function then calls [Manifold::CalculateNormals](https://github.com/elalish/manifold/blob/11387fb871077953cedd3e08c0c70afe5969b3f0/src/manifold.cpp#L710-L714) where the equivalent parameter, `minSharpAngle`, is of type `double` (formerly `float`).

This type mismatch limits the function's potential by restricting the precision with which sharp angles are handled.

NickUfer:
Fixes #1083

### DIFF
--- a/bindings/c/include/manifold/manifoldc.h
+++ b/bindings/c/include/manifold/manifoldc.h
@@ -215,7 +215,7 @@ double manifold_min_gap(ManifoldManifold *m, ManifoldManifold *other,
                         double searchLength);
 ManifoldManifold *manifold_calculate_normals(void *mem, ManifoldManifold *m,
                                              int normal_idx,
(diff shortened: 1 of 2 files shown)


pca006132 (approved):

pca006132:
Thanks!

PR #1086, opened 2024-12-01
avoid recursion in csg tree

### LINKED ISSUE #989: accidental quadratic runtime in csg tree

pca006132:
Fixes #989.

This refactors the code into a single loop that does dfs on the csg tree, and perform flattening on-the-fly. 

There are still some failures, I guess this is due to shared nodes being evaluated in a weird way... Will try to fix it soon. But for now I think the logic is mostly there and should be ready for review.

For comparison:
```
// before
n = 20
nTri = 91814, time = 2.28608 sec
n = 30
nTri = 279334, time = 20.6878 sec

// after
n = 20
nTri = 91814, time = 1.18475 sec
n = 30
nTri = 279334, time = 5.67487 sec
```

### DIFF
--- a/src/csg_tree.cpp
+++ b/src/csg_tree.cpp
@@ -624,6 +624,8 @@ CsgNodeType CsgOpNode::GetNodeType() const {
     case OpType::Intersect:
       return CsgNodeType::Intersection;
   }
+  // unreachable...
+  return CsgNodeType::Leaf;
 }
 
 mat3x4 CsgOpNode::GetTransform() const { return transform_; }
--- a/src/csg_tree.cpp
+++ b/src/csg_tree.cpp
@@ -330,129 +330,11 @@ Manifold::Impl CsgLeafNode::Compose(
   return combined;
 }
 
-CsgOpNode::CsgOpNode() {}
-
-CsgOpNode::CsgOpNode(const std::vector<std::shared_ptr<CsgNode>> &children,
-                     OpType op)
-    : impl_(Impl{}) {
-  auto impl = impl_.GetGuard();
-  impl->children_ = children;
-  SetOp(op);
-}
-
-CsgOpNode::CsgOpNode(std::vector<std::shared_ptr<CsgNode>> &&children,
-                     OpType op)
-    : impl_(Impl{}) {
-  auto impl = impl_.GetGuard();
-  impl->children_ = children;
-  SetOp(op);
-}
-
-std::shared_ptr<CsgNode> CsgOpNode::Boolean(
-    const std::shared_ptr<CsgNode> &second, OpType op) {
-  std::vector<std::shared_ptr<CsgNode>> children;
-
-  auto isReused = [](const auto &node) { return node->impl_.UseCount() > 1; };
-
-  auto copyChildren = [&](const auto &list, const mat3x4 &transform) {
-    for (const auto &child : list) {
-      children.push_back(child->Transform(transform));
-    }
-  };
-
-  auto self = std::dynamic_pointer_cast<CsgOpNode>(shared_from_this());
-  if (IsOp(op) && !isReused(self)) {
-    auto impl = impl_.GetGuard();
-    copyChildren(impl->children_, transform_);
-  } else {
-    children.push_back(self);
-  }
-
-  auto secondOp = std::dynamic_pointer_cast<CsgOpNode>(second);
-  auto canInlineSecondOp = [&]() {
-    switch (op) {
-      case OpType::Add:
-      case OpType::Intersect:
-        return secondOp->IsOp(op);
-      case OpType::Subtract:
-        return secondOp->IsOp(OpType::Add);
-      default:
-        return false;
-    }
-  };
-
-  if (secondOp && canInlineSecondOp() && !isReused(secondOp)) {
-    auto secondImpl = secondOp->impl_.GetGuard();
-    copyChildren(secondImpl->children_, secondOp->transform_);
-  } else {
-    children.push_back(second);
-  }
-
-  return std::make_shared<CsgOpNode>(children, op);
-}
-
-std::shared_ptr<CsgNode> CsgOpNode::Transform(const mat3x4 &m) const {
-  auto node = std::make_shared<CsgOpNode>();
-  node->impl_ = impl_;
-  node->transform_ = m * Mat4(transform_);
-  node->op_ = op_;
-  return node;
-}
-
-std::shared_ptr<CsgLeafNode> CsgOpNode::ToLeafNode() const {
-  if (cache_ != nullptr) return cache_;
-  // turn the children into leaf nodes
-  GetChildren();
-  auto impl = impl_.GetGuard();
-  auto &children_ = impl->children_;
-  if (children_.size() > 1) {
-    switch (op_) {
-      case CsgNodeType::Union:
-        BatchUnion();
-        break;
-      case CsgNodeType::Intersection: {
-        std::vector<std::shared_ptr<const Manifold::Impl>> impls;
-        for (auto &child : children_) {
-          impls.push_back(
-              std::dynamic_pointer_cast<CsgLeafNode>(child)->GetImpl());
-        }
-        children_.clear();
-        children_.push_back(std::make_shared<CsgLeafNode>(
-            BatchBoolean(OpType::Intersect, impls)));
-        break;
-      };
-      case CsgNodeType::Difference: {
-        // take the lhs out and treat the remaining nodes as the rhs, perform
-        // union optimization for them
-        auto lhs = std::dynamic_pointer_cast<CsgLeafNode>(children_.front());
-        children_.erase(children_.begin());
-        BatchUnion();
-        auto rhs = std::dynamic_pointer_cast<CsgLeafNode>(children_.front());
-        children_.clear();
-        Boolean3 boolean(*lhs->GetImpl(), *rhs->GetImpl(), OpType::Subtract);
-        children_.push_back(
-            std::make_shared<CsgLeafNode>(std::make_shared<Manifold::Impl>(
-                boolean.Result(OpType::Subtract))));
-      };
-      case CsgNodeType::Leaf:
-        // unreachable
-        break;
-    }
-  } else if (children_.size() == 0) {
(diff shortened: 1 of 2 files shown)


src/csg_tree.cpp
```diff
@@ -566,76 +444,188 @@ void CsgOpNode::BatchUnion() const {
...
         impls.push_back(
-            std::dynamic_pointer_cast<CsgLeafNode>(children_[start + set[0]])
+            std::static_pointer_cast<CsgLeafNode>(children[start + set[0]])
                 ->GetImpl());
       } else {
         std::vector<std::shared_ptr<CsgLeafNode>> tmp;
         for (size_t j : set) {
           tmp.push_back(
-              std::dynamic_pointer_cast<CsgLeafNode>(children_[start + j]));
+              std::dynamic_pointer_cast<CsgLeafNode>(children[start + j]));
```
elalish, line 453: why is this cast dynamic while the one above is static?

src/csg_tree.cpp
```diff
@@ -566,76 +444,188 @@ void CsgOpNode::BatchUnion() const {
...
+  node->transform_ = m * Mat4(transform_);
+  node->op_ = op_;
+  return node;
+}
+
+struct CsgStackFrame {
+  bool finalize;
+  OpType parent_op;
+  mat3x4 transform;
+  std::vector<std::shared_ptr<CsgLeafNode>> *parent_children;
```
elalish, line 506: what's the relationship between `parent_children` and `left_children` and `right_children`? And why is just the parent a pointer?

src/csg_tree.cpp
```diff
@@ -566,76 +444,188 @@ void CsgOpNode::BatchUnion() const {
...
+                std::vector<std::shared_ptr<CsgLeafNode>> *parent_children,
+                std::shared_ptr<const CsgOpNode> op_node)
+      : finalize(finalize),
+        parent_op(parent_op),
+        transform(transform),
+        parent_children(parent_children),
+        op_node(op_node) {}
+};
+
+std::shared_ptr<CsgLeafNode> CsgOpNode::ToLeafNode() const {
```
elalish, line 521: I never fully understood what the previous version of this algorithm did either - perhaps some comments on the general steps this is recursing through?

elalish (commented):
Thanks! Looks like a good direction, though a bit out of my wheelhouse. I'll be curious to understand where the bug is...

src/csg_tree.cpp
pca006132: ah, I should change these to static.

src/csg_tree.cpp
pca006132: `parent_children` is the list of children of the parent (or ancestor), while `left_children` and `right_children` are the sets of children of the current node.

Will explain it better with some comment.

elalish: I think I might call `parent_children` `siblings`, if I'm understanding you correctly?

pca006132:
CondensedMatter16 hits #970 after this PR. I guess this is due to the code now performs optimization more aggressively, and there are more things that got feed into BatchUnion...

pca006132:
Will add documentation tmr, a bit tired today.

pca006132:
Anyway, decided to write the documentation now :)

About the original bug: The idea is that the flattening takes O(n) time in the old implementation, which makes it problematic because in the worst case we can have O(n) flattening operations as well, which gives us O(n^2) time. As an example, consider the following:

```
Union(a, Union(b, Union(c, Union(d)))
=> Union(a, Union(b, Union(c, d)))       // 1 copy
=> Union(a, Union(b, c, d))              // 2 copy
=> Union(a, b, c, d)                     // 3 copy
```

The implementation here doesn't attempt to do flattening when we construct the node. In fact, it is impossible to avoid this quadratic behavior if we perform (all) flattening when we construct the nodes, because there will be O(n^2) nodes if we do that. The algorithm only does one pass over the tree, and do flattening on-the-fly. Each node is moved only once. This makes it O(n).

src/csg_tree.cpp
```diff
@@ -565,77 +444,239 @@ void CsgOpNode::BatchUnion() const {
...
+      stack.pop_back();
+    } else {
+      auto add_children = [&](std::shared_ptr<CsgNode> &node, OpType op,
+                              mat3x4 transform, auto *children) {
+        if (node->GetNodeType() == CsgNodeType::Leaf)
+          children->push_back(std::static_pointer_cast<CsgLeafNode>(
+              node->Transform(transform)));
+        else
+          stack.push_back(std::make_shared<CsgStackFrame>(
+              false, op, transform, children,
```
elalish, line 640: I don't quite follow this logic: either we add the node to the set of children, or we add a frame to the stack that contains this node *and* all the children? Do the children belong to this node or not (when the function is called)?

src/csg_tree.cpp
```diff
@@ -565,77 +444,239 @@ void CsgOpNode::BatchUnion() const {
...
+            break;
+        }
+      frame->op_node->cache_ = std::static_pointer_cast<CsgLeafNode>(
+          impl->children_[0]->Transform(frame->op_node->transform_));
+      if (frame->destination != nullptr)
+        frame->destination->push_back(std::static_pointer_cast<CsgLeafNode>(
+            frame->op_node->cache_->Transform(frame->transform)));
+      stack.pop_back();
+    } else {
+      auto add_children = [&](std::shared_ptr<CsgNode> &node, OpType op,
```
elalish, line 633: I prefer to be explicit about what's captured in a lambda - it helps me to understand the scope of the function.

src/csg_tree.cpp
```diff
@@ -565,77 +444,239 @@ void CsgOpNode::BatchUnion() const {
...
+                       i == 0 ? &frame->left_children : &frame->right_children);
+      } else {
+        // op_node use_count == 2 because it is both inside one CsgOpNode
+        // and in our stack.
+        bool skipFinalize = frame->destination != nullptr &&
+                            frame->op_node->op_ == frame->parent_op &&
+                            frame->op_node.use_count() <= 2 &&
+                            frame->op_node->impl_.UseCount() == 1;
+        if (skipFinalize) stack.pop_back();
+        auto transform =
```
elalish, line 656: const? And for short types, I prefer to avoid `auto`.

src/csg_tree.cpp
```diff
@@ -565,77 +444,239 @@ void CsgOpNode::BatchUnion() const {
...
+          add_children(impl->children_[i], OpType::Add, la::identity,
+                       i == 0 ? &frame->left_children : &frame->right_children);
+      } else {
+        // op_node use_count == 2 because it is both inside one CsgOpNode
+        // and in our stack.
+        bool skipFinalize = frame->destination != nullptr &&
+                            frame->op_node->op_ == frame->parent_op &&
+                            frame->op_node.use_count() <= 2 &&
+                            frame->op_node->impl_.UseCount() == 1;
+        if (skipFinalize) stack.pop_back();
```
elalish, line 655: We just set `frame->finalize = true;` above - is that still true? Or does this skipping refer to its child or parent?

src/csg_tree.cpp
```diff
@@ -565,77 +444,239 @@ void CsgOpNode::BatchUnion() const {
...
+  // keep using the old `destination` pointer for the grandchildren. Hence,
+  // removing a node by flattening takes O(1) time. We also need to store the
+  // parent operation type for checking if the node is eligible for flattening,
+  // and transform matrix because we need to re-apply the transformation to the
+  // children.
+  //
+  // `Subtract` is handled differently from `Add` and `Intersect`. It is treated
+  // as two `Add` nodes, `left_children` and `right_children`, that should be
+  // subtracted later. This allows flattening children `Add` nodes. For normal
+  // `Add` and `Intersect`, we only use `left_children`.
```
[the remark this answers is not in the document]
pca006132: sure.

src/csg_tree.cpp
```diff
@@ -565,77 +444,239 @@ void CsgOpNode::BatchUnion() const {
...
+    // `BatchBoolean` or `Subtract` on this is already a no-op...
+    if (frame->op_node->cache_) {
+      frame->destination->push_back(std::static_pointer_cast<CsgLeafNode>(
+          frame->op_node->cache_->Transform(frame->transform)));
+      stack.pop_back();
+      continue;
+    }
+    auto impl = frame->op_node->impl_.GetGuard();
+    if (frame->finalize) {
+      if (frame->op_node->cache_ == nullptr) switch (frame->op_node->op_) {
```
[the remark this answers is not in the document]
pca006132: Yeah, I forgot to remove this check after adding the check above.

src/csg_tree.cpp
pca006132: The frame is removed here, so we will skip the current frame. Anyway, I can make it explicit.

src/csg_tree.cpp
pca006132: You can view adding a stack frame as "we will call this node recursively".
Maybe I should add a recursion pseudocode to make things clearer, this code is quite indirect...

pca006132: See if it is better with the recursive pseudocode...

elalish (approved):
Thanks, that's much clearer!

pca006132:
maybe wait for me to merge it, I just thought of some minor changes.

pca006132:
After thinking about it, the fast path for testing `cache_` is probably moot, as it requires the exact same node with the exact same transformation, which users will probably never write (why would they do boolean operations on two things that are exactly the same? Not about the mesh result, but the input expressions are the same). Instead, I just test for single children case, which indicates this node is probably shared and is already evaluated, this allows for node sharing with different transforms, which is a legitimate usecase.

PR #1090, opened 2024-12-03
refactor(1089): update index types from int to size_t

### LINKED ISSUE #1089: Inconsistent use of int and size_t in C Bindings for std::vector Types
The types `ManifoldCrossSectionVec`, `ManifoldPolygon`, `ManifoldPolygons` and `ManifoldManifoldVec` represent `std::vector` in the C bindings. However, their API functions mix the use of `size_t` and `int` for lengths and indices. For example:

ManifoldCrossSectionVec:
```c
size_t manifold_cross_section_vec_length(ManifoldCrossSectionVec *csv); // uses size_t
ManifoldCrossSection *manifold_cross_section_vec_get(void *mem, ManifoldCrossSectionVec *csv, int idx); // uses int
```

SimplePolygon:
```c
size_t manifold_simple_polygon_length(ManifoldSimplePolygon *p); // uses size_t
ManifoldVec2 manifold_simple_polygon_get_point(ManifoldSimplePolygon *p, int idx); // uses int
```

ManifoldPolygons:
```c
size_t manifold_polygons_length(ManifoldPolygons *ps); // uses size_t
ManifoldSimplePolygon *manifold_polygons_get_simple(void *mem, ManifoldPolygons *ps, int idx); // uses int
```

ManifoldManifoldVec:
```c
size_t manifold_manifold_vec_length(ManifoldManifoldVec *ms); // uses size_t
void manifold_manifold_vec_set(ManifoldManifoldVec *ms, int idx, ManifoldManifold *m); // uses int
ManifoldManifold *manifold_manifold_vec_get(void *mem, ManifoldManifoldVec *ms, int idx); // uses int
```

### Problem
The use of `int` for indices is problematic because `int` cannot represent the full range of values that `size_t` can. This mismatch can lead to:

- **Potential bugs**: Negative or out-of-range values for indices.
- **Type safety issues**: Implicit conversions and sign mismatches.
- **Platform-specific inconsistencies**: Differences in the size of `int` and `size_t` across platforms.

### Suggested Solution
All `int` parameters used for indices in the API should be updated to `size_t` to ensure consistency with the `std::vector` representation and avoid these issues.
Would be open to contribute a fix 😄

NickUfer:
Fixes #1089

### DIFF
--- a/bindings/c/cross.cpp
+++ b/bindings/c/cross.cpp
@@ -59,12 +59,12 @@ size_t manifold_cross_section_vec_length(ManifoldCrossSectionVec *csv) {
 }
 
 ManifoldCrossSection *manifold_cross_section_vec_get(
-    void *mem, ManifoldCrossSectionVec *csv, int idx) {
+    void *mem, ManifoldCrossSectionVec *csv, size_t idx) {
   auto cs = (*from_c(csv))[idx];
   return to_c(new (mem) CrossSection(cs));
 }
 
-void manifold_cross_section_vec_set(ManifoldCrossSectionVec *csv, int idx,
+void manifold_cross_section_vec_set(ManifoldCrossSectionVec *csv, size_t idx,
                                     ManifoldCrossSection *cs) {
   (*from_c(csv))[idx] = *from_c(cs);
 }
--- a/bindings/c/include/manifold/manifoldc.h
+++ b/bindings/c/include/manifold/manifoldc.h
@@ -30,14 +30,14 @@ ManifoldPolygons *manifold_polygons(void *mem, ManifoldSimplePolygon **ps,
                                     size_t length);
 size_t manifold_simple_polygon_length(ManifoldSimplePolygon *p);
 size_t manifold_polygons_length(ManifoldPolygons *ps);
-size_t manifold_polygons_simple_length(ManifoldPolygons *ps, int idx);
+size_t manifold_polygons_simple_length(ManifoldPolygons *ps, size_t idx);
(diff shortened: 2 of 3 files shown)


include/manifold/manifold.h
```diff
@@ -270,7 +270,7 @@ class Manifold {
   size_t NumVert() const;
   size_t NumEdge() const;
   size_t NumTri() const;
-  size_t NumProp() const;
+  int NumProp() const;
```
elalish, line 273: Why do you want to turn this into `int`?

NickUfer: Because NumProp is of type int: https://github.com/elalish/manifold/blob/f1a86930164b360765c5e04d633aae4d617acdd1/src/impl.h#L37C5-L37C21

elalish: I think the right thing is to update that to `size_t` rather than vice-versa.

NickUfer: Okay, I thought about trying a simpler refactor, but if you say it is fine to change that to `size_t`, I'll do that later :smile:

elalish: Certainly beats changing our public API 😄

pca006132: Yeah, we have things that are not yet `size_t` but probably should be. They are mostly internal, though.

pca006132: Sounds good, I think we can keep the internal things unchanged for now.

elalish (approved):
Thanks!

PR #1094, opened 2024-12-04
run tests for nix

pca006132:


flake.nix
```diff
@@ -106,6 +106,7 @@
               "-DBUILD_SHARED_LIBS=ON"
               "-DMANIFOLD_PAR=${if parallel then "ON" else "OFF"}"
             ];
+            doCheck = true;
```
elalish, line 109: build_mxe still looks like it's not testing - also build_mac_..._subdir, unless that's intentional?

elalish, line 109: And any idea why build_nix_tbb is so slow? It looks like it's building all of TBB before it even gets to Manifold...

pca006132: mxe cannot be tested: we are building some windows exe on linux. at least we cannot test it without another job on windows...

bulid_mac_..._subdir... is just to test the cmake consumer, see if we can build when put inside a subdirectory in a larger project.

pca006132: I added the latest tbb version for emscripten, forgot why I also use it for other native builds... fixed.

PR #1099, opened 2024-12-04
finer timing in ManifoldCAD.org

elalish:
I've been playing around with timing our mesh smoothing SDF vs increasing SDF resolution, but the timings were giving me odd results: see this ManifoldCAD [example](https://tinyurl.com/2226r5e4). Turns out most of the time was being spent in file export, not the actual Manifold library. So here I've split the timing into two parts: Manifold time and export time. 

However, during the process I found that GLB export is next to instant - the real time killer is 3MF export, particularly the zip. I'm thinking now it would be better to do the 3MF export on save rather than on run. I wonder if we can just pass the whole `to3mf` JSON object out of the worker? Makes the API a bit less clean, but...

### DIFF
--- a/bindings/wasm/examples/editor.js
+++ b/bindings/wasm/examples/editor.js
@@ -434,16 +434,11 @@ function disableCancel() {
   runButton.classList.remove('red', 'cancel');
 }
 
-let t0 = performance.now();
-
 function finishRun() {
   disableCancel();
-  const t1 = performance.now();
   const log = consoleElement.textContent;
   // Remove "Running..."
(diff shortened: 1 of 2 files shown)


pca006132:
Sorry, but I really want to make a meme for this:

image

elalish:
I guess I'll go ahead an merge this for now, though I think now the better course will be to revert this, have the worker only generate a GLB, then have the Save3MF button convert the GLB to a 3MF. This'll simplify our worker API too, which will make it nicer to put out as its own npm package. But that'll take longer, so it'll have to wait.

PR #1104, opened 2024-12-08
boolean debugging

pca006132:
1. Dump problematic boolean operands when we perform boolean operations.
2. Check for self-intersections after boolean operations.

I don't think this can be merged now, because the self-intersection check causes quite a few failures. Some are probably fine (epsilon-valid because they are just touching), but some are probably not. Use the verbose option to print the information, because it can be quite long.

@elalish maybe you should have a look and play with this.


src/csg_tree.cpp
```diff
@@ -109,9 +109,51 @@ std::shared_ptr<CsgNode> CsgLeafNode::Transform(const mat3x4 &m) const {
...
+    dump_lock.lock();
+    std::cout << "self-intersection detected" << std::endl;
+    dump_lock.unlock();
+    throw logicErr("self intersection detected");
+  }
+#endif
   return std::make_shared<CsgLeafNode>(std::make_shared<Manifold::Impl>(impl));
 }
 
+std::shared_ptr<CsgLeafNode> SimpleBoolean(const Manifold::Impl &a,
```
pca006132, line 123: This is just a helper function to avoid duplicating the debug code everywhere.

src/csg_tree.cpp
```diff
@@ -109,9 +109,51 @@ std::shared_ptr<CsgNode> CsgLeafNode::Transform(const mat3x4 &m) const {
 CsgNodeType CsgLeafNode::GetNodeType() const { return CsgNodeType::Leaf; }
 
 std::shared_ptr<CsgLeafNode> ImplToLeaf(Manifold::Impl &&impl) {
+#ifdef MANIFOLD_DEBUG
+  if (impl.IsSelfIntersecting()) {
+    dump_lock.lock();
+    std::cout << "self-intersection detected" << std::endl;
```
pca006132, line 115: Should probably clean up these error messages later.

src/impl.cpp
```diff
@@ -686,4 +688,76 @@ SparseIndices Manifold::Impl::VertexCollisionsZ(VecView<const vec3> vertsIn,
     return collider_.Collisions<false, false>(vertsIn);
 }
 
+#ifdef MANIFOLD_DEBUG
+std::ostream& operator<<(std::ostream& stream, const Manifold::Impl& impl) {
+  stream << std::setprecision(17);  // for double precision
+  stream << "# ======= begin mesh ======" << std::endl;
+  stream << "# tolerance = " << impl.tolerance_ << std::endl;
```
pca006132, line 695: The idea is that things that are unique to `MeshGL(64)` and not defined in the `.obj` file format are defined using comments.

src/impl.cpp
```diff
@@ -686,4 +688,76 @@ SparseIndices Manifold::Impl::VertexCollisionsZ(VecView<const vec3> vertsIn,
...
+    stream << "f " << impl.halfedge_[i].startVert + 1 << " "
+           << impl.halfedge_[i + 1].startVert + 1 << " "
+           << impl.halfedge_[i + 2].startVert + 1 << std::endl;
+  stream << "# ======== end mesh =======" << std::endl;
+  return stream;
+}
+#endif
+
+#ifdef MANIFOLD_EXPORT
+MeshGL64 ImportMeshGL64(std::istream& stream) {
```
pca006132, line 710: This function imports the `.obj` file. Note that this is a quick-and-dirty implementation that doesn't really support the actual specification, but good enough that it can parse our outputs. I do not plan to make it spec compliant, nor make the format stable, so this will probably keep internal to us.

src/properties.cpp
```diff
@@ -202,6 +202,55 @@ bool Manifold::Impl::Is2Manifold() const {
...
+    for (int i : {0, 1, 2}) {
+      tri_x[i] = vertPos_[halfedge_[3 * x + i].startVert];
+      tri_y[i] = vertPos_[halfedge_[3 * y + i].startVert];
+    }
+    // if triangles x and y share a vertex, return true to skip the
+    // check. we relax the sharing criteria a bit to allow for at most
+    // distance epsilon squared
+    for (int i : {0, 1, 2})
+      for (int j : {0, 1, 2})
+        if (linalg::distance2(tri_x[i], tri_y[j]) <= epsilonSq) return true;
```
pca006132, line 234: Note that it is not topological here, we allow pairs sharing a common edge to pass the test, because otherwise there will be some meshes that we want to support but can never pass this test...

test/boolean_complex_test.cpp
```diff
@@ -1016,4 +1020,26 @@ TEST(BooleanComplex, SimpleOffset) {
   EXPECT_EQ(c.Status(), Manifold::Error::NoError);
 }
 
+TEST(BooleanComplex, UnionDifferenceSelfIntersect) {
```
pca006132, line 1023: This is a simple failure case, and a test case to exercise the parser as well.

src/csg_tree.cpp
```diff
@@ -109,9 +109,51 @@ std::shared_ptr<CsgNode> CsgLeafNode::Transform(const mat3x4 &m) const {
 CsgNodeType CsgLeafNode::GetNodeType() const { return CsgNodeType::Leaf; }
 
 std::shared_ptr<CsgLeafNode> ImplToLeaf(Manifold::Impl &&impl) {
+#ifdef MANIFOLD_DEBUG
+  if (impl.IsSelfIntersecting()) {
```
elalish, line 113: Maybe also put this behind the `intermediate_checks` flag? I don't think there's a safe way to test this in general.

src/csg_tree.cpp
pca006132: Ah make sense, forgot to remove that.

pca006132:
> So the big question: does this successfully spit out the Boolean that leads to the first self-intersection in our problem Minkowski cases? If so, this will be super helpful. If not, we probably need to rethink it.

This seems to work for the `BooleanComplex.SimpleOffset` case, but you need to disable the status check for it to output a pair of smaller operands.

src/properties.cpp
```diff
@@ -202,6 +202,55 @@ bool Manifold::Impl::Is2Manifold() const {
...
+
+    if (DistanceTriangleTriangleSquared(tri_x, tri_y) == 0.0) {
```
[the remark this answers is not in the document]
pca006132: I think we can just try to move the triangle along the normal vector of the conflicting triangle by epsilon. Not sure if there is a good way to figure out the signed distance, but this can probably check if we can move by epsilon amount to make them disjoint?

pca006132:
Here are the failures we have when intermediate checks are enabled (via a new `-c` flag)

```
[ RUN      ] BooleanComplex.SelfIntersect
self intersections detected
LHS self-intersecting: 1
RHS self-intersecting: 1
unknown file: Failure
C++ exception with description "self intersection detected" thrown in the test body.

[  FAILED  ] BooleanComplex.SelfIntersect (39 ms)
[ RUN      ] BooleanComplex.GenericTwinBooleanTest7081
self intersections detected
LHS self-intersecting: 1
RHS self-intersecting: 0
unknown file: Failure
C++ exception with description "self intersection detected" thrown in the test body.

[  FAILED  ] BooleanComplex.GenericTwinBooleanTest7081 (6798 ms)
[ RUN      ] BooleanComplex.GenericTwinBooleanTest7863
self intersections detected
LHS self-intersecting: 0
RHS self-intersecting: 1
unknown file: Failure
C++ exception with description "self intersection detected" thrown in the test body.

[  FAILED  ] BooleanComplex.GenericTwinBooleanTest7863 (8 ms)
[ RUN      ] BooleanComplex.Havocglass8Bool
[       OK ] BooleanComplex.Havocglass8Bool (2 ms)
[ RUN      ] BooleanComplex.CraycloudBool
[       OK ] BooleanComplex.CraycloudBool (1 ms)
[ RUN      ] BooleanComplex.HullMask
self intersections detected
LHS self-intersecting: 0
RHS self-intersecting: 0
unknown file: Failure
C++ exception with description "self intersection detected" thrown in the test body.

[  FAILED  ] BooleanComplex.HullMask (29 ms)
[ RUN      ] BooleanComplex.SimpleOffset
self intersections detected
LHS self-intersecting: 0
RHS self-intersecting: 0
unknown file: Failure
C++ exception with description "self intersection detected" thrown in the test body.

[  FAILED  ] BooleanComplex.SimpleOffset (1330 ms)
[ RUN      ] BooleanComplex.OffsetTriangulationFailure
-----------------------------------
Triangulation failed! Precision = 2.01711e-08
Error in file: /home/pca006132/code/manifold/src/polygon.cpp (118): 'std::all_of(triangles.begin(), triangles.end(), [&vertPos, epsilon](const ivec3 &tri) { return CCW(vertPos[tri[0]], vertPos[tri[1]], vertPos[tri[2]], epsilon) >= 0; })' is false: triangulation is not entirely CCW!
Polygon 0 2.01711e-08 2
11
5001.56 -19653.4
5002.53 -19653.5
5001.56 -19653.4
5001.56 -19653.4
5001.56 -19653.4
5001.56 -19653.4
5001.56 -19653.4
5001.56 -19653.4
5002.53 -19653.5
5002.53 -19653.5
5001.56 -19653.4
12
5003.5 -19653.6
5039.82 -19373.5
5039.82 -19373.5
5039.82 -19373.5
5039.82 -19373.5
5039.82 -19373.5
5039.82 -19373.5
5039.82 -19373.5
5039.82 -19373.5
5039.82 -19373.5
5001.56 -19653.4
5002.53 -19653.5
# ... 
show(array([
  [5001.56, -19653.4],
  [5002.53, -19653.5],
  [5001.56, -19653.4],
  [5001.56, -19653.4],
  [5001.56, -19653.4],
  [5001.56, -19653.4],
  [5001.56, -19653.4],
  [5001.56, -19653.4],
  [5002.53, -19653.5],
  [5002.53, -19653.5],
  [5001.56, -19653.4],
]))
show(array([
  [5003.5, -19653.6],
  [5039.82, -19373.5],
  [5039.82, -19373.5],
  [5039.82, -19373.5],
  [5039.82, -19373.5],
  [5039.82, -19373.5],
  [5039.82, -19373.5],
  [5039.82, -19373.5],
  [5039.82, -19373.5],
  [5039.82, -19373.5],
  [5001.56, -19653.4],
  [5002.53, -19653.5],
]))
produced this triangulation:
4661, 2943, 2993
3003, 4666, 4658
4658, 4659, 2172
2993, 2995, 2999
2999, 3001, 3003
2999, 3003, 4658
2172, 4661, 2993
2172, 2993, 2999
2172, 2999, 4658
2989, 4657, 5071
2987, 2989, 5071
2985, 2987, 5071
2983, 2985, 5071
2991, 2983, 5071
5070, 2173, 2453
2991, 5071, 5070
2453, 5767, 3084
5070, 2453, 3084
2991, 5070, 3084
LHS self-intersecting: 0
RHS self-intersecting: 0
unknown file: Failure
C++ exception with description "Error in file: /home/pca006132/code/manifold/src/polygon.cpp (118): 'std::all_of(triangles.begin(), triangles.end(), [&vertPos, epsilon](const ivec3 &tri) { return CCW(vertPos[tri[0]], vertPos[tri[1]], vertPos[tri[2]], epsilon) >= 0; })' is false: triangulation is not entirely CCW!" thrown in the test body.

[  FAILED  ] BooleanComplex.OffsetTriangulationFailure (15 ms)
```

It now checks if the two operands are self-intersecting up to epsilon.

pca006132:
The main issue is that the operands causing these issues are typically quite large, may require some randomized algorithm to reduce their sizes.

pca006132:
Managed to obtain a smaller self-intersecting result case with openscad. The two operands are fine, but the result is not.

Btw, I found that the problem with verbose output is that we also print running time and algorithm statistics with verbose enabled, and things are unreadable when run in multiple threads... Wonder if we should split that into some debug flag, as users should never use that verbose output (except during debugging).

pca006132:
It seems that the offending pairs look like this by peforming decomposition:

```diff
diff --git a/test/boolean_complex_test.cpp b/test/boolean_complex_test.cpp
index 2cef5f6..a16bd67 100644
--- a/test/boolean_complex_test.cpp
+++ b/test/boolean_complex_test.cpp
@@ -1045,7 +1045,7 @@ TEST(BooleanComplex, DISABLED_OffsetTriangulationFailure) {
   ManifoldParams().intermediateChecks = intermediateChecks;
 }
 
-TEST(BooleanComplex, DISABLED_OffsetSelfIntersect) {
+TEST(BooleanComplex, OffsetSelfIntersect) {
   const bool intermediateChecks = ManifoldParams().intermediateChecks;
   ManifoldParams().intermediateChecks = true;
   std::string file = __FILE__;
@@ -1063,10 +1063,17 @@ TEST(BooleanComplex, DISABLED_OffsetSelfIntersect) {
     std::cout << err.what() << std::endl;
     FAIL();
   }
-  Manifold x(a);
-  Manifold y(b);
-  Manifold result = x + y;
-  EXPECT_EQ(result.Status(), Manifold::Error::NoError);
+
+  auto xs = Manifold(a).Decompose();
+  auto ys = Manifold(b).Decompose();
+
+  for (const auto &x : xs) {
+    for (const auto &y : ys) {
+      Manifold result = x + y;
+      EXPECT_EQ(result.Status(), Manifold::Error::NoError);
+    }
+  }
+
   ManifoldParams().intermediateChecks = intermediateChecks;
 }
 
```

image

But for some reason after extracting the mesh, I cannot reproduce the issue. Perhaps this is related to rounding error or face normals?

pca006132:
It can now be reproduced after allowing setting epsilon.

test/boolean_complex_test.cpp
```diff
@@ -1020,26 +1019,53 @@ TEST(BooleanComplex, SimpleOffset) {
...
+  } catch (std::exception& err) {
+    std::cout << err.what() << std::endl;
+    FAIL();
+  }
+
+  Manifold result = a + b;
+  EXPECT_EQ(result.Status(), Manifold::Error::NoError);
+  ManifoldParams().intermediateChecks = intermediateChecks;
+
+  ManifoldParams().intermediateChecks = intermediateChecks;
```
elalish, line 1068: duplicate

src/properties.cpp
```diff
@@ -231,9 +231,20 @@ bool Manifold::Impl::IsSelfIntersecting() const {
...
+      // try to move the triangles around the normal of the other face
+      std::array<vec3, 3> tmp_x, tmp_y;
+      for (int i : {0, 1, 2}) tmp_x[i] = tri_x[i] + epsilon_ * faceNormal_[y];
+      if (DistanceTriangleTriangleSquared(tmp_x, tri_y) > 0.0) return true;
+      for (int i : {0, 1, 2}) tmp_x[i] = tri_x[i] - epsilon_ * faceNormal_[y];
+      if (DistanceTriangleTriangleSquared(tmp_x, tri_y) > 0.0) return true;
+      for (int i : {0, 1, 2}) tmp_y[i] = tri_y[i] + epsilon_ * faceNormal_[x];
+      if (DistanceTriangleTriangleSquared(tri_x, tmp_y) > 0.0) return true;
+      for (int i : {0, 1, 2}) tmp_y[i] = tri_y[i] - epsilon_ * faceNormal_[x];
+      if (DistanceTriangleTriangleSquared(tri_x, tmp_y) > 0.0) return true;
```
elalish, line 246: 👍

test/boolean_complex_test.cpp
```diff
@@ -1016,4 +1019,53 @@ TEST(BooleanComplex, SimpleOffset) {
...
+  try {
+    f.open(dir + "/models/Offset1.obj");
+    a = Manifold::ImportMeshGL64(f);
+    f.close();
+    f.open(dir + "/models/Offset2.obj");
+    b = Manifold::ImportMeshGL64(f);
+    f.close();
+  } catch (std::exception& err) {
+    std::cout << err.what() << std::endl;
+    FAIL();
```
elalish, line 1038: Why do we need a try catch here?

elalish (approved):
Looks good, thanks!

PR #1110, opened 2024-12-13
Fixed unreferenced vert seg fault

### LINKED ISSUE #1109: SIGSEGV in manifold.RefineToLength() after mesh.Merge() with tolerance
Load attached model via assimp, then got SIGSEV at RefineToLength in folowing code sequence:
    mesh.tolerance = 1e-5;
    mesh.Merge();
    manifold::Manifold manifold(mesh);
    manifold = manifold.RefineToLength(1.0);

It works without Merge(). It also work for tolerances below 1e-5 (e.g. 1e-6). But fails for tolerances >= 1e-5. I believe when Merge() begins to actually merge something.

Linux. Library from today (13.12.2024), build with these options:
cmake -DCMAKE_BUILD_TYPE=Release -DBUILD_SHARED_LIBS=OFF -DMANIFOLD_PAR=OFF -DMANIFOLD_CROSS_SECTION=OFF .. && make

Please find attached an archive with problematic model in .obj format.
[B2179.zip](https://github.com/user-attachments/files/18124557/B2179.zip)

elalish:
Fixes #1109 

I'm not getting a repro locally on my Macbook - let's see if the CI can reproduce.


elalish:
Interesting - I updated the test and it still passes locally (and the mac CI), but the CI does show the seg fault now on linux. @pca006132 can you repro locally? Any idea what's going on?

pca006132:
Yes, it says index out of range in subdivision.cpp:670.

elalish:
Okay, I got a repro using address sanitizer. The problem was that `SimplifyTopology` after `Merge` was resulting in a couple of unreferenced verts, which messed up the indexing in `Subdivide`.

src/sdf.cpp
```diff
@@ -525,7 +525,9 @@ Manifold Manifold::LevelSet(std::function<double(vec3)> sdf, Box bounds,
   triVerts.resize(index[0]);
 
   pImpl_->CreateHalfedges(triVerts);
+  pImpl_->RemoveUnreferencedVerts();
   pImpl_->CleanupTopology();
+  pImpl_->RemoveUnreferencedVerts();
```
elalish, line 530: This seems wrong - having both of these was the only way I could fix the seg fault in `BooleanComplex.MeshRelation`, but it seems it is still happening on the CI anyway. I haven't been able to figure out what's going wrong here because even with address sanitizer I can't make it fail when run alone - only when the whole test suite is run, which doesn't really work for debug. 

Any thoughts @pca006132? I'm beginning to think we should get a bit more crisp about where unreferenced verts can get created and where they get removed. But maybe there's some other problem...

pca006132: No idea why it only fails when we run the whole test suite. I think we should probably have some validation code that track this unreferenced vertex information. Should be more reliable than asan.

src/impl.cpp
```diff
@@ -448,6 +431,8 @@ void Manifold::Impl::CreateHalfedges(const Vec<ivec3>& triVerts) {
       if (h0.startVert != h1.endVert || h0.endVert != h1.startVert) break;
       if (halfedge_[NextHalfedge(pair0)].endVert ==
           halfedge_[NextHalfedge(pair1)].endVert) {
+        h0 = {-1, -1, -1};
+        h1 = {-1, -1, -1};
```
elalish, line 435: I think this got lost accidentally during the `face` removal - it meant opposed triangles were not actually getting removed. I randomly noticed this on an SDF a couple weeks ago, so that should be fixed now too.

src/sdf.cpp
elalish: Turns out we ran into exactly this problem before, in #894. I forgot that address sanitizer doesn't check for uninitialized memory.

src/impl.h
```diff
@@ -215,8 +211,14 @@ struct Manifold::Impl {
     CreateFaces();
 
     SimplifyTopology();
+    RemoveUnreferencedVerts();
```
elalish, line 214: This was the fix for the reported bug: it seems that in some cases `SimplifyTopology` strands unreferenced verts that it does not set to NAN. Ideally it shouldn't do that - probably worth looking into in the future; it needs an overhaul anyway.

PR #1112, opened 2024-12-16
update fuzztest

pca006132:
It seems that it now builds fine with the latest version of fuzztest.


test/manifold_fuzz.cpp
```diff
@@ -0,0 +1,99 @@
...
+    ElementOf({Transform::Translate, Transform::Rotate, Transform::Scale}),
+    Vec3Domain);
+auto CsgDomain = VectorOf(PairOf(VectorOf(TransformDomain).WithMaxSize(20),
+                                 ElementOf({0, 1})))
+                     .WithMaxSize(100);
+
+void SimpleCube(
+    const std::vector<std::pair<
+        std::vector<std::pair<Transform, std::array<double, 3>>>, int>>
+        &inputs) {
```
elalish, line 39: This is a bit hard to read - maybe a vector of structs?

test/manifold_fuzz.cpp
```diff
@@ -0,0 +1,99 @@
...
+          break;
+        case Transform::Rotate:
+          cube = cube.Rotate(std::get<0>(transform.second),
+                             std::get<1>(transform.second),
+                             std::get<2>(transform.second));
+          break;
+        case Transform::Scale:
+          cube = cube.Scale({std::get<0>(transform.second),
+                             std::get<1>(transform.second),
+                             std::get<2>(transform.second)});
```
elalish, line 60: Might this be simpler if we just use `cube.Transform`?

test/manifold_fuzz.cpp
```diff
@@ -0,0 +1,99 @@
...
+      printf("timeout after %dms...\n", 10000);
+      pthread_cancel(tid.load());
+    }
+
+    EXPECT_FALSE(faulted.load());
+    if (faulted.load()) break;
+  }
+}
+
+FUZZ_TEST(ManifoldFuzz, SimpleCube).WithDomains(CsgDomain);
```
elalish, line 89: Does this use the same random seed every time or different? If it fails, will it spit out all the info we need to create a deterministic test?

test/CMakeLists.txt
```diff
@@ -69,4 +69,9 @@ if(MANIFOLD_FUZZ)
   target_link_libraries(polygon_fuzz PUBLIC manifold)
   link_fuzztest(polygon_fuzz)
   gtest_discover_tests(polygon_fuzz)
+
+  add_executable(manifold_fuzz manifold_fuzz.cpp)
+  target_link_libraries(manifold_fuzz PUBLIC manifold)
+  link_fuzztest(manifold_fuzz)
+  gtest_discover_tests(manifold_fuzz)
```
elalish, line 76: Are we going to enable `MANIFOLD_FUZZ` on the CI?

elalish (approved):
Thanks!

test/manifold_fuzz.cpp
pca006132: sure

test/manifold_fuzz.cpp
pca006132: I'm using three different transforms to try to get the fuzzer to produce something that is simpler. We can look into that later after the current issue is fixed, I guess.

test/manifold_fuzz.cpp
pca006132: It is different every time, but we can override that to replay the fuzzing process. It will spit out the info needed for the test when using the minimizer script, I will document it.

test/CMakeLists.txt
pca006132: No, this is something that runs indefinitely, and non-deterministically. So this is probably something we want to run in our own machines.

test/manifold_fuzz.cpp
```diff
@@ -89,11 +96,11 @@ void SimpleCube(
 FUZZ_TEST(ManifoldFuzz, SimpleCube).WithDomains(CsgDomain);
 
 TEST(ManifoldFuzz, SimpleCubeRegression) {
-  SimpleCube({{{{static_cast<Transform>(1),
+  SimpleCube({{{{static_cast<TransformType>(1),
                  {-0.10000000000000001, 0.10000000000000001, -1.}}},
                1},
               {{}, 1},
-              {{{static_cast<Transform>(1),
+              {{{static_cast<TransformType>(1),
```
elalish, line 103: I still can't read this. Let's use `true` and `false` for booleans instead of 0 and 1. And what does `static_cast<TransformType>(1)` mean? Can we somehow get these to say `Translation`, `Rotation`, etc?

test/boolean_test.cpp
```diff
@@ -421,3 +421,15 @@ TEST(Boolean, Precision2) {
   cube2 = cube2.Translate(vec3(scale * kPrecision));
   EXPECT_FALSE((cube ^ cube2).IsEmpty());
 }
+
+TEST(Boolean, SimpleCubeRegression) {
```
pca006132, line 425: @elalish it is now here.

PR #1115, opened 2024-12-18
Allow Clipper2 to be supplied externally when using Manifold as a submodule

kintel:
* Checks `Clipper2_FOUND` before attempting to find Clipper2
* Add imported Clipper2 library _only_ when trying pkg_config

### DIFF
--- a/cmake/manifoldDeps.cmake
+++ b/cmake/manifoldDeps.cmake
@@ -84,20 +84,7 @@ if(MANIFOLD_CROSS_SECTION)
       pkg_check_modules(Clipper2 Clipper2)
     endif()
   endif()
-  if(Clipper2_FOUND)
-    add_library(Clipper2 SHARED IMPORTED)
-    set_property(
-      TARGET Clipper2
-      PROPERTY IMPORTED_LOCATION ${Clipper2_LINK_LIBRARIES}
-    )
-    if(WIN32)
-      set_property(
-        TARGET Clipper2
-        PROPERTY IMPORTED_IMPLIB ${Clipper2_LINK_LIBRARIES}
-      )
-    endif()
-    target_include_directories(Clipper2 INTERFACE ${Clipper2_INCLUDE_DIRS})
-  else()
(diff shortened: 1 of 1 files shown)


cmake/manifoldDeps.cmake
```diff
@@ -78,26 +78,27 @@ endif()
...
-    )
-    if(WIN32)
-      set_property(
-        TARGET Clipper2
-        PROPERTY IMPORTED_IMPLIB ${Clipper2_LINK_LIBRARIES}
-      )
-    endif()
-    target_include_directories(Clipper2 INTERFACE ${Clipper2_INCLUDE_DIRS})
-  else()
+      if(Clipper2_FOUND)
```
elalish, line 85: Can this condition ever be hit? Two lines up we say that clipper2 was not found.

pca006132: pkg-config can set this to true after the call to `pkg_check_modules`.

elalish: Ah, got it, thanks.

elalish:
What are you waiting for to make this not a Draft?

kintel:
> What are you waiting for to make this not a Draft?

My own manual testing across platforms, which takes some time. I mostly wanted to make sure this is conceptually ok with you, plus make sure Manifold's tests pass.

kintel:
This is good to go now!

pca006132:
LGTM. I'm wondering if we should do something similar for other dependencies, e.g. tbb.

kintel:
Yeah, it could make sense to align the deps lookups. I didn't run into the same issue with tbb as we're always supplying it as a system dependency from the OpenSCAD side.

kintel:
Another idea would be to hide all the details in custom Find macros, but I didn't want to touch too much as it's so time consuming to test it..

pca006132 (approved):

elalish (approved):
Thanks!

PR #1121, opened 2024-12-23
setup dependabot

pca006132:
Setup dependabot so we can update nanobind as soon as a new version is released.

Wonder if there is a way to check if the config is correct...

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,6 @@
+version: 2
+updates:
+  - package-ecosystem: "pip"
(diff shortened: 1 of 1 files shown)


elalish (approved):
So this will make PRs to update python deps?

pca006132:
> So this will make PRs to update python deps?

Yes. I'm thinking if we should do the same to npm packages as well. In fact, I think github security is complaining about some dependencies being outdated with vulnerabilities (but doesn't matter too much for us because that is just a static webpage)

PR #1123, opened 2024-12-24
CI updates

pca006132:
1. Use consistent names for all CI jobs.
2. Removed standalone C binding test (as it is tested everywhere already).
3. Build `MANIFOLD_DEBUG=ON` variant for wasm as well (not uploaded, just for tests)
4. Check for windows dll location when `BUILD_SHARED_LIBS=ON`.


pca006132:
@elalish this should be included in the patch release due to fixing msvc dll stuff.

elalish (approved):
Thanks! Yes, let's go ahead and get this in.

PR #1124, opened 2024-12-24
wasm: update package-lock

pca006132:
Seems to please dependabot.

### DIFF
--- a/bindings/wasm/examples/worker.test.ts
+++ b/bindings/wasm/examples/worker.test.ts
@@ -13,7 +13,7 @@
 // limitations under the License.
 
 import {WebIO} from '@gltf-transform/core';
-import assert from 'node:assert';
+import {strict as assert} from 'assert';
 import {afterEach, expect, suite, test} from 'vitest';
 
 import {readMesh, setupIO} from './gltf-io';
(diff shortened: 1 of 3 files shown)


bindings/wasm/examples/worker.test.ts
```diff
@@ -55,71 +55,71 @@ afterEach(async () => {await new Promise(resolve => setTimeout(resolve, 500))})
 suite('Examples', () => {
   test('Intro', async () => {
     const result = await runExample('Intro');
-    expect(result.genus).to.equal(5, 'Genus');
-    expect(result.volume).to.be.closeTo(203164, 1, 'Volume');
-    expect(result.surfaceArea).to.be.closeTo(62046, 1, 'Surface Area');
+    expect(result?.genus).to.equal(5, 'Genus');
```
elalish, line 58: Why all the ?s - can result actually be null?

elalish (approved):

bindings/wasm/examples/worker.test.ts
pca006132: Yes: https://github.com/elalish/manifold/blob/209350fd1ee35749fecf2f6bd58a3e96cf8f0291/bindings/wasm/examples/worker.ts#L732

PR #1127, opened 2024-12-27
csg_tree: avoid recursive drop

### LINKED ISSUE #1125: avoid recursive drop for csg tree
Thinking about it, I forgot to do this. Perhaps this is causing wasm build to behave weirdly when tbb is enabled.

Should be a easy fix.

pca006132:
Fixes #1125.


elalish (approved):
Thanks!

PR #1129, opened 2024-12-30
Add const keyword to CrossSection::Compose parameter

tomasf:
Add a missing `const` to the `crossSections` parameter of`CrossSection::Compose`.


elalish (approved):
Thanks!

PR #1143, opened 2025-02-12
Fix misc

pca006132:
While debugging #1040 I took the time to fix a bunch of `-Wextra` warnings.

### DIFF
--- a/include/manifold/vec_view.h
+++ b/include/manifold/vec_view.h
@@ -106,8 +106,7 @@ class VecView {
                   size_t length = std::numeric_limits<size_t>::max()) {
     if (length == std::numeric_limits<size_t>::max())
       length = this->size_ - offset;
-    ASSERT(length >= 0, std::out_of_range("Vec::view out of range"));
-    ASSERT(offset + length <= this->size_ && offset >= 0,
+    ASSERT(offset + length <= this->size_,
            std::out_of_range("Vec::view out of range"));
     return VecView<T>(this->ptr_ + offset, length);
   }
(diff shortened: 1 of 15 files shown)


src/iters.h
```diff
@@ -22,7 +23,7 @@ template <typename F, typename Iter>
 struct TransformIterator {
  private:
   Iter iter;
-  F f;
+  std::optional<F> f;
```
pca006132, line 26: This is optional so we can use the `emplace` function to avoid calling the copy assignment operator of `f`, which doesn't exist for lambdas. They only have copy constructors.

include/manifold/vec_view.h
```diff
@@ -106,8 +106,7 @@ class VecView {
                   size_t length = std::numeric_limits<size_t>::max()) {
     if (length == std::numeric_limits<size_t>::max())
       length = this->size_ - offset;
-    ASSERT(length >= 0, std::out_of_range("Vec::view out of range"));
```
pca006132, line 109: Because `length` is unsigned, it cannot be smaller than 0.

src/boolean_result.cpp
```diff
@@ -359,8 +359,8 @@ void AppendPartialEdges(Manifold::Impl &outR, Vec<char> &wholeHalfedgeP,
     // reference is now to the endVert instead of the startVert, which is one
     // position advanced CCW. This is only valid if this is a retained vert; it
     // will be ignored later if the vert is new.
-    const TriRef forwardRef = {forward ? 0 : 1, -1, faceLeftP};
-    const TriRef backwardRef = {forward ? 0 : 1, -1, faceRightP};
+    const TriRef forwardRef = {forward ? 0 : 1, -1, faceLeftP, -1};
```
pca006132, line 362: The `faceID` was never initialized here, so I set it to -1. I think this is fine because we should copy the face IDs later?

elalish (approved):
Thanks for the cleanup!

PR #1147, opened 2025-02-17
improved decimation

### LINKED ISSUE #1017: better handling of tolerance for simplification
Basically things left from #1007

elalish:
Fixes #1017

Added `Simplify()` which is a lot like `SetTolerance()`, but with a clearer name that matches our `CrossSection` API. 

We now have the beginnings of a halfway-decent decimator; in 3 seconds we can take a sphere with 8 million triangles and reduce it to under 2,500, looking like this:
image

I've also updated our edge collapse to only operate on intersected edges when it runs as part of the Boolean cleanup, which has been requested by several people, and should increase performance a bit too. I tried to apply that logic to the edge swapping as well, but it broke a few tests, so I'll come back to that.

### DIFF
--- a/src/edge_op.cpp
+++ b/src/edge_op.cpp
@@ -48,11 +48,10 @@ struct ShortEdge {
   VecView<const Halfedge> halfedge;
   VecView<const vec3> vertPos;
   const double tolerance;
-  const size_t firstNewVert;
 
   bool operator()(int edge) const {
     const Halfedge& half = halfedge[edge];
-    if (half.pairedHalfedge < 0 || half.startVert < firstNewVert) return false;
+    if (half.pairedHalfedge < 0) return false;
     // Flag short edges
     const vec3 delta = vertPos[half.endVert] - vertPos[half.startVert];
     return la::dot(delta, delta) < tolerance * tolerance;
@@ -322,7 +321,7 @@ void Manifold::Impl::SimplifyTopology(size_t firstNewVert) {
   {
     ZoneScopedN("CollapseShortEdge");
     numFlagged = 0;
-    ShortEdge se{halfedge_, vertPos_, epsilon_, firstNewVert};
+    ShortEdge se{halfedge_, vertPos_, epsilon_};
     s.run(nbEdges, se, [&](size_t i) {
       const bool didCollapse = CollapseEdge(i, scratchBuffer);
       if (didCollapse) numFlagged++;
--- a/src/edge_op.cpp
+++ b/src/edge_op.cpp
@@ -48,10 +48,13 @@ struct ShortEdge {
   VecView<const Halfedge> halfedge;
   VecView<const vec3> vertPos;
   const double tolerance;
+  const int firstNewVert;
 
   bool operator()(int edge) const {
     const Halfedge& half = halfedge[edge];
-    if (half.pairedHalfedge < 0) return false;
+    if (half.pairedHalfedge < 0 ||
+        (half.startVert < firstNewVert && half.endVert < firstNewVert))
+      return false;
     // Flag short edges
     const vec3 delta = vertPos[half.endVert] - vertPos[half.startVert];
     return la::dot(delta, delta) < tolerance * tolerance;
@@ -94,10 +97,15 @@ struct SwappableEdge {
   VecView<const vec3> vertPos;
   VecView<const vec3> triNormal;
   const double tolerance;
+  const int firstNewVert;
 
   bool operator()(int edge) const {
     const Halfedge& half = halfedge[edge];
     if (half.pairedHalfedge < 0) return false;
+    if (half.startVert < firstNewVert && half.endVert < firstNewVert &&
+        halfedge[NextHalfedge(edge)].endVert < firstNewVert &&
+        halfedge[NextHalfedge(half.pairedHalfedge)].endVert < firstNewVert)
+      return false;
 
     int tri = edge / 3;
     ivec3 triEdge = TriOf(edge);
@@ -318,7 +326,7 @@ void Manifold::Impl::SimplifyTopology(int firstNewVert) {
   {
     ZoneScopedN("CollapseShortEdge");
     numFlagged = 0;
-    ShortEdge se{halfedge_, vertPos_, epsilon_};
+    ShortEdge se{halfedge_, vertPos_, epsilon_, firstNewVert};
     s.run(nbEdges, se, [&](size_t i) {
       const bool didCollapse = CollapseEdge(i, scratchBuffer);
       if (didCollapse) numFlagged++;
@@ -353,7 +361,8 @@ void Manifold::Impl::SimplifyTopology(int firstNewVert) {
   {
     ZoneScopedN("RecursiveEdgeSwap");
     numFlagged = 0;
-    SwappableEdge se{halfedge_, vertPos_, faceNormal_, tolerance_};
+    SwappableEdge se{halfedge_, vertPos_, faceNormal_, tolerance_,
+                     firstNewVert};
     std::vector<int> edgeSwapStack;
     std::vector<int> visited(halfedge_.size(), -1);
     int tag = 0;
--- a/src/edge_op.cpp
+++ b/src/edge_op.cpp
@@ -61,7 +61,7 @@ struct ShortEdge {
 struct FlagEdge {
   VecView<const Halfedge> halfedge;
   VecView<const TriRef> triRef;
-  const size_t firstNewVert;
+  const int firstNewVert;
 
   bool operator()(int edge) const {
     const Halfedge& half = halfedge[edge];
@@ -301,7 +301,7 @@ void Manifold::Impl::CleanupTopology() {
  * Rather than actually removing the edges, this step merely marks them for
  * removal, by setting vertPos to NaN and halfedge to {-1, -1, -1, -1}.
  */
-void Manifold::Impl::SimplifyTopology(size_t firstNewVert) {
+void Manifold::Impl::SimplifyTopology(int firstNewVert) {
   if (!halfedge_.size()) return;
 
   CleanupTopology();
--- a/src/edge_op.cpp
+++ b/src/edge_op.cpp
@@ -317,36 +317,35 @@ void Manifold::Impl::SimplifyTopology() {
     numFlagged = 0;
     ShortEdge se{halfedge_, vertPos_, epsilon_};
     s.run(nbEdges, se, [&](size_t i) {
-      CollapseEdge(i, scratchBuffer);
+      const bool didCollapse = CollapseEdge(i, scratchBuffer);
+      if (didCollapse) numFlagged++;
       scratchBuffer.resize(0);
-      numFlagged++;
     });
-  }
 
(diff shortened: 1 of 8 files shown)


test/boolean_test.cpp
```diff
@@ -138,6 +138,14 @@ TEST(Boolean, Cubes) {
 #endif
 }
 
+TEST(Boolean, Simplify) {
+  Manifold cube = Manifold::Cube().Refine(10);
+  Manifold result = cube + cube.Translate({1, 0, 0});
+  EXPECT_EQ(result.NumTri(), 2000);
```
elalish, line 144: I need to figure out why I'm not getting the number of faces I expect here, but it's late, so I'll leave this for tonight.

src/manifold.cpp
```diff
@@ -415,6 +415,29 @@ Manifold Manifold::SetTolerance(double tolerance) const {
...
+    impl->tolerance_ = tolerance;
+    impl->CreateFaces();
+    impl->SimplifyTopology();
+    impl->Finish();
+  }
+  impl->tolerance_ = oldTolerance;
+  return Manifold(impl);
+  // const double oldTolerance = GetCsgLeafNode().GetImpl()->tolerance_;
+  // Manifold result = SetTolerance(tolerance);
+  // result.GetCsgLeafNode().GetImpl()->tolerance_ = oldTolerance;
```
elalish, line 437: @pca006132 What do you think is the right pattern to reduce duplication between these functions? The one I've commented out here didn't compile.

test/properties_test.cpp
```diff
@@ -77,14 +77,14 @@ TEST(Properties, Tolerance) {
 }
 
 TEST(Properties, ToleranceSphere) {
-  const int n = 100;
+  const int n = 1000;
```
elalish, line 80: This used to fail dramatically, but now works well.

src/manifold.cpp
elalish: eh, I guess this is fine - only a few lines duplicated really.

src/edge_op.cpp
```diff
@@ -567,7 +570,7 @@ void Manifold::Impl::CollapseEdge(const int edge, std::vector<int>& edges) {
   const vec3 pNew = vertPos_[endVert];
   const vec3 pOld = vertPos_[toRemove.startVert];
   const vec3 delta = pNew - pOld;
-  const bool shortEdge = la::dot(delta, delta) < tolerance_ * tolerance_;
+  const bool shortEdge = la::dot(delta, delta) < epsilon_ * epsilon_;
```
pca006132, line 573: Should this be epsilon? I thought we want to remove edges smaller than tolerance (tolerance is the shortest edges that the user want).

src/edge_op.cpp
```diff
@@ -317,36 +320,35 @@ void Manifold::Impl::SimplifyTopology() {
     numFlagged = 0;
     ShortEdge se{halfedge_, vertPos_, epsilon_};
```
pca006132, line 321: Should this `ShortEdge` check limit to newly added edges? Adding the check to this seems fine.

src/edge_op.cpp
```diff
@@ -317,36 +320,35 @@ void Manifold::Impl::SimplifyTopology() {
...
-              << std::endl;
-  }
+    if (ManifoldParams().verbose && numFlagged > 0) {
+      std::cout << "collapsed " << numFlagged << " colinear edges" << std::endl;
+    }
 #endif
+  }
 
   {
     ZoneScopedN("RecursiveEdgeSwap");
```
pca006132, line 354: Edge swap as well: Should we swap old edges? Adding the check makes 4 tests fail with too many degenerate triangles, wondering why.

pca006132: OK I guess I know why, should check for both candidates. Will try that later...

pca006132:
```diff
diff --git a/src/edge_op.cpp b/src/edge_op.cpp
index 6ed9e1fd..b5a5f065 100644
--- a/src/edge_op.cpp
+++ b/src/edge_op.cpp
@@ -48,10 +48,13 @@ struct ShortEdge {
   VecView<const Halfedge> halfedge;
   VecView<const vec3> vertPos;
   const double tolerance;
+  const int firstNewVert;
 
   bool operator()(int edge) const {
     const Halfedge& half = halfedge[edge];
-    if (half.pairedHalfedge < 0) return false;
+    if (half.pairedHalfedge < 0 ||
+        (half.startVert < firstNewVert && half.endVert < firstNewVert))
+      return false;
     // Flag short edges
     const vec3 delta = vertPos[half.endVert] - vertPos[half.startVert];
     return la::dot(delta, delta) < tolerance * tolerance;
@@ -94,10 +97,15 @@ struct SwappableEdge {
   VecView<const vec3> vertPos;
   VecView<const vec3> triNormal;
   const double tolerance;
+  const int firstNewVert;
 
   bool operator()(int edge) const {
     const Halfedge& half = halfedge[edge];
     if (half.pairedHalfedge < 0) return false;
+    if (half.startVert < firstNewVert && half.endVert < firstNewVert &&
+        halfedge[NextHalfedge(edge)].endVert < firstNewVert &&
+        halfedge[NextHalfedge(half.pairedHalfedge)].endVert < firstNewVert)
+      return false;
 
     int tri = edge / 3;
     ivec3 triEdge = TriOf(edge);
@@ -318,7 +326,7 @@ void Manifold::Impl::SimplifyTopology(int firstNewVert) {
   {
     ZoneScopedN("CollapseShortEdge");
     numFlagged = 0;
-    ShortEdge se{halfedge_, vertPos_, epsilon_};
+    ShortEdge se{halfedge_, vertPos_, epsilon_, firstNewVert};
     s.run(nbEdges, se, [&](size_t i) {
       const bool didCollapse = CollapseEdge(i, scratchBuffer);
       if (didCollapse) numFlagged++;
@@ -353,7 +361,8 @@ void Manifold::Impl::SimplifyTopology(int firstNewVert) {
   {
     ZoneScopedN("RecursiveEdgeSwap");
     numFlagged = 0;
-    SwappableEdge se{halfedge_, vertPos_, faceNormal_, tolerance_};
+    SwappableEdge se{halfedge_, vertPos_, faceNormal_, tolerance_,
+                     firstNewVert};
     std::vector<int> edgeSwapStack;
     std::vector<int> visited(halfedge_.size(), -1);
     int tag = 0;
```

pca006132:
OK parallelized pinched verts splitting:

```c++
  ZoneScoped;

#if MANIFOLD_PAR == 1
  if (halfedge_.size() > 1e5) {
    std::mutex mutex;
    std::vector<size_t> problematic;
    std::vector<size_t> largestEdge(NumVert(),
                                    std::numeric_limits<size_t>::max());
    auto halfedgeCount = halfedge_.size();
    tbb::combinable<std::vector<bool>> store(
        [halfedgeCount]() { return std::vector<bool>(halfedgeCount, false); });
    tbb::parallel_for(
        tbb::blocked_range<size_t>(0, halfedge_.size()),
        [&store, &mutex, &problematic, &largestEdge, this](const auto& r) {
          auto& local = store.local();
          for (auto i = r.begin(); i < r.end(); ++i) {
            if (local[i]) continue;
            local[i] = true;
            const int vert = halfedge_[i].startVert;
            size_t largest = i;
            ForVert(i, [&local, &largest](int current) {
              local[current] = true;
              largest = std::max(largest, static_cast<size_t>(current));
            });
            auto expected = std::numeric_limits<size_t>::max();
            if (!reinterpret_cast<std::atomic<size_t>*>(largestEdge.data() +
                                                        vert)
                     ->compare_exchange_strong(expected, largest) &&
                expected != largest) {
              // we know that there is another loop...
              mutex.lock();
              problematic.push_back(largest);
              mutex.unlock();
            }
          }
        });

    std::vector<bool> halfedgeProcessed(halfedge_.size(), false);
    for (size_t i : problematic) {
      if (halfedgeProcessed[i]) continue;
      vertPos_.push_back(vertPos_[halfedge_[i].startVert]);
      const int vert = NumVert() - 1;
      ForVert(i, [this, vert, &halfedgeProcessed](int current) {
        halfedgeProcessed[current] = true;
        halfedge_[current].startVert = vert;
        halfedge_[halfedge_[current].pairedHalfedge].endVert = vert;
      });
    }
  } else
#endif
  {
    std::vector<bool> vertProcessed(NumVert(), false);
    std::vector<bool> halfedgeProcessed(halfedge_.size(), false);
    for (size_t i = 0; i < halfedge_.size(); ++i) {
      if (halfedgeProcessed[i]) continue;
      int vert = halfedge_[i].startVert;
      if (vertProcessed[vert]) {
        vertPos_.push_back(vertPos_[vert]);
        vert = NumVert() - 1;
      } else {
        vertProcessed[vert] = true;
      }
      ForVert(i, [this, &halfedgeProcessed, vert](int current) {
        halfedgeProcessed[current] = true;
        halfedge_[current].startVert = vert;
        halfedge_[halfedge_[current].pairedHalfedge].endVert = vert;
      });
    }
  }
}
```

pca006132:
And thinking about it, edge deduplication shouldn't require sorting all the entries. We are guaranteed that there are no pinched vertices, so we can just check locally around the starting vertex to search for edge deduplication. That should allow us to eliminate the expensive sorting operation for rare edge duplication cases.

src/edge_op.cpp
elalish: Yeah, we should look into this one more - I was surprised by those failures too.

src/edge_op.cpp
elalish: I tried, but it fails when two retained verts are at the same position in space - that edge can't be collapsed, even though it's zero length.

src/edge_op.cpp
elalish: No, actually the check was already changed to epsilon above, but this line was forgotten. Really the flagged edge collapse should be a superset of the short edge collapse, but the actually the short edges relax a few things. I think that's really only valid for degenerate triangles, rather than "within tolerance" - all the longer edges should really go through the additional flagged edge checks.

elalish:
I'm impressed with your pinched vert parallelization - I looked at that and found I couldn't figure out how to parallelize it. Care to open a PR with that?

pca006132:
> I'm impressed with your pinched vert parallelization - I looked at that and found I couldn't figure out how to parallelize it. Care to open a PR with that? And the deduplication stuff too?

Done.

And I think my patch for `SwappableEdge` (above the parallelization stuff) is working. We can go from there.

pca006132:
Hmm, weird:

```
[ RUN      ] BooleanComplex.CraycloudBool
/home/pca006132/code/manifold/test/boolean_complex_test.cpp:915: Failure
Value of: res.IsEmpty()
  Actual: false
Expected: true
```

With my patch, this one is failing. Probably because I forgot to turn on `MANIFOLD_EXPORT` when testing yesterday.

src/edge_op.cpp
elalish: Nice job, that was easier than I'd feared.

elalish:
I tweaked that test slightly and it passes - I think the behavior is expected. Thanks for the patch!

PR #1150, opened 2025-02-19
Less alloc

pca006132:
Just found out https://github.com/KDE/heaptrack which can show the allocations. It turns out `std::function` requires allocation, and `std::stable_sort` also performs allocation.

This tries to get rid of most temporary allocations caused by these `std::function` and `std::stable_sort` on small vectors. This also use small vector to avoid allocation when the halfedge vector for `PairUp` is small.

Benchmark shows that this gives modest improvement, but this probably depends on the system allocator. Maybe the difference will be larger for, e.g. emscripten and windows. Will get some more detailed statistics later.

### DIFF
--- a/src/boolean_result.cpp
+++ b/src/boolean_result.cpp
@@ -18,6 +18,7 @@
 
 #include "./boolean3.h"
 #include "./parallel.h"
+#include "./small_vector.h"
 #include "./utils.h"
 
 #if (MANIFOLD_PAR == 1) && __has_include(<tbb/concurrent_map.h>)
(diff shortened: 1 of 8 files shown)


pca006132:
It turns out the improvement is just around 1% (consistent), and around 4% in more extreme cases, probably because our code is already pretty fast.

Maybe I should just implement some of the simpler fixes, e.g. make the `std::function` into a template when it is simple.

elalish:
Yeah, that's a lot of code for a small improvement - these kinds of things may also add up in WASM binary size eventually. Can you do some of the simpler ones without increasing code complexity much?

pca006132:
It seems that I added the most beneficial change (`impl.cpp`) to another PR. I guess this can be closed for now until we want to revisit it later.

PR #1152, opened 2025-02-19
edge_op: quit CollapseEdge early if possible

pca006132:
The idea is to avoid orbiting around `endVert` by moving it after the code that may early return. This can help to speed up `CollapseEdge` in some cases, e.g. `Samples.Sponge4`.


elalish (approved):
Good call.

PR #1160, opened 2025-02-21
kdtree (2D) for collider

pca006132:
Simple kdtree for collider. This reduces memory consumption (for performance) as well as speed up triangulation a bit. The implementation is very simple for now. Naming is subject to changes, I don't know what is a good name for this.

On my computer, the 104 polygon tests took 1355ms instead of 1554ms previously.

This is doing a lot of sorting, but from my testing building the index takes a negligible amount of time in our triangulator, so it should be fine. I thought about using quadtree, but it feels like kdtree can be more balanced. I also tried making a separate vector for the split coordinates for less cache access. Not helpful either, probably because our polygons are usually quite small.

Multiway kdtree can improve performance slightly but not by much, and the code is considerably more complex, so I don't think it is a good idea to use that. In case you are interested:

<details>

```c++
template <typename P, bool sortX = true>
void BuildTwoDTreeImpl(VecView<P> points) {
  if (points.size() <= 8) return;
  std::sort(points.begin(), points.end(), [](const P& a, const P& b) {
    return sortX ? a.pt.x < b.pt.x : a.pt.y < b.pt.y;
  });
  size_t mid = points.size() / 2;
  size_t low = mid / 2;
  size_t high = (points.size() - mid) / 2 + mid;
  const P a = points[low];
  const P b = points[mid];
  const P c = points[high];

  std::move_backward(points.begin() + mid + 1, points.begin() + high,
                     points.begin() + high + 1);
  std::move_backward(points.begin() + low + 1, points.begin() + mid,
                     points.begin() + mid + 2);
  std::move_backward(points.begin(), points.begin() + low,
                     points.begin() + low + 3);
  points[0] = a;
  points[1] = b;
  points[2] = c;

  BuildTwoDTreeImpl<P, !sortX>(points.view(high + 1));
  BuildTwoDTreeImpl<P, !sortX>(points.view(mid + 2, high - mid - 1));
  BuildTwoDTreeImpl<P, !sortX>(points.view(low + 3, mid - low - 1));
  BuildTwoDTreeImpl<P, !sortX>(points.view(3, low));
}

template <typename P>
void BuildTwoDTree(VecView<P> points) {
  ZoneScoped;
  BuildTwoDTreeImpl<P, true>(points);
}

template <typename P, typename F>
void QueryDTreeImpl(VecView<P> points, Rect r, Rect parent, F f, bool sortX) {
  if (points.size() <= 8) {
    for (const auto& p : points)
      if (r.Contains(p.pt)) f(p);
    return;
  }
  // first three points are the split points
  for (const auto& p : points.view(0, 3))
    if (r.Contains(p.pt)) f(p);

  auto coord = [sortX](vec2& v) -> double* { return sortX ? &v.x : &v.y; };
  size_t mid = points.size() / 2;
  size_t low = mid / 2;
  size_t high = (points.size() - mid) / 2 + mid;

  Rect current = parent;
  *coord(current.max) = *coord(points[0].pt);
  if (r.DoesOverlap(current)) {
    QueryDTreeImpl(points.view(3, low), r, current, f, !sortX);
  }
  *coord(current.min) = *coord(points[0].pt);
  *coord(current.max) = *coord(points[1].pt);
  if (r.DoesOverlap(current)) {
    QueryDTreeImpl(points.view(low + 3, mid - low - 1), r, current, f,
                     !sortX);
  }
  *coord(current.min) = *coord(points[1].pt);
  *coord(current.max) = *coord(points[2].pt);
  if (r.DoesOverlap(current)) {
    QueryDTreeImpl(points.view(mid + 2, high - mid - 1), r, current, f,
                     !sortX);
  }
  *coord(current.min) = *coord(points[2].pt);
  *coord(current.max) = *coord(parent.max);
  if (r.DoesOverlap(current)) {
    QueryDTreeImpl(points.view(high + 1), r, current, f, !sortX);
  }
}

template <typename P, typename F>
void QueryDTree(VecView<P> points, Rect r, F f) {
  ZoneScoped;
  QueryDTreeImpl(points, r,
                   Rect(vec2(-std::numeric_limits<double>::infinity()),
                        vec2(std::numeric_limits<double>::infinity())),
                   f, true);
}
```

</details>


pca006132:
we should probably mark nodes as removed while ear clipping.

src/polygon.cpp
```diff
@@ -218,6 +219,11 @@ std::vector<ivec3> TriangulateConvex(const PolygonsIdx &polys) {
   return triangles;
 }
 
+struct PointWrapper {
+  vec2 pt;
+  size_t idx;
```
elalish, line 224: I think we already have this: `PolyVert`?

src/twodtree.h
elalish, line 1: Maybe `tree_2d.h`?

src/twodtree.h
```diff
@@ -0,0 +1,108 @@
...
+#include <algorithm>
+
+#include "./utils.h"
+#include "manifold/common.h"
+#include "manifold/optional_assert.h"
+#include "manifold/vec_view.h"
+
+namespace manifold {
+
+template <typename P, bool sortX = true>
```
elalish, line 26: I'll take a wild guess that the next version of this will be 3D, splitting X, Y, Z, X, Y, Z. Maybe `bool sortX` should be `int sortAxis`?

src/twodtree.h
```diff
@@ -0,0 +1,108 @@
...
+template <typename P>
+void BuildTwoDTree(VecView<P> points) {
+  ZoneScoped;
+  // don't even bother...
+  if (points.size() <= 8) return;
+  BuildTwoDTreeImpl<P, true>(points);
+}
+
+template <typename P, typename F>
+void QueryTwoDTree(VecView<P> points, Rect r, F f) {
```
elalish, line 51: So, if this is faster than the collider for 2D, do you think a version of it will also be faster for 3D? It would be nice to have something universal. Or failing that, to at least have them use the same kind of API so it's easier to compare them? Is that reasonable?

pca006132: I need to think a bit about this.

pca006132:
For some reason, I can't reproduce the CI failure locally. The exact polygon did not produce invalid triangulation here. Not sure if this is something caused by the unstable sort, so I am switching over to the stable sort we have and that should make it deterministic.

pca006132:
OK, it seems that we do need to expand the bounding box by epsilon.

src/polygon.cpp
```diff
@@ -502,32 +502,26 @@ class EarClip {
...
-      Box earBox = Box{vec3(center.x - radius, center.y - radius, 0),
-                       vec3(center.x + radius, center.y + radius, 0)};
-      earBox.Union(vec3(pos, 0));
-      collider.collider.Collisions(VecView<const Box>(&earBox, 1),
-                                   collider.ind);
+      Rect earBox = Rect(vec2(center.x - radius, center.y - radius),
+                         vec2(center.x + radius, center.y + radius));
+      earBox.Union(pos);
+      earBox.min -= epsilon;
+      earBox.max += epsilon;
```
elalish, line 509: Good call - surprised this wasn't a problem before.

src/polygon.cpp
```diff
@@ -502,32 +502,26 @@ class EarClip {
...
-            if (!Clipped(test) && test->mesh_idx != mesh_idx &&
-                test->mesh_idx != lid &&
-                test->mesh_idx != rid) {  // Skip duplicated verts
-              double cost = Cost(test, openSide, epsilon);
-              if (cost < -epsilon) {
-                cost = DelaunayCost(test->pos - center, scale, epsilon);
-              }
-              return cost;
-            }
-            return std::numeric_limits<double>::lowest();
```
elalish, line 528: Is this case handled inside the query function now?

elalish (approved):

src/polygon.cpp
pca006132: I think for lowest, it just means it will not change the cost (because we take the max).

src/polygon.cpp
pca006132: We do need it previously, just that I forgot to add it back. Previously, we enlarge the bbox of each vertex, but now we enlarge the bbox of the query.

PR #1162, opened 2025-02-22
Misc changes

### LINKED ISSUE #1161: python CI test for all python versions
before pushing to pypi

pca006132:
1. Avoid uploading to PyPI when manual triggering the python build CI. This should allow us to test the release PR before actually doing a release. Closes #1161.
2. Update dependencies, some are quite old now.
3. Re-enable `Samples.CondensedMatter16` test with triangulation check disabled. It can produce bad triangulation, but the output looks kind of fine.

To test 1, we need to merge this PR first because manually triggering CI uses the file in the master branch iirc.

### DIFF
--- a/.github/workflows/build_wheels.yml
+++ b/.github/workflows/build_wheels.yml
@@ -55,6 +55,7 @@ jobs:
       url: https://pypi.org/p/manifold3d
     permissions:
       id-token: write  # IMPORTANT: this permission is mandatory for trusted publishing
+    if: ${{ github.event_name == 'release' }}
     steps:
     - uses: actions/download-artifact@v4
       with:
--- a/test/boolean_test.cpp
+++ b/test/boolean_test.cpp
@@ -432,12 +432,10 @@ TEST(Boolean, Precision2) {
 
 TEST(Boolean, DISABLED_SimpleCubeRegression) {
   ManifoldParams().intermediateChecks = true;
-  ManifoldParams().processOverlaps = false;
   Manifold result =
       Manifold::Cube().Rotate(-0.10000000000000001, 0.10000000000000001, -1.) +
       Manifold::Cube() -
       Manifold::Cube().Rotate(-0.10000000000000001, -0.10000000000066571, -1.);
   EXPECT_EQ(result.Status(), Manifold::Error::NoError);
   ManifoldParams().intermediateChecks = false;
-  ManifoldParams().processOverlaps = true;
 }
--- a/test/samples_test.cpp
+++ b/test/samples_test.cpp
@@ -295,13 +295,17 @@ TEST(Samples, Sponge4) {
 #endif
 #endif
 
-TEST(Samples, DISABLED_CondensedMatter16) {
+TEST(Samples, CondensedMatter16) {
+  // FIXME: Triangulation can be invalid
+  bool old = PolygonParams().processOverlaps;
+  PolygonParams().processOverlaps = true;
   Manifold cm = CondensedMatter(16);
   CheckGL(cm);
 #ifdef MANIFOLD_EXPORT
   if (options.exportModels)
     ExportMesh("condensedMatter16.glb", cm.GetMeshGL(), {});
 #endif
+  PolygonParams().processOverlaps = old;
 }
 
 TEST(Samples, CondensedMatter64) {
--- a/RELEASE_CHECKLIST.md
+++ b/RELEASE_CHECKLIST.md
@@ -6,7 +6,7 @@
 1. Make a new branch called the version, e.g. v2.3.0.
 1. Use VSCode to search and replace the old version with the new - so far in test-cmake.sh, flake.nix, pyproject.toml, and package.json.
 1. Also update CMakeLists.txt version by searching for "set(MANIFOLD_VERSION_".
-1. Commit, push, open a PR, verify tests pass, merge.
+1. Commit, push, open a PR, verify tests pass, manually trigger PyPI CI, merge.
 1. On Github, draft a new release, make a new tag with the version number, add release notes, and publish.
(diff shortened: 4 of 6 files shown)


pca006132:
@elalish FYI this is the bad polygon causing the triangulation error:

image

There are two simple polygons here, both are winding in CCW direction. Probably a bug in our boolean.

pca006132:
btw, should we keep both `ManifoldParams` and `PolygonParams`? I don't think we are using `processOverlaps` in `ManifoldParams`, and it can get a bit confusing.

elalish:
>btw, should we keep both ManifoldParams and PolygonParams?

👍  Let's unify them.

test/boolean_test.cpp
```diff
@@ -432,12 +432,10 @@ TEST(Boolean, Precision2) {
...
 TEST(Boolean, DISABLED_SimpleCubeRegression) {
   ManifoldParams().intermediateChecks = true;
-  ManifoldParams().processOverlaps = false;
   Manifold result =
       Manifold::Cube().Rotate(-0.10000000000000001, 0.10000000000000001, -1.) +
       Manifold::Cube() -
       Manifold::Cube().Rotate(-0.10000000000000001, -0.10000000000066571, -1.);
   EXPECT_EQ(result.Status(), Manifold::Error::NoError);
   ManifoldParams().intermediateChecks = false;
-  ManifoldParams().processOverlaps = true;
```
elalish, line 442: What's the trouble with this test?

test/samples_test.cpp
```diff
@@ -295,13 +295,17 @@ TEST(Samples, Sponge4) {
 #endif
 #endif
 
-TEST(Samples, DISABLED_CondensedMatter16) {
+TEST(Samples, CondensedMatter16) {
+  // FIXME: Triangulation can be invalid
+  bool old = PolygonParams().processOverlaps;
+  PolygonParams().processOverlaps = true;
```
elalish, line 301: As for the bad triangulation - did you output that poly with full precision? It says they exactly share a vertex position, but I'm guessing they actually overlap slightly? Because otherwise I think the triangulator should work. When you find polys like this and they turn out to actually be epsilon-valid, go ahead and add them to our polygon test corpus.

.github/workflows/build_wheels.yml
```diff
@@ -55,6 +55,7 @@ jobs:
       url: https://pypi.org/p/manifold3d
     permissions:
       id-token: write  # IMPORTANT: this permission is mandatory for trusted publishing
+    if: ${{ github.event_name == 'release' }}
```
elalish, line 58: Want to add instructions to our release checklist?

elalish (approved):
Thanks!

test/boolean_test.cpp
pca006132: It produces self-intersecting results (more than epsilon):

image

test/samples_test.cpp
pca006132: Just updated, but it is indeed sharing the exact same vertex position:

```
Polygon 0 2.306422737411849e-11 2
18
3.493267459305203 8.14635
3.188192877073223 7.673607122926777
3.287098975966871 7.673607122926777
3.287098975966871 7.673607122926777
3.304015665365264 7.686855769635112
3.493267459305203 8.14635
3.188192877073223 7.673607122926777
3.493267459305203 8.14635
3.493267459305203 8.14635
3.493267459305203 8.14635
3.493267459305203 8.14635
3.493267459305203 8.14635
3.493267459305203 8.14635
3.5889649365369 7.673607122926777
3.493267459305203 8.14635
3.47348551191873 7.693480092989279
3.49005883764325 7.673607122926777
3.5889649365369 7.673607122926777
5
3.401099121578296 7.730144750127856
3.419703476450267 7.724547521102497
3.493267459305203 8.14635
3.380015762439546 7.724547521102497
3.401099121578296 7.730144750127856
# ... 
show(array([
  [3.493267459305203, 8.14635],
  [3.188192877073223, 7.673607122926777],
  [3.287098975966871, 7.673607122926777],
  [3.287098975966871, 7.673607122926777],
  [3.304015665365264, 7.686855769635112],
  [3.493267459305203, 8.14635],
  [3.188192877073223, 7.673607122926777],
  [3.493267459305203, 8.14635],
  [3.493267459305203, 8.14635],
  [3.493267459305203, 8.14635],
  [3.493267459305203, 8.14635],
  [3.493267459305203, 8.14635],
  [3.493267459305203, 8.14635],
  [3.5889649365369, 7.673607122926777],
  [3.493267459305203, 8.14635],
  [3.47348551191873, 7.693480092989279],
  [3.49005883764325, 7.673607122926777],
  [3.5889649365369, 7.673607122926777],
]))
show(array([
  [3.401099121578296, 7.730144750127856],
  [3.419703476450267, 7.724547521102497],
  [3.493267459305203, 8.14635],
  [3.380015762439546, 7.724547521102497],
  [3.401099121578296, 7.730144750127856],
]))
```

pca006132:
> > btw, should we keep both ManifoldParams and PolygonParams?
> 
> 👍 Let's unify them.

Hmm, looking at it, we want to enable intermediate checks for polygons but not manifold. Maybe we can make a `checkSelfIntersection` flag after unifying them.

And is this considered a breaking change? I think some users may use these params and removing `PolygonParams` may break things. Maybe update the `PolygonParams` function to give the `ManifoldParams` but also mark is as deprecated? In case people expecting these two to be different, they will be notified.

I think I will do that in another PR.

elalish:
Yeah, good call on deprecation, though honestly I hope no one is changing those values in production.

elalish (approved):

test/samples_test.cpp
elalish: That looks like a good test for the corpus. I need to come back and debug that one.

PR #1166, opened 2025-02-25
Early return when result size too large

### LINKED ISSUE #1164: Crash with large number of intersections
When I run the following code, manifold crashes.  I am not 100% sure, but when I was debugging in the context of Blender, I found that the crash was CopyFaceEdges(), I think called by Filter11(). It seems that p2q1 has data of size 4,255,073,176 and the CopyFaceEdges arithmetic is with ints, which appears to overflow. In the particular crash I looked at, 

```
i = 398913110
j = 0
q1 = 4594
offset = 350,573,744
idx = 3 * (i + offset) = 2,248,460,562
```
and idx is too big.

I can understand that maybe there are just too many intersections to handle here, but it would be good if the program returned an error rather than crashing.

Test code:
```
#include <iostream>
#include "manifold/manifold.h"

using namespace manifold;

void man_big()
{
    Manifold cyl = Manifold::Cylinder(2.0, 1.0, 0.0, 1024);
    Manifold sph = Manifold::Sphere(1.0, 3252).Translate(vec3(0.0, 0.0, 1.04));
    std::cout << "cyl has " << cyl.NumTri() << " trianges\n";
    std::cout << "sph has " << sph.NumTri() << " trianges\n";
    Manifold result = cyl - sph;
    MeshGL mgl = result.GetMeshGL();
    std::cout << "result has " << mgl.NumTri() << " triangles\n";
}

int main() {
    man_big();
    return 0;
}
```

pca006132:
Closes #1164.

I have looked at the code for cases where the result size can be too large, and it seems that not a lot can cause integer overflow if we limit the input size (halfedge count).

@howardt can you check if this can avoid the crash?

### DIFF
--- a/src/boolean3.cpp
+++ b/src/boolean3.cpp
@@ -497,6 +497,8 @@ Boolean3::Boolean3(const Manifold::Impl &inP, const Manifold::Impl &inQ,
   // Symbolic perturbation:
   // Union -> expand inP
   // Difference, Intersection -> contract inP
+  constexpr size_t INT_MAX_SZ =
+      static_cast<size_t>(std::numeric_limits<int>::max());
 
 #ifdef MANIFOLD_DEBUG
   Timer broad;
@@ -530,6 +532,11 @@ Boolean3::Boolean3(const Manifold::Impl &inP, const Manifold::Impl &inQ,
   SparseIndices p2q0 = inP.VertexCollisionsZ(inQ.vertPos_, true);  // inverted
   p2q0.Sort();
   PRINT("p2q0 size = " << p2q0.size());
+
+  if (3 * p1q2_.size() + 3 * p2q1_.size() > INT_MAX_SZ) {
+    valid = false;
+    return;
+  }
 
   // Find involved edge pairs from Level 3
   SparseIndices p1q1 = Filter11(inP_, inQ_, p1q2_, p2q1_);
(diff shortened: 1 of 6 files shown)


src/boolean3.cpp
```diff
@@ -531,6 +533,11 @@ Boolean3::Boolean3(const Manifold::Impl &inP, const Manifold::Impl &inQ,
   p2q0.Sort();
   PRINT("p2q0 size = " << p2q0.size());
 
+  if (3 * p1q2_.size() + 3 * p2q1_.size() > INT_MAX_SZ) {
```
pca006132, line 536: This is the specific expression that can overflow integers.

src/boolean3.cpp
```diff
@@ -573,6 +580,11 @@ Boolean3::Boolean3(const Manifold::Impl &inP, const Manifold::Impl &inQ,
       Intersect12(inQ, inP, s20, p2q0, s11, p1q1, z20, xyzz11, p2q1_, false);
   PRINT("x21 size = " << x21_.size());
 
+  if (x12_.size() > INT_MAX_SZ || x21_.size() > INT_MAX_SZ) {
```
pca006132, line 583: x12 and x21 can be at most the size of the sparse indices, which is at most n^2 and can cause overflow elsewhere.

elalish (approved):
Thanks!

PR #1167, opened 2025-02-25
unified ManifoldParams and PolygonParams

pca006132:
The old way of specifying parameters for manifold and polygon separately is quite confusing. This merges them, and removed `PolygonParams` all together.

This is a small breaking that hopefully no one is using, except our own testing code. That said, the next release should still bump the minor version (3.1.0) together with other changes in decimation etc.

### DIFF
--- a/src/polygon.cpp
+++ b/src/polygon.cpp
@@ -27,6 +27,8 @@
 namespace {
 using namespace manifold;
 
+constexpr int TRIANGULATOR_VERBOSE_LEVEL = 2;
+
 constexpr double kBest = -std::numeric_limits<double>::infinity();
 
 // it seems that MSVC cannot optimize la::determinant(mat2(a, b))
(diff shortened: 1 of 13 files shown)


src/polygon.cpp
```diff
@@ -528,7 +527,7 @@ class EarClip {
 
     void PrintVert() const {
 #ifdef MANIFOLD_DEBUG
-      if (!params.verbose) return;
+      if (!ManifoldParams().verbose) return;
```
elalish, line 530: We'll need to separate the verbosity or have levels or something. The triangulator is orders of magnitude more verbose than the Boolean is.

elalish (changes requested):
Any idea why those tests are failing? I didn't notice anything functionally wrong in my read-through.

src/polygon.cpp
pca006132: it is now an int

elalish (approved):
Thanks!

PR #1179, opened 2025-03-06
radix sort impl

### LINKED ISSUE #1176: use radix sort when the comparator is just a key mapping
Radix sort is faster, but we are not using it as much as possible due to some sort operations requiring custom comparators. We can generalize the radix sort implementation to allow a key mapping function.

The merge function is probably something we can look into as well, it seems that the std implementation is not doing much optimization with it.

pca006132:
Closes #1176. This is a rather large change to the algorithm implementation, but it scales much better now.

```
                               1K    5K   10K   50K  100K    1M   10M 
sort_Rnd                     2.41  1.09  1.01  4.96  8.53 17.33 11.96 
sort_Rnd_size                1.08  1.08  1.04  8.76 15.27 28.58 28.37 
sort_Rnd_size_merge          2.24  1.07  1.02  4.81  9.15 15.66 10.51 
sort_roughly_Asc             1.25  1.03  0.99  1.40  2.30  2.73  2.64 
sort_roughly_Asc_size        1.00  1.06  1.02  1.09  1.62  6.52 12.00 
sort_roughly_Asc_size_merge  1.29  1.00  1.00  1.54  2.03  2.66  2.61 
sort_Des                     1.01  0.95  0.94  1.12  1.87  2.54  2.57 
sort_Des_size                1.02  0.98  0.99  0.86  1.34  4.33  5.64 
sort_Des_size_merge          1.01  0.94  0.96  2.13  2.73  3.28  2.95 
```

Speedup for radix sort for larger data size is increasing. Previously, it is not increasing due to merge being quite sequential:

```
                               1K   10K  100K    1M   10M 
sort_Rnd                     2.39  1.17 10.01 17.68 11.36 
sort_Rnd_size                1.67  2.86 12.66 22.35 11.90 
sort_Rnd_size_merge          2.42  1.04  9.25 16.13 10.71 
sort_roughly_Asc             1.31  1.02  2.13  3.14  2.69 
sort_roughly_Asc_size        1.13  1.50  3.16  4.86  7.29 
sort_roughly_Asc_size_merge  1.23  0.99  2.28  2.87  2.57 
sort_Des                     0.96  0.95  1.89  2.70  2.53 
sort_Des_size                1.26  1.37  2.05  2.63  2.94 
sort_Des_size_merge          0.93  0.91  2.36  3.03  2.82 
```

The new result is not as fast for smaller input instances, but that can probably be tuned.

On my computer, `Samples.CondensedMatter64` runtime reduced from 7785ms to 5365ms due to the huge sparse indices it has to sort. For other benchmarks, the improvement is not as significant (e.g. `Samples.Sponge4` from 2682ms to 2433ms). For perfTest, there is not much improvement, the collider is the slowest part there.

### DIFF
--- a/src/parallel.h
+++ b/src/parallel.h
@@ -177,12 +177,7 @@ void radix_sort_with_key_rec(I start, I end, J dest, int bytes, KeyFn &keyfn,
           prev = x;
           hists[i][getByte(curr, bytes)]++;
         }
-        metadata[i] = {.localMin = localMin,
-                       .localMax = localMax,
-                       .prefixMax = localMax,
-                       .suffixMin = localMin,
-                       .sorted = sorted,
-                       .canSkip = false};
+        metadata[i] = {localMin, localMax, localMax, localMin, sorted, false};
       });
     }
     taskgroup.wait();
--- a/src/parallel.h
+++ b/src/parallel.h
@@ -126,24 +126,9 @@ void radix_sort_with_key_rec(I start, I end, J dest, int bytes, KeyFn &keyfn,
                              bool writeback) {
   using T = std::remove_reference_t<decltype(*start)>;
   using U = std::invoke_result_t<KeyFn, T>;
-  constexpr size_t MIN_BLOCK_SIZE = 10'000;
+  constexpr size_t MIN_BLOCK_SIZE = 100'000;
   size_t n = std::distance(start, end);
-
-  if ((n <= kSeqThreshold && bytes >= 1) || (n < 2'000)) {
-    if (writeback) {
-      std::stable_sort(start, end, [&keyfn](const T &a, const T &b) {
-        return keyfn(a) < keyfn(b);
-      });
-    } else {
-      std::copy(start, end, dest);
-      std::stable_sort(dest, dest + n, [&keyfn](const T &a, const T &b) {
-        return keyfn(a) < keyfn(b);
-      });
-    }
-    return;
-  }
   auto getByte = ByteGetter(keyfn);
-
   unsigned int hardware_concurrency = std::thread::hardware_concurrency();
   size_t blocks =
       std::max(std::min(hardware_concurrency,
--- a/src/parallel.h
+++ b/src/parallel.h
@@ -17,6 +17,8 @@
 
 #pragma once
 
+#include <thread>
+
 #include "./iters.h"
 #if (MANIFOLD_PAR == 1)
 #include <tbb/combinable.h>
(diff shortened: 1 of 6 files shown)


src/parallel.h
```diff
@@ -168,132 +170,255 @@ struct CopyIfScanBody {
...
+
+template <typename I, typename J, typename KeyFn>
+void radix_sort_with_key_rec(I start, I end, J dest, int bytes, KeyFn &keyfn,
+                             bool writeback) {
+  using T = std::remove_reference_t<decltype(*start)>;
+  using U = std::invoke_result_t<KeyFn, T>;
+  constexpr size_t MIN_BLOCK_SIZE = 10'000;
+  size_t n = std::distance(start, end);
+
+  if ((n <= kSeqThreshold && bytes >= 1) || (n < 2'000)) {
```
pca006132, line 232: For sufficiently small cases, sequential sort is faster.

src/parallel.h
```diff
@@ -168,132 +170,255 @@ struct CopyIfScanBody {
...
+  if (blocks > 1) {
+    size_t blockSize = n / blocks;
+    std::vector<std::array<size_t, 256>> hists(blocks + 1);
+
+    // local min, local max, prefix max, suffix min, sorted, can skip
+    std::vector<RadixBlockMetadata<U>> metadata(blocks);
+
+    auto taskgroup = tbb::task_group();
+    for (size_t i = 0; i < blocks; ++i) {
+      taskgroup.run([&, i]() {
```
pca006132, line 262: We compute per-block metadata on the fly, which can be used to determine if the prefix or suffix of the input is already sorted and do not require further processing.

pca006132:
Thinking about it, I think I have more things I can play with before deciding on the actual changes. LSB-radix sort isn't really that bad, in fact its sequential performance is great when the input is quite random, and maybe I can somehow change it to make it better for roughly sorted distributions.

elalish:
I'm trying to read your perf tables - is the first column the name of the benchmark being tested? Are the values (time for old sort / time for this sort)? Also keep in mind that I'm planning to remove most of the SpareIndices usage - though we still sort in plenty of other places.

pca006132:
Yes, but sort performance can depend a lot on the input distribution. I think we can withhold this for a while until we finish removing `SparseIndices` sort, and do optimization using the performance at that time.

starseeker:
@pca006132 Is there an issue or PR for the "removing SparceIndices sort" goal?  Just thinking it would be handy to reference it here if there is...

pca006132:
Oh sparse indices was removed entirely, so this is for everything else, e.g. halfedges.

starseeker:
Ah, so the blocker is already removed?  Nice.

pca006132:
> Ah, so the blocker is already removed? Nice.

This PR is stalled mainly because I don't have time to improve it. I have some idea for making radix sort faster, but I need time and energy to implement that.

pca006132:
It seems that the performance of radix sort doesn't really matter now, we don't have that much things to sort, and bandwidth is the most limiting part. Changing sort implementation doesn't change that.

PR #1185, opened 2025-03-09
Don't simplify unless requested

### LINKED ISSUE #1138: Improve performance of MeshGL -> Manifold construction
It feels like there is room to improve the performance of the MeshGL -> Manifold construction.

When I construct two MeshGL's, each containing a sphere with 1,046,528 triangles, it takes 138.9 ms to construct an Impl from a MeshGL for each of them.  The boolean itself (a difference of one sphere offset a bit from the other), and conversion back into MeshGL takes 242 ms.  So it is a bit surprising to me that the time to convert 2 inputs into Impl's takes comparable time to the boolean operation itself.

The MeshGL's I construct have no vertex properties beyond the 3 for the coordinates. I set faceID, but don't populate runIndex and runOriginalID.  There are no mergeFrom/mergeTo items populated. I put some timers into the Impl.h file around various phases of the contruction, and it printed this for doing one sphere: 

Impl constructor: copy vert properties took 1.07067 ms
Impl constructor: set up triRef took 42 ns
Impl constructor: set up triVerts took 2.70829 ms
Impl constructor: CreateHalfEdges took 15.3821 ms
Impl constructor: IsManifold took 0.430167 ms
Impl constructor: CalculateBBox took 0.248125 ms
Impl constructor: SplitPinchedVerts took 22.2563 ms
Impl constructor: CalculateNormals took 3.83696 ms
Impl constructor: InitializeOriginal took 0.627458 ms
Impl constructor: CreateFaces took 22.3918 ms
Impl constructor: SimplifyTopology took 36.091 ms
Impl constructor: RemoveUnreferencedVerts took 0.460083 ms
Impl constructor: Finish took 32.6864 ms
Impl constructor from MeshGLP, total took 138.872 ms

These times are all on my Macbook Pro M3 Max : 16 cores (12 performance, 4 efficiency), 64GB memory.

The high running phases are 
	SimplifyTopology (36 ms)
	Finish (33 ms)
	CreateFaces (22 ms)
	SplitPInchedVerts (22 ms)
	CreateHalfEdges (15 ms)

Do these numbers look reasonable, or is there room to improve?

elalish:
Fixes #1138 - really that thread, while interesting, has grown way too long and gone on too many tangents. Some has already been fixed, this fixes some more - please make separate issues for whatever's left.

Now that #1147 has landed which makes the Boolean only decimate the intersected parts of meshes, this PR finishes the job by removing our auto-decimation from the rest of our functions, so users can just call it manually if they want through `Simplify`. I've also deprecated the `cleanupTriangles` option, since it's effectively always false now. We're still collapsing degenerate triangles and fixing topology at the epsilon level, but not touching colinear edges, coplanar faces, or anything that's only within tolerance. @stephomi, @starseeker, @howardt I believe you all may have asked for things along this line at various points, so it would be great to have some early eyes on this and get feedback. 

Good new / bad news is this improved our testing and thus found that `Simplify` doesn't handle properties correctly at all. It looks like I now need to add a significant refactor here to fix that, but on the plus side I think I can remove our `CreateProperties` function entirely, which has gone through quite a bit of optimization due to slowing things down a lot in the past. I think I have a cleaner way now.


pca006132:
btw this may be related to simplification: https://github.com/openscad/openscad/issues/5738

src/boolean_result.cpp
```diff
@@ -809,6 +810,40 @@ Manifold::Impl Boolean3::Result(OpType op) const {
   PRINT(n12 << " new verts from edgesP -> facesQ");
   PRINT(n21 << " new verts from facesP -> edgesQ");
 
+  const int numProp = std::max(inP_.NumProp(), inQ_.NumProp());
+  if (numProp > 0) {
+    // Ideally, switch over from triProperties vec to Halfedge.startProp - then
+    // the prop index will be available in the append edges functions below.
```
elalish, line 816: @pca006132 I'm struggling to get this one finished - are you interested in taking a look? If you look at my last commit, I've added comments on the refactor I have planned. I think it's pretty straight forward, and I bet you'd optimize it better than I would. 

I ask because this is probably the last major PR to land before doing a release, right? We have a lot of good stuff now (mostly yours) and I don't want to delay it. Let me know what you think - and is my description clear at all?

pca006132: Sure, I can have a look at it later tonight.

pca006132: So this is kind of similar to what CreateProperties is doing, but you want to do it in parallel, and probably handle edges differently?

About the inclusion number array, should it be indexed by the vertex id, halfedge id, prop vert id or something?

elalish: Yes, index by prop vert. Every prop vert has exactly one vert, so that makes things nice. The idea is to keep track of properties directly by index just like the verts, instead of needing to look anything up with maps or hash tables.

src/boolean_result.cpp
```diff
@@ -809,6 +810,40 @@ Manifold::Impl Boolean3::Result(OpType op) const {
...
+                 for (const int i : {0, 1, 2}) {
+                   prop2VertQ[inQ_.meshRelation_.triProperties[tri][i]].store(
+                       inQ_.halfedge_[3 * tri + i].startVert);
+                 }
+               });
+    // Scan the properties inclusion numbers, e.g. + i03[prop2VertP[prop]].
+    //
+    // New verts create either 2 or 3 new properties: one for the face, plus one
+    // for an edge where start and end verts share properties, or plus two for
+    // an edge that doesn't share both properties.
```
elalish, line 837: This is the answer to question 2-i. It might be clearer if I said "prop verts" instead of "properties" - that's what I mean.

src/boolean_result.cpp
```diff
@@ -809,6 +810,40 @@ Manifold::Impl Boolean3::Result(OpType op) const {
...
+  if (numProp > 0) {
+    // Ideally, switch over from triProperties vec to Halfedge.startProp - then
+    // the prop index will be available in the append edges functions below.
+    Vec<std::atomic<int>> prop2VertP(inP_.NumPropVert());
+    for_each_n(autoPolicy(inP_.NumTri()), countAt(0), inP_.NumTri(),
+               [&](size_t tri) {
+                 for (const int i : {0, 1, 2}) {
+                   prop2VertP[inP_.meshRelation_.triProperties[tri][i]].store(
+                       inP_.halfedge_[3 * tri + i].startVert);
+                 }
```
elalish, line 823: Another piece of 2-i: No - every propVert maps to exactly one vertex, hence this vector being possible. Whenever the same value shows up multiple times in this list, that's a vertex that's associated with multiple propVerts.

elalish: This is an important feature (because it's also implicitly a restriction in MeshGL), so to answer question 1 - we definitely need to duplicate rather than reuse the index. Again, this duplication is actually exceedingly rare in practice since it's only for handing self-intersecting meshes.

src/boolean_result.cpp
```diff
@@ -809,6 +810,40 @@ Manifold::Impl Boolean3::Result(OpType op) const {
...
+               });
+    // Scan the properties inclusion numbers, e.g. + i03[prop2VertP[prop]].
+    //
+    // New verts create either 2 or 3 new properties: one for the face, plus one
+    // for an edge where start and end verts share properties, or plus two for
+    // an edge that doesn't share both properties.
+    //
+    // Copy included inP and inQ properties to outR, and duplicate for new verts
+    // accoding to the inclusion numbers. Calculate new outR properties using
+    // the Barycentric function from CreateProperties.
```
elalish, line 841: For 2-iii: It's best to think of propVerts as being on a halfedge or triangle - the paired halfedge usually, but doesn't always, shares the same propVerts. Only if both ends of the paired halfedges share propVerts can you interpolate that as a single new shared propVert - otherwise you'll be creating two new propVerts, one for each halfedge. It's not possible to have two ends of the same halfedge have properties that don't represent the same channels of data. The channels of data are consistent across each meshID, or equivalently, each triangle run in MeshGL. So they are also consistent over a triangle or halfedge.

src/boolean_result.cpp
```diff
@@ -809,6 +810,40 @@ Manifold::Impl Boolean3::Result(OpType op) const {
...
+                 }
+               });
+    // Scan the properties inclusion numbers, e.g. + i03[prop2VertP[prop]].
+    //
+    // New verts create either 2 or 3 new properties: one for the face, plus one
+    // for an edge where start and end verts share properties, or plus two for
+    // an edge that doesn't share both properties.
+    //
+    // Copy included inP and inQ properties to outR, and duplicate for new verts
+    // accoding to the inclusion numbers. Calculate new outR properties using
```
elalish, line 840: For 2-ii: Yes, we copy all retained propVerts, just like we do verts. But the number isn't going to explode any more than the number of verts does. Well, technically you'll get double the number of new propVerts as new verts, but considering in general I think of these problems as scaling such that the number of new verts is O(sqrt(n)) for n: the number of input verts, I don't think that's a huge problem. Plus it's fundamentally necessary in order to represent e.g. UV coordinates.

elalish (commented):
I made separate threads to answer each of your questions.

src/boolean_result.cpp
pca006132: Can you define what share propVerts means? If each propVert maps to exactly one vertex.

And for these new vertices, where do their propVerts come from? Say, if each vertex of the associated triangle originally maps to 2 propVert, we should have 2 propVert interpolated using barycentric coord, plus some propVerts interpolated using propVerts associated to endpoints of the edge?

elalish: By share propVerts, I mean the paired halfedges point to the same prop index. Basically verts *must* be manifold, while propVerts *may* be manifold. The edge is prop-manifold if the propVerts are shared.

The second part isn't quite right. Any given triangle only has one set of three propVerts; new propVerts on either its edges or its center all get interpolated barycentricly using that single interpolant. So e.g. 6 triangles ajoin a vertex - this vertex might have 2 propVerts, meaning e.g. 3 of those triangles point to one propVert and 3 point to the other.

pca006132: If two paired halfedges (with different starting vertices) can point to the same prop vert, what does it mean by each prop vert maps to one vert?

Or does this mean each prop vert maps to one edge (can be one or two halfedges)?

src/boolean_result.cpp
```diff
@@ -809,6 +810,40 @@ Manifold::Impl Boolean3::Result(OpType op) const {
...
+    for_each_n(autoPolicy(inQ_.NumTri()), countAt(0), inQ_.NumTri(),
+               [&](size_t tri) {
+                 for (const int i : {0, 1, 2}) {
+                   prop2VertQ[inQ_.meshRelation_.triProperties[tri][i]].store(
+                       inQ_.halfedge_[3 * tri + i].startVert);
+                 }
+               });
+    // Scan the properties inclusion numbers, e.g. + i03[prop2VertP[prop]].
+    //
+    // New verts create either 2 or 3 new properties: one for the face, plus one
```
pca006132, line 835: And also, how are these new prop verts computed and attached to halfedges?

For the face prop vert, it is barycentric interpolation from the three prop vert of the intersecting face, but which new halfedge to attach to?

For prop vert corresponding to an edge (two different cases), how should we interpolate them and where should we attach the prop vert to?

Sorry for a lot of questions, it may be easier for you to directly implement it rather than explaining them.

src/boolean_result.cpp
fire: UV coordinates :o

elalish:
Okay, not sure why I let this languish so long - there was really no reason to let it be blocked by #1232, since that issue with Simplify is unrelated to this change, I just happened to notice it here. Let's get this in since a bunch of users are waiting for it.

fire:
Can we estimate when a new release of manifold will happen?

elalish:
I want to do https://github.com/elalish/manifold/issues/1217 first, which I don't think should be too hard, hopefully. So, ideally a week or two? We just bought a house and need to hunt down enough furniture to move my family into it, so my time is a bit stretched at the moment.

PR #1186, opened 2025-03-09
avoid sparse indices

### LINKED ISSUE #972: Refactor to avoid using SparseIndices
See https://github.com/elalish/manifold/pull/931#issuecomment-2353465022

At least avoid it as much as possible. This should improve both the memory usage and performance.

pca006132:
Start removing sparse indices usage in `Boolean3.cpp`. This is mainly for discussion about invariants that I am not sure about. Closes #972.

The code is really ugly now, but it seems to be working. I will clean it up later. The refactor here is surprisingly simple: just change the binary search to a function call. There is more to be done to make it faster, though.

---

Performance:

Previously:

benchmark-master

Now:

benchmark-avoid-sparseindices

For some numbers:

```
Previously:
---------------------------------------------------
Threads= 1
Total time
  avg : 179.569
  min : 2.692
  max : 10754.046
Converting MeshGL to Manifold
  avg : 11.869
  min : 1.015
  max : 76.816
Manifold boolean
  avg : 167.559
  min : 1.089
  max : 10702.296
Converting Manifold to MeshGL
  avg : 0.141
  min : 0.003
  max : 1.686

Threads= 16
Total time
  avg : 34.759
  min : 2.903
  max : 1688.115
Converting MeshGL to Manifold
  avg : 6.419
  min : 1.078
  max : 24.742
Manifold boolean
  avg : 28.170
  min : 1.021
  max : 1670.252
Converting Manifold to MeshGL
  avg : 0.170
  min : 0.003
  max : 1.480

---------------------------------------------------
Now:
---------------------------------------------------
Threads= 1
Total time
  avg : 84.570
  min : 2.537
  max : 2886.896
Converting MeshGL to Manifold
  avg : 12.471
  min : 1.016
  max : 77.300
Manifold boolean
  avg : 71.948
  min : 0.943
  max : 2831.661
Converting Manifold to MeshGL
  avg : 0.151
  min : 0.003
  max : 1.758

Threads= 16
Total time
  avg : 17.625
  min : 2.154
  max : 270.270
Converting MeshGL to Manifold
  avg : 6.517
  min : 1.059
  max : 25.309
Manifold boolean
  avg : 10.920
  min : 0.737
  max : 250.945
Converting Manifold to MeshGL
  avg : 0.189
  min : 0.003
  max : 1.606
```

Not all models are equal in this case, when the sparse indices are relatively small, e.g. for our perfTest where we use high resolution spheres, the speedup is minimal (nTri = 8388608, time = 1.80269 sec).


src/boolean3.cpp
```diff
@@ -517,87 +421,36 @@ Boolean3::Boolean3(const Manifold::Impl &inP, const Manifold::Impl &inQ,
...
-  z02.clear();
-  z20.clear();
-
-  Vec<int> p0 = p0q2.Copy(false);
-  p0q2.Resize(0);
-  Vec<int> q0 = p2q0.Copy(true);
-  p2q0.Resize(0);
   // Sum up the winding numbers of all vertices.
-  w03_ = Winding03(inP, p0, s02, false);
+  w03_ = Winding03(inP, inQ, p0q2, expandP_, true);
```
pca006132, line 451: @elalish I'm thinking if it is possible to somehow integrate this winding number computation into `Intersect12`, but not sure if that is sound. Will it ever miss anything?

src/boolean3.cpp
```diff
@@ -451,41 +341,55 @@ struct Kernel12 {
...
-      autoPolicy(p1q2.size(), 1e4), countAt(0_uz), p1q2.size(),
-      Kernel12({x12, v12, p0q2.AsVec64(), s02, z02, p1q1.AsVec64(), s11, xyzz11,
-                inP.halfedge_, inQ.halfedge_, inP.vertPos_, forward, p1q2}));
+  F02 f02{inP.vertPos_,
+          inQ.halfedge_,
+          inQ.vertPos_,
+          expandP,
+          forward ? inP.vertNormal_ : inQ.vertNormal_,
+          forward};
+  F11 f11{forward ? inP.vertPos_ : inQ.vertPos_,
```
pca006132, line 358: I will definitely fix these later, it is just interesting that it works within 30 minutes of trying :)

src/boolean3.cpp
```diff
@@ -448,44 +337,149 @@ struct Kernel12 {
...
 
-  for_each_n(
-      autoPolicy(p1q2.size(), 1e4), countAt(0_uz), p1q2.size(),
-      Kernel12({x12, v12, p0q2.AsVec64(), s02, z02, p1q1.AsVec64(), s11, xyzz11,
-                inP.halfedge_, inQ.halfedge_, inP.vertPos_, forward, p1q2}));
+  inQ.collider_.Collisions<false, Box, Kernel12Recorder>(PEdgeBB.cview(),
+                                                         recorder);
 
-  p1q2.KeepFinite(v12, x12);
+  Kernel12Tmp result = recorder.get();
```
pca006132, line 447: This is the place where things is broken. p1q2's order is now non-deterministic, and it seems that `Sponge4` will randomly report

```
manifold/src/polygon.cpp (124): 'std::all_of(triangles.begin(), triangles.end(), [&vertPos, epsilon](const ivec3 &tri) { return CCW(vertPos[tri[0]], vertPos[tri[1]], vertPos[tri[2]], epsilon) >= 0; })' is false: triangulation is not entirely CCW!" thrown in the test body.
```

but I don't see anything wrong in the code apart from the different order. The major issue here is that the error is pretty rare (20%?), and never happens when I make the collider sequential, so it is probably related to our collider code. I am not sure if this is an old bug or a new one.

src/vec.h
```diff
@@ -140,32 +140,31 @@ class Vec : public VecView<T> {
     std::swap(capacity_, other.capacity_);
   }
 
-  inline void push_back(const T &val, bool seq = false) {
```
elalish, line 143: 💯

src/boolean3.cpp
elalish: I would guess it's triggering a bug in our triangulator, rather then being actually wrong. If you can manage to dump the polygon into a new test when it happens to fail, that would be ideal. However, we have lots of people asking us for determinism, so I think regardless we'll need to sort or something.

src/boolean3.cpp
```diff
@@ -512,92 +506,30 @@ Boolean3::Boolean3(const Manifold::Impl &inP, const Manifold::Impl &inQ,
...
   Timer intersections;
   intersections.Start();
 #endif
 
-  // Level 2
-  // Build up XY-projection intersection of two edges, including the z-value for
-  // each edge, keeping only those whose intersection exists.
-  Vec<int> s11;
-  Vec<vec4> xyzz11;
-  std::tie(s11, xyzz11) = Shadow11(p1q1, inP, inQ, expandP_);
```
elalish, line 556: I love seeing all this removed! That broad phase timer above is probably useless now and can be removed as well.

src/boolean3.cpp
elalish: Not sure if you already answered this for yourself, but w03 is calculated for every vertex. It doesn't need to be - I used to only calculate it for the verts of intersected edges and faces - you can flood fill the values to the rest because they're always consistent. I removed that because my flood-filling algorithm was slower than just calculating them all, but I may have just been doing it poorly.

src/sparse.h
```diff
@@ -146,8 +146,8 @@ class SparseIndices {
         data_.size() / sizeof(int32_t));
   }
 
-  inline void Add(int p, int q, bool seq = false) {
-    data_.extend(sizeof(int64_t), seq);
+  inline void Add(int p, int q) {
```
elalish, line 149: Removing `SparseIndices` entirely sounds great - anything that results in fewer lines of code overall is likely worthwhile.

src/collider.h
```diff
@@ -163,14 +163,16 @@ struct FindCollision {
   VecView<const T> queries;
   VecView<const Box> nodeBBox_;
   VecView<const std::pair<int, int>> internalChildren_;
-  Recorder recorder;
+  Recorder& recorder;
 
-  inline int RecordCollision(int node, const int queryIdx, SparseIndices& ind) {
+  using LocalT = typename Recorder::LocalT;
+
+  inline int RecordCollision(int node, const int queryIdx, LocalT& local) {
```
elalish, line 170: For updating other collider usages - I leave it up to you. If it's less work, do it here - otherwise keep them separate for ease of review.

src/boolean3.cpp
pca006132: I think calculating them all is probably good for now, it is pretty fast.

src/boolean3.cpp
pca006132: ok, just precision issue

src/boolean3.cpp
```diff
@@ -88,35 +92,23 @@ struct CopyFaceEdges {
...
-
 inline bool Shadows(double p, double q, double dir) {
   return p == q ? dir < 0 : p < q;
 }
 
 inline std::pair<int, vec2> Shadow01(
     const int p0, const int q1, VecView<const vec3> vertPosP,
     VecView<const vec3> vertPosQ, VecView<const Halfedge> halfedgeQ,
-    const double expandP, VecView<const vec3> normalP, const bool reverse) {
+    const double expandP, VecView<const vec3> normal, const bool reverse) {
```
pca006132, line 102: changed the naming because it is not really for `P`, but depends on `reverse`. When `reverse` is false, it is for `P`, but otherwise it is for `Q`.

src/boolean3.cpp
```diff
@@ -126,70 +118,26 @@ inline std::pair<int, vec2> Shadow01(
...
-  VecView<vec4> xyzz;
-  VecView<int> s;
   VecView<const vec3> vertPosP;
   VecView<const vec3> vertPosQ;
   VecView<const Halfedge> halfedgeP;
   VecView<const Halfedge> halfedgeQ;
   const double expandP;
-  VecView<const vec3> normalP;
-  const SparseIndices &p1q1;
+  VecView<const vec3> normal;
```
pca006132, line 136: Similarly here.

src/boolean3.cpp
```diff
@@ -448,44 +337,137 @@ struct Kernel12 {
...
+
+std::tuple<Vec<int>, Vec<vec3>> Intersect12(const Manifold::Impl &inP,
+                                            const Manifold::Impl &inQ,
+                                            SparseIndices &p1q2, double expandP,
+                                            bool forward) {
   ZoneScoped;
-  Vec<int> x12(p1q2.size());
-  Vec<vec3> v12(p1q2.size());
+  // a: 1 (edge), b: 2 (face)
+  const Manifold::Impl &a = forward ? inP : inQ;
```
pca006132, line 416: Now, instead of swapping `inP` and `inQ` parameters, I just define new temporaries a and b, depending on `forward`. This should hopefully be cleaner now.

src/manifold.cpp
```diff
@@ -514,22 +514,6 @@ size_t Manifold::NumDegenerateTris() const {
   return GetCsgLeafNode().GetImpl()->NumDegenerateTris();
 }
 
-/**
- * This is a checksum-style verification of the collider, simply returning the
- * total number of edge-face bounding box overlaps between this and other.
- *
- * @param other A Manifold to overlap with.
- */
-size_t Manifold::NumOverlaps(const Manifold& other) const {
```
pca006132, line 523: I just removed this function because it is not used, and to be honest, I don't think the number of overlaps from the collider result is meaningful.

test/polygons/polygon_corpus.txt
```diff
@@ -1694,3 +1694,272 @@ CondensedMatter64 19 2.30642e-11 1
 -0.06842195833766776 9.6799396338048
 -0.06842195833766818 9.679939633804798
 -0.06842195833766825 9.679939633804798
+DISABLED_Sponge4b 291 5.000000000000000909e-13 27
```
pca006132, line 1697: The weird triangulation failure. It seems that it requires really high precision to reproduce the error.

pca006132:
Sorting the result in boolean3 only adds minor runtime, so we can have determinism there without hurting performance.

@elalish I think this is ready for full review, e.g., naming, organization, etc.

pca006132:
This change removed 413 LoC (excluding the triangulator test case), completely removed some of the logic related to sparse indices. The logic in boolean3 is also a lot more straightforward now.

src/boolean3.cpp
```diff
@@ -247,46 +167,26 @@ struct Kernel11 {
...
-
-  return std::make_tuple(s11, xyzz11);
+    return std::make_pair(s11, xyzz11);
+  }
 };
 
 struct Kernel02 {
-  VecView<int> s;
-  VecView<double> z;
   VecView<const vec3> vertPosP;
```
pca006132, line 180: Maybe we should change the name of these P and Q, because we will swap them. Not sure what is the best name here.

elalish: I like your renaming of them to e.g. `vertPosA` and `halfedgeB` to handle the swapping.

src/boolean3.cpp
elalish: Okay, so this one is weird - I managed to confuse myself a bunch of times, which means my naming is terrible, and part of the problem is I added on this `expandP` thing late. At a global level, only P is expanded (or contracted if false), according to its own normals. The normals of Q are never used for anything. This is weird because this function reverses P and Q sometimes, but the normals are always the same. I *think* I've used them right, but a double-check would be great. As for naming... I don't know.

elalish (approved):
Looking awesome, thanks!

src/collider.h
```diff
@@ -269,6 +241,16 @@ constexpr inline uint32_t SpreadBits3(uint32_t v) {
 /** @ingroup Private */
 class Collider {
  public:
+  template <typename F>
+  struct SimpleRecorder {
+    using LocalT = F;
```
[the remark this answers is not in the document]
pca006132: Oh, it is just some local storage type.

src/parallel.h
```diff
@@ -378,11 +378,13 @@ void for_each(ExecutionPolicy policy, Iter first, Iter last, F f) {
                 "You can only parallelize RandomAccessIterator.");
 #if (MANIFOLD_PAR == 1)
   if (policy == ExecutionPolicy::Par) {
-    tbb::parallel_for(tbb::blocked_range<Iter>(first, last),
-                      [&f](const tbb::blocked_range<Iter> &range) {
-                        for (Iter i = range.begin(); i != range.end(); i++)
-                          f(*i);
-                      });
+    tbb::this_task_arena::isolate([&]() {
```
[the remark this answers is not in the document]
pca006132: This prevents the current thread from executing work that is outside this scope. Otherwise, thread local variables can be modified by another thread... See https://uxlfoundation.github.io/oneTBB/main/tbb_userguide/work_isolation.html

pca006132:
@elalish I changed the naming a bit, but for `Interpolate` and `Intersect` I will leave the naming to you. Not sure if I should change them to `aL, aR, bL, bR` or just simply `l1, r1, l2, r2`.

pca006132:
and I will leave it for you to merge it, if you think the naming is fine now

src/collider.h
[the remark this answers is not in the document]
elalish: Why not just leave it as `F` or `T`? What is special about the type that makes it local?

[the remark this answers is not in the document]
pca006132: It has to be a bit more specific because each thread has its own instance of this value, and the implementation has to deal with this thread local thing. Maybe I should just write a comment about this.

elalish:
Hmm, I think I wasn't clear about the naming. Would you mind just reverting the `fix naming` commit and then we can merge this? Let's talk through the naming separately, because I'm not sure this clarifies it for me.

pca006132:
Sure

PR #1187, opened 2025-03-12
make boolean deterministic

### LINKED ISSUE #1177: Determinism / Repeatability?
This is a mostly question more than an issue.

Does  Manifold make any guarantees about the repeatability of the results?  That is, if you run it again or run it on a different machine, will you get the same results (where "same" includes the order of the vertices and output faces, as well as the actual geometry of the output)?

I am guessing the answer is "no", but would be happily surprised to learn otherwise.  If the answer is "no", a second question would be: if there is no parallelism used to get the answer (either because it is turned off, or because the problem size is quite small), might we expect such determinism? I suspect this mostly comes down to: are any sorts used stable sorts, and do you avoid hashing on pointers?  If the answer is still "no" for small problems / parallelism turned off, then how hard would it be to make it deterministic/repeatable/portable in those cases?

The reason I care is two-fold.

(1) It makes regression testing much easier, and Blender's regression tests typically rely on such a guarantee. If Manifold is repeatable on small cases, that would be sufficient for this.

(2) There are cases where non-repeatability can affect the visual appearance of a render or animation.  E.g., if you get different results frame-to-frame in an animation, the animation can flicker. Or, if the user has relied on element indices in making procedural postprocessing of the Boolean result (using our procedural system "Geometry Nodes") then they will not get repeatable results. They are not supposed to do this latter thing, so this is less important; but nevertheless, some users do.

pca006132:
Closes #1177

### DIFF
--- a/src/boolean_result.cpp
+++ b/src/boolean_result.cpp
@@ -657,6 +657,39 @@ void CreateProperties(Manifold::Impl &outR, const Manifold::Impl &inP,
     }
   }
 }
+
+void ReorderHalfedges(VecView<Halfedge> &halfedges) {
+  // halfedges in the same face are added in non-deterministic order, so we have
+  // to reorder them for determinism
+
+  // step 1: reorder within the same face, such that the halfedge with the
+  // smallest starting vertex is placed first
+  for_each(autoPolicy(halfedges.size() / 3), countAt(0),
+           countAt(halfedges.size() / 3), [&halfedges](size_t tri) {
+             std::array<Halfedge, 3> face = {halfedges[tri * 3],
+                                             halfedges[tri * 3 + 1],
+                                             halfedges[tri * 3 + 2]};
+             int index = 0;
+             for (int i : {1, 2})
+               if (face[i].startVert < face[index].startVert) index = i;
+             for (int i : {0, 1, 2})
+               halfedges[tri * 3 + i] = face[(index + i) % 3];
+           });
+  // step 2: fix paired halfedge
+  for_each(autoPolicy(halfedges.size() / 3), countAt(0),
+           countAt(halfedges.size() / 3), [&halfedges](size_t tri) {
+             for (int i : {0, 1, 2}) {
+               Halfedge &curr = halfedges[tri * 3 + i];
+               int oppositeFace = curr.pairedHalfedge / 3;
+               int index = -1;
+               for (int j : {0, 1, 2})
+                 if (curr.startVert == halfedges[oppositeFace * 3 + j].endVert)
+                   index = j;
+               curr.pairedHalfedge = oppositeFace * 3 + index;
+             }
+           });
+}
+
 }  // namespace
 
 namespace manifold {
@@ -846,6 +879,8 @@ Manifold::Impl Boolean3::Result(OpType op) const {
   outR.Face2Tri(faceEdge, halfedgeRef);
   halfedgeRef.clear();
   faceEdge.clear();
+
+  ReorderHalfedges(outR.halfedge_);
 
 #ifdef MANIFOLD_DEBUG
   triangulate.Stop();
--- a/src/boolean_result.cpp
+++ b/src/boolean_result.cpp
@@ -198,6 +198,7 @@ std::tuple<Vec<int>, Vec<int>> SizeOutput(
 struct EdgePos {
   double edgePos;
   int vert;
+  int collisionId;
   bool isStart;
 };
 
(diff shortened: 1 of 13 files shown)


src/boolean_result.cpp
```diff
@@ -658,6 +658,39 @@ void CreateProperties(Manifold::Impl &outR, const Manifold::Impl &inP,
     }
   }
 }
+
+void ReorderHalfedges(VecView<Halfedge> &halfedges) {
+  // halfedges in the same face are added in non-deterministic order, so we have
+  // to reorder them for determinism
```
elalish, line 664: There's a `CreateHalfedges` call at the end of `Face2Tri` - would it be more efficient to merge this logic in there?

src/edge_op.cpp
```diff
@@ -160,32 +160,21 @@ struct FlagStore {
...
+    for (const auto& tmp : stores) {
+      sizes.push_back(total_size);
+      total_size += tmp.size();
     }
+    result.resize(total_size);
+    for_each_n(ExecutionPolicy::Seq, countAt(0), stores.size(), [&](size_t i) {
+      std::copy(stores[i].begin(), stores[i].end(), result.begin() + sizes[i]);
+    });
+    stable_sort(autoPolicy(result.size()), result.begin(), result.end());
+    for (size_t x : result) f(x);
```
elalish, line 177: Simpler *and* more deterministic - I love it!

elalish (commented):
How much of #1177 does this solve?

pca006132:
Only nondeterminism in csg tree is left, I think. And it shouldn't be too hard to fix, but will need some benchmark on that.

src/edge_op.cpp
pca006132: It was supposed to be deterministic, I just somehow messed it up :P

src/boolean_result.cpp
pca006132: Probably, but `CreateHalfedges` is called in multiple places and this reordering is only concerned with boolean result. I think we can try to optimize this later when this is causing slowness.

elalish (approved):

pca006132:
I think I can adapt @howardt's simple check for determinism and throw it into places like Sponge4 and the ember test. If they all pass, we can be quite confident that we are deterministic. Will do that maybe tmr.

pca006132:
Tested this on top of #1186, and the slowdown is around 1% up to 4% with benchmarks in ember test. I expect future patches to convert stable sort to radix sort for example to easily get that back.

pca006132:
Found one missing case: In `DedupeEdge` is quite complicated, we need to ensure the same set of edges are marked duplicates. It is fixed now, and so far the 999 ember test cases are deterministic, at least with two executions.

pca006132:
OK this took a while, but I can no longer find anything non-deterministic now. SDF is out of scope though.

pca006132:
I think if we remove the usage of fma and acos (with approximation), our boolean is probably deterministic across machines as well. But that can be a bit harder to prove...

test/samples_test.cpp
```diff
@@ -291,6 +291,9 @@ TEST(Samples, Sponge4) {
     ExportMesh("mengerSponge.glb", out, options);
   }
 #endif
+  Manifold sponge2 = MengerSponge(4);
+  std::pair<Manifold, Manifold> cutSponge2 = sponge2.SplitByPlane({1, 1, 1}, 0);
+  CheckGLEquiv(cutSponge.first.GetMeshGL(), cutSponge2.first.GetMeshGL());
```
elalish, line 296: 👍

src/csg_tree.cpp
```diff
@@ -353,50 +351,44 @@ std::shared_ptr<CsgLeafNode> BatchBoolean(
...
-    results.pop_back();
-    std::pop_heap(results.begin(), results.end(), cmpFn);
-    auto b = std::move(results.back());
-    results.pop_back();
-    // boolean operation
-    auto result = SimpleBoolean(*a->GetImpl(), *b->GetImpl(), operation);
-    if (results.size() == 0) return result;
-    results.push_back(result);
-    std::push_heap(results.begin(), results.end(), cmpFn);
+    for (size_t i = 0; i < 4 && results.size() > 1; i++) {
```
elalish, line 365: Did this fix the CSG tree non-determinism too?

pca006132: Yes, batch boolean is now deterministic.

elalish:
Go ahead and merge it when you feel ready. This is great!

pca006132:
I will wait for #1186, considering it is close to merging, it is easier to fix conflicts.

PR #1190, opened 2025-03-13
misc changes

pca006132:
1. Removed the use of acos and fma in boolean code. Our boolean should hopefully be deterministic across machines now, please let us know if this is not.
2. Made mesh dump function public. This will write an OBJ file in a custom format to an ostream in full precision, which can be imported via the ImportMesh64 function. This is for debug purposes only because the format is subject to changes, and the ImportMesh64 cannot read OBJ files exported by other programs (and not intended to be used that way, user should just use meshio).

### DIFF
--- a/src/impl.cpp
+++ b/src/impl.cpp
@@ -19,6 +19,7 @@
 #include <map>
 #include <optional>
 
+#include "./csg_tree.h"
 #include "./hashtable.h"
 #include "./mesh_fixes.h"
 #include "./parallel.h"
@@ -27,6 +28,7 @@
 #ifdef MANIFOLD_EXPORT
 #include <string.h>
 
+#include <iomanip>
 #include <iostream>
 #endif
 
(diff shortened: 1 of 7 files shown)


src/impl.cpp
```diff
@@ -561,7 +568,7 @@ void Manifold::Impl::CalculateNormals() {
       // should just exclude it from the normal calculation...
       if (!la::isfinite(currEdge[0]) || !la::isfinite(prevEdge[0])) return;
       double dot = -la::dot(prevEdge, currEdge);
-      double phi = dot >= 1 ? 0 : (dot <= -1 ? kPi : std::acos(dot));
+      double phi = dot >= 1 ? 0 : (dot <= -1 ? kPi : acos(dot));
```
elalish, line 571: maybe we should call these floats, since our acos function isn't giving us double precision?

elalish (approved):
Thanks!

src/impl.cpp
elalish: I guess what I mean is, our normals aren't really as precise as doubles anymore right? It's probably fine to leave it as is, but I was just thinking maybe it would be good to remind ourselves that these have more error than e.g. our positions.

PR #1192, opened 2025-03-14
Fix face removal (#1191)

### LINKED ISSUE #1191: regression on meshgl with touching faces
It seems that this is due to simplification or something. I checked that reverting the previous 3 PRs (determinism and parallelization of CreateHalfedges) doesn't solve the problem, so it is probably something earlier.

```scad
// Two cubes sharing a face.
// This is a non-manifold object since the collapsed vertices are not distinct.
polyhedron(
  points=[
    [0, 0, 0],
    [1, 0, 0],
    [0, 1, 0],
    [1, 1, 0],
    [0, 0, 1],
    [1, 0, 1],
    [0, 1, 1],
    [1, 1, 1],

    [2, 0, 0],
    [2, 1, 0],
    [2, 0, 1],
    [2, 1, 1],
  ],
  faces=[
    [6,7,5],[6,5,4],
    [0,1,3],[0,3,2],
    [4,5,1],[4,1,0],
    [5,7,3],[5,3,1],
    [7,6,2],[7,2,3],
    [4,0,2],[4,2,6],
    [7,11,10],[7,10,5],
    [1,8,9],[1,9,3],
    [5,10,8],[5,8,1],
    [10,11,9],[10,9,8],
    [11,7,3],[11,3,9],
    [5,1,3],[5,3,7],
  ]);
```

Expected:

Image

What we got:

Image

pca006132:
Closes #1191.

The bulk of the code is handling removed faces. Instead of letting them sit in the vector, I remove them and reorder properties dependent on this. This avoids the need to make other functions accept removed halfedges.

While writing this, something that comes to my mind is that we are not really trying our best to handle cases that are not 2-manifold. The current matching method may produce pairs that cross-cross together. I think one way to handle this is to sort the duplicated halfedges according to the face normals, so we will not produce any criss-crossing pairs. Probably for another issue, though.

### DIFF
--- a/src/impl.cpp
+++ b/src/impl.cpp
@@ -312,6 +312,7 @@ void Manifold::Impl::CreateHalfedges(const Vec<ivec3>& triVerts) {
   // which are removed later by RemoveUnreferencedVerts() and Finish().
   const int numEdge = numHalfedge / 2;
 
+  constexpr int removedHalfedge = -2;
   const auto body = [&](int i, int consecutiveStart, int segmentEnd) {
     const int pair0 = ids[i];
     Halfedge& h0 = halfedge_[pair0];
@@ -322,7 +323,7 @@ void Manifold::Impl::CreateHalfedges(const Vec<ivec3>& triVerts) {
       if (h0.startVert != h1.endVert || h0.endVert != h1.startVert) break;
       if (halfedge_[NextHalfedge(pair0)].endVert ==
           halfedge_[NextHalfedge(pair1)].endVert) {
-        h0.pairedHalfedge = h1.pairedHalfedge = -2;
+        h0.pairedHalfedge = h1.pairedHalfedge = removedHalfedge;
         // Reorder so that remaining edges pair up
         if (k != i + numEdge) std::swap(ids[i + numEdge], ids[k]);
         break;
@@ -374,7 +375,7 @@ void Manifold::Impl::CreateHalfedges(const Vec<ivec3>& triVerts) {
   for_each_n(policy, countAt(0), numEdge, [this, &ids, numEdge](int i) {
     const int pair0 = ids[i];
     const int pair1 = ids[i + numEdge];
-    if (halfedge_[pair0].pairedHalfedge != -2) {
+    if (halfedge_[pair0].pairedHalfedge != removedHalfedge) {
       halfedge_[pair0].pairedHalfedge = pair1;
       halfedge_[pair1].pairedHalfedge = pair0;
     } else {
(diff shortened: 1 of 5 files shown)


elalish (commented):
Thanks for digging into this!

src/impl.cpp
```diff
@@ -348,23 +357,111 @@ void Manifold::Impl::CreateHalfedges(const Vec<ivec3>& triVerts) {
...
            });
 #else
-  for (int i = 0; i < numEdge; ++i) body(i, numEdge);
+  int consecutiveStart = 0;
+  for (int i = 0; i < numEdge; ++i)
+    consecutiveStart = body(i, consecutiveStart, numEdge);
 #endif
 
+  if (!removedHalfedges.empty()) {
+    // Remove the marked halfedges
```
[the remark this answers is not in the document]
pca006132: Yeah make sense, I just added the checks elsewhere and keep the removed halfedges in-place.

pca006132:
Weird, I can't reproduce the failure.

pca006132:
OK some uninitialized memory stuff. Here we go again.

src/impl.cpp
```diff
@@ -306,25 +311,30 @@ void Manifold::Impl::CreateHalfedges(const Vec<ivec3>& triVerts) {
...
     while (1) {
       const int pair1 = ids[k];
-      Halfedge h1 = halfedge_[pair1];
+      Halfedge& h1 = halfedge_[pair1];
       if (h0.startVert != h1.endVert || h0.endVert != h1.startVert) break;
       if (halfedge_[NextHalfedge(pair0)].endVert ==
           halfedge_[NextHalfedge(pair1)].endVert) {
-        h0 = {-1, -1, -1};
-        h1 = {-1, -1, -1};
+        h0.pairedHalfedge = h1.pairedHalfedge = -2;
```
elalish, line 325: perhaps we should start using named constants?

test/sdf_test.cpp
```diff
@@ -42,7 +42,7 @@ TEST(SDF, SphereShell) {
       },
       {vec3(-1.1), vec3(1.1)}, 0.01, 0, 0.0001);
 
-  EXPECT_NEAR(sphere.Genus(), 11500, 1000);
+  EXPECT_NEAR(sphere.Genus(), 14235, 1000);
```
elalish, line 45: Does this still look good, or look better? I recall working to remove little bits of infinitely thin geometry on this one.

elalish (approved):

test/sdf_test.cpp
pca006132: I don't see infinitely thin stuff, but I see something that looks like self-intersection:

image

This is probably related to the other thing I mentioned, as there are many  non-2-manifold edges here.

elalish: Ah, right - I think my slick pre-edge collapse thing in the SDF can create even manifold edges. We should probably look at that again and see if there's a better way. I believe without that it's all guaranteed 2-manifold.

Issue #1173, opened 2025-02-28
MeshGL Merge() can crash with access violation


strich:
When using MeshGL.Merge() in my application it causes an instant crash of the application due to an access violation. My application is in C# and so uses bindings and a fence to perform the operation. It may be the cause of the problem rather than the underlying Manifold framework, or it may be my specific use of it too.

Crashdump stacktrace:
![Image](https://github.com/user-attachments/assets/be9f3d2a-cc00-4cec-9178-a70d1c304567)

MeshGL.cs bindings:
```csharp
public MeshGL Merge() => new(manifold_meshgl_merge(_pointer));

[DllImport(DllName, CallingConvention = CallingConvention.Cdecl)]
public static extern IntPtr manifold_meshgl_merge(IntPtr m);
```

Some C# code showing use of Merge():
```csharp
    public static ManifoldNET.MeshGL GetManifoldMeshFromMesh(MeshFilter meshFilter)
    {
        var vertProperties = new List<float>();
        //var triVerts = new List<uint>();
        var localToWorld = meshFilter.transform.localToWorldMatrix;
        var mesh = meshFilter.sharedMesh;
        var noUVs = mesh.uv.Length != mesh.vertices.Length;
        for(int i = 0; i < mesh.vertices.Length; i++)
        {
            var v = mesh.vertices[i];
            var n = mesh.normals[i];
            var uv = noUVs ? Vector2.zero : mesh.uv[i];

            // Convert to world space
            v = localToWorld.MultiplyPoint3x4(v);

            // Convert Unity coordinates to Manifold's coordinate system (Y-Up → Z-Up)
            v = new Vector3(v.x, v.z, -v.y);
            n = new Vector3(n.x, n.z, -n.y);

            vertProperties.Add(v.x);
            vertProperties.Add(v.y);
            vertProperties.Add(v.z);
            vertProperties.Add(n.x);
            vertProperties.Add(n.y);
            vertProperties.Add(n.z);
            vertProperties.Add(uv.x);
            vertProperties.Add(uv.y);
        }

        var triVerts = mesh.triangles.Select(t => (uint)t).ToArray();

        // Ensure triangles are properly wound
        //for(int i = 0; i < triVerts.Count; i += 3)
        //{
        //    (triVerts[i + 1], triVerts[i + 2]) = (triVerts[i + 2], triVerts[i + 1]);
        //}

        var meshGL = new ManifoldNET.MeshGL(vertProperties.ToArray(), triVerts, 8);
[...]

PR #1196, opened 2025-03-24
Changed the delete memory operation to a destructure in the Merge functions as Manifold doesn't own this memory

strich:
As requested at https://github.com/elalish/manifold/issues/1173#issuecomment-2748381121.


pca006132:
Thanks, but will need to fix the format error (extra space) before we can merge this.

PR #1201, opened 2025-03-28
use more precise acos function

pca006132:
Fixes the issue in https://github.com/openscad/openscad/pull/5781

### DIFF
--- a/src/impl.cpp
+++ b/src/impl.cpp
@@ -35,20 +35,94 @@
 namespace {
 using namespace manifold;
 
-// Absolute error <= 6.7e-5
-float fastacos(float x) {
-  float negate = float(x < 0);
-  x = std::abs(x);
-  float ret = -0.0187293;
-  ret = ret * x;
-  ret = ret + 0.0742610;
-  ret = ret * x;
-  ret = ret - 0.2121144;
-  ret = ret * x;
-  ret = ret + 1.5707288;
-  ret = ret * sqrt(1.0 - x);
-  ret = ret - 2 * negate * ret;
-  return negate * 3.14159265358979 + ret;
+/**
+ * Returns arc cosine of 𝑥.
+ *
+ * @return value in range [0,M_PI]
+ * @return NAN if 𝑥 ∈ {NAN,+INFINITY,-INFINITY}
+ * @return NAN if 𝑥 ∉ [-1,1]
+ */
+double sun_acos(double x) {
+  /*
+   * Origin of acos function: FreeBSD /usr/src/lib/msun/src/e_acos.c
+   * Changed the use of union to memcpy to avoid undefined behavior.
+   * ====================================================
+   * Copyright (C) 1993 by Sun Microsystems, Inc. All rights reserved.
+   *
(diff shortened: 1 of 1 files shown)


src/impl.cpp
```diff
@@ -35,20 +35,94 @@
...
+  auto R = [](double z) {
+    double p, q;
+    p = z * (pS0 + z * (pS1 + z * (pS2 + z * (pS3 + z * (pS4 + z * pS5)))));
+    q = 1.0 + z * (qS1 + z * (qS2 + z * (qS3 + z * qS4)));
+    return p / q;
+  };
+  double z, w, s, c, df;
+  uint64_t xx;
+  uint32_t hx, lx, ix;
+  memcpy(&xx, &x, sizeof(xx));
```
pca006132, line 91: The original implementation uses a union of double and `uint64_t`, but technically that is undefined behavior. The standard says that you can only access the last written union variant. Reinterpret cast of pointers will also be undefined behavior. Hence, this uses `memcpy`, but this is typically optimized to nop or register copy by a good enough compiler. (register copy is required when using SSE, because you cannot do integer operations on vector registers...)

pca006132: OK that is just C vs C++. In C type pruning through union is fine, but with C++ it is not fine. We use c++ here, so memcpy is good.

pca006132:
@elalish any idea about the macos build failure? Is it somehow changing the library search path?

src/impl.cpp
```diff
@@ -16,6 +16,7 @@
 
 #include <algorithm>
 #include <atomic>
+#include <cstring>
```
elalish, line 19: What do we need cstring for?

elalish (approved):
It's crazy to me that this acos is more uniform across architectures than the one provided in the standard. Good find.

elalish:
> @elalish any idea about the macos build failure? Is it somehow changing the library search path?

That's weird - I just pulled your branch and did the test cmake thing locally and it's working fine on my mac. Maybe some problem with Actions?

elalish:
Yeah, I don't think it's anything you did: it failed on my last PR when it merged to master, even though it passed first: https://github.com/elalish/manifold/actions/runs/14115518457

src/impl.cpp
pca006132: memcpy

pca006132:
I wonder if they updated macos version?

I can't debug this, maybe you or @kintel have some idea about this.

pca006132:
> It's crazy to me that this acos is more uniform across architectures than the one provided in the standard. Good find.

I think the standard did not define what implementation to be used, nor precision requirements. Libm typically do platform specific optimizations I guess.

kintel:
> I wonder if they updated macos version?
> 
> I can't debug this, maybe you or @kintel have some idea about this.


CMake on macOS got updated to 4.0.0 - I suspect that may have triggered the build issue
I noticed that you're getting CMake from a custom GitHub action, while in OpenSCAD, we get CMake from Homebrew. ..so there may be an easy way to downgrade to 3.x for now. I'm not exactly ecstatic to be an early adopter of V4.0.0 of CMake ;)

PR #1205, opened 2025-03-29
Fix triangulator handling of splinters

### LINKED ISSUE #970: Potential boolean bug

elalish:
Fixes #970 

First is just the tests - the polygon is epsilon-valid, but fails to properly triangulate because it doesn't clip the splinters. I also added @pca006132's original test that produced the problem polygon, but that passes on my machine - perhaps simplification changes improved this? I'm pushing it here to see if some other arch has trouble with it.

Okay, I think I have it fixed now, and the algorithm simplified to boot.

And we no longer have any disabled tests!


src/polygon.cpp
```diff
@@ -411,20 +409,9 @@ class EarClip {
...
     bool IsConvex(double epsilon) const {
-      const int convexity = CCW(left->pos, pos, right->pos, epsilon);
-      if (convexity != 0) {
-        return convexity > 0;
-      }
-      if (la::dot(left->pos - pos, right->pos - pos) <= 0) {
-        return true;
-      }
-      return left->InsideEdge(left->right, epsilon, true);
+      return CCW(left->pos, pos, right->pos, epsilon) >= 0;
```
elalish, line 414: This is interesting - I had been operating under the assumption that I needed to effectively determine a correct perturbation of the epsilon-valid polygon to make it truly valid (non-self-intersecting). This was in fact necessary for my previous triangulator, but since ear clipping proceeds in a naturally manifold way, I tried just letting it clip all degenerate ears. It seems to work great, and vastly simplifies the code. I think the keyholing step may be the only part that's really sensitive to this kind of thing.

src/properties.cpp
```diff
@@ -215,46 +215,47 @@ std::mutex dump_lock;
  * Note that this is not checking for epsilon-validity.
  */
 bool Manifold::Impl::IsSelfIntersecting() const {
-  const double epsilonSq = epsilon_ * epsilon_;
+  const double ep = 2 * epsilon_;
```
elalish, line 218: This is the main change - in line with our degenerate checks and such, we increase epsilon to account for error within these checks themselves. The rest of this function is just renames because I was confused by x and y not being directions.

test/boolean_complex_test.cpp
```diff
@@ -1020,38 +1018,21 @@ TEST(BooleanComplex, SimpleOffset) {
...
-  f.close();
+  Manifold a = ReadMesh64("Offset1.obj");
+  Manifold b = ReadMesh64("Offset2.obj");
   Manifold result = a + b;
   EXPECT_EQ(result.Status(), Manifold::Error::NoError);
   ManifoldParams().selfIntersectionChecks = selfIntersectionChecks;
 }
 
-TEST(BooleanComplex, DISABLED_OffsetSelfIntersect) {
+TEST(BooleanComplex, OffsetSelfIntersect) {
```
elalish, line 1031: These disabled tests are now fixed!

src/polygon.cpp
```diff
@@ -786,7 +771,8 @@ class EarClip {
           above * CCW(start->pos, vert->pos, connector->pos, epsilon_);
       if (vert->pos.x > start->pos.x - epsilon_ &&
           vert->pos.y * above > start->pos.y * above - epsilon_ &&
-          (inside > 0 || (inside == 0 && vert->pos.x < connector->pos.x)) &&
+          (inside > 0 || (inside == 0 && vert->pos.x < connector->pos.x &&
+                          vert->pos.y * above < connector->pos.y * above)) &&
```
elalish, line 775: Here's the second fix - for the disabled sponge test.

pca006132:
Interesting. I think you can try to run the fuzz test. It should report cases that are causing errors in either self-intersection check or triangulation. Let me know if you have trouble running the fuzz test.

Btw we don't need to run the script for the fuzz test now, just compile it with fuzzing enabled and run
```
$ ./test/manifold_fuzz --gtest_filter=Boolean.SimpleCubeRegression
```

pca006132:
Got a failure case:

```
TEST(ManifoldFuzz, SimpleCubeRegression) {
  SimpleCube(
    {CubeOp{{Transform{static_cast<TransformType>(1), {-0.22291143964604709, 0.10000000000000001, -0.15746689630359223}}}, true}, CubeOp{{Transform{static_cast<TransformType>(1), {-0.23746613978666686, 0.10000000000000001, -0.10000000039702957}}}, false}, CubeOp{{}, true}, CubeOp{{Transform{static_cast<TransformType>(1), {-0.10000000000000001, 0.10131119384128703, 0.10000000000000028}}}, true}, CubeOp{{Transform{static_cast<TransformType>(2), {0.10000000000000001, 0.10000000001019704, -0.10000000000000001}}}, true}, CubeOp{{Transform{static_cast<TransformType>(1), {0.10000000097050545, 0.1000000000000008, 0.10000000000000001}}}, false}, CubeOp{{}, true}}
  );
}
```

will translate it into something readable later...

elalish:
No, I can't build with the fuzzer - first I needed to change the clang check to `if(NOT CMAKE_CXX_COMPILER_ID MATCHES "(C|c?)lang")`, and python stubgen had some error, so I had to turn off pybind. Now I'm down to polygon_fuzz.cpp and it says `gettid()` isn't defined and it doesn't like the `pthread_cancel` call either. 

I'm going to go ahead and merge this and we can mess with the fuzzer as a follow-on.

elalish:
@starseeker sure - care to make a PR?

PR #1206, opened 2025-03-29
Fix Cross Section Mirror and add Test

### LINKED ISSUE #1204: Cross Section Mirror behaving unexpectedly
It appears that CrossSection's mirror method is mirroring over the wrong axis. Consider the following example using ManifoldCAD (I have also experienced this behavior using the c-bindings):
```
// For visually identifying each quadrant
const reference_geometry = 
       Manifold.cylinder(5,  1, 1, 90).translate([ 5,  5, 0])
  .add(Manifold.cylinder(10, 1, 1, 90).translate([-5,  5, 0]))
  .add(Manifold.cylinder(15, 1, 1, 90).translate([-5, -5, 0]))
  .add(Manifold.cylinder(20, 1, 1, 90).translate([ 5, -5, 0]))

const tri: SimplePolygon = [[0, 0], [5, 5], [0, 10]];
const mirrored_tri = CrossSection.ofPolygons(tri).mirror([1.0, 1.0]);

// Display the before as the thinner triangle
const tri_manifold = 
       Manifold.extrude(tri, 1)
  .add(Manifold.extrude(mirrored_tri, 5));

const result = reference_geometry.add(tri_manifold);
```
Running this, I get the following:

Image

The mirrored triangle is ending up in the 3rd quadrant as if it had been mirrored over the axis I'd expect to be given by (-1, 1), instead of the provided axis of (1, 1). I quickly glanced at the source and wasn't immediately familiar with the process the code was using to calculate the reflection matrix. However, one thing that did stand out as suspect to me was that it looks like the code takes the absolute value of each component of the supplied axis before doing anything else: `auto n = la::normalize(la::abs(ax));` This seems incorrect to me, assuming I'm understanding the code, because it would result in (-1, 1) and (1,1) being treated identically.

michael-j-faulkner:
Fixes #1204 

Removes an extraneous `la::abs` call that would result in incorrect behavior for axes with mixed sign components (i.e. (-1, 1)) and adds a test to catch it.

Note: what initially caused me to think something was wrong was that mirroring something in quadrant 1 across the axis given by (1,1) ended up in quadrant 3. After looking at the code more intently, it appears this is actually intended behavior and the provided axis vector should be interpreted as being the normal to the axis we are mirroring over, not tangent to the axis. 

I believe this is done to avoid having to calculate any angles and seems like a reasonable thing to do; however, I was caught off guard because in 2D I would typically think of defining a line in terms of it's tangent vector. Unclear if this is just me having some bad intuition or if some additional clarification could also be useful.

### DIFF
--- a/src/cross_section/cross_section.cpp
+++ b/src/cross_section/cross_section.cpp
@@ -519,9 +519,9 @@ CrossSection CrossSection::Scale(const vec2 scale) const {
 }
 
 /**
- * Mirror this CrossSection over the arbitrary axis described by the unit form
- * of the given vector. If the length of the vector is zero, an empty
- * CrossSection is returned. This operation can be chained. Transforms are
+ * Mirror this CrossSection over the arbitrary axis whose normal is described by
+ * the unit form of the given vector. If the length of the vector is zero, an
+ * empty CrossSection is returned. This operation can be chained. Transforms are
  * combined and applied lazily.
  *
  * @param ax the axis to be mirrored over
--- a/src/cross_section/cross_section.cpp
+++ b/src/cross_section/cross_section.cpp
@@ -530,7 +530,7 @@ CrossSection CrossSection::Mirror(const vec2 ax) const {
   if (la::length(ax) == 0.) {
     return CrossSection();
   }
-  auto n = la::normalize(la::abs(ax));
+  auto n = la::normalize(ax);
   auto m = mat2x3(mat2(la::identity) - 2.0 * la::outerprod(n, n), vec2(0.0));
   return Transform(m);
 }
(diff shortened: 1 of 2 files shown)


michael-j-faulkner:
I'm not familiar enough to know what's going wrong with the Mac TBB checks given how little I changed, so I leave it to others, assuming those are actual errors that need fixing.

pca006132:
Those are unrelated, we will fix them soon to unblock these PRs.

pca006132:
you can merge the master branch and push now

elalish (changes requested):
The reason the mirror is defined by its normal is that this way the API is consistent between 2D and 3D - in 3D it takes two tangents to define a plane, but only one normal. If that wasn't clear from the API docs, would you mind updating them?

elalish (approved):
Perfect, thanks!

PR #1207, opened 2025-03-29
set cmake version in macos action

pca006132:


pca006132:
@kintel is right, this seems to be caused by cmake version issue. I guess cmake 4.0 changed how rpath is being resolved. Setting the version to the latest version before 4.0, until someone write a proper fix for this.

kintel:
@pca006132 Let me know if you learn anything about CMake 4 - at some point in the nearish future we'll have to deal with this for OpenSCAD too. Will ping you if we make any changes over there.

PR #1212, opened 2025-03-31
In-loop Status calls were a work-around - see if #1205 solved it.

starseeker:


starseeker:
Cool - looks like that may have gotten it?

elalish (approved):
Awesome, thanks for checking!

PR #1216, opened 2025-04-02
Fix `triangulate` js bindings

jirihon:
There was a missing 3rd parameter for the `Module._Triangulate` function. Emscripten is not able to fill the default parameter values for the bindings automatically.


elalish (approved):
Thanks! Want to check if this made it into the C and Python bindings too?

jirihon:
I'm sorry, but I'm not familiar with the C and Python bindings. But from quick look at the code I guess Python is ok, but C bindings are also missing the `allowConvex` parameter.

PR #1219, opened 2025-04-07
Make API for high precision OBJ I/O method

starseeker:
When reading or writing debugging output, glTF's limitation to float data is insufficient for full capture of the active data.  pca006132 prepared a special OBJ based I/O mechanism to deal with such cases - this PR makes that method an always-available I/O mechanism rather than just debugging only code.

Another advantage this code offers is that it does not rely on the assimp library - even minimalist Manifold library builds are still able to read and write debugging data.


include/manifold/manifold.h
```diff
@@ -221,6 +221,100 @@ struct MeshGLP {
         halfedgeTangent[offset], halfedgeTangent[offset + 1],
         halfedgeTangent[offset + 2], halfedgeTangent[offset + 3]);
   }
+
```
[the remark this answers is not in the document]
starseeker: The intent was to have both Manifold and MeshGL/MeshGL64 be readable and writable as OBJ info - my understanding was that the original code was just for the Manifold type, and MeshGLP is encapsulated as a header-only class with no .cpp implementation?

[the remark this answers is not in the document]
pca006132: The reason we made it header-only is due to template and the methods are relatively short. For longer methods like ReadOBJ and Write, we declare the template specialization in the header and implement them in a `.cpp` file.

include/manifold/manifold.h
```diff
@@ -13,12 +13,12 @@
...
+#include <fstream>
 #include <functional>
-#include <memory>
-
-#ifdef MANIFOLD_EXPORT
+#include <iomanip>
 #include <iostream>
-#endif
+#include <memory>
+#include <string>
```
elalish, line 21: So, we were intentionally keeping things like `<string>` from being included in the base library because they have a tendency to bloat WASM builds considerably. I haven't tested this recently - can you show us the difference in WASM build size before and after this change?

include/manifold/manifold.h
```diff
@@ -430,17 +452,35 @@ class Manifold {
...
+   * it captures information needed for debugging.  The only API guarantee is
+   * that the ReadOBJ method in a given build/release will read in the output
+   * of the WriteOBJ method produced by that release.
+   */
+  static Manifold ReadOBJ(std::istream& stream);
+  static Manifold ReadOBJ(std::string& filename);
+  static Manifold ReadOBJ(const char* filename);
+  std::ostream& WriteOBJ(std::ostream& stream) const;
+  bool WriteOBJ(std::string& filename) const;
+  bool WriteOBJ(const char* filename) const;
```
elalish, line 475: Do we need to export both `Manifold` and `MeshGL`? I'd prefer the simplicity of choosing one or the other. I'd also like to avoid any code duplication.

elalish (changes requested):
Thanks for looking into this.

include/manifold/manifold.h
starseeker: Well, I guess we could go with just Manifold - my rational for adding MeshGL was to handle cases like the one I'm working on, where the input to offset is in fact not manifold.  However, I suppose strictly speaking the only thing needed for that test-wise would be reading and the code is small enough that it could just be part of the test code itself.

include/manifold/manifold.h
starseeker: Um.  I'm not set up to build WASM, so that'd take some time for me to figure out.  What about using a different guard than MANIFOLD_EXPORT?  What I'd like to do is have this I/O capability without needing to rope in assimp - or, if we want to use MANIFOLD_EXPORT for this, maybe switch the MeshIO guard to MANIFOLD_ASSIMP instead?

starseeker: I can see wanting to keep <string> out of the base build if size is an issue - even if it happens to be not-large this go around (I'm guessing that's pretty unlikely) there's no guarantee going forward it would stay that way.  I would like to have the ability to read and write the OBJs both in debug and release build configs and not need to patch to avoid roping in assimp, but beyond that I've not got any really strong preferences.

starseeker:
@elalish Would [91e8cf8](https://github.com/elalish/manifold/pull/1219/commits/91e8cf8e60f9f5578fc7555c0795e5078aa302c3) (or something along those lines) be OK?

src/impl.cpp
```diff
@@ -17,21 +17,19 @@
 #include <algorithm>
 #include <atomic>
 #include <cstring>
+#include <fstream>
+#include <iomanip>
+#include <iostream>
 #include <map>
 #include <optional>
+#include <string>
```
elalish, line 25: We still need to guard these includes.

src/impl.cpp
```diff
@@ -716,16 +718,31 @@ std::ostream& operator<<(std::ostream& stream, const Manifold::Impl& impl) {
...
+  if (!filename.length()) return false;
+  std::ofstream ofile;
+  ofile.open(filename);
+  if (!ofile.is_open()) return false;
+  WriteOBJ(ofile);
+  ofile.close();
+  return true;
+}
+
+bool Manifold::WriteOBJ(const char* filename) const {
```
elalish, line 735: Is this necessary? I thought a std::string would accept a literal anyway.

elalish (changes requested):
Thanks for looking into this.

include/manifold/manifold.h
starseeker: One of my potential use cases for this ability is to be able to automatically dump out reproduction inputs if/when a problem case arises - there's no telling if/when that will happen, so ideally I'd like to not have to worry about recompiling Manifold if I run into an issue.  It's possible I'm misinterpreting what MANIFOLD_DEBUG is for - I had the impression it was to toggle on code that could potentially have performance implications if left on all the time, but maybe there's no real penalty for enabling it beyond library size?

src/impl.cpp
starseeker: Hmm - I'd have to check a raw const char * to a std::string method would work, but it doesn't matter - I can just make the std::string on my end.  I'll yank 'em.

include/manifold/manifold.h
elalish: MANIFOLD_DEBUG should just be about removing code from the release build - it shouldn't actually run anything extra on its own, but just allow things like the verbose flag and extra checks flag to do something when also enabled. I should update our docs to make that more clear. I just looked through the code and found two places that's not currently true, so we'll fix those. 

Meanwhile, yes I think this is a perfect function to have guarded by `MANIFOLD_DEBUG`. Since it'll be public API, I guess just guard the implementation and on the `#else` side just return an error Manifold or a boolean flag.

starseeker: The README indicates the MANIFOLD_DEBUG option incurs about 20% runtime overhead?

starseeker:
Is std::filesystem a problem in WASM?

starseeker:
Although I guess the TBB:ON DEBUG:ON WASM build succeeded... odd...

elalish:
The build succeeded, just a couple of tests failed:

```
[  FAILED  ] 2 tests, listed below:
[  FAILED  ] BooleanComplex.OffsetTriangulationFailure
[  FAILED  ] BooleanComplex.OffsetSelfIntersect
```

We've had some trouble with those two before - they must still be slightly flaky. Go ahead and disable them for now. @pca006132 I guess this must be a small run-to-run non-determinism that's left, at least for WASM-TBB-OFF?

include/manifold/manifold.h
elalish: I'm hoping that's really when `extraChecks` is enabled. If not, we'll make it that way.

starseeker:
I think I scrubbed the stream includes back down to the minimum.

PR #1223, opened 2025-04-10
Remove per-error hacks in favor of a "disable Werror" option

starseeker:
For users who are just looking to build and aren't interested in failing the build due to warnings, provide a MANIFOLD_STRICT option to control it.

By default it will remain on, since developers will generally want to fix errors found, but for user projects or those working with older compilers setting -DMANIFOLD_STRICT=OFF will allow them to avoid build failure when warnings are spurious.

Also should eliminate most of the motivation to suppress specific warnings per-file, although that may still be desirable in some specific instances if the newest compilers are the ones generating verifiably incorrect/unavoidable errors.


elalish (approved):
Thanks!

PR #1225, opened 2025-04-11
added MANIFOLD_ASSERT

### LINKED ISSUE #1224: MANIFOLD_DEBUG performance implications
Testing Manifold v3 integration, I tried two builds - one with MANIFOLD_DEBUG ON and one with OFF.  It does show a performance delta:

MANIFOLD_DEBUG=OFF:
3m20.706s

MANIFOLD_DEBUG=ON:
4m46.463s

elalish:
Fixes #1224 

We incur about a 20% overhead for enabling assertions, which `MANIFOLD_DEBUG` did. Here I've added `MANIFOLD_ASSERT` to control that explicitly. `MANIFOLD_DEBUG` controls the inclusion of all string and stream related headers, and I've now centralized those includes.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -37,6 +37,7 @@ set(
 # Primary user facing options
 option(MANIFOLD_CROSS_SECTION "Build CrossSection for 2D support" ON)
 option(MANIFOLD_DEBUG "Enable debug tracing/timing" OFF)
+option(MANIFOLD_ASSERT "Enable assertions - requires MANIFOLD_DEBUG" OFF)
 option(MANIFOLD_STRICT "Treat compile warnings as fatal build errors" ON)
 option(
   MANIFOLD_DOWNLOADS
--- a/include/manifold/manifold.h
+++ b/include/manifold/manifold.h
@@ -16,11 +16,6 @@
 #include <functional>
 #include <memory>  // needed for shared_ptr
 
-#if defined(MANIFOLD_EXPORT) || defined(MANIFOLD_DEBUG)
-#include <iostream>
-#include <string>
-#endif
-
 #include "manifold/common.h"
 #include "manifold/vec_view.h"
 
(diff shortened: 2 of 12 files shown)


include/manifold/manifold.h
```diff
@@ -16,11 +16,6 @@
 #include <functional>
 #include <memory>  // needed for shared_ptr
 
-#if defined(MANIFOLD_EXPORT) || defined(MANIFOLD_DEBUG)
-#include <iostream>
-#include <string>
-#endif
-
```
starseeker, line 23: Just a thought - should we be hiding the ReadOBJ/WriteOBJ methods behind an #ifdef MANIFOLD_DEBUG as well, since they use std::istream and std::ostream?

starseeker: Actually, nevermind - I'll fix that myself in a separate PR - should be orthogonal to this.

CMakeLists.txt
```diff
@@ -37,6 +37,7 @@ set(
 # Primary user facing options
 option(MANIFOLD_CROSS_SECTION "Build CrossSection for 2D support" ON)
 option(MANIFOLD_DEBUG "Enable debug tracing/timing" OFF)
+option(MANIFOLD_ASSERT "Enable assertions - requires MANIFOLD_DEBUG" OFF)
```
pca006132, line 40: I remember there is a fancy dependent option thing that may be useful here. At least we may want to issue an error if the user only enabled `MANIFOLD_ASSERT` without enabling `MANIFOLD_DEBUG`?

.clang-format
```diff
@@ -1 +1,4 @@
 BasedOnStyle: Google
+PointerAlignment: Left
+ReferenceAlignment: Left
+DerivePointerAlignment: false
```
elalish, line 4: I got tired of clang-format going back and forth on things between versions, and I think this'll help. I had to reformat the whole code base because apparently it was not being consistent about where the `&` was.

CMakeLists.txt
elalish: Good idea - want to throw that into a follow-up PR?

PR #1226, opened 2025-04-11
Fix MANIFOLD_DEBUG guards for Read/Write OBJ api

starseeker:


include/manifold/manifold.h
```diff
@@ -480,8 +477,10 @@ class Manifold {
    * ofile.close();
    * @endcode
    */
+#ifdef MANIFOLD_DEBUG
   static Manifold ReadOBJ(std::istream& stream);
   bool WriteOBJ(std::ostream& stream) const;
+#endif
```
elalish, line 483: I'm honestly curious, from a consumer's perspective - it this more convenient? This way if you write code that uses these and then change your CMake flag, it'll suddenly stop compiling. The old way it would just fail to run. Which do we think is better behavior?

pca006132: Good question. I am a PL guy, so I want more static guarantees. But from a user perspective, failing the compilation may not be the best way...

starseeker: If the headers are going to be behind the MANIFOLD_DEBUG guard, don't these methods need to be too?  Otherwise user code might not compile anyway due to not having the necessary includes?

elalish: Oh right, the point is to not include the stream definitions, so we do have to guard them this way. I suppose finding out about problems earlier (compilation) is better than later (runtime).

src/impl.cpp
```diff
@@ -720,7 +720,6 @@ std::ostream& operator<<(std::ostream& stream, const Manifold::Impl& impl) {
  * Write function.
  */
 Manifold Manifold::ReadOBJ(std::istream& stream) {
```
elalish, line 722: Don't we have to guard the implementations the same way as the headers?

elalish: Maybe pull master and check - I feel like this won't compile as is with `MANIFOLD_DEBUG=OFF`, or else I'm missing something about how these compilers work.

starseeker: Yes we do - there is an ifdef MANIFOLD_DEBUG guarding the whole block on line 685.  I forgot to remove it when I added the method internal versions in the previous commit, so I didn't need to re-add it here.

starseeker: Come to think of it - do all the CI tests have MANIFOLD_DEBUG=ON?

elalish (approved):
Ah, got it, thanks!

PR #1228, opened 2025-04-18
Update README.md

james-bern:
Adding Conversation to users of Manifold.


elalish:
I fixed the format CI - you should be able to pull master and push again and it should work.

PR #1229, opened 2025-04-19
Add Fillet

Trzeth:
Idea
Figure 1. General idea

Fillet operation means to create rounding corner with selected edges. 
**Algorithm Input**: target mesh, radius of rounding corner and selected edges.
**Implement**: Main idea is create a properly cylinder and boolean with target mesh to create the rounding corner.
1. (Figure1. first row) With given radius and tangent planes, the position of cylinder can  be calculate. 
2. (Figure1. second row) Height of cylinder is related 1 neighbourhood, this is done by SplitByPlane.
3. (Unfinished) Boolean to get the intersect edge, and union parts to generate the final output.

Corner cases
Figure 2. Corner cases

There are some corner cases this implement can't handle.
1. (Fig 2. Case 1) If the request radius is larger than interproximal angle, the position of cylinder can't be calculate.
2. (Fig 2. Case 2)  If the selected edge is part of straight line, the rounding corner lack a transition surface to sharp edge.

Snipaste_2025-04-18_11-55-17

Siemens NX fillet operation

Snipaste_2025-04-19_11-24-48

Blender bevel

As for case 1, most software only deal with adjacent faces, can't handle this case with pure geometric way. But, this can be solve by MinkowskiSum with outward and inward offset.

Snipaste_2025-04-18_11-21-35

Siemens NX fillet operation

Snipaste_2025-04-19_11-23-49

Blender bevel

As for case 2, Blender create transition surface while NX can't handle it.


include/manifold/manifold.h
```diff
@@ -422,6 +422,18 @@ class Manifold {
                          const std::vector<Smoothness>& sharpenedEdges = {});
   ///@}
 
+  /** @name Fillet
+   * Create a rounding corner
+   */
+  ///@{
+  static Manifold Fillet(const MeshGL&, double radius,
+                         const std::vector<size_t>& selectedEdges = {});
```
elalish, line 430: So, the API here is tricky - I can see why you're basing this on MeshGL input, since that allows you to select edges, and that's what I did for one of the Smoothing constructors. However, I think I'd prefer to avoid edge-selection APIs as much as we can. I think I'd prefer a Fillet command that fillets all edges beyond a threshold angle, and then it can be non-static. Do you think that's doable?

Trzeth: I think the fillet in pure geometric way won't be that robustness, See corner case 1. It heavily rely on the request radius and adjacent relationship. If so, the result might be out of control.
But I think it would be extremely valuable if there is a full robust fillet method, and i'll keep looking for it.

elalish: Yeah, Manifold's core ethos is robustness - we want find an algorithm that doesn't have edges cases, rather than trying to fix edges cases one by one. In computational geometry, edge cases tend to be the norm, rather than the exception.

Trzeth:
Snipaste_2025-04-24_17-48-24

https://github.com/openscad/openscad/issues/5337 Implement roof() without CGAL

might be able to handle together.

elalish:
Actually I have a straight skeleton algorithm I implemented in Matlab ages ago that I've been meaning to port if you're interested. It is good for polygon offsets, though I'm not sure how that would relate to polygon fillets or especially to mesh fillets?

Trzeth:
> Actually I have a straight skeleton algorithm I implemented in Matlab ages ago that I've been meaning to port if you're interested. It is good for polygon offsets, though I'm not sure how that would relate to polygon fillets or especially to mesh fillets?

@elalish 
Yes! I would like to see your straight skeleton algorithm if you have time to port!

Besides, I think it might require using Clipper to maintain robustness , as otherwise, construct new point could lead to floating-point precision issues and numerical instability. Do you think my consideration exist?

And yes, it's more relevant to offsets. Since this is a sudden thought, I want to keep offset thread clean so put it here .

elalish:
I don't have time to port it at the moment - it'll probably take me a week just to dredge up the old Matlab script from my archives. If you'd be interested in reading through it and trying to port it to C++, I'll be happy to give pointers. The main thing was using SVD to calculate skeleton verts from sets of three edges, so as to also get a numerical uncertainty. I used this to merge verts, which keeps the numerics stable. A Hilbert curve is a great test case, and shows why degeneracy is so common in the straight skeleton.

Trzeth:
Thanks! I'll give it a try.

pca006132:
image

Notes about today's meeting. Consider the polygon with a hole in the middle. If we define fillet using rolling ball (blue) directly, we will get two rounded squares at the bottom because the middle section can't fit a ball inside.

The issue here is that this changes the topology in a potentially weird way, so maybe we can try not to remove stuff when removing it can change the contour. With this interpretation, we get the result at the top.

Trzeth:
My current thought is to process vertex one by one. If there is no solution in the neighbor, we will find it via adjacent vertex circles overlapping. It can simply be proven. This gives a better result than rolling sphere method ends up nothing, see the second figure.

However, I still need to find a way to search for a solution for adjacent overlapping circles. I'm currently thinking about using an inward offset to find a possible solution area.

What do you think about it? CC @elalish @pca006132 

未命名绘图 drawio

未命名绘图-第 3 页 drawio

elalish:
Our `Collider` BVH should give you an efficient way to search for overlapping circles. How about you focus on writing this up as code here so we can start to test it on real polygons. No need for it to handle every edge case yet, but if it runs, we'll be able to discuss it more easily with examples. 

You might consider writing it in such a way that it can incrementally create the fillets, so that if it runs into trouble, we can dump out the intermediate result in a form we can visualize.

pca006132:
And what do you mean by adjacent vertex circle overlapping? Can you give a more detailed example? Or explain why the 3 circles in your first figure are different in sizes.

Trzeth:
未命名绘图-第 1 页 drawio
Sorry for the unclear description, here is what I'm thinking.

Since we are focusing on 2d convex vertex fillet, I am processing convex vertex one by one, calculating the intersect point and bridge with an arc. See Figure 1, this is when the fillet radius is relatively small.
Assume process vertex one by one, I was worried that the adjacent vertex would create overlapping arcs with a large radius, which would result in a strange two arcs combined shape, see Figure 2. But it can be simply proven if overlap exists, there is no solution in the neighborhood. 
See Figure 3, with an increased radius, two circles will end up cutting three edges, after that there is no solution in the neighborhood. So we can rule out e2, that's what I'm trying to say before.

As we rule out e2, we have to find a solution between e1 and e3. The solution might meet the concave vertex and split the shape into two parts, that's what I currently achieved. See Figure 5.

未命名绘图-第 4 页 drawio

But there are also situations where no solution existed between e1 and e3, and need to do a global search for a solution. Last week, I was trying to find it locally but have no idea.
The last meeting @pca006132 suggests to split the polygon into regions and cutting out the no solution part. But I found it hard to properly split it. I tried splitting into a convex polygon, but since the region is partly open, it's hard to get the maximum radius of the region, see Figure 7. Then I was trying to tweak around bisector, but the thought is still unclear.

Another idea is using offset to get a possible solution region nearly, which is also not so clear.

Maybe I should just start coding and get some results at least.

pca006132:
>  The last meeting @pca006132 suggests to split the polygon into regions and cutting out the no solution part. But I found it hard to properly split it. I tried splitting into a convex polygon, but since the region is partly open, it's hard to get the maximum radius of the region, see Figure 7. Then I was trying to tweak around bisector, but the thought is still unclear.

No, maybe I should explain it a bit better.

image

Here, the blue area is where the fixed-radius circle can reach. The issue is that it has a different topology than the thing we started with. So we fill in the green area to make the topology the same as the original, and the final result is the colored region.

image

Here, it is mostly similar to the example above. However, the two red regions on the right are not significant in terms of topology, so we can simply remove them.

The question here is can we figure out a simple way to compute this. Maybe we can split the edges, compute if each segment can touch the circle, from this we can find the edges that are in the same blue component. And for the remaining edges, we can check if they connect different blue components, and if yes we keep them.

> Assume process vertex one by one, I was worried that the adjacent vertex would create overlapping arcs with a large radius, which would result in a strange two arcs combined shape, see Figure 2. But it can be simply proven if overlap exists, there is no solution in the neighborhood.

I think if you have something like figure 2, this means you can't have a circle that touches both e1 and e2 and fits within the polygon. Similarly for e2 and e3. This means that e2 should go.

pca006132:
Some trivial stuff related to the algorithm that you may consider:

image

For every oriented edge, we check the red region for other edges. If there are no other edges, we must be able to fit circles inside the red region, so the edge must be kept. Otherwise, there are several cases: tangent--tangent, tangent--point, point--point. For all cases, we need to ensure that no other edges can intersect the circle, or it will be invalid. For tangents, we split the edge and do the check recursively. For points, we add an edge to join them (and split the polygon). For the degenerate case where there is another edge that touches the circle, we ignore that because it doesn't matter.

pca006132:
@elalish For keeping topology, apart from not connecting different contours, we also need to make sure the chamfer edge is not splitting the contour into two parts where both contain edges that should be kept. This can be done quickly by identifying edges that should be kept and computing consecutive runs of edges that may be discarded in each contour. When we attempt to insert the chamfer edge, we do a binary search to find the run enclosing the edge starting and ending index (the search is slightly more complex than a binary search, but I don't plan to sketch everything here...). Only when we find such a run, the chamfer is fine and can be added.

elalish:
Is some or all of this ready for an initial code review?

Trzeth:
Sorry, it takes longer than i expected. Tomorrow should be ok.

Trzeth:
I think it's ready for review here. https://github.com/Trzeth/manifold/pull/2

cc @elalish @pca006132

project/main.cpp
```diff
@@ -0,0 +1,423 @@
...
+      }
+    }
+  }
+
+  // Construct Reuslt
+
+  return newPoly;
+}
+
+int main() {
```
elalish: Okay, first thing's first - we need to clean up and separate code to make this reviewable. It looks like you have a half-baked 3D fillet algorithm in fillet.cpp and then a 2D fillet algorithm you'd like reviewed in main.cpp. Split these apart to separate branches - don't ask us to review anything that isn't ready or isn't part of the same idea.

project/main.cpp
```diff
@@ -0,0 +1,423 @@
...
+              vec2{6, 5}, vec2{4, 0}, vec2{2, 3}}},
+      TShape{{vec2{0, 0}, vec2{0, 5}, vec2{2, 5}, vec2{0, 8}, vec2{4, 8},
+              vec2{3, 5}, vec2{5, 5}, vec2{5, 0}}};
+
+  const manifold::Polygons poly = TShape;
+  const double radius = 2.3;
+
+  std::vector<PolygonTest> result{poly,
+                                  PolygonTest(VertexByVertex(radius, poly)),
+                                  PolygonTest(RollingBall(radius, poly))};
```
elalish: Second - I think you have both an algorithm and some tests hiding in this odd file? Can you please reorganize it to match our repo? e.g. build the 2D algorithm as an API on `CrossSection` and build the tests using GTest in our testing directory just like all the others?

project/vis.py
```diff
@@ -0,0 +1,161 @@
+import matplotlib.pyplot as plt
+import numpy as np
+from typing import List, Dict, Any
+
+
+
+def plot_polygon(ax, points_list, label, color, closed=True, linestyle='-', marker='o', markersize=5, linewidth=1.5, show_indices=False, index_offset=0.2):
```
elalish: Move this file to the scripts directory - no need for a new project directory.

.vscode/tasks.json
```diff
@@ -0,0 +1,66 @@
...
+    },
+    {
+      "label": "Build, Run & Visualize",
+      // This is the main task that runs all other tasks in sequence.
+      "dependsOn": [
+        "1. Compile C++",
+        "2. Run C++ to Generate Data",
+        "3. Run Python Visualizer"
+      ],
+      // Ensures the tasks run one after another in the order specified above.
```
elalish: Once this is built within our library's existing structure, this shouldn't be necessary anymore. Maybe just some comments in the python file or a README to running the visualizer.

src/impl.h
```diff
@@ -373,6 +373,9 @@ struct Manifold::Impl {
   void CreateTangents(std::vector<Smoothness>);
   void Refine(std::function<int(vec3, vec4, vec4)>, bool = false);
 
+  // fillet.cpp
+  void Fillet(double radius, const std::vector<size_t>& selectedEdges);
```
elalish, line 377: Pull all these changes out to their own branch, as they're not ready for review, right?

Trzeth:
I think it's ready now. Separated the current 2d fillet to https://github.com/Trzeth/manifold/pull/2 this branch, we can move there.
@elalish

elalish:
Thanks - as I just put in your other PR, please open a new one in this repo. And make sure to look through the "changed files" tab to see that what you're proposing for review makes sense as a cohesive idea.

PR #1231, opened 2025-04-20
Add PythonOpenSCAD and AnchorSCAD as 'users'.

owebeeone:
PythonOpenSCAD 2.0,0 now provides mesh render support.  AnchorSCAD (anchorscad-core) will follow.


elalish (approved):
👍

PR #1232, opened 2025-04-20
Remove triProperties

elalish:
Finally finished this refactor - been meaning to deal with this for ages. Also did a bunch of renaming and various other cleanup.


elalish:
@pca006132 Okay, I have this working in that it's passing all the tests, which means I can now start removing triProperties. However, both the CI and my local machine fail on the WASM tests with a heap memory corruption. Do you have any idea what might be happening?

Issue #1233, opened 2025-04-21
Non-determinism between architectures


howardt:
If I run the code below on a Macbook Pro with an arm64 chip (Apple M3 Max), and then on a Debian Linux machine with an AMD Ryzen 9 9950X chip, I get different ordering of the output faces. I didn't try on my Windows machine but I know that it is different from the Mac one. I had thought you had made the library deterministic (and that included: deterministic across architectures? maybe I misunderstood.  This problem messes up my regression test for Blender and is holding up my release of the Manifold solver for Blender. Is there any way to make the ordering consistent across operating system and chip architecture variants?

code:
```
#include <iostream>
#include <iomanip>
#include "manifold/manifold.h"

using namespace manifold;

void dump_meshgl(const MeshGL &mesh)
{
    std::cout << "MeshGL\n";
    std::cout << "NumVert = " << mesh.NumVert() << "\n";
    std::cout << "NumTri = " << mesh.NumTri() << "\n";
    std::cout << "numProp = " << mesh.numProp << "\n";
    std::cout << "vertProperties =";
    std::cout << std::fixed << std::setprecision(4);
    const int verts_per_row = 3;
    int nprop = mesh.numProp;
    for (int i = 0; i < mesh.vertProperties.size(); i++) {
        if ((i % (verts_per_row * nprop)) == 0) {
            std::cout << "\n[" << i << "] ";
        }
        if ((i % mesh.numProp == 0)) {
            std::cout << " / ";
        }
        std::cout << mesh.vertProperties[i] << " ";
    }
    std::cout << "\n";
    if (mesh.mergeFromVert.size() > 0) {
        std::cout << "vert merges: ";
        for (int i = 0; i < mesh.mergeFromVert.size(); i++) {
            if ((i % 10) == 0) {
                std::cout << "\n";
            }
            std::cout << mesh.mergeFromVert[i] << " -> "
              << mesh.mergeToVert[i] << ", ";
        }
        std::cout << "\n";
    }
    std::cout << "tris";
    const int tris_per_row = 5;
    const int ntrivert = mesh.triVerts.size();
    for (int i = 0; i < ntrivert; i++) {
        if ((i % (tris_per_row * 3)) == 0) {
            std::cout << "\n[" << i << "] ";
        }
        if ((i % 3) == 0) {
            std::cout << " / ";
        }
        std::cout << mesh.triVerts[i] << " ";
    }
    std::cout << "\n";
}

void cubecube()
{
[...]

PR #1234, opened 2025-04-22
Fix determinism

Trzeth:
#1233 

unstable sort cause this

https://github.com/elalish/manifold/blob/9756892b0907aa24c0534256aa349e3ab7dd1a5c/src/manifold.cpp#L97-L101


elalish:
Good find! And thanks for adding the test.

PR #1236, opened 2025-04-23
Update ts definitions

Kaladum:
Hi,

thanks for the cool lib! I have been using it for a bit now, and it's really handy!

I just found two small inconveniences in the TS bindings.

- Multiple functions accept arrays of parameters. The TS definitions require these arrays to be modifiable even though the arrays are never modified. I have marked these arrays as read-only.
- The split and splitByPlane methods always return exactly two results (according to the docs). Therefore, they should return a tuple with two items and not an array of items with an arbitrary length. I have changed that.

Best wishes,
Kaladum

### DIFF
--- a/bindings/wasm/manifold-encapsulated-types.d.ts
+++ b/bindings/wasm/manifold-encapsulated-types.d.ts
@@ -821,7 +821,7 @@ export class Manifold {
    *
    * @param cutter
    */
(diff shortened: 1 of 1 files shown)


elalish (approved):
Thank you! I always appreciate a drive-by type cleanup.

PR #1241, opened 2025-04-29
use input faceID

### LINKED ISSUE #1238: Undesired merging of input faces
I guess it is expected that Manifold might merge some coplanar input faces before or after doing its thing?

I would have hoped that if I gave separate faceIDs to some input triangles, that they would stay separate on output, but this appears not to be the case.  E.g., if one argument is a cube where each face is subdivided into 16 squares (all coplanar), then on output they may have all turned into just two triangles.
This is undesired behavior for my users.  What is a way to prevent this from happening?

elalish:
Fixes #1238 

@howardt I believe this fixes the rest of the problem, so now you can use different `faceID`s to maintain edges through simplification. This definitely makes the API more consistent, and caused me to find and fix a few other bugs along the way. Let me know if anything is still missing.

### DIFF
--- a/include/manifold/manifold.h
+++ b/include/manifold/manifold.h
@@ -413,9 +413,9 @@ class Manifold {
   Manifold RefineToTolerance(double) const;
   Manifold SmoothByNormals(int normalIdx) const;
   Manifold SmoothOut(double minSharpAngle = 60, double minSmoothness = 0) const;
-  static Manifold Smooth(MeshGL,
+  static Manifold Smooth(const MeshGL&,
                          const std::vector<Smoothness>& sharpenedEdges = {});
-  static Manifold Smooth(MeshGL64,
+  static Manifold Smooth(const MeshGL64&,
                          const std::vector<Smoothness>& sharpenedEdges = {});
   ///@}
 
--- a/include/manifold/manifold.h
+++ b/include/manifold/manifold.h
@@ -149,10 +149,11 @@ struct MeshGLP {
   /// This matrix is stored in column-major order and the length of the overall
   /// vector is 12 * runOriginalID.size().
   std::vector<Precision> runTransform;
-  /// Optional: Length NumTri, contains the source face ID this
-  /// triangle comes from. When auto-generated, this ID will be a triangle index
-  /// into the original mesh. This index/ID is purely for external use (e.g.
-  /// recreating polygonal faces) and will not affect Manifold's algorithms.
+  /// Optional: Length NumTri, contains the source face ID this triangle comes
+  /// from. Simplification will maintain all edges between triangles with
+  /// different faceIDs. Input faceIDs will be maintained to the outputs, but if
+  /// none are given, they will be filled in with Manifold's coplanar face
+  /// calculation based on mesh tolerance.
   std::vector<I> faceID;
   /// Optional: The X-Y-Z-W weighted tangent vectors for smooth Refine(). If
   /// non-empty, must be exactly four times as long as Mesh.triVerts. Indexed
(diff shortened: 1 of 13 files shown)


include/manifold/manifold.h
```diff
@@ -412,9 +413,9 @@ class Manifold {
   Manifold RefineToTolerance(double) const;
   Manifold SmoothByNormals(int normalIdx) const;
   Manifold SmoothOut(double minSharpAngle = 60, double minSmoothness = 0) const;
-  static Manifold Smooth(const MeshGL&,
+  static Manifold Smooth(MeshGL,
```
pca006132, line 416: Looking at the code, it seems that we will need to copy it because we changed the `faceID` array. I think a better approach would be to keep the old API, do the copy internally, and try to avoid that copying later, e.g. by making another `MeshGL` that uses `VecView` or something.

src/constructors.cpp
```diff
@@ -47,14 +47,28 @@ namespace manifold {
...
                "when supplying tangents, the normal constructor should be used "
                "rather than Smooth().");
 
+  const std::vector<uint32_t> faceIDin = meshGL.faceID;
+  meshGL.faceID.resize(meshGL.NumTri());
+  std::iota(meshGL.faceID.begin(), meshGL.faceID.end(), 0);
+
   std::shared_ptr<Impl> impl = std::make_shared<Impl>(meshGL);
   impl->CreateTangents(impl->UpdateSharpenedEdges(sharpenedEdges));
+  // Restore the original faceID
```
pca006132, line 62: Maybe move this out into a function.

src/face_op.cpp
```diff
@@ -271,33 +271,42 @@ void Manifold::Impl::FlattenFaces() {
...
+                                           refA.faceID < refB.faceID)));
+  };
+
   Vec<size_t> newHalf2Old(halfedge_.size());
   sequence(newHalf2Old.begin(), newHalf2Old.end());
-  stable_sort(
-      newHalf2Old.begin(), newHalf2Old.end(),
-      [&edgeFace](size_t a, size_t b) { return edgeFace[a] < edgeFace[b]; });
+  stable_sort(newHalf2Old.begin(), newHalf2Old.end(), comp);
   newHalf2Old.resize(std::find_if(countAt(0_uz), countAt(halfedge_.size()),
```
pca006132, line 307: btw I'm suddenly thinking if we should change this `find_if` into `lower_bound` to reduce time complexity.

src/face_op.cpp
```diff
@@ -271,33 +271,42 @@ void Manifold::Impl::FlattenFaces() {
...
                }
              });
 
+  auto comp = [&edgeTri, this](size_t a, size_t b) {
+    if (edgeTri[a] == edgeTri[b]) return false;
+    if (edgeTri[a] == kRemove) return false;
+    if (edgeTri[b] == kRemove) return true;
+    const TriRef& refA = meshRelation_.triRef[edgeTri[a]];
+    const TriRef& refB = meshRelation_.triRef[edgeTri[b]];
+    return refA.meshID < refB.meshID || (refA.meshID == refB.meshID &&
```
pca006132, line 298: One neat trick I just found is that we can use `std::tie` to make a tuple and a tuple comparison to do lexicographic comparisons.

src/impl.h
```diff
@@ -155,37 +155,41 @@ struct Manifold::Impl {
     }
 
     Vec<TriRef> triRef;
-    if (!meshGL.runOriginalID.empty()) {
```
pca006132, line 158: So if I understand this correctly, this is mainly just removing the conditional?

howardt:
I built our manifold library from the commit of this PR, and I confirm that I can remove the code that set all the OriginalIDs and with this PR, I get the expected result (things with separate faceIDs are not merged). Thanks! I look forward to this being merged into your main branch.

src/impl.h
elalish: Yes, I realized this block was basically a better `InitializeOriginal`, so instead of conditionally doing this block or that initialization, I removed the other and am using this one all the time, which fixed a bug.

src/face_op.cpp
elalish: Good call - I updated a couple of similar functions I found in sort.cpp too.

src/constructors.cpp
elalish: Yeah, not sure why I didn't do that before...

PR #1244, opened 2025-04-30
Add test target to CMake

diizzyy:
Add manifold_test as test target for unit testing

### DIFF
--- a/test/CMakeLists.txt
+++ b/test/CMakeLists.txt
@@ -60,7 +60,7 @@ endif()
 target_compile_options(manifold_test PRIVATE ${MANIFOLD_FLAGS})
 exportbin(manifold_test)
 
-add_test(test_all manifold_test)
+add_test(NAME manifold_test COMMAND manifold_test)
 
 if(MANIFOLD_FUZZ)
   fuzztest_setup_fuzzing_flags()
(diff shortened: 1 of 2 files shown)


test/CMakeLists.txt
```diff
@@ -60,7 +60,7 @@ endif()
 target_compile_options(manifold_test PRIVATE ${MANIFOLD_FLAGS})
 exportbin(manifold_test)
 
-add_test(test_all manifold_test)
+add_test(NAME manifold_test COMMAND manifold_test)
```
elalish, line 63: Thanks! What does this change? Would you mind adding a line in the README to explain how to use it?

diizzyy: Hi,

It generates the target test, (make test etc)
See https://cmake.org/cmake/help/book/mastering-cmake/chapter/Testing%20With%20CMake%20and%20CTest.html#how-does-cmake-facilitate-testing

I don't think there's much to add as this is common practice when using CMake

Best regards,
Daniel

elalish: Well, as much as I use CMake, I still don't know a million of its details like this one. It looks from that link like this allows someone to run `make test` instead of `./build/test/manifold_test`, correct? I think the README should be updated to tell people that - perhaps even update our CI scripts to work that way as a small simplification.

diizzyy: Indeed, you can now run "make test" (or whatever generator you're using)

diizzyy: PR for change in readme, https://github.com/elalish/manifold/pull/1245

pca006132:
Maybe we can just use ctest for our CI? And mention this in our readme. I don't think we should say `make test`, people can use things like ninja.

PR #1247, opened 2025-05-01
Fix intersection behavior in CrossSection::BatchBoolean

tomasf:
`CrossSection::BatchBoolean` builds a big list of clip paths from the tail of the `crossSections` list and then performs the boolean operation with the head. This works fine for union and difference, but when _intersecting more than two_ cross sections, it produces incorrect results.

This PR adds a special case for `OpType::Intersect` where `C2::BooleanOp` is instead called repeatedly for every part of the tail.

I'm not a C++ guy, so take the code with a scoop of salt.

### DIFF
--- a/src/cross_section/cross_section.cpp
+++ b/src/cross_section/cross_section.cpp
@@ -380,8 +380,8 @@ CrossSection CrossSection::BatchBoolean(
   if (op == OpType::Intersect) {
     auto res = subjs->paths_;
     for (size_t i = 1; i < crossSections.size(); ++i) {
-      res = C2::BooleanOp(C2::ClipType::Intersection, C2::FillRule::Positive, res,
-                          crossSections[i].GetPaths()->paths_, precision_);
+      res = C2::BooleanOp(C2::ClipType::Intersection, C2::FillRule::Positive,
+                          res, crossSections[i].GetPaths()->paths_, precision_);
     }
     return CrossSection(shared_paths(res));
   }
--- a/src/cross_section/cross_section.cpp
+++ b/src/cross_section/cross_section.cpp
@@ -376,6 +376,16 @@ CrossSection CrossSection::BatchBoolean(
     return crossSections[0];
 
   auto subjs = crossSections[0].GetPaths();
+
+  if (op == OpType::Intersect) {
+    auto res = subjs->paths_;
+    for (size_t i = 1; i < crossSections.size(); ++i) {
+      res = C2::BooleanOp(C2::ClipType::Intersection, C2::FillRule::Positive, res,
+                          crossSections[i].GetPaths()->paths_, precision_);
+    }
+    return CrossSection(shared_paths(res));
+  }
+
   int n_clips = 0;
   for (size_t i = 1; i < crossSections.size(); ++i) {
     n_clips += crossSections[i].GetPaths()->paths_.size();


Trzeth:
Yes, I think you are right.
Here is my test example. A square intersect with three non-overlapping circles.
The desired result should be only one half circle, but come out three.
image

```js
const result = CrossSection.intersection(
  [
    CrossSection.square(100)
    ,CrossSection.circle(30,30).translate(-10,30)
    ,CrossSection.circle(20,30).translate(110,20)
    ,CrossSection.circle(40,30).translate(50,110)
  ]).extrude(1)
```

tomasf:
> Yes, I think you are right. Here is my test sample. A square intersect with three non-overlapping circles. The desired result should be only one half circle, but come out three.

I actually think the correct result for your code should be empty. There's no area covered by all four shapes. The result should be equivalent to:

```javascript
const result = CrossSection.intersection([
  CrossSection.square(100),
  CrossSection.intersection([
    CrossSection.circle(30,30).translate(-10,30),
    CrossSection.intersection([
      CrossSection.circle(20,30).translate(110,20),
      CrossSection.circle(40,30).translate(50,110)
    ])
  ])
]).extrude(1)
```

tomasf:
This code makes a Reuleaux triangle:
```javascript
const size = 10.0
const offset = size / Math.sqrt(3)
const result = CrossSection.intersection([
  CrossSection.circle(size)
    .translate([offset, 0]),
  CrossSection.circle(size)
    .translate([offset, 0])
    .rotate(120),
  CrossSection.circle(size)
    .translate([offset, 0])
    .rotate(240)
]).extrude(1)
```

## Expected
Screenshot 2025-05-02 at 09 07 43

## Actual
Screenshot 2025-05-02 at 09 07 56

Trzeth:
My bad, the output CrossSection indeed should be empty.
I believe your code can fix it but I'm not sure it is the best way.
Because what manifold is just collect CrossSections path and send to Clipper2 to handle it. Do you think this problem might come from Clipper2?

tomasf:
> My bad, the output CrossSection indeed should be empty. I believe your code can fix it but I'm not sure it is the best way. Because what manifold is just collect CrossSections path and send to Clipper2 to handle it. Do you think this problem might come from Clipper2?

I'm also not sure it's the best way. As far as I can tell, Clipper2 has no direct way to intersect more than two shapes, so Manifold would need to make multiple intersection calls, one way or another.

Trzeth:
Actually Clipper2 can handle multi shapes. But the different is Clipper2 handles the input as **a whole set**, not process one by one with order as we thought. So manifold just keeps this feature. 

So I think this is not a bug. Because this apparently can be useful under some circumstances, and no other exposed api can replace this.
But it is indeed misleading since the behavior is not as the same as Manifold::BatchBoolean.

Rather than throw this feature completely, maybe the CrossSection::BatchBoolean can have another name? Like whole boolean or something? 
Or change the Manifold::BatchBoolean behavior not doing it one by one but as a whole.

```js
// Process CrossSection as a whole, output three parts
const result = CrossSection.intersection(
  [
    CrossSection.square(100)
    ,CrossSection.circle(30,30).translate(-10,30)
    ,CrossSection.circle(20,30).translate(110,20)
    ,CrossSection.circle(40,30).translate(50,110)
  ]).extrude(1)

// Process Manifold one by one, output nothing
 const result = Manifold.intersection([
    CrossSection.square(100).extrude(1)
    ,CrossSection.circle(30,30).translate(-10,30).extrude(1)
    ,CrossSection.circle(20,30).translate(110,20).extrude(1)
    ,CrossSection.circle(40,30).translate(50,110).extrude(1)
  ])
```

tomasf:
> So I think this is not a bug. Because this apparently can be useful under some circumstances, and no other exposed api can replace this. But it is indeed misleading since the behavior is not as the same as Manifold::BatchBoolean.

As far as I can tell, the current behavior is the same as an intersection with a union of the tail. For your example, the proper replacement for the same output is:

```javascript 
const result = CrossSection.intersection(
  [
    CrossSection.square(100),
    CrossSection.union([
      CrossSection.circle(30,30).translate(-10,30),
      CrossSection.circle(20,30).translate(110,20),
      CrossSection.circle(40,30).translate(50,110)
    ])
  ]).extrude(1)
```

Trzeth:
You are right, I think it's the best way to fix the misleading behaviour.

pca006132 (approved):

elalish (commented):
Thanks! One thing I would have asked for is a test for this - note that codecov shows your addition doesn't have any coverage.

PR #1252, opened 2025-05-06
Make OriginalID tracking optional and opt-in

### LINKED ISSUE #1217: opt-in mesh id for primitives
Faces with different mesh ids can't be collapsed, this is to allow users to identify face to mesh mapping. However, most of the time users don't need this, so this is hurting performance. We should make primitives and maybe MeshGL import using a default -1 mesh id, which can be collapsed with other faces with mesh id -1.

elalish:
Fixes #1217

This could be construed as a breaking change, though I think it's minor enough I'm going to let it slide in a minor release. The idea is that mesh ID tracking can slow things down for complicated trees of Booleans with lots of coplanar faces, and we shouldn't bother by default so folks who are not using properties or faceIDs can skip all of that. However, IDs are still necessary to make properties and such work, since we need to retain edges separating input meshes to avoid interpolating between inconsistent properties. 

With this change, if you're tracking IDs / materials / properties, you'll need to call `AsOriginal()` on each Manifold, but it's natural to do so since on construction you already have to be calling `OriginalID()` to record the value in whatever material map you use. 

`AsOriginal()` also now takes an optional id parameter - with none given or negative it has the old behavior of making a new unique ID. But now you can set zero to opt back out of mesh ID tracking and allow more simplification. Likewise you can set multiple manifolds to have the same ID. This would not be a good idea if using vertex properties, but could be helpful if, e.g. your materials consist entirely of uniform values, like a color.


elalish:
@pca006132 I'm curious if we can create a test that actually runs faster with this change - I've yet to find one. I think that having to run `MarkCoplanar` after each Boolean op may be taking more time that is gained from having fewer faces to operate on. I'll be in favor of this if it actually aids performance, but otherwise the slight increase in API complexity may not be justified.

jirihon:
> I don't think you need to call `AsOriginal` in that case. We are just opting out of `OriginID` tracking for primitives.

It seems ok for primitives only. Thanks for clarification.

elalish:
Superseded by #1790.

PR #1256, opened 2025-05-12
wasm: use string status

### LINKED ISSUE #1213: status() doesn't work in JS bindings
I was actually trying to update the docs, but found that `status()` is missing from our TS types, though it is present in our JS bindings. However, calling it only produces `{}` - I'm not sure why. The error enum is defined, so it seems like it ought to work. Any ideas @pca006132?

pca006132:
Closes #1213.

### DIFF
--- a/bindings/wasm/bindings.js
+++ b/bindings/wasm/bindings.js
@@ -224,8 +224,8 @@ Module.setup = function() {
     removeFunction(wasmFuncPtr);
 
     const status = out.status();
-    if (status.value !== 0) {
-      throw new Module.ManifoldError(status.value);
+    if (status !== 'NoError') {
+      throw new Module.ManifoldError(status);
     }
     return out;
   };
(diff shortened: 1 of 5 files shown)


bindings/wasm/bindings.js
```diff
@@ -415,36 +415,36 @@ Module.setup = function() {
   Module.ManifoldError = function ManifoldError(code, ...args) {
     let message = 'Unknown error';
     switch (code) {
-      case Module.status.NonFiniteVertex.value:
+      case 'NonFiniteVertex':
```
elalish, line 418: Could we get rid of this whole switch statement if we used the descriptor strings as the status strings directly? e.g. use "Non-finite vertex" instead of "NonFiniteVertex"?

pca006132: Yes, but not sure if it is the best approach. I feel like the status name should be separated from the actual error message, and while we are using strings, what we want is just the status name.

elalish: To me the only separation between status name and message is whether it contains spaces. C++ is annoying because enums can't have spaces and can't be coerced into strings, so you end up needing a bunch of duplication. If we're just using strings here anyway, seems like we can short circuit that. But you're welcome to merge this if you disagree - it's not a big deal.

pca006132: I think keeping closer to the C++ API is the right thing to do here, it minimizes surprises.

Issue #1262, opened 2025-05-26
Cracks after simplification


tomasf:
I input an OBJ file with a large number of vertices and `Simplify` with a tolerance of 0.005. Large cracks appear in the result.

```c++
#include <fstream>
#include <manifold/manifold.h>

using namespace manifold;

int main(int argc, const char * argv[]) {
    const char *inputFile = "simplify-input.obj";
    const char *outputFile = "simplify-output.obj";

    std::ifstream ifile;
    ifile.open(inputFile);
    Manifold model = Manifold::ReadOBJ(ifile);
    ifile.close();

    model = model.Simplify(0.005);

    std::ofstream ofile;
    ofile.open(outputFile);
    model.WriteOBJ(ofile);
    ofile.close();

    return 0;
}
```

## Before simplification
https://www.dropbox.com/scl/fi/fntvzo3k05hw8yiq1415d/simplify-input.obj?rlkey=2lugs155lhruhhwit9d2v5p94&st=9uu9qf82&dl=1 (108 MB)

<img width="925" alt="Image" src="https://github.com/user-attachments/assets/3edb6ea1-d9d3-496a-93bc-aad0704e5a76" />

## After simplification
https://www.dropbox.com/scl/fi/me354g7rgz1yb7x0l1if1/simplify-output.obj?rlkey=zs4yq4vxt6aqxn92620yfkwb0&st=5o4ucj8o&dl=1 (201 KB)

<img width="925" alt="Image" src="https://github.com/user-attachments/assets/2c1e3ce6-517a-4cdb-b6de-ee57d5097907" />

pca006132:
It looks to me that your original mesh has a crack (maybe touching faces?) and simplify somehow enlarged that.

This is not ideal, we should look into it. But I'm also thinking if it is possible to avoid this in all cases, even when it is formed by two meshes that are side by side for example.

elalish:
Would you care to make a PR adding a test? Ideally starting from the cylinder/box so we can see exactly what's going on? How big are those cracks relative to your 0.005 tolerance?

tomasf:
I finally managed to reproduce it directly against Manifold in C++, and I created a test case that checks the result against expected volume and surface area: #1263

[...]

PR #1263, opened 2025-05-27
Add test for Simplify adding cracks

tomasf:
See #1262

`Simplify` added weird cracks to my model. This failing test reproduces this case and checks the volume and surface area of the output of `Simplify`.


tomasf:
Makes sense. Done!

elalish (approved):
Thanks!

PR #1271, opened 2025-06-01
Use union-find for winding number

### LINKED ISSUE #1267: Try flood fill for faster winding number computation

pca006132:
Based on #1269, closes #1267. I tried to play with single-threaded union-find, it was slightly slower than parallel computation of all winding numbers (1 thread vs 20 threads!), so I grabbed a lock-free implementation of union-find instead. It is now about 5 times faster than the old approach.

Instead of flood fill, I just compute one value for each connected component, and find that value when we need the winding number (in `boolean_result.cpp`). This reduces the memory consumption quite a bit.

This code is not yet finished, there are minor changes needed, e.g. use the lock-free version of union-find everywhere and maybe change to using `std::unordered_map` instead of the sorted array and `std::set`. But I don't think there should be much performance difference. I am putting it here for review first, while I need to work on something else.

Performance-wise:

- high resolution sphere (nTri = 8388608, 32 threads): 1.51907s -> 1.43618s
- high resolution sphere (nTri = 8388608, 20 threads, another PC): 1.60584s -> 1.42585s
- Ember benchmark single threaded: 74.368ms -> 63.667ms
- Ember benchmark 16 threads: 16.720ms -> 15.787ms


pca006132:
Thinking about it, flood fill may be fine, will try and check the performance.

src/boolean3.cpp
```diff
@@ -394,60 +393,113 @@ std::tuple<Vec<int>, Vec<vec3>> Intersect12(const Manifold::Impl& inP,
...
+#endif
+  {
+    for (int v = 0; v < static_cast<int>(a.vertPos_.size()); v++)
+      components.insert(UnionFindKeyValue{uF.cfind(v), v});
+  }
+  Vec<UnionFindKeyValue> w03;
+  w03.reserve(components.size());
+  for (const auto& component : components) w03.push_back(component);
+
+  Vec<int> verts(w03.size());
```
elalish: Can we avoid constructing this vector entirely and directly look up the component values instead?

src/boolean3.cpp
```diff
@@ -394,60 +393,113 @@ std::tuple<Vec<int>, Vec<vec3>> Intersect12(const Manifold::Impl& inP,
...
+        [&](int v) {
+          componentsShared.local().insert(UnionFindKeyValue{uF.find(v), v});
+        });
+    componentsShared.combine_each([&](const std::set<UnionFindKeyValue>& data) {
+      components.insert(data.begin(), data.end());
+    });
+  } else
+#endif
+  {
+    for (int v = 0; v < static_cast<int>(a.vertPos_.size()); v++)
```
elalish: can we skip the cast by iterating a uint instead?

src/boolean3.cpp
```diff
@@ -394,60 +393,113 @@ std::tuple<Vec<int>, Vec<vec3>> Intersect12(const Manifold::Impl& inP,
...
+           countAt(a.halfedge_.size()), [&](size_t edge) {
+             const Halfedge& he = a.halfedge_[edge];
+             if (!he.IsForward()) return;
+             // check if the edge is broken
+             auto it = std::lower_bound(
+                 p1q2.begin(), p1q2.end(), edge,
+                 [index](const std::array<int, 2>& a, size_t e) {
+                   return a[index] < static_cast<int>(e);
+                 });
+             if (it == p1q2.end() || (*it)[index] != static_cast<int>(edge))
```
elalish: why are these casts necessary?

src/boolean3.cpp
pca006132: Yeah, I was coding at midnight and forgot this.

src/boolean3.cpp
pca006132: I think yes, forgot why I did this.

src/boolean3.cpp
pca006132: Will not be needed if I iterate over signed indices.

src/utils.h
```diff
@@ -136,61 +136,100 @@ class ConcurrentSharedPtr {
...
-struct UnionFind {
-  Vec<I> parents;
-  // we do union by rank
-  // note that we shift rank by 1, rank 0 means it is not connected to anything
-  // else
-  Vec<R> ranks;
-
-  UnionFind(I numNodes) : parents(numNodes), ranks(numNodes, 0) {
-    sequence(parents.begin(), parents.end());
+// from https://github.com/wjakob/dset, changed to add connected component
```
starseeker: Since dset uses zlib license, should that be mentioned here?  Or at least somewhere?

starseeker: In particular - "3. This notice may not be removed or altered from any source distribution."

src/boolean3.cpp
elalish: Also, we're reusing `a` - can we pick a more descriptive name? It's hard to tell here what `a` and `e` are or why casting is needed.

src/utils.h
elalish: Agreed, and thanks for pointing that out.

elalish (changes requested):
Couple of small things, and then you can feel free to merge this.

src/utils.h
pca006132: Ah, I somehow forgot restriction 3, I think I kept 1 and 2 in mind. Will fix.

src/boolean3.cpp
pca006132: OK will fix, this is not intended.

elalish (approved):
Thanks!

pca006132:
@elalish btw do we want a patch release for these two PRs? 25% performance improvement is quite a bit.

elalish:
Yeah, that's fair. Maybe give it a week to see if we can fix any bugs while we're at it? I was going to look at that swap edges one, and maybe a triangulator bug too.

PR #1273, opened 2025-06-04
Make GCC floating-point operations standard conforming

### LINKED ISSUE #1266: Boolean.SimpleCubeRegression fails under -march=x86-64-v3 with gcc14
Just a note, should investigate later. Simply enlarging bounding boxes does not help.

pca006132:
Closes #1266. GCC defaults to `-ffp-contract=fast`, i.e., it will optimize things into FMA. It will also perform things with excess precision if that is faster, and is now disabled via `-fexcess-precision=standard`. I don't see any performance degradation.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -188,7 +188,7 @@ else()
     # disable fp math optimizations, e.g. FMA for supported architectures, as
     # this changes floating-point results
     # clang defaults to -ffp-contract=off so we don't have to set that
-    list(APPEND MANIFOLD_FLAGS -ffp-contract=off)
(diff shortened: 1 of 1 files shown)


pca006132:
I'm thinking if it will make sense to dump all the MeshGL64 into obj files (in our special format) and check if the results are the same across all platforms, for better determinism.

elalish (approved):
Thanks! Dumping all the MeshGL64s sounds like a major bloat to the git size of the library, but perhaps we should pick a few?

pca006132:
Probably, will open an issue about that.

PR #1279, opened 2025-06-06
faster halfedge pairing for large meshes

pca006132:
For larger vertex count, we separate the ids into slices for halfedges with the same smaller vertex.
We first copy them there (as HalfedgePairData), and then do sorting locally for each slice.
This helps with memory locality and is faster for larger meshes.

For large meshes, this reduces the time for preparing IDs by about half. EMBER benchmark results:

```
Old:
Total time
  avg : 17.035
  min : 2.130
  max : 177.936
Converting MeshGL to Manifold
  avg : 6.502
  min : 1.222
  max : 30.621
Manifold boolean
  avg : 10.353
  min : 0.828
  max : 157.637
Converting Manifold to MeshGL
  avg : 0.180
  min : 0.006
  max : 1.446

New
Total time
  avg : 14.775
  min : 1.699
  max : 168.734
Converting MeshGL to Manifold
  avg : 5.904
  min : 0.961
  max : 23.307
Manifold boolean
  avg : 8.686
  min : 0.619
  max : 151.385
Converting Manifold to MeshGL
  avg : 0.185
  min : 0.004
  max : 1.291
```

Basically improves on both average and worst-case time. And yields significant improvement for `perfTest` as well (~0.1s).

And bonus for triangulation, basically just avoid `std::function` and tuned the loop iteration slightly. Gives ~17% performance improvement for complex triangulation, e.g. Zebra and Zebra3.

### DIFF
--- a/src/impl.cpp
+++ b/src/impl.cpp
@@ -372,8 +372,8 @@ struct PrepHalfedges {
   void operator()(const int tri) {
     const ivec3& props = triProp[tri];
     for (const int i : {0, 1, 2}) {
-      const int j = (i + 1) % 3;
-      const int k = (i + 2) % 3;
+      const int j = NextHalfedge(i);
+      const int k = NextHalfedge(j);
       const int e = 3 * tri + i;
       const int v0 = useProp ? props[i] : triVert[tri][i];
       const int v1 = useProp ? props[j] : triVert[tri][j];
@@ -409,17 +409,19 @@ void Manifold::Impl::CreateHalfedges(const Vec<ivec3>& triProp,
       // For small vertex count, it is faster to just do sorting
       Vec<uint64_t> edge(numHalfedge);
       auto setEdge = [&edge](int e, int v0, int v1) {
-        edge[e] = uint64_t(v0 < v1 ? 1 : 0) << 63 |
-                  ((uint64_t)std::min(v0, v1)) << 32 | std::max(v0, v1);
+        edge[e] = static_cast<uint64_t>(v0 < v1 ? 1 : 0) << 63 |
+                  (static_cast<uint64_t>(std::min(v0, v1))) << 32 |
+                  static_cast<uint64_t>(std::max(v0, v1));
       };
-      if (triVert.empty())
+      if (triVert.empty()) {
         for_each_n(policy, countAt(0), numTri,
                    PrepHalfedges<true, decltype(setEdge)>{halfedge_, triProp,
                                                           triVert, setEdge});
-      else
+      } else {
         for_each_n(policy, countAt(0), numTri,
                    PrepHalfedges<false, decltype(setEdge)>{halfedge_, triProp,
                                                            triVert, setEdge});
+      }
       sequence(ids.begin(), ids.end());
       stable_sort(ids.begin(), ids.end(), [&edge](const int& a, const int& b) {
         return edge[a] < edge[b];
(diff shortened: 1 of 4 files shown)


src/impl.cpp
```diff
@@ -351,6 +351,39 @@ void Manifold::Impl::DedupePropVerts() {
...
+  VecView<Halfedge> halfedges;
+  const VecView<ivec3> triProp;
+  const VecView<ivec3> triVert;
+  F& f;
+
+  void operator()(const int tri) {
+    const ivec3& props = triProp[tri];
+    for (const int i : {0, 1, 2}) {
+      const int j = (i + 1) % 3;
+      const int k = (i + 2) % 3;
```
elalish: We have a `Next3` method for this, I believe.

src/impl.cpp
```diff
@@ -366,35 +399,74 @@ void Manifold::Impl::CreateHalfedges(const Vec<ivec3>& triProp,
...
+  int vertCount = static_cast<int>(vertPos_.size());
+  Vec<int> ids(numHalfedge);
+  {
+    ZoneScopedN("PrepHalfedges");
+    if (vertCount < (1 << 18)) {
+      // For small vertex count, it is faster to just do sorting
+      Vec<uint64_t> edge(numHalfedge);
+      auto setEdge = [&edge](int e, int v0, int v1) {
+        edge[e] = uint64_t(v0 < v1 ? 1 : 0) << 63 |
+                  ((uint64_t)std::min(v0, v1)) << 32 | std::max(v0, v1);
```
elalish: static_cast?

src/impl.cpp
```diff
@@ -366,35 +399,74 @@ void Manifold::Impl::CreateHalfedges(const Vec<ivec3>& triProp,
...
+      };
+      if (triVert.empty())
+        for_each_n(policy, countAt(0), numTri,
+                   PrepHalfedges<true, decltype(setEdge)>{halfedge_, triProp,
+                                                          triVert, setEdge});
+      else
```
elalish: brackets might be a good idea.

src/impl.cpp
```diff
@@ -403,6 +475,8 @@ void Manifold::Impl::CreateHalfedges(const Vec<ivec3>& triProp,
   const auto body = [&](int i, int consecutiveStart, int segmentEnd) {
     const int pair0 = ids[i];
     Halfedge& h0 = halfedge_[pair0];
+    const int startVert = h0.startVert;
+    const int endVert = h0.endVert;
```
elalish: is this change necessary?

src/tree2d.cpp
```diff
@@ -37,14 +37,16 @@ namespace manifold {
...
   };
-  CmpFn cmpy = [](const PolyVert& a, const PolyVert& b) {
+  auto cmpy = [](const PolyVert& a, const PolyVert& b) {
     return a.pos.y < b.pos.y;
   };
-  manifold::stable_sort(points.begin(), points.end(), sortX ? cmpx : cmpy);
+  if (sortX)
+    manifold::stable_sort(points.begin(), points.end(), cmpx);
+  else
+    manifold::stable_sort(points.begin(), points.end(), cmpy);
```
elalish, line 49: Is this different than before?

elalish (approved):
Thanks! Just a few minor comments.

src/tree2d.cpp
pca006132: they are the same

src/impl.cpp
pca006132: no, it was for something else before and I found it doesn't work

src/impl.cpp
```diff
@@ -372,8 +372,8 @@ struct PrepHalfedges {
   void operator()(const int tri) {
     const ivec3& props = triProp[tri];
     for (const int i : {0, 1, 2}) {
-      const int j = (i + 1) % 3;
-      const int k = (i + 2) % 3;
+      const int j = NextHalfedge(i);
+      const int k = NextHalfedge(j);
```
elalish: I think Next3 is faster than NextHalfedge when i < 3 - avoiding the mod?

src/tree2d.cpp
elalish: In that case, I think I like the ternary better - a bit cleaner.

pca006132: Oh sorry, by the same I mean the behavior. Performance is quite different.

elalish: Weird, I'd like to understand why that is - does the old way somehow stop the compiler from optimizing?

elalish (approved):

Issue #1281, opened 2025-06-09
Python binding lacks new `Manifold::Simplify()` method


jonathanhogg:
The new `Manifold::Simplify()` method added in v3.1.0 doesn't seem to have made it into the Python bindings.

elalish:
Strange, I see it in there. It's even being tested in `all_apis.py`. Can you show us what you mean?

jonathanhogg:
Sure:

```
nuisance@ python3 -m venv test
nuisance@ test/bin/pip3 install manifold3d
Collecting manifold3d
  Using cached manifold3d-3.1.1-cp313-cp313-macosx_11_0_arm64.whl.metadata (18 kB)
Collecting numpy>=1.26.0b1 (from manifold3d)
  Using cached numpy-2.3.0-cp313-cp313-macosx_14_0_arm64.whl.metadata (62 kB)
Using cached manifold3d-3.1.1-cp313-cp313-macosx_11_0_arm64.whl (734 kB)
Using cached numpy-2.3.0-cp313-cp313-macosx_14_0_arm64.whl (5.1 MB)
Installing collected packages: numpy, manifold3d
Successfully installed manifold3d-3.1.1 numpy-2.3.0
nuisance@ test/bin/python3
Python 3.13.4 (main, Jun  3 2025, 15:34:24) [Clang 17.0.0 (clang-1700.0.13.3)] on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>> import manifold3d
>>> manifold3d.Manifold.batch_boolean
<nanobind.nb_func object at 0x10100a350>
>>> manifold3d.Manifold.simplify
Traceback (most recent call last):
  File "<python-input-4>", line 1, in <module>
    manifold3d.Manifold.simplify
AttributeError: type object 'manifold3d.Manifold' has no attribute 'simplify'
>>> 
```

I *think* you're maybe seeing `CrossSection.simplify()`?

Fix looks like all it needs is:

```patch
diff --git a/bindings/python/examples/all_apis.py b/bindings/python/examples/all_apis.py
index aaa20f7..47157c4 100644
--- a/bindings/python/examples/all_apis.py
+++ b/bindings/python/examples/all_apis.py
@@ -92,6 +92,7 @@ def all_manifold():
     i = m.original_id()
     p = m.get_tolerance()
     pp = m.set_tolerance(0.0001)
+    m = m.simplify(0.001)
     c = m.project()
     m = m.refine(2)
[...]

PR #1282, opened 2025-06-10
Add missing Manifold.simplify() method to Python API (#1281)

jonathanhogg:
* add binding to manifold3d.cpp

* add test call in all_apis.py


pca006132:
Thanks.

PR #1291, opened 2025-06-25
fix pyi vert_properties type

### LINKED ISSUE #1290: The .pyi stub file appears to be incorrect
The .pyi file in the package is as follows:

```pyi
class Mesh:
    @property
    def vert_properties(self) -> Annotated[ArrayLike, dict(dtype='float32', order='C', writable=False)]: ...
```

The .pyi file generated via `nanobind-stubgen manifold3d` is as follows:

```pyi
class Mesh:
    @property
    def vert_properties(self) -> numpy.typing.NDArray:
```

Although the latter is inferior in many ways, at least regarding `vert_properties`, it appears more usable—Pylance correctly handles cases like `mesh.vert_properties[:, :3]`.

Trzeth:
Fix #1290.

Before:
PyCharm
image
VSCode
image

After
image
image


elalish (approved):
Thanks!

pca006132:
I think there are more stubs that should be fixed, but this is probably ok for now.

Trzeth:
I'm gonna check further. Let me solve it in this pr.

Trzeth:
It's good to go.

PR #1296, opened 2025-07-13
skip halfedges that are removed

pca006132:
Fixes https://github.com/godotengine/godot/issues/107994.

I looked around, and it seems to me that this is the only function that has this bug because it is relatively new. I tried to make this a test case by exporting to our special obj format, but we need mergeToVert and we do not yet support writing that to obj. I will leave it for later.


elalish:
Good find, thank you!

PR #1300, opened 2025-07-15
add cmake check for fp related options

pca006132:

PR #1302, opened 2025-07-15
Fix downstream openscad build

pca006132:


pca006132:
Will fix and merge this after the downstream CI is green.

Trzeth:
I think this was caused by compiling with an old version of gcc, so was https://github.com/elalish/manifold/pull/1300 .
Since downstream maybe still using gcc 11, our ci all run on 13 and 14.

pca006132:
> I think this was caused by compiling with an old version of gcc, so was #1300 . Since downstream maybe still using gcc 11, our ci all run on 13 and 14.

It is not *due to* #1300, these two are bugs. We should add a test with older gcc to our CI, maybe by having an older Ubuntu.

Issue #1303, opened 2025-07-16
add older gcc to CI


pca006132:
That downstream CIs are using.

PR #1308, opened 2025-07-18
add old version gcc to CI

Trzeth:
for #1303

### DIFF
--- a/.github/workflows/manifold.yml
+++ b/.github/workflows/manifold.yml
@@ -326,9 +326,9 @@ jobs:
         brew install googletest
         pip install trimesh pytest
     - uses: actions/checkout@v4
+    - uses: jwlawson/actions-setup-cmake@v2
       with:
         cmake-version: 4.0.2
-    - uses: jwlawson/actions-setup-cmake@v2
     - name: test cmake consumer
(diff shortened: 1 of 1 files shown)


Trzeth:
`-DMANIFOLD_DEBUG=ON` causes gcc 11 compile failed, so strange.

pca006132:
> `-DMANIFOLD_DEBUG=ON` causes gcc 11 compile failed, so strange.

It is normal. It is like adding tests and found that something is not working.

For this particular case, it seems that this is related to printing `vec3`. I am about to go to the airport so no time for debugging now.

Trzeth:
Safe trip. It compiles in 3.0.0, looking into it.

Trzeth:
gtest need `operator<<` to perform `EXPECT_EQ` for `vec3`, but compilor would only search in certain namespaces, like global and linalg, and it will ignore `using namespace manifold` in  gcc11.
So moving this back to `namespace linalg` fix it.

But I found @elalish move it from linalg.h to common.h in 09fb03c for clear strucutre. Should I just put it in common.h or move it back?

pca006132:
Moving it back sounds fine.

PR #1310, opened 2025-07-21
fix CheckHalfedges()

### LINKED ISSUE #1309: Infinite Loop or Crash in SplitPinchedVerts

Image

Trzeth:
fixes #1309

This fix will properly generate `logicErr exception` if `ManifoldParams().intermediateChecks` is set to `true`.

What cause this is a halfedge, both startVert, endVert and pairedHalfedge were set to -1, lead to a invalid access in `SplitPinchedVerts()` and infinte loop.

### DIFF
--- a/src/properties.cpp
+++ b/src/properties.cpp
@@ -73,8 +73,8 @@ struct CheckHalfedges {
 
   bool operator()(size_t edge) const {
     const Halfedge halfedge = halfedges[edge];
-    if (halfedge.startVert == -1 || halfedge.endVert == -1) return true;
     if (halfedge.pairedHalfedge == -1) return false;
+    if (halfedge.startVert == -1 || halfedge.endVert == -1) return true;
 
     const Halfedge paired = halfedges[halfedge.pairedHalfedge];
     bool good = true;
--- a/src/properties.cpp
+++ b/src/properties.cpp
@@ -73,8 +73,11 @@ struct CheckHalfedges {
 
   bool operator()(size_t edge) const {
     const Halfedge halfedge = halfedges[edge];
-    if (halfedge.pairedHalfedge == -1) return false;
+    if (halfedge.startVert == -1 && halfedge.endVert == -1 &&
+        halfedge.pairedHalfedge == -1)
+      return false;
     if (halfedge.startVert == -1 || halfedge.endVert == -1) return true;
+    if (halfedge.pairedHalfedge == -1) return false;
 
     const Halfedge paired = halfedges[halfedge.pairedHalfedge];
     bool good = true;


Trzeth:
https://github.com/elalish/manifold/blob/e4c8defdff03a850d948a812f5b19ea6c51c2700/src/properties.cpp#L74-L86

hummm, I'm not familiar with the logic in this part. The case where `pairedHalfedge == -1` isn't handled in `SplitPinchedVerts`, but the check is skipped by `startVert== -1 || endVert == -1`. But it seems other logic is also depends on it. Any ideas on how I should handle it properly? 

I think `pairedHalfedge == -1` marks the half-edge that should be removed, but I'm not sure.

pca006132:
This doesn't really *fix* the issue? This is for debugging.

I think the point here is that if a halfedge is marked as removed, all its fields should be set to -1. Maybe we should check if this holds.

Trzeth:
Sorry, let me rephrase this. What cause this is a halfedge produced by `Face2Tri()`, both startVert, endVert and pairedHalfedge were set to -1, lead to a invalid access in `SimplifyTopology()` and infinte loop.

@pca006132 Do you think this type of halfedge should be removed before `SimplifyTopology()`? Or it just means this's a self-intersect case, but `IsManifold()` check missed it.

pca006132:
I don't think this is produced by Face2Tri, it is something left from some code that is run before. Face2Tri should not mark new halfedges to be removed on its own. And I don't think it matters where these are created: they are valid and our code should deal with and ignore them properly.

> Do you think this type of halfedge should be removed before `SimplifyTopology()`? Or it just means this's a self-intersect case, but `IsManifold()` check missed it.

No and no. These halfedges are fine.

elalish:
Yeah, we have a bunch of functions (mostly in SimplifyTopology) that mark halfedges for removal. What shouldn't be possible is getting to them from a valid halfedge, e.g. having a mixture of valid and invalid edges in a triangle or a valid halfedge paired with an invalid one. If `IsManifold` isn't catching those problems, we may need to tighten it up.

pca006132:
Maybe what we should do instead is

```diff
diff --git i/src/properties.cpp w/src/properties.cpp
index d66b0559..d69e49ec 100644
--- i/src/properties.cpp
+++ w/src/properties.cpp
@@ -73,7 +73,13 @@ struct CheckHalfedges {
 
   bool operator()(size_t edge) const {
     const Halfedge halfedge = halfedges[edge];
-    if (halfedge.startVert == -1 || halfedge.endVert == -1) return true;
+    if (halfedge.startVert == -1 && halfedge.endVert == -1 &&
+        halfedge.pairedHalfedge == -1)
+      return true;
+    if (halfedges[NextHalfedge(edge)].startVert == -1 ||
+        halfedges[NextHalfedge(NextHalfedge(edge))].startVert == -1) {
+      return false;
+    }
     if (halfedge.pairedHalfedge == -1) return false;
 
     const Halfedge paired = halfedges[halfedge.pairedHalfedge];
```

Triangles should be all-or-nothing. This catches the bug. But this will not work with the polygonal mesh check before `Face2Tri`. There are also some tests failing with this after disabling the polygonal mesh check.

Trzeth:
I found two partially removed triangles in `Face2Tri()` output. But after I got rid of them, still received a self-intersecting output.

So, I tried to export the self-intersecting input without any other processing. The self-intersection points differed by less than 1e-6, so I changed the tolerance to 1e-5. Even after calling `Simplify()` manually, these points still not get merged.

Therefore, I think I should investigate what is happening within the `Simplify()` function instead of tracing back the two partially removed triangles. Or is there anything I did wrong in the below code?

```
  Manifold arg1 = Manifold(mgl_1());
  arg1 = arg1.AsOriginal().SetTolerance(1e-5);
  arg1 = arg1.Simplify(1e-5);
  MeshGL marg1 = arg1.GetMeshGL();
  manifold::ExportMesh("marg1S.obj", marg1, {});
```

pca006132:
Self-intersection is fine.

Trzeth:
image
The attached image shows two partially removed triangles that cause the infinite loop. One of them is almost a line (indicated by the yellow arrow), and it seems a flipped version of this triangle also exists. Any suggestion which part I should specially focused on?

pca006132:
So do you mean the halfedges still fail the test I mentioned above? Or it passes but we still have an infinite loop?

Trzeth:
Yes, your test catches the bug. I'm now trying to find the source that produces the partially removed triangles.

pca006132:
If we look at the polygonal mesh, do we have polygons that are partially removed?

i.e. is the problem introduced before Face2Tri?

Trzeth:
I think it doesn't. (assuming start end paired = -1)

It's also worth to mention that none of our test case have these pre-marked removed triangles(either full or partially) between `Face2Tri()` and `SimplifyTopology()` , was it forget to add into testcase or it's not logical correct?
But `SimplifyTopology()` can handle pre-marked **full** removed triangle is one thing I can be sure.

Trzeth:
image

`Tri(242 243 83)` and `Tri(83 243 495)` are the partially removed triangles. `Vert 242` and `243` positions are almost identical.
Multiple pairs of half-edges created between (242 and 243), (243 and 83), (242, 83) by `Face2Tri()`. The red label indicates a half-edge that remains, while black indicates one that has been marked for removal.

I think maybe something wrong happened in `CreateHalfedges()` ?

Below images shows these two triangles in 3d view, and their halfedge connnections before triangulate.

image

image

elalish:
Hang on, when I added @pca006132's change to `IsManifold`, it's telling the input is Not Manifold, so this looks like it's not a Boolean bug, but just our manifold check not being picky enough about our inputs.

elalish:
Oh wait no, @Trzeth you're right, I see what you mean now. We already validate their input has no removed triangles, so when it finds them on import, it means we did it ourselves in `CreateHalfedges`. Need to dig a little more.

Trzeth:
I want to confirm my understanding. `Face2Tri` will take CCW polygon as input and triangulates it. Most of the resulting triangles is CCW, but any self-intersection parts will produce CW triangles. And it might be multi edge between two vertices. 

Then `CreateHalfedges` takes this triangulation as its input, and outputs halfedges that represents a manifold topology where **all triangles have been unified to CCW**.

Please correct me if there is anything wrong.

elalish:
Close, but there's no way to correct CW triangles - if the input was self-intersecting, you're stuck with them. All it tries to do is remove pairs of triangles like [4,5,6] & [6,5,4], since those just make an infinitely thin sheet.

Trzeth:
Close in favor of #1313.

PR #1188, opened 2025-03-13
parallel CreateHalfedges

pca006132:
This turns out to be quite simple to parallelize. Not a huge gain though.

PR #1313, opened 2025-07-24
reverted #1188

### LINKED ISSUE #1309: Infinite Loop or Crash in SplitPinchedVerts

elalish:
Fixes #1309 

Following the discussion from #1310, I've tightened up our Manifoldness check to catch the problem before it gets to the Boolean. I've also added @howardt's test case. I fixed the problem by reverting #1188 - @pca006132 since that's yours, would you like to take a look at what's different there? This does cause one SDF test to fail, since I believe it's not removing all the facing triangles. I think there's another PR somewhere related to this function, but I haven't found it yet.


pca006132:
Will look at it later tonight.

pca006132:
I think I don't have enough time to review this recently, so if you can fix the CI we can merge it first and I will revisit it later.

src/impl.cpp
```diff
@@ -476,61 +476,27 @@ void Manifold::Impl::CreateHalfedges(const Vec<ivec3>& triProp,
...
+    int k = i + numEdge;
     while (1) {
       const int pair1 = ids[k];
-      Halfedge& h1 = halfedge_[pair1];
+      Halfedge h1 = halfedge_[pair1];
       if (h0.startVert != h1.endVert || h0.endVert != h1.startVert) break;
       if (halfedge_[NextHalfedge(pair0)].endVert ==
           halfedge_[NextHalfedge(pair1)].endVert) {
-        h0.pairedHalfedge = h1.pairedHalfedge = kRemovedHalfedge;
+        h0 = {-1, -1, -1};
```
elalish, line 489: Okay, I'm getting closer. #1188 wasn't the problem - reverting that only worked because the version I've reverted to here is a no-op: note that I forgot the `&` on `h0` and `h1` so nothing actually happens. I'm now looking at #1192, which added the `consecutiveStart` logic that I don't really understand. Do you remember @pca006132? I put in a print here and got this:

```
Removing halfedge pair: 1446 and 1471, verts: 274, 66, 275
Removing halfedge pair: 1470 and 1447, verts: 275, 66, 274
Removing halfedge pair: 1407 and 1090, verts: 242, 83, 243
Removing halfedge pair: 1089 and 1408, verts: 243, 83, 242
Removing halfedge pair: 1431 and 1408, verts: 243, 83, 242 <--- problem 
Removing halfedge pair: 1436 and 1410, verts: 242, 240, 243
Removing halfedge pair: 1412 and 1434, verts: 243, 240, 242
Removing halfedge pair: 1409 and 1091, verts: 243, 242, 83
Removing halfedge pair: 1435 and 1411, verts: 243, 242, 240
Removing halfedge pair: 1448 and 1472, verts: 275, 274, 66
Removing halfedge pair: 1474 and 1456, verts: 275, 274, 276
Removing halfedge pair: 1455 and 1475, verts: 276, 274, 275
Removing halfedge pair: 1473 and 1457, verts: 276, 275, 274
```

pca006132: I am finally back to HK and will have time to look into this. iirc it is just to make sure parallel partitioning will partition at correct points, because we need to scan a range for stuff.

elalish: I dunno, it's used by the serial version too...

elalish:
closing in favor of #1319

PR #1317, opened 2025-07-29
Fix `faceID` documentation in wasm bindings

jirihon:
To prevent future misunderstandings from wasm bindings users like me. See #1316


elalish (approved):
Thank you! One of these days we really need to get some automation around these docs like we have for the Python bindings...

PR #1319, opened 2025-08-02
Fix #1309

### LINKED ISSUE #1309: Infinite Loop or Crash in SplitPinchedVerts

pca006132:
Fixes #1309. Not sure if there are still hidden issues here, see comments.

### DIFF
--- a/src/boolean_result.cpp
+++ b/src/boolean_result.cpp
@@ -878,11 +878,6 @@ Manifold::Impl Boolean3::Result(OpType op) const {
 #endif
 
   // Level 6
-
-  // if (ManifoldParams().intermediateChecks)
-  //   DEBUG_ASSERT(outR.IsManifold(), logicErr, "polygon mesh is not
-  //   manifold!");
-
   outR.Face2Tri(faceEdge, halfedgeRef);
   halfedgeRef.clear();
   faceEdge.clear();


src/boolean_result.cpp
```diff
@@ -879,8 +879,8 @@ Manifold::Impl Boolean3::Result(OpType op) const {
 
   // Level 6
 
-  if (ManifoldParams().intermediateChecks)
-    DEBUG_ASSERT(outR.IsManifold(), logicErr, "polygon mesh is not manifold!");
+  // if (ManifoldParams().intermediateChecks)
```
pca006132: The problem with this is that we are checking for polygon mesh manifoldness, and the tightened manifoldness check is not suitable for that (we assume a triangular mesh).

src/impl.cpp
```diff
@@ -484,8 +484,9 @@ void Manifold::Impl::CreateHalfedges(const Vec<ivec3>& triProp,
       const int pair1 = ids[k];
       Halfedge& h1 = halfedge_[pair1];
       if (h0.startVert != h1.endVert || h0.endVert != h1.startVert) break;
-      if (halfedge_[NextHalfedge(pair0)].endVert ==
-          halfedge_[NextHalfedge(pair1)].endVert) {
+      if (h1.pairedHalfedge != kRemovedHalfedge &&
```
pca006132, line 487: The bug is that we still match removed halfedges when checking for opposed triangles. This now fixes the issue.

src/impl.cpp
```diff
@@ -484,11 +484,38 @@ void Manifold::Impl::CreateHalfedges(const Vec<ivec3>& triProp,
...
+          halfedge_[NextHalfedge(pair0)].endVert ==
+              halfedge_[NextHalfedge(pair1)].endVert) {
+        // Reorder so that remaining edges pair up.
+        // This makes sure ids[k] is filled with something legal after moving
+        // its value to ids[i + numEdge].
+        // Note that this is intentionally done before marking h0 and h1 as
+        // removed, so step can work correctly (eventually stop at i + numEdge)
+        if (k != i + numEdge) {
```
pca006132: This is something I think the old code is probably wrong, but not sure if the current version is needed.

The old code swaps `ids[i+numEdge]` and `ids[k]`, which means that if we have a long run of halfedges with the same start/end vertex, we will invalidate the invariant that these halfedges are ordered by triangle ids. This now fixes the issue, but one has to be careful *not* to move removed halfedges, which leads to this somewhat complex code.

That said, this may or may not be directly related to the correctness of this algorithm. What we should ensure is that if we have the three halfedges of the same triangle, they always pair with the same opposed triangle. I need to think a bit about it...

pca006132: Yeah both are correct. We only care about the relative order between halfedges with the same tri verts. We always match the first pair and remove it (marked as removed), so we never change this relative order.

src/boolean_result.cpp
elalish: I think it's fine to just remove this instead of commenting it out. I can't remember the last time it caught anything anyway.

src/impl.cpp
elalish: Nice, I thought I tried this, but I must have gotten it wrong.

elalish (approved):
Thanks!

PR #1321, opened 2025-08-05
deterministic batch boolean

### LINKED ISSUE #1320: nondeterministic result, multiple unions, lots of coplanarity
Not sure if this is the same as nondeterministic result #1275 but reporting this here in case not. This is an active Blender manifold boolean bug report.

With a union of 19 cubes, many have some coplanarity with others:

Image

which has a result like this:

Image

Doing this twice in a row yields different results. The vertex positions seem to be the same -- just the output triangles are different.

The attached C++ file (remove the .txt extension) shows the problem.

[mesh_union.cpp.txt](https://github.com/user-attachments/files/21569463/mesh_union.cpp.txt)

I think PAR=1 is necessary to get the non-determinism.

pca006132:
Fixes #1320. This one is simpler than expected, so it is quick :)

### DIFF
--- a/src/csg_tree.cpp
+++ b/src/csg_tree.cpp
@@ -342,7 +342,9 @@ std::shared_ptr<CsgLeafNode> BatchBoolean(
   std::vector<std::shared_ptr<CsgLeafNode>> tmp;
 #if MANIFOLD_PAR == 1
   tbb::task_group group;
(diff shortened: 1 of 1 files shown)


elalish (approved):
Nice!

howardt:
Thanks!

PR #1323, opened 2025-08-05
fix odr violation

pca006132:
We should not define these in the header file.


pca006132:
> Oh right, I wonder what an inline static member in a header file does? Not sure what I was thinking there... Was there some compiler message that alerted you to this?

address sanitizer reports this as an error :)

pca006132:
I don't think inline does anything here, and static is kind of like a global variable, so this is just odr violation.

PR #1329, opened 2025-08-06
Repackage WASM dynamic evaluation and mesh export

tonious:
I've been noodling with ManifoldCAD for a while now.  I've recently considered building a command line tool.  After a little experimentation, it might just be easier to do it in place. 

My goals for this refactor:

- [x] Extract and separate dynamic evaluation and mesh export logic.
- ~~Declare types (and interfaces) in modules, and import them ManifoldCAD editor.~~
- [x] Make GLTF and 3MF export usable within node.js (typescript optional)
- [x] Same for dynamic evaluation.
- [x] Introduce a command line version of ManifoldCAD

While I'm an experienced developer, I'm a novice contributor.  Please let me know if this approach fits this project, or if there's a better way to do it.  I mean, there's always a better way, please point me in the right direction :D


pca006132:
It will be helpful if you can give an example of how to use your tool, including the script and how the CLI commands work.

tonious:
I'm picturing a very simple CLI. E.g.:

```
manifold-cli ./mymodel.js -o mymodel.3mf
```

Fundamentally, I want to be able to do what I do in the browser, just without the browser.  All of that logic is here, it's just not packaged in an accessible way.  Packaging it for node should not (must not!) detract from browser based ManifoldCAD in any way.

Dynamic evaluation (setting up the environment and executing a script from a string) and model export (GLTF in particular) are currently tightly coupled to the web version.   They work by adding methods to the `ManifoldTopLevel` module.  My first, naive approach is do the same thing.

```
const module = await Module();
module.setup();
evalSetup(module);  // Add evaluateCadToManifold and related methods to module.
GLTFSetup(module);  // Add exportModels and related methods to module.

function evaluateCADToModel(code: string) {
  const {globalDefaults, manifold} = module.evaluateCADToManifold(code);
  return await module.exportModels(globalDefaults, manifold);
}
```

On the other hand, would it make more sense to aim for more of an object oriented basis?

```
const module = await Module();
module.setup();

const evaluator = new Evaluator(module);
const {globalDefaults, manifold} = evaluator.createManifold(code);

const exporter = new GLTFExporter(module, io);
await exporter.write(globalDefaults, manifold);

```

Or, on the third hand, getting to a simple CLI isn't very far from where this PR is already.  Should I just do that, and leave any larger thoughts to the future?

pca006132:
> I want to be able to do what I do in the browser, just without the browser.

Ah, make sense. I don't have too much opinion about how the implementation should be, and I think it shouldn't matter too much because this tool aims to be a tool instead of a library (or do we?). @elalish will probably have a stronger opinion about the js implementation :)

elalish:
Thanks, this is interesting, and always good to hear from our JS users. I mostly have to guess at the use cases. I'm always a fan of small steps, as it makes review easier, so if that's easy, please do. I think your object oriented approach looks pretty clean. The main thing is I want to keep our bundle size down by keeping optional parts a la carte. And points for automation and less code to maintain here.

tonious:
Export is purely cut and paste.

I got a little fancier with Evaluate, and exposed `context` as a property.  This uncouples it from export; it only sets up Manifold itself now:

```typescript
  evaluate(code: string) {
    const exposedFunctions = toplevelConstructors.concat(toplevel).map(
        (name) => [name, (this.module as any)[name]]);
    const context = {
      ...Object.fromEntries(exposedFunctions),
      module: this.module,
      ...this.context
    };

    const evalFn = new Function(
        ...Object.keys(context),
        this.beforeScript + '\n' + code + '\n' + this.afterScript + '\n');
    return evalFn(...Object.values(context));
  }
```

This means that methods from export (`show`, `only`, &c.) need to be explicitly passed in.  That happens in [worker.ts](https://github.com/elalish/manifold/blob/319bcd3939894d4913a034b935651a6e622a7464/bindings/wasm/examples/worker.ts#L31-L40).  The upshot is that one can use the exporter without the evaluator, or vice versa.

```typescript
// Setup the evaluator and it's context.
const evaluator = new Evaluator(module);

evaluator.context = {
  glMatrix,
  GLTFNode: exporter.GLTFNode,
  setMaterial: exporter.setMaterial,
  setMorphStart: exporter.setMorphStart,
  setMorphEnd: exporter.setMorphEnd,
  show: exporter.show,
  only: exporter.only,
};

...

export async function evaluateCADToModel(code: string) {
  const globalDefaults = {};

  evaluator.context.globalDefaults = globalDefaults;
  const manifold = evaluator.evaluate(code);

  return await exporter.exportModels(
      globalDefaults as GlobalDefaults, manifold);
}
```

Evaluate now has TS docs, but export scares me a little.  I guess I don't need to document everything about it, though.

As for releasing this, I'm not sure -- I figured I'd just get it working and then drop that decision on you.  My rough thoughts on the matter:

  - Evaluate is relatively self contained.  It could fit into the main package quite well
  - Including export means that we'd also require `gltf-transform`, `3mf-export` and `fflate` for everybody downstream.  Not sure how I feel about that, but chances are, that's the piece users will want.
  - In any case, there will need to be a `build:transpile` job.  I'm messing around with that now.

I think I'll do up `manifold-cli` as node flavoured JavaScript to demonstrate interoperability.  There's value in that as part of the examples subproject even if users have to go searching for it.

I'm a little less sanguine about moving type definitions around.  It'd reduce the coupling between the editor and everything else, but it's starting to seem like a job for another day.

tonious:
Progress! 😁

```
examples % ./bin/manifold-cli intro.js intro.glb
Manifold took 0.04 seconds
Exporting GLB & 3MF took 0.05 seconds
examples % examples % ls -l intro*
-rw-r--r--@ 1 tony  staff  63172 Aug  7 19:27 intro.glb
-rw-r--r--@ 1 tony  staff    140 Aug  7 18:48 intro.js
examples %
```

bindings/wasm/examples/lib/evaluate.ts
```diff
@@ -0,0 +1,193 @@
...
+    this.addMembers('Manifold', manifoldMemberFunctions, false);
+    this.addMembers('Manifold', manifoldStaticFunctions, true);
+    this.addMembers('CrossSection', crossSectionMemberFunctions, false);
+    this.addMembers('CrossSection', crossSectionStaticFunctions, true);
+
+    for (const name of toplevelConstructors) {
+      //@ts-ignore
+      const originalFn = module[name];
+      //@ts-ignore
+      this.module[name] = (...args: any) => {
```
tonious: Changed to an arrow function to not clobber `this`.

bindings/wasm/examples/lib/evaluate.ts
```diff
@@ -0,0 +1,193 @@
...
+ *
+ * @param module A Manifold WASM instance, already set up.
+ * @property context Additional objects inserted into the evaluation
+ * context.
+ * @property beforeScript Boilerplate script run before the supplied
+ * code.
+ * @property afterScript Boilerplate code run after the supplied code.
+ */
+export class Evaluator {
+  context: any = {};
```
tonious: I didn't bother with setters/getters here.  I think just being able to directly set `context`, `beforeScript` and `afterScript` is sufficient.

bindings/wasm/examples/lib/evaluate.ts
```diff
@@ -0,0 +1,193 @@
...
+   * or a `Manifold` object.  Changing `afterScript` will affect this
+   * behaviour.
+   */
+  evaluate(code: string) {
+    const exposedFunctions = toplevelConstructors.concat(toplevel).map(
+        (name) => [name, (this.module as any)[name]]);
+    const context = {
+      ...Object.fromEntries(exposedFunctions),
+      module: this.module,
+      ...this.context
```
tonious: Does not automatically include `glMatrix` or `exportModels` anymore; that's up to the caller.

bindings/wasm/examples/package.json
```diff
@@ -26,7 +29,7 @@
   "devDependencies": {
     "@types/node": "^22.10.2",
     "@vitest/ui": "^1.6.0",
-    "typescript": "5.2.2",
+    "typescript": "^5.7.0",
```
tonious: TypeScript 5.7.0 has some additional settings that make including modules easier in nodejs.  Specifically `rewriteRelativeImportExtensions`.

bindings/wasm/examples/tsconfig.json
```diff
@@ -16,9 +16,12 @@
     "noFallthroughCasesInSwitch": true,
     "declaration": true,
     "skipLibCheck": true,
+    "allowImportingTsExtensions": true,
+    "rewriteRelativeImportExtensions": true,
```
tonious, line 20: Makes including modules easier in nodejs.  I had to enable `allowImportingTsExtensions` to get it to work cleanly.

bindings/wasm/examples/package.json
```diff
@@ -5,7 +5,9 @@
   "type": "module",
   "scripts": {
     "dev": "vite",
-    "build": "tsc --noEmit && vite build",
+    "build": "npm run build:vite && npm run build:transpile",
```
tonious: This made sense to me, but I'm open to other conventions.

bindings/wasm/examples/bin/manifold-cli
```diff
@@ -0,0 +1,67 @@
...
+
+  return await exporter.exportModels(globalDefaults, manifold);
+}
+
+const program = new Command();
+
+program
+  .name('manifold-cli')
+  .description('A simple CLI for ManifoldCAD')
+  .argument('<infile.js>', 'Input file')
```
tonious: I could make this a subcommand?  E.g. `manifold-cli build intro.js intro.glb`.  That way, we could eventually other commands like `manifold-cli make-manifold leaky.glb notleaky.glb`.

bindings/wasm/examples/lib/tsconfig.json
```diff
@@ -0,0 +1,27 @@
+{
+  "compilerOptions": {
+    "target": "es2022",
+    "module": "es2022",
+    "moduleResolution": "node",
```
tonious: I'm assuming that building the `lib` directory in particular means a node user, while a vite user will be in the parent directory.

bindings/wasm/examples/lib/evaluate.ts
tonious: The caller may accidentally overwrite context if they're not careful.  I think that follows the principle of least astonishment.

bindings/wasm/examples/lib/evaluate.ts
```diff
@@ -0,0 +1,193 @@
...
+    };
+
+    const evalFn = new Function(
+        ...Object.keys(context),
+        this.beforeScript + '\n' + code + '\n' + this.afterScript + '\n');
```
tonious: Same as before, just now with properties.

bindings/wasm/examples/worker.ts
```diff
@@ -12,791 +12,50 @@
...
-import {strToU8, Zippable, zipSync} from 'fflate'
-import * as glMatrix from 'gl-matrix';
+import * as glMatrix from 'gl-matrix'
 
-import Module, {CrossSection, Manifold, ManifoldToplevel, Mesh, Vec3} from './built/manifold';
-import {Properties, setupIO, writeMesh} from './gltf-io';
-import {GLTFMaterial, Quat} from './public/editor';
+import Module from './built/manifold';
+import {Evaluator} from './lib/evaluate';
+import * as exporter from './lib/export';
```
tonious: I don't particularly like how `evaluate` and `export` are imported and set up differently.

bindings/wasm/examples/bin/manifold-cli
```diff
@@ -0,0 +1,67 @@
...
+    if (!is3mf && !isGlb) {
+      throw new Error('Output file must be either .glb or .3mf');
+    }
+
+    const code = await readFile(inPath);
+
+    const { glbURL, threeMFURL } = await evaluateCADToModel(code);
+    const res = await fetch(is3mf ? threeMFURL : glbURL);
+    const binary = await res.arrayBuffer();
+    await writeFile(outPath, Buffer.from(binary));
```
tonious: I understand why this is necessary for a web worker, but it's awkward in node.

tonious:
I've tried a couple of partial refactors of the export code, and I'm reading through the GLTF spec to get a little more understanding of what's going on.  Here's a thought dump, so future me can point and laugh (that guy just has it in for me).

From my understanding, any properties (animation, materials, &c.) in ManifoldCAD external to manifold itself are semantically GLTF.  When a `GLTFNode` is constructed it adds itself to an internal list for later export.  Which means even though I've tagged this file 'export' for now, it is just as much an extended state store. 

In its current form, the `result` object, `GLTFNode` objects, and exporting to 3MF and GLTF all overlap to some degree.  To untangle it, I think the immediate future is something loosely along the lines of:

  * `manifoldToGLTFDoc(manifold, options)`: Manifold object to a GLTF-transform document.
  * `GLTFNodesToGLTFDoc(nodes[], options)`: GLTF Nodes to a complete GLTF-transform document.
  * `GLTFDocToGLB(doc)`: Convert a GLTF-transform document into a blob in GLB format.
  * `GLTFDocTo3MF(doc)`: Convert a GLTF-transform document into a blob in 3MF format.
  * Setters and getters for various properties. (`setProperty()`, `hasGLTFNodes()`, `getGLTFNode()`, &c.)

Less refined thoughts for later:

  * Leave space such that importing models won't require another refactor.
    * With so much external state being in GLTF, it might be possible to import models _for display only_.  I.e.: not present in manifold, but included in the final GLTF document.  This could be an interesting way to show a ManifoldCAD part fitting into an assembly.
  * I really like the approach of using GLTF semantics.  Not only does it provide easy serialization, it also means that there's a larger design context right at hand.  I wonder if it's possible to somehow keep even more of the ManifoldCAD state in a dynamic GLTF-transform document, and skip the various internal maps and lists?
  * Continuing to trim modules down to easily digestible chunks:
    * State storage, model export and model import as separate modules.
    * Maybe break out materials and animations into their own modules.

hrgdavor:
@tonious i sent you an email few days ago, but am not sure u got it. I try to avoid spamming here. i hope @elalish does not mind. i see someone that wants to code-cad in JS, so I try to interest them in jscad too,  trying to keep tour dwindling jscad community alive.

tonious:
Right! I meant to get back to you, @hrgdavor.  I'll go do that :)

elalish:
Thanks for your description, and I'm glad you like the GLTF semantics (I was in the PBR working group for 5 years or so, so I've got a soft spot for them). You're right that it could use a refactor, and I think your approach sounds quite reasonable. I also feel pretty confident letting you work through it because I think our examples form a pretty decent test set of the features we currently support - if they continue working as they did before, you should be in pretty good shape.

One thing that I recall tended to tie the code together a bit is the support for the debugging functions: `only()` and `show()`. Hopefully they don't give you too much trouble, but let me know if anything about the current code is unclear.

Are you waiting for more feedback, or should I wait until you're further along to review more thoroughly? Are there any parts that are ready to merge?

tonious:
I might add some more test coverage when I get a little further (3mf in particular), but end-to-end tests have definitely kept me honest.   I really don't want to break any models.

`only()` and `show()` did indeed throw me for a bit of a loop, but made sense once I realized what I was looking at.  Eventually I can see breaking that out into some kind debug module that fits with both export and materials.  But that's pretty far away.

IMHO, this branch as it stands could be committed.  I'm happy with the evaluator, it stands alone nicely.
So if you like, review this for merge, and I'll open a new PR once I'm further along with export?

bindings/wasm/examples/bin/manifold-cli
elalish: My feeling is that it would be better to have multiple CLIs than one with a complicated interface. Perhaps we should name this one `manifold-cad` and perhaps the latter could be `manifold-fix`?

Will this allow us to run e.g. `npx manifold-run intro.js intro.glb`? Can we add some docs to the README?

bindings/wasm/examples/lib/evaluate.ts
```diff
@@ -0,0 +1,193 @@
+// Copyright 2022 The Manifold Authors.
```
elalish: Should probably update the year.

bindings/wasm/examples/lib/evaluate.ts
```diff
@@ -0,0 +1,193 @@
...
+ *
+ * This class provides some simple garbage collection.  It does this by
+ * intercepting calls to a white-list of functions, tracking new
+ * instances of `Manifold` and `CrossSection`.  This way, users don't
+ * have to care about calling `delete` manually.  Note that this only
+ * fixes memory leak across different runs: the memory will only be freed
+ * when `cleanup()` is called.
+ *
+ * @param module A Manifold WASM instance, already set up.
+ * @property context Additional objects inserted into the evaluation
```
elalish: @param or @property? Also, module look vestigial?

bindings/wasm/examples/package.json
```diff
@@ -5,7 +5,9 @@
...
+    "build:vite": "tsc --noEmit && vite build",
+    "build:transpile": "npx tsc -p lib --outDir built/lib",
```
elalish: What's npx for here?

bindings/wasm/examples/worker.ts
```diff
@@ -12,791 +12,55 @@
...
-const GLOBAL_DEFAULTS = {
-  roughness: 0.2,
-  metallic: 1,
-  baseColorFactor: [1, 1, 0] as [number, number, number],
-  alpha: 1,
-  unlit: false,
-  animationLength: 1,
-  animationMode: 'loop'
+evaluator.context = {
+  ...evaluator.context,
```
elalish: If this line was forgotten, would everything break? Perhaps we should have something like `evaluator.addFunctions()` instead?

bindings/wasm/examples/worker.ts
elalish: What would you like to change?

bindings/wasm/examples/bin/manifold-cli
```diff
@@ -0,0 +1,66 @@
...
+const evaluateCADToModel = async (code) => {
+  const globalDefaults = {};
+
+  evaluator.context.globalDefaults = globalDefaults;
+  const manifold = evaluator.evaluate(code);
+
+  return await exporter.exportModels(globalDefaults, manifold);
+}
+
+const program = new Command();
```
elalish: This seems like the right time to have a discussion about npm packaging. We currently have one published module: `manifold-3d`. Looking at another library Iike, `gltf-transform`, the module is both a library and a CLI via npx. This makes me feel like perhaps we should include `evaluate.ts` and this CLI in the `manifold-3d` package - what do you think?

I've also been intending to release an import/export npm package, but that feels pretty separate (especially with all the dependencies). The work you're doing seems like the right impetus to get that figured out too (in a following PR). Of course people might also just want `manifold-gltf` or `gltf-io`, but I think in the modern era of tree-shaking it won't be bad having them in one package? 

I'll be easily swayed - packaging is a bit outside my wheelhouse.

elalish (changes requested):
I think this looks great - let's just agree on the CLI API before we merge.

bindings/wasm/examples/worker.ts
tonious: If this was omitted, the evaluator would still include all of the manifold functions, but not the additional context specified here.

bindings/wasm/examples/bin/manifold-cli
tonious: My opinion has been fluttering back and forth a bit, which usually means it doesn't matter.  Your point about tree-shaking is dead on.  Web packaging tools will include the minimal subset, and the overhead is negligible for server side or local apps.

So I'll move everything up a level, and make sure I can build the package properly.  I'll also make some notes where API surfaces are likely to change (export).

bindings/wasm/examples/package.json
tonious: I just copied and pasted the command line apparently without enough thought 😁

bindings/wasm/examples/lib/evaluate.ts
tonious: Good call.  That'll let me hide some complexity.

tonious:
More progress :grin:

```
tmp % npx manifold-cad intro.js intro.glb
Manifold took 0.04 seconds
Exporting GLB & 3MF took 0.05 seconds
tmp %
```

bindings/wasm/examples/worker.ts
elalish: Even if just line 32 was gone but the rest of the expression remained? What is in that existing context that you're unpacking and repacking?

tonious: Oh, I see!  Originally I had split this up into successive updates.  Here, the destructuring is vestigial, and I'll get rid of it.

bindings/wasm/examples/worker.ts
tonious: IMHO, `evaluate` and `export` should either both be object oriented, or both built as modules. Right now I'm leaning towards object oriented, but that'll be another PR as discussed.

bindings/wasm/examples/worker.ts
tonious: I added some accessor methods to evaluate while I was at it.

tonious:
Okay, so:

  * `examples/lib` raised to `/lib`
  * Package files updated accordingly.
  * Dependencies updated to latest stable.
  * `lib/evaluate.ts` added to `tsdoc`.
  * CLI added to readme.
  * context accessors added to `lib/evaluate.ts`
  * Copyright dates bumped on files I've touched.

I think I'm ready for another review, @elalish 😀

pca006132:
@elalish do you want to merge this before the next patch release?

elalish:
Oh, good question. Actually, are you online for our call? A little late, but I'm here if you'd like to chat.

pca006132:
can do.

tonious:
Lemme know if you want any changes or explanations.

In other news, I've started teasing apart the big ball of GLTF state.  3MF export works nicely, and I've fixed a few compatibility issues on that front.

elalish:
And I'm looking forward to your next one!

elalish:
Looks like some CI path trouble - mind taking a look? https://github.com/elalish/manifold/actions/runs/17204714858/job/48802687271

PR #1341, opened 2025-08-22
Fix install with static link and using builtin dependencies

### LINKED ISSUE #1338: can't find Clipper2 from static manifold link
I built Manifold statically on modern MacOS with the following commands:
```
 cmake .. -DCMAKE_BUILD_TYPE=Release -DBUILD_SHARED_LIBS=OFF -DMANIFOLD_USE_BUILTIN_TBB=ON -DMANIFOLD_USE_BUILTIN_NANOBIND=ON -DMANIFOLD_TEST=OFF -DCMAKE_OSX_DEPLOYMENT_TARGET=12.0 -DCMAKE_INSTALL_PREFIX=/usr/local
make -j10
sudo make install
```
It tells me that it will use the built-in Clipper2. It installed fine to /usr/local, or so I thought, but it actually failed to copy the clipper lib, _deps/clipper2-build/libClipper2.a . When I attempted to compile a project using Manifold I got this error:
```
CMake Error at /usr/local/lib/cmake/manifold/manifoldTargets.cmake:61 (set_target_properties)
The link interface of target "manifold::manifold" contains:
    Clipper2::Clipper2
  but the target was not found.
```
I went to the _deps/clipper2-build folder and ran `sudo make install`. That appeared to copy over all the necessary clipper files. Unfortunately, it doesn't work. Cmake persists in telling me Clipper2 was not found. Any ideas? I assume that I don't need to call find_package(Clipper2)? I don't do that on my Linux or Windows builds, which builds succeed.

pca006132:
Fixes #1338. Not sure if it really works, need testing on CI and probably by @BrannonKing.


pca006132:
`pkg-config` will not work for static lib for now. I can make it work, but it is annoying...

PR #1347, opened 2025-08-31
Simplify 180° Edges

zalo:
Adds a routine to simplify antiparallel edges, which should largely eliminate the slivers resulting from CSG on coplanar faces.

Test cases coming soon...

Previous discussion here: https://github.com/elalish/manifold/pull/666#issuecomment-3239755814

### DIFF
--- a/src/edge_op.cpp
+++ b/src/edge_op.cpp
@@ -95,6 +95,29 @@ struct FlagEdge {
   }
 };
 
+struct AntiparallelEdge {
+  VecView<const Halfedge> halfedge;
+  VecView<const vec3> triNormal;
+  const int firstNewVert;
+  const double angleThreshold;  // Threshold for considering normals antiparallel
+
+  bool operator()(int edge) const {
+    const Halfedge& half = halfedge[edge];
+    if (half.pairedHalfedge < 0) return false;
+    // Only process edges where at least one vertex is new
+    if (half.startVert < firstNewVert && half.endVert < firstNewVert) return false;
+    
+    // Check if the adjacent triangle normals are antiparallel (180 degrees apart)
+    const int tri1 = edge / 3;
+    const int tri2 = half.pairedHalfedge / 3;
+    const double dotProduct = la::dot(triNormal[tri1], triNormal[tri2]);
+    
+    // If normals are pointing in opposite directions (dot product close to -1)
+    // This indicates a sliver where triangles are nearly coplanar but facing opposite directions
+    return dotProduct < angleThreshold;  // e.g., -0.99 for nearly antiparallel
+  }
+};
+
 struct SwappableEdge {
   VecView<const Halfedge> halfedge;
   VecView<const vec3> vertPos;
@@ -237,6 +260,7 @@ void Manifold::Impl::SimplifyTopology(int firstNewVert) {
 
   CleanupTopology();
   CollapseShortEdges(firstNewVert);
+  CollapseAntiparallelEdges(firstNewVert);
   CollapseColinearEdges(firstNewVert);
   SwapDegenerates(firstNewVert);
 }
@@ -246,6 +270,7 @@ void Manifold::Impl::RemoveDegenerates(int firstNewVert) {
 
   CleanupTopology();
   CollapseShortEdges(firstNewVert);
+  CollapseAntiparallelEdges(firstNewVert);
   SwapDegenerates(firstNewVert);
 }
 
@@ -304,6 +329,30 @@ void Manifold::Impl::CollapseColinearEdges(int firstNewVert) {
     }
 #endif
   }
+}
+
+void Manifold::Impl::CollapseAntiparallelEdges(int firstNewVert) {
+  ZoneScopedN("CollapseAntiparallelEdges");
+  FlagStore s;
+  size_t numFlagged = 0;
+  const size_t nbEdges = halfedge_.size();
+  std::vector<int> scratchBuffer;
+  scratchBuffer.reserve(10);
+  
+  // Use -0.9999 as threshold - edges where normals are nearly opposite (angle > ~179.982 degrees)
+  // This targets slivers where triangles are nearly coplanar but facing opposite directions
+  AntiparallelEdge ae{halfedge_, faceNormal_, firstNewVert, -0.9999};
+  s.run(nbEdges, ae, [&](size_t i) {
+    const bool didCollapse = CollapseEdge(i, scratchBuffer);
+    if (didCollapse) numFlagged++;
+    scratchBuffer.resize(0);
+  });
+
+#ifdef MANIFOLD_DEBUG
+  if (ManifoldParams().verbose > 0 && numFlagged > 0) {
+    std::cout << "collapsed " << numFlagged << " antiparallel edges" << std::endl;
+  }
+#endif
 }
 
 void Manifold::Impl::SwapDegenerates(int firstNewVert) {
--- a/src/impl.h
+++ b/src/impl.h
@@ -342,6 +342,7 @@ struct Manifold::Impl {
   void RemoveDegenerates(int firstNewVert = 0);
   void CollapseShortEdges(int firstNewVert = 0);
   void CollapseColinearEdges(int firstNewVert = 0);
+  void CollapseAntiparallelEdges(int firstNewVert = 0);
   void SwapDegenerates(int firstNewVert = 0);
   void DedupeEdge(int edge);
   bool CollapseEdge(int edge, std::vector<int>& edges);


zalo:
Hmm, this _almost_ just works (via asOriginal + Simplify), except in the cases where two of the edges on a sliver are valid, so it doesn't want to collapse those edges...  What would be the correct move in this case?  @elalish @pca006132 

https://github.com/user-attachments/assets/dd6bbba7-1a58-44e7-8338-1aeeb3d117ba

The best my mind can conjure up is:
- Find antiparallel edges where the other two edges of the triangle are valid
- Test each adjacent triangle with the raycast thickness test or a reflex edge test to see which triangle is only on the sliver
- Thicken the only sliver triangle and CSG subtract it

CSG ops in edge_ops.cpp seems a bit taboo, though, so I'm hoping there is a better way 😅

src/edge_op.cpp
```diff
@@ -306,6 +331,30 @@ void Manifold::Impl::CollapseColinearEdges(int firstNewVert) {
   }
 }
 
+void Manifold::Impl::CollapseAntiparallelEdges(int firstNewVert) {
+  ZoneScopedN("CollapseAntiparallelEdges");
+  FlagStore s;
+  size_t numFlagged = 0;
+  const size_t nbEdges = halfedge_.size();
+  std::vector<int> scratchBuffer;
+  scratchBuffer.reserve(10);
```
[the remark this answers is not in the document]
zalo: Everyone else is doing it 😄

src/edge_op.cpp
```diff
@@ -306,6 +331,30 @@ void Manifold::Impl::CollapseColinearEdges(int firstNewVert) {
...
+  
+  // Use -0.9999 as threshold - edges where normals are nearly opposite (angle > ~179.982 degrees)
+  // This targets slivers where triangles are nearly coplanar but facing opposite directions
+  AntiparallelEdge ae{halfedge_, faceNormal_, firstNewVert, -0.9999};
```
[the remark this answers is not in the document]
zalo: Should this be configurable as a separate tolerance variable in the `Simplify` function?

zalo:
I added the beginnings of the thick sliver removal process (managed to save on needing raycasts), but it seems like I was right that higher level CSG functions are not generally called from within `Impl`...  

Say my edge collapse routine generates a set of Manifolds to subtract from the current Manifold... how do I smuggle them out of `Impl`?  Is there a better way?

elalish:
I'm going to say hard pass on the extra Boolean ops to fix slivers. That's a bit too ugly for my taste, and makes me fear infinite recursion. Let me consider...

src/edge_op.cpp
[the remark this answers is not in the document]
elalish: No, I need another configurable tolerance like I need another hole in my head. What units does this tolerance have? There should be a way to use our existing tolerance, since this is doing conceptually the same thing.

src/edge_op.cpp
```diff
@@ -306,6 +339,124 @@ void Manifold::Impl::CollapseColinearEdges(int firstNewVert) {
...
+#endif
+
+      // Check which triangle on this halfedge has two other edges that are
+      // concave
+      // const Halfedge& half = halfedge_[i];
+      size_t concaveEdgesTri1 = 0;
+      const size_t tri1 = i / 3;
+      for (size_t e = 0; e < 3; e++) {
+        const size_t edgeIndex = tri1 * 3 + e;
+        if (i / 3 == edgeIndex) continue;
```
elalish: Do you mean `if(i == edgeIndex)`?

src/edge_op.cpp
```diff
@@ -306,6 +339,124 @@ void Manifold::Impl::CollapseColinearEdges(int firstNewVert) {
...
+                << std::endl;
+//}
+#endif
+
+      // Check which triangle on this halfedge has two other edges that are
+      // concave
+      // const Halfedge& half = halfedge_[i];
+      size_t concaveEdgesTri1 = 0;
+      const size_t tri1 = i / 3;
+      for (size_t e = 0; e < 3; e++) {
```
elalish: I prefer doing these short loops like `for (const size_t e : {0, 1, 2})`.

test/boolean_test.cpp
```diff
@@ -505,3 +505,49 @@ TEST(Boolean, BatchBoolean) {
...
+        Manifold::Cube({1, 1, 1}, true).Rotate(0, 0, alpha * i));
+  }
+  Manifold outerGear = Manifold::BatchBoolean(outerCubes, OpType::Add);
+  outerGear = outerGear.Scale({2, 2, 1});
+
+  // Create inner gear - same pattern but not scaled
+  std::vector<Manifold> innerCubes;
+  for (int i = 0; i < N; ++i) {
+    innerCubes.push_back(
+        Manifold::Cube({1, 1, 1}, true).Rotate(0, 0, alpha * i));
```
elalish: Can we skip the duplication and just do `gear.Scale({2,2,1})-gear`?

test/boolean_test.cpp
```diff
@@ -505,3 +505,49 @@ TEST(Boolean, BatchBoolean) {
...
+  // https://github.com/BrunoLevy/thingiCSG/blob/main/DATABASE/Basic/nasty_gear_1.scad
+
+  const int N = 50;  // Number of rotations for the gear pattern
+  const double alpha = 360.0 / N;
+
+  // Create outer gear - many rotated cubes unioned together
+  std::vector<Manifold> outerCubes;
+  for (int i = 0; i < N; ++i) {
+    outerCubes.push_back(
+        Manifold::Cube({1, 1, 1}, true).Rotate(0, 0, alpha * i));
```
elalish, line 521: This isn't introducing any rounding error in Z, so I'm a little surprised we're having trouble with slivers. Seems like symbolic perturbation has failed - perhaps we should debug that first. Does it also repro with N = 2 or 4? Also, have you tried forcing evaluation of the gear, before you subtract them from eachother, e.g. `OuterGeart.NumVert()`? I'm curious if the Boolean reordering is giving us trouble.

src/edge_op.cpp
```diff
@@ -306,6 +339,124 @@ void Manifold::Impl::CollapseColinearEdges(int firstNewVert) {
...
+             vertPos_[halfedge_[tri1 * 3 + 1].startVert] + offset,
+             vertPos_[halfedge_[tri1 * 3 + 2].startVert] + offset,
+             vertPos_[halfedge_[tri1 * 3].startVert] - offset,
+             vertPos_[halfedge_[tri1 * 3 + 1].startVert] - offset,
+             vertPos_[halfedge_[tri1 * 3 + 2].startVert] - offset});
+
+        //*this = this->Boolean(thickSliver, OpType::Subtract);
+      }
+
+      if (concaveEdgesTri2 == 2) {
```
elalish: I don't think I'm willing to auto-fix topology like this - it's in the same category as thin webs, where the two sides can easily not be neighbors at all.

src/edge_op.cpp
```diff
@@ -95,6 +95,33 @@ struct FlagEdge {
...
+    // Check if the adjacent triangle normals are antiparallel (180 degrees
+    // apart)
+    const int tri1 = edge / 3;
+    const int tri2 = half.pairedHalfedge / 3;
+    const double dotProduct = la::dot(triNormal[tri1], triNormal[tri2]);
+
+    // If normals are pointing in opposite directions (dot product close to -1)
+    // This indicates a sliver where triangles are nearly coplanar but facing
+    // opposite directions
+    return dotProduct < angleThreshold;  // e.g., -0.99 for nearly antiparallel
```
elalish, line 121: I wonder if our `CreateFaces()` function should handle this case? I'm a little surprised it doesn't currently...

src/edge_op.cpp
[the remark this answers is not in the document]
zalo: Presently, this is just the dot product threshold for slivers; every sliver edge I’ve tested has a folding angle of “Pi” radians (to like 9 decimal places), so a configurable threshold might be silly for such obvious degeneracies…

src/edge_op.cpp
zalo: I respect the principle, but I’d be curious to hear of “edge” cases 😄

Since this routine only gloms onto 180* folding edges, it should only activate when they are neighbors 🧐

test/boolean_test.cpp
zalo: I think @pca006132 once figured out how to get triangulation errors / slivers to arise with just two strategically placed rotated cubes, but I think it’s non-deterministic, so a high N is just creating many chances for it to occur (since coplanar simplification is also having trouble, so the top and bottom surfaces are quite crunchy).

I’ll report back soon with more detailed pictures of what happens when N is low or the gear is pre-evaluated…

src/edge_op.cpp
zalo: That’s deep enough into the heart of the CSG routine that I’m unfamiliar 😅

I was thinking that this would be a lot easier to catch right when the faces are created though… would have to see how many unit tests break… when I put the antiparallel edge collapse routine into “removeDegenerateTriangles” as well, it only broke the “mergeEmpty” unit test, because it reduced the mesh to 0 verts too early.  So maybe it’s a free win?

zalo:
> I'm going to say hard pass on the extra Boolean ops to fix slivers. That's a bit too ugly for my taste, and makes me fear infinite recursion. Let me consider...

My first principled thought was to take the folded edge, split it at the midpoint, and move the new vert down to the opposite vertex of the edge (eliminating the triangle with two concave edges), but for many of the real webbed slivers, this would fold another triangle back on itself 180° and start the nightmare all over again, but worse since this one couldn’t be fixed that way. 🫠

CSG seemed clean since it already has retriangulation logic built in… but perhaps catching them at the point of creation will work better…

Though, I note, slivers can be made of many triangles, and it’s not always obvious whether a triangle is part of a sliver until its outer neighbors are removed first… the only way I can think of to get them all in sort of one shot is with thickness raycasting, which sucks…

test/boolean_test.cpp
zalo: > Does it also repro with N = 2 or 4? Also, have you tried forcing evaluation of the gear, before you subtract them from eachother, e.g. `OuterGeart.NumVert()`? I'm curious if the Boolean reordering is giving us trouble.

Slivers don't repro until N = 10.  Also, I was wrong earlier, these slivers are deterministic.

Forcing evaluation of the gear before subtraction doesn't change the number or location of the slivers :-(

zalo:
Just 95% finished the implementation; simplified and moved the CSG subtraction part to the API Simplify call in `manifold.cpp` to get it to work.

This now gets rid of all of the slivers, save for this mysterious double webbed sliver... which I'll probably need some slightly more advanced test (raycasting?) to suss out…

https://github.com/user-attachments/assets/798957d6-b2b1-4898-bbdc-aa89e45dbebf

While this might be a dead path from a code aesthetics perspective, it'll be useful for me to see if something like this fixes the issues with the Minkowski / Offset / Sweep PRs...

zalo:
Ooh, I’m not crazy!  Geogram’s Minkowski also creates slivers and he has to clean them up too!
https://x.com/brunolevy01/status/1967664413311373355?s=46

elalish:
So, what's the next step here? Think you can get the tests to pass? Are you waiting for feedback?

zalo:
Honestly, as work ebbs and flows, this project got moved to my back burner 💀 

I think detecting 180° edges can be an indicator that the slivering bug has occurred within the kernel, so perhaps if asserts are added at every stage of the CSG pipeline to check for them, it might be possible to track down specifically where it occurs... but I haven't the faintest idea where or how to do that 🫠 

I never came up with any better ideas than using CSG to clean up the webbed slivers...   I can remove the CSG parts from this PR so just the AntiparallelEdge Collapse routine can improve the quality of the outputs to the state shown in https://github.com/elalish/manifold/pull/1347#issuecomment-3240241436 if that would be an acceptable intermediate improvement.

elalish:
Sure, that sounds reasonable for now. Thanks.

mmiscool:
I am very interested to try this when it makes it's way to wasm npm build.

zalo:
> Sure, that sounds reasonable for now. Thanks.

@elalish For what it's worth, I did remove the CSG Simplification (which was causing the tests to fail).  

To recap: This post-process in this PR will only partially cleanup meshes; webbed slivers will remain (since I don't know how to trivially collapse them 😅 ).

Also, this routine will only work if you call `myCleanedManifold = myDirtyManifold.as_original().simplify(0.000000001)` on your mesh 😅

zalo:
Also, here's a testing script that shows both the original nasty_gear, the simplified version that this PR produces, and the final CSG cleanup version (implemented in Python): [nasty_gear_cleanup.py](https://github.com/user-attachments/files/23686609/nasty_gear_cleanup.py)

Hmm, part of me is wondering if newer manifold releases perform this process implicitly now...

zalo:
Oooh, interesting, it seems the latest release of Manifold seems to make this PR redundant; [here's the nasty gear and reproduction in manifoldcad.org.](https://manifoldcad.org/#Nasty%20Gear%3Ccode%3Eimport%20%7BManifold%7D%20from%20'manifold-3d%2FmanifoldCAD'%3B%0A%0Afunction%20multi_rot_cube(N%3D10.0)%7B%0A%20%20%20let%20alpha%20%3D%20360.0%2FN%3B%0A%20%20%20let%20cubes%20%3D%20Manifold.cube(%5B1%2C%201%2C%201%5D%2C%20true).rotate(%5B0%2C%200%2C%200%5D)%3B%0A%20%20%20for(let%20i%20%3D%201%3B%20i%20%3C%20N%3B%20i%2B%2B)%7B%0A%20%20%20%20cubes%20%3D%20cubes.add(Manifold.cube(%5B1%2C%201%2C%201%5D%2C%20true).rotate(%5B0%2C%200%2C%20alpha*i%5D))%3B%0A%20%20%20%7D%0A%20%20%20return%20cubes%3B%0A%7D%0A%0A%2F%2F%20The%20original%20nasty%20gear%2C%20it%20looks%20especially%20bad%20on%20web%20for%20some%20reason%0Alet%20result%20%3D%20multi_rot_cube(50.0).scale(%5B2%2C%202%2C%201%5D).subtract(multi_rot_cube(50.0))%3B%0A%0A%2F%2F%20This%20line%20appears%20to%20perform%20the%20same%20function%20as%20the%20antiparallel%20edge%20collapse%20PR%20now%0Aresult%20%3D%20result.asOriginal().simplify(0.000001)%3B%0A%0Aexport%20default%20result%3B%0A) @mmiscool 

Unfortunately, it also suffers from the webbing artifact 😓

mmiscool:
Yeah. I upgraded to the latest version of the library in the hope that the "webbing" with zero thickness model areas would go away but no dice.

pca006132:
Interesting, but I don't remember we did anything related recently. Maybe this PR is getting too old.

PR #1349, opened 2025-09-01
fix single point import

### LINKED ISSUE #1348: Error status for single point could be improved
If one makes a MeshGL with a single point in it and no triangles, and then tries to convert it to Manifold, it returns with a Status of Manifold::Error::NonFiniteVertex.

Clearly this is silly to try, but it would be nicer if the error message were something like Manifold::Error::NotManifold (?).  Or else return an empty Manifold, if that is allowed and if isolated vertices in the input are ignored (I don't really know what the expectation is there).

Not high priority, but a user complained about an inscrutible error message from me.

Attached is a sample program.

[point.cpp.txt](https://github.com/user-attachments/files/22068720/point.cpp.txt)

Trzeth:
fix #1348

### DIFF
--- a/src/impl.h
+++ b/src/impl.h
@@ -68,12 +68,12 @@ struct Manifold::Impl {
     const uint32_t numVert = meshGL.NumVert();
     const uint32_t numTri = meshGL.NumTri();
 
-    if (!numVert && !numTri) {
+    if (numVert == 0 && numTri == 0) {
       MakeEmpty(Error::NoError);
       return;
     }
 
-    if ((numVert < 4) || (numTri < 4)) {
+    if (numVert < 4 || numTri < 4) {
       MakeEmpty(Error::NotManifold);
       return;
     }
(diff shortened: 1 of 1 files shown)


src/impl.h
```diff
@@ -68,6 +68,16 @@ struct Manifold::Impl {
     const uint32_t numVert = meshGL.NumVert();
     const uint32_t numTri = meshGL.NumTri();
 
+    if (!numVert && !numTri) {
```
elalish: I prefer the readability of `numVert == 0`.

src/impl.h
```diff
@@ -68,6 +68,16 @@ struct Manifold::Impl {
...
+      MakeEmpty(Error::NoError);
+      return;
+    }
+
+    if ((numVert < 4) || (numTri < 4)) {
```
elalish: More parentheses than necessary, I think.

PR #1351, opened 2025-09-02
Refactor WASM state and export.

tonious:
Hello!  Here's some more TypeScript refactoring.  The goal is to generalize ManifoldCAD (the tool) functionality to be accessible outside of ManifoldCAD (e.g.: in Node.js) and hopefully make it more maintainable.

Goals for this PR:

 - [x] Split ManifoldCAD state into digestible chunks.
 - [ ] ~~Refactor those chunks for consistency.~~
 - [ ] ~~Demonstrate ManifoldCAD helpers and GLTF semantics (e.g.: `show`, `only`) outside of the evaluator.~~
 - [x] Improve `3mf` export.
 - [ ] ~~Set up a pattern for eventual support of additional formats.~~
 - [ ] ~~Improve WASM binding documentation.~~

With the additional criteria of *not* breaking any existing ManifoldCAD designs.

Like my [previous PR](https://github.com/elalish/manifold/pull/1329), this is going to touch lots of bits.  Please feel free to jump in with questions, comments and rants.


bindings/wasm/lib/export.ts
```diff
@@ -477,185 +174,147 @@ function cloneNodeNewMaterial(
...
   const mm2m = 1 / 1000;
   const wrapper = doc.createNode('wrapper')
                       .setRotation([-halfRoot2, 0, 0, halfRoot2])
                       .setScale([mm2m, mm2m, mm2m]);
   doc.createScene().addChild(wrapper);
+  return wrapper
+}
+
+export function manifoldToGLTFDoc(
+    manifold: Manifold, defaults: GlobalDefaults) {
```
elalish: Could this function just create a GLTFNode from the manifold and then call `GLTFNodesToGLTFDoc` instead? Reduce duplication a bit - not a big deal though.

bindings/wasm/lib/gltf-transform/export-3mf.ts
```diff
@@ -0,0 +1,216 @@
...
+  license?: string;
+  modificationDate?: string;
+
+  /**
+   * Convert a GLTF-Transform document to a 3MF model.
+   *
+   * @param doc The GLTF document to convert.
+   * @returns A blob containing the converted model.
+   */
+  asBlob(doc: Document) {
```
elalish: as3MFblob perhaps?

bindings/wasm/lib/gltf-transform/export-3mf.ts
```diff
@@ -0,0 +1,216 @@
...
+      if (!manifoldPrimitive && !clonedFrom) {
+        // GLTF Mesh, no manifold primitive,
+        // not an instance of another mesh.
+        // We should handle this case, but for now we do not.
+        console.log('skipping non-ManifoldCAD mesh')
+      }
+    }
+
+    // Some 3MF parsers (like PrusaSlicer) expect child nodes
+    // to be defined before their parents.
```
elalish: 👍  - I wrote the spec, but I can't remember if that's a requirement or not...

bindings/wasm/lib/gltf-transform/export-3mf.ts
```diff
@@ -0,0 +1,216 @@
...
+        // Most 3MF parsers will not accept a number in scientific notation.
+        // Transforms are serialized to a string, containing 12 numbers
+        // separated by spaces.  If we force a number to a string here,
+        // 3mf-export passes it through.
+        transform: node.getMatrix().map(n => n.toFixed(to3mf?.precision ?? 7))
+      };
+      const parent = node.getParentNode();
+
+      if (parent) {
+        // This is a child node, add it to it's parent.
```
elalish: nit: its

bindings/wasm/lib/gltf-transform/export-3mf.ts
elalish, line 1: Out of curiosity, why the gltf-transform folder? Inside are some 3mf things, and other gltf-transform related files are outside. Maybe just flatten?

elalish (approved):
This looks excellent - very nice job reorganizing everything into separate files. Do you want to merge this PR before moving on with the rest of your list? 

Also, since you're doing such fine work in this area, might you be interested in the import side (glTF and/or 3MF)? That's the obvious piece ManifoldCAD is missing. And one other functionality I've always wanted is loading in other people's ManifoldCAD script functions as libraries (a la OpenSCAD modules).

bindings/wasm/lib/export.ts
tonious: Makes sense.  It's really more of a helper for when users _aren't_ using GLTFNodes from the outset.

bindings/wasm/lib/gltf-transform/export-3mf.ts
tonious: I haven't formalized the interface here, but I was thinking about multiple exporters.  Something like:

```typescript
const filename = 'mymodel.3mf';
const exporter = exporters.find(ex => ex.extension === path.extname(filename))!;
const blob = await exporter.asBlob(doc);
```

And similarly:

```typescript
await exporter.write(filename, doc);
```

bindings/wasm/lib/gltf-transform/export-3mf.ts
tonious: Yeah.  Sorry.  I speak English like a Canadian :sob:

bindings/wasm/lib/gltf-transform/export-3mf.ts
tonious: It's not a requirement AFAICT, and shouldn't actually be necessary for any DOM oriented parser.  But it's also not a hardship.

bindings/wasm/lib/gltf-transform/export-3mf.ts
tonious: I'd intended to put files in here that depend on gltf-transform, but not the rest of Manifold.  E.g.: the GLTF extension, converting from GLTF to other formats and vice versa.

Either way, I'm running into namespace issues.

tonious:
I really appreciate the vote of confidence -- I actually hadn't considered this merge-worthy just yet.  It makes sense to do tighter iterations, so I'll clean this up and get it in.


Model imports are definitely on my list.  I'm considering doing this two ways:

- Importing a solid model into ManifoldCAD directly is obviously the primary use case.
- I'd also like to be able to import something and just stick it into the GLTF document.  Picture being able to see a ManifoldCAD part in a multi-model assembly, without having to process the other models in Manifold itself.

Module imports likewise.  This is actually why I wanted nodejs support in the first place 😁 
Jscad does a really good job of this in a web context.

tonious:
Naming things is hard.  `exporter` doesn't really cover what it does any more.  does `builder` make more sense?  `model-state`?  `gltf-state`?

bindings/wasm/lib/gltf-transform/export-3mf.ts
tonious: Changed it up slightly: some exporters might have multiple extensions, like `.gltf` and `.glb`.

```typescript
const exporter = exporters.find(ex => ex.extensions.includes(path.extname(filename)));
```

I'm using this in the cli, [here](https://github.com/elalish/manifold/blob/0e3f89bd05bbb4ee91f7c3ef6239ffa90a02eb8b/bindings/wasm/bin/manifold-cad#L48-L52) and [here](https://github.com/elalish/manifold/blob/0e3f89bd05bbb4ee91f7c3ef6239ffa90a02eb8b/bindings/wasm/bin/manifold-cad#L89-L95).

bindings/wasm/lib/gltf-transform/export-3mf.ts
elalish: Someone should still tell Prusa to fix his bug though...

elalish (approved):
How about `scene-builder`? It's building a doc, which is really a node hierarchy with materials and animation, which happens to be GLTF-compatible, but we also use for 3MF. I've generally heard that called a scene.

LGTM - Let me know when you're ready to merge.

elalish:
>I'd also like to be able to import something and just stick it into the GLTF document. Picture being able to see a ManifoldCAD part in a multi-model assembly, without having to process the other models in Manifold itself.

Yeah, this makes perfect sense. Seems like import should mirror export: 
```
const binary = await readFile(filename);
const doc = await importer.fromStream(binary);
const manifold = meshToManifold(doc.getRoot().listMeshes()[0]);
```
Or perhaps even one extra step through our own mesh format, since that would give users a chance to run `merge` and add IDs and such:
```
const mesh = readMesh(doc.getRoot().listMeshes()[0]);
const manifold = Manifold(mesh);
```

bindings/wasm/lib/gltf-transform/export-3mf.ts
tonious: I've found other viewers that don't like intermediate nodes in between the root and any meshes, and at least one that doesn't like transformation matrices.  I didn't take any real notes, so this is anecdotal, not science.  These were just model viewer web sites -- I wanted to make sure I didn't break the model.

The XSL schema came in handy.  At some point that'd make a good integration test.

tonious:
This can be merged any time.

Thanks!

elalish (approved):
Thanks!

elalish:
Looks like the CI failed again - mind taking a look? https://github.com/elalish/manifold/actions/runs/17476258780/job/49636749621

tonious:
There's always something.

#1353

PR #1360, opened 2025-09-19
Make evaluator asynchronous, and other small TS tweaks.

tonious:
Hello!

This PR is pretty small, because I've been playing with building things _in_ manifold, rather than manifold itself.

* `Evaluator.evaluate` is now asynchronous.  This allows the use of any async functions, but most importantly, `import()`.
* Explicitly exposed`triangulate` to evaluated models.
* Broke out `euler2quat` into it's own file, as it didn't really fit anywhere else.
* Encapsulated the debug and animation builders in behind the scene builder.  This makes usage just a little cleaner:

```typescript
import {Evaluator} from '../lib/evaluate.js';
import * as scenebuilder from '../lib/scene-builder.js';

// Setup the evaluator and context.
const evaluator = new Evaluator();

evaluator.addContextMethodWithCleanup('show', scenebuilder.show);
evaluator.addContextMethodWithCleanup('only', scenebuilder.only);
evaluator.addContextMethodWithCleanup('setMaterial', scenebuilder.setMaterial);
evaluator.addContext({
  GLTFNode: scenebuilder.GLTFNode,
  setMorphStart: scenebuilder.setMorphStart,
  setMorphEnd: scenebuilder.setMorphEnd,
});
```

### DIFF
--- a/bindings/wasm/lib/evaluate.ts
+++ b/bindings/wasm/lib/evaluate.ts
@@ -12,8 +12,6 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-import * as glMatrix from 'gl-matrix'
-
 import {CrossSection, Manifold} from '../manifold-encapsulated-types';
 import type {ManifoldToplevel} from '../manifold.d.ts';
 import Module from '../manifold.js';
@@ -21,9 +19,6 @@ import Module from '../manifold.js';
 // Instantiate Manifold WASM
 const manifoldwasm = await Module();
 manifoldwasm.setup();
-
-// Faster on modern browsers than Float32Array
-glMatrix.glMatrix.setMatrixArrayType(Array);
 
 // manifold static methods (that return a new manifold)
 const manifoldStaticFunctions = [
@@ -214,7 +209,6 @@ export class Evaluator {
     const context = {
       ...Object.fromEntries(exposedFunctions),
       module: this.module,
-      glMatrix,
       ...this.context
     };
     const AsyncFunction =
--- a/bindings/wasm/bin/manifold-cad
+++ b/bindings/wasm/bin/manifold-cad
@@ -17,6 +17,7 @@ import {readFile, writeFile} from 'node:fs/promises';
 import {resolve, extname} from 'node:path';
 
 import {Command} from 'commander';
+import glMatrix from 'gl-matrix';
 
 import {Evaluator} from '../lib/evaluate.js';
 import * as scenebuilder from '../lib/scene-builder.js';
@@ -25,6 +26,7 @@ import {ExportGLTF} from '../lib/export-gltf.js';
 
 // Setup the evaluator and context.
 const evaluator = new Evaluator();
+evaluator.addContext({glMatrix});
 
 // These are methods that generate Manifold
 // or CrossSection objects.  Tell the evaluator to intercept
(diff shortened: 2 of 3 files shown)


bindings/wasm/lib/evaluate.ts
```diff
@@ -12,6 +12,8 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+import * as glMatrix from 'gl-matrix'
```
elalish: Did we have this as a necessary dependency before? Do you think any users would rather use a different math library?

bindings/wasm/lib/scene-builder.ts
```diff
@@ -21,10 +21,15 @@ import {GLTFMaterial} from '../examples/public/editor';
 import {Manifold} from '../manifold-encapsulated-types';
 import {Vec3} from '../manifold-global-types';
 
-import {addAnimationToDoc, addMotion, cleanupAnimation, euler2quat, getMorph, morphEnd, morphStart, setMorph} from './animation.ts';
-import {getDebugGLTFMesh, getMaterialByID} from './debug.ts'
+import {addAnimationToDoc, addMotion, cleanup as cleanupAni, cleanupAnimation, getMorph, morphEnd, morphStart, setMorph} from './animation.ts';
```
elalish: What's the difference between cleanupAni and cleanupAnimation?

bindings/wasm/lib/euler2quat.ts
elalish, line 1: Maybe just call this math.js, in case we need other functions later?

bindings/wasm/lib/scene-builder.ts
tonious: One cleans up the module, and the other removes unused animation attributes from the GLTF document.

I'll just go rename those....

bindings/wasm/lib/evaluate.ts
tonious: It is already used by the editor (another package, I know) and `euler2quat`.  I think it's okay to include it here as it's pretty small -- 860kb on disk, less when packaged.

There are other packages, and some of them have nicer APIs.  It's easy enough to add them through `addContext()`.  In fact, if a user adds anything named `glMatrix` through `addContext()` it will override this.  It just seemed to be a sensible default to me.

This is a weakly held opinion -- I'll move it back if you prefer.

bindings/wasm/lib/scene-builder.ts
tonious: "Refactor `animation.ts` such that accessors and samplers are made on demand" is already on mental to-do list, fwiw.

bindings/wasm/lib/evaluate.ts
elalish: 860kb? Is that a typo, because that sounds pretty huge for a vector library. I used it because a friend of mine wrote it, but he doesn't seem to have time to maintain it anymore, and the API feels a bit clunky - I think with modern V8 there's not much performance need to reuse every vec3 anyway. Since I found linalg for C++ I'm starting to have a higher bar for math libraries. 

The trouble with including glMatrix, is then if we were ever to try and remove it, it would break everyone's scripts. I think it's better to copy a little more math into euler2quat to avoid a significant dependency.

tonious: 860kb for the whole npm package.  52kb for the minified file.  Still seems large, though.

Removing it entirely breaks a few of the example scripts, so I've reverted this to as it was before -- added to context in `worker.ts` (and `bin/manifold-cad`).

Issue #1368, opened 2025-09-29
Debian Package?


nomike:
Hi!

I'm one of the PythonSCAD developers and currently mainly working on getting builds and tests more stable and preparing for adding the program to various Linux distros.

I was wondering whether you consider submitting manifold to Debian and Fedora so we can use it for PythonSCAD (and probably OpenSCAD as those peeps are working towards a release as well).

I could maybe help out with the technical side of creating .deb and .rpm packages.

t-paul:
We can also check with @knielsen, if there's time to look at this.

pca006132:
I think the ideal way is to find someone familiar with the distro (better if they are already package maintainers) to do that.

@t-paul I think you have a lot more experience dealing with distro packaging, what do you think is the best way here?

pca006132:
Oh, apparently you replied before I refresh :)

nomike:
> We can also check with [@knielsen](https://github.com/knielsen), if there's time to look at this.

Seem like a good idea as they are already package maintainer of the OpenSCAD package.

nomike:
I explored this a bit, but it looks like clipper2 is not yet available on Debian.
So we might need to get this packaged first.
There is clipper already, but not clpper2 unfortunately.

t-paul:
True, but note that this is `libpolyclipping-dev`, the package called `clipper` is something totally different.

knielsen:
I'd like to find some time to work on some Debian packaging issues. Bit busy just now, but maybe a bit later.

Could involve getting to package manifold (and clipper2), as well as lib3mf2 and the code generation dependency, and getting the current master in git building as a package as well in preparation for next release or packaging a snapshot.

t-paul:
Heh, same here. Maybe I'll have a look at OBS which could serve as a staging ground to get the packaging worked on by multiple people. Or does Debian have something like that with access by non-DDs?

nomike:
I created two PRs for starters...

https://github.com/AngusJohnson/Clipper2/pull/1017
https://github.com/elalish/manifold/pull/1369

t-paul:
I'm not a Debian Developer, but I believe I've read that including the `debian` folder in the upstream repo is not recommended. I suppose the reason for that is to decouple the packaging from the upstream source. In case it's going to be maintained by someone from the upstream team, it's fine.

nomike:
That's a good point. I did pull out the clipper2 Debian stuff into it's own repository:
[...]

PR #1369, opened 2025-09-29
Add Debian packaging support

nomike:
- Add debian/ directory with control, rules, and install files
- Update .gitignore to exclude Debian build artifacts
- Document Debian package build process in README
- Create changelog and copyright files for packaging

This surely needs reviewing and polishing, but it's a start.
Feel free to adapt/edit/steal everything.


pca006132:
1. I think we should distinguish between language packages and distro packages. Should put them in two different tables instead of one, or use badges for distros. And we should not say vcpkg or debian package, as debian package only works for debian and derivatives.
2. I am still uncertain about putting distro related stuff into our main repo. First, this should be put into a directory for distros, with debian as one of the subdirs. Second, I am not sure how will this affect our release. E.g. if there are patches needed to make the build work with some distro due to dependency issues, should that stay here? And how will this affect releases? We release something and then change the distro files? The source shipped to distros are having outdated distro files?

nomike:
Oh. I forgot to close this PR. 
There is some discussion ongoing in #1368.
I have now extracted that Debian stuff into its own repository.

PR #1375, opened 2025-10-06
Draft: Remediate top level async calls.

tonious:
Adapting some of @Loosetooth's amazing work (see  #1373)

This PR:

  * Breaks out instantiation of the Manifold module into `wasm.ts`.
  * Instantiation now takes place in a function, rather than at the top level.
  * The evaluator can be passed a wasm instance.  If it has no instance at evaluation time, it will create one then.
  * The evaluator checks for any loose garbage-collectible objects when setting a module.


bindings/wasm/lib/evaluate.ts
```diff
@@ -102,44 +99,34 @@ export class Evaluator {
   afterScript: string =
       'return typeof result === "undefined" ? undefined : result;';
 
-  protected module: ManifoldToplevel = manifoldwasm;
+  protected module: ManifoldToplevel|undefined;
```
elalish, line 102: I generally prefer `|null = null` - tends to be somewhat easier to read and debug.

bindings/wasm/lib/wasm.ts
```diff
@@ -0,0 +1,52 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+import type {ManifoldToplevel} from '../manifold.d.ts';
+import Module from '../manifold.js';
+
+// Instantiate Manifold WASM
+let manifoldwasm: ManifoldToplevel|undefined;
+let wasmUrl: string|undefined = undefined;
```
elalish: prefer `null`

elalish (approved):
This looks great, thank you! Is it actually still a draft for any reason?

tonious:
I wanted @Loosetooth to have a chance to double check my work from their end.  They've kindly done so.  This can go in any time.

For what it's worth, there's still more work to be done here in JS/TS land, but I think this is a good point to cut an NPM release.  The evaluator API is reasonably stable.  There might be some flux in the scene builder, but I think it's going to be enhancements over breaking changes.

bindings/wasm/lib/evaluate.ts
elalish: Might want to do a global find of `undefined` at some point...

elalish:
Yes, agreed that we should get this published to npm sooner than later. But let's deal with https://github.com/cookiecad/manifold-vscode-extension/pull/1#issuecomment-3378584351 first, since that feels like a potential breaking change if we don't figure it out now.

PR #1380, opened 2025-10-09
Refactor worker.ts

tonious:
Spurred on by a [discussion](https://github.com/cookiecad/manifold-vscode-extension/pull/1) about formalizing the web worker, here's an attempt.

* Worker moved from `/examples` to `/lib`
* `worker-wrapper.ts` removed -- this implementation is polymorphic.
* Messages to and from worker now have a consistent structure, with defined interfaces.  That said, I'm not sure I've done that in an idiomatic way.
* Evaluating the model and exporting it to a blob are now separate calls.  This allows implementations to only create a 3MF file on demand.

Still considering:
  * Philosophically, the worker is where the evaluator and scene builder come together.  I'm not sure 'worker' is the right name.  Maybe 'document builder'? (`worker` is close enough.)
  * Using the npm package results in statements like `import {Evaluator} from 'manifold-3d/lib/evaluate';`.  I wonder if that 'lib' could be removed.  Is there a more natural way to structure these package exports? (Naaah.  Too much work :grin:)
  * Packaging `editor.d.ts`.  Ideally, it should be generated or inferred somehow. (moved to `/types/manifoldCAD.d.ts`, sans `gl-matrix`.)

As always, I'm open to comments, suggestions or rants, as the case may merit.

### DIFF
--- a/bindings/wasm/examples/editor.js
+++ b/bindings/wasm/examples/editor.js
@@ -284,8 +284,15 @@ declare const module: ManifoldToplevel;
 `;
 }
 
-async function getEditorDTS() {
-  const global = await fetch('/editor.d.ts').then(response => response.text());
+async function getManifoldCadDTS() {
+  const global =
+      await fetch('/manifoldCAD.d.ts').then(response => response.text());
+  return `${global.replace(/^import.*$/gm, '')}`;
+}
+
+async function getGlMatrixDTS() {
+  const global =
+      await fetch('/gl-matrix.d.ts').then(response => response.text());
   return `${global.replace(/^import.*$/gm, '')}`;
 }
 
@@ -297,7 +304,9 @@ require(['vs/editor/editor.main'], async function() {
   monaco.languages.typescript.typescriptDefaults.addExtraLib(
       await getManifoldDTS());
   monaco.languages.typescript.typescriptDefaults.addExtraLib(
-      await getEditorDTS());
+      await getManifoldCadDTS());
+  monaco.languages.typescript.typescriptDefaults.addExtraLib(
+      await getGlMatrixDTS());
   editor = monaco.editor.create(
       document.getElementById('editor'),
       {language: 'typescript', automaticLayout: true});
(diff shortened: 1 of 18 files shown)


bindings/wasm/lib/worker.ts
```diff
@@ -0,0 +1,288 @@
...
+  scenebuilder?.cleanup();
+}
+
+/**
+ * Transform a model from code to a GLTF document.
+ *
+ * @param code A string containing the code to evaluate.
+ * @returns A gltf-transform Document.
+ */
+export async function evaluate(code: string) {
```
elalish: I'd prefer to use the TS return type syntax on function declarations, especially important ones like this, rather than having it inferred.

bindings/wasm/lib/worker.ts
```diff
@@ -0,0 +1,288 @@
...
+      (Math.round((t1 - t0) / 10) / 100).toLocaleString()} seconds`);
+  return blobURL;
+}
+
+/**
+ * This module is polymorphic.  If it's run as a web worker,
+ * set up message handlers and logging.
+ */
+const initializeWebWorker =
+    function() {
```
elalish: why `function()` here and `=>` above`?

bindings/wasm/lib/worker.ts
```diff
@@ -0,0 +1,288 @@
...
+    }
+  };
+
+  const handleExport = async (message: MessageToWorker.Export) => {
+    const blobURL = await exportBlobURL(gltfdoc!, message.extension);
+
+    self.postMessage({
+      type: 'blob',
+      extension: message.extension,
+      filename: `manifold${message.extension}`,
```
elalish: The filename seems a bit arbitrary to be part of the API - the extension seems like plenty of information.

bindings/wasm/examples/editor.js
```diff
@@ -448,42 +449,59 @@ const output = {
...
+    } else if (message?.type === 'done') {
       setScript('safe', 'true');
+      manifoldWorker.postMessage({type: 'export', extension: '.glb'});
+      manifoldWorker.postMessage({type: 'export', extension: '.3mf'});
+
+      finishRun();
+      runButton.disabled = true;
+
+    } else if (message?.type === 'blob') {
+      if (message.extension === '.glb') {
```
elalish: I would tend to expect the extension to be only `glb`, excluding the dot. What seems least surprising, considering this is public API?

bindings/wasm/lib/worker.ts
```diff
@@ -0,0 +1,288 @@
...
+
+  const t1 = performance.now();
+  log(`Exporting ${extension.toUpperCase()} took ${
+      (Math.round((t1 - t0) / 10) / 100).toLocaleString()} seconds`);
+  return blobURL;
+}
+
+/**
+ * This module is polymorphic.  If it's run as a web worker,
+ * set up message handlers and logging.
```
elalish: Since this file is the main interface now, we should probably put a big comment at the top of this file explaining its overall usage - either calling functions directly or else indirectly via messages if we're in a different context.

bindings/wasm/lib/worker.ts
tonious: No reason.  I'll stick with an arrow function.

bindings/wasm/examples/editor.js
tonious: I'd personally expect the dot.  But that's not a strong opinion, and I'm happy to change it.

elalish:
Looks like we missed a path somewhere. Care to take a look? https://github.com/elalish/manifold/actions/runs/18436993955

PR #1389, opened 2025-10-15
Fall back to packaged type files if not built.

tonious:
Suggestion from @Loosetooth, [over here](https://github.com/elalish/manifold/pull/1387#issuecomment-3408447532).


elalish (approved):
Nice! Maybe a comment somewhere about what this is a fallback for? And what kind of problems one might expect if it were to break in the future?

tonious:
This just makes life a little easier for us lowly javascript devs when we forget to build a fresh set of wasm bindings :joy:

Not that I've ever done that myself.... 🤥

Issue #1388, opened 2025-10-15
ImportMesh not compiling with assimp (ASSIMP_DOUBLE_PRECISION=ON)


surrim:
When assimp was compiled with `ASSIMP_DOUBLE_PRECISION=ON` the `ImportMesh` function will not compile because of narrowing doubles to floats.
To fix it use `aiVector3f` instead of `aiVector3D`:

```diff
diff --git a/src/meshIO/meshIO.cpp b/src/meshIO/meshIO.cpp
index 748e014..46c85d5 100644
--- a/src/meshIO/meshIO.cpp
+++ b/src/meshIO/meshIO.cpp
@@ -147,7 +147,7 @@ MeshGL ImportMesh(const std::string& filename, bool forceCleanup) {
   for (size_t i = 0; i < scene->mNumMeshes; ++i) {
     const aiMesh* mesh_i = scene->mMeshes[i];
     for (size_t j = 0; j < mesh_i->mNumVertices; ++j) {
-      const aiVector3D vert = mesh_i->mVertices[j];
+      const aiVector3f vert = mesh_i->mVertices[j];
       if (isYup)
         mesh_out.vertProperties.insert(mesh_out.vertProperties.end(),
                                        {vert.z, vert.x, vert.y});
```

elalish:
Thanks - care to open a PR?

PR #1391, opened 2025-10-17
using static_cast<float> in ImportMesh to support assimp with double-precision

surrim:
When assimp was compiled with ASSIMP_DOUBLE_PRECISION=ON the ImportMesh function will not compile because of narrowing doubles to floats.
To fix it I used aiVector3f instead of aiVector3D.

This fixes https://github.com/elalish/manifold/issues/1388#issuecomment-3407964647

### DIFF
--- a/src/meshIO/meshIO.cpp
+++ b/src/meshIO/meshIO.cpp
@@ -147,13 +147,17 @@ MeshGL ImportMesh(const std::string& filename, bool forceCleanup) {
   for (size_t i = 0; i < scene->mNumMeshes; ++i) {
     const aiMesh* mesh_i = scene->mMeshes[i];
     for (size_t j = 0; j < mesh_i->mNumVertices; ++j) {
-      const aiVector3D vert = mesh_i->mVertices[j];
+      const aiVector3t<float> vert = mesh_i->mVertices[j];
       if (isYup)
-        mesh_out.vertProperties.insert(mesh_out.vertProperties.end(),
-                                       {vert.z, vert.x, vert.y});
+        mesh_out.vertProperties.insert(
+            mesh_out.vertProperties.end(),
+            {static_cast<float>(vert.z), static_cast<float>(vert.x),
+             static_cast<float>(vert.y)});
       else
-        mesh_out.vertProperties.insert(mesh_out.vertProperties.end(),
-                                       {vert.x, vert.y, vert.z});
+        mesh_out.vertProperties.insert(
+            mesh_out.vertProperties.end(),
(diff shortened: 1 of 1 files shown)


surrim:
```
/home/runner/work/manifold/manifold/src/meshIO/meshIO.cpp:150:13: error: ‘aiVector3f’ does not name a type; did you mean ‘aiVector3t’?
  150 |       const aiVector3f vert = mesh_i->mVertices[j];
      |             ^~~~~~~~~~
      |             aiVector3t
```

Works fine with GCC 15 and latest v5/v6 assimp. Cannot reproduce with GCC 15.2.
Defined [here](https://github.com/assimp/assimp/blob/cc6ab1408fa0dc7b386f5b8e5883c3cedcae950b/include/assimp/vector3.h#L154C1-L154C38), should be defined `#ifdef __cplusplus`, which should be the case.

elalish:
Looks like a gcc 14 problem. Are you missing an include? What defines aiVector?

surrim:
> Looks like a gcc 14 problem. Are you missing an include? What defines aiVector?

As I wrote, when aiVector3D is known, aiVector3f should be also available.
You can try to `#include <assimp/vector3.h>`, but I don't want to guess. What version of assimp is used for this build?

Trzeth:
aiVector3f definition add on assimp 5.4.0.
CI's ubuntu 24.04 is using 5.3.1 from apt, 22.04 uses 5.2.2.

You can fix this by checking ASSIMP_VERSION_MAJOR and ASSIMP_VERSION_MINOR.

surrim:
> You can fix this by checking ASSIMP_VERSION_MAJOR and ASSIMP_VERSION_MINOR.

The only way to find out the assimp version is using functions like [`ASSIMP_API unsigned int aiGetVersionMajor (void)`](https://github.com/assimp/assimp/blob/6a08c39e3a91ef385e76515cfad86aca4bfd57ff/include/assimp/version.h#L82), which are *non-constexpr*.

`ASSIMP_VERSION_*` is only used to compile assimp, not available in header files. Another way is to get the version number with pgk-config/CMake, but I'm not very familiar with that.

Trzeth:
Maybe just explicit cast it into float?

```
const aiVector3t<float> vert{float(mesh_i->mVertices[j].x), float(mesh_i->mVertices[j].y), float(mesh_i->mVertices[j].z)};
```

surrim:
Using `aiVector3t<float>` because `aiVector3f` didn't exist before assimp 5.4.0.

PR #1392, opened 2025-10-20
Isomorphic Javascript Libraries

tonious:
This has turned into a bigger refactor than I had expected, but I'm fairly happy with how it's going so far.  This builds on the current remote package support, but uses esbuild to dynamically bundle code in the browser.

  * [x] Bundle local files.
  * [x] Bundle remote packages.
  * Better error handling:
    * [x] Inline source maps used to rewrite stack traces.
    * [x] Bundler and runtime exceptions are re-formatted and rethrown.
  * [x] Expose evaluation context to imported modules in evaluator (via `import ... from "manifold-3d/manifoldCAD"`)
  * [x] Expose evaluation context to modules in node (via `import ... from "manifold-3d/manifoldCAD"`)
  * Tests... Aren't quite there yet.
     * ~~Need a good set of tests around `result` logic.  It will be easy to break.~~
     * [x] Tests for `isManifoldCAD()` and `export default`
  * Helpers:  
    * ~~`import.meta.main` support (allow a module to know if it's the entry point or not).~~ `export default` meets our needs
    * [x] `isManifoldCAD()` support (allow a module to know if it's in manifoldCAD or not).

And of course, all of this needs documentation.  Buuut, I think this might be stable enough to play with.


tonious:
Also, tests are now entirely in the main module, rather than examples.

elalish:
I'm starting to review this now - if possible can you avoid force-pushing your branch? It makes it much easier to do follow-up reviews if I can just look at the latest commits that have changed stuff, rather than rewriting the history. Don't worry about merge commits and stuff in the branch history - we squash on merge anyway, so it all gets cleaned up.

bindings/wasm/lib/bundle.ts
```diff
@@ -0,0 +1,269 @@
+import {Plugin} from 'esbuild';
```
elalish: We should copy in our copyright header.

bindings/wasm/lib/bundler.ts
```diff
@@ -0,0 +1,269 @@
...
+  }
+};
+
+/**
+ * These content delivery networks provide NPM modules as ES modules,
+ * whether they were published that way or not.
+ */
+export const cdnUrlHelpers: {[key: string]: (specifier: string) => string} = {
+  'esm.sh': (specifier) => `https://esm.sh/${specifier}`,
+  'jsDelivr': (specifier) => `https://cdn.jsdelivr.net/npm/${specifier}/+esm`,
```
elalish, line 39: This is cool, but feels fragile. Should we just have users supply us this function? Maybe something like `esmUrl(package) => Url`?

bindings/wasm/lib/bundle.ts
```diff
@@ -0,0 +1,269 @@
...
+    }
+    await esbuild.initialize(esbuildOptions);
+    esbuild_initialized = true;
+  }
+
+  return {
+    // Create a bundle in memory.
+    bundle: true,
+    write: false,
+    treeShaking: false,  // FIXME ensure that 'result' and dependencies are not
```
elalish: `result` was my first hacky way of getting this working. If it's making things annoying, I'm fine with removing that and saying the only way to produce output is via GLTFNode. I don't have a strong opinion, but consider it an option.

bindings/wasm/lib/evaluate.ts
```diff
@@ -244,24 +285,28 @@ export class Evaluator {
...
-    this.setupGarbageCollection();
-
-    const exposedFunctions =
-        toplevel.map((name) => [name, (this.module as any)[name]]);
-    const context = {
-      ...Object.fromEntries(exposedFunctions),
-      module: this.module,
-      ...this.context
-    };
+    const context = await this.getFullContext()
```
elalish: I wonder if there's a way to make clang-format give us line-ending `;` automatically?

bindings/wasm/lib/manifoldCAD.ts
```diff
@@ -0,0 +1,67 @@
...
+/**
+ * These are the objects and functions that are available to manifoldCAD
+ * scripts.
+ *
+ * This is an isomorphic module.  When imported within manifoldCAD, the bundler
+ * will swap it out for an identical module built from the evaluator context.
+ * When imported as an ES module, it will implicitly instantiate a worker with
+ * an evaluator, and export that context.  This allows modules to behave
+ * identically when running on manifoldCAD.org, through the CLI, or through
+ * nodejs.
```
elalish: 💯

bindings/wasm/lib/evaluate.ts
elalish, line 1: Should this file be called `evaluator.ts`?

bindings/wasm/lib/bundler.ts
elalish, line 1: Should this be called `bundler.ts`?

bindings/wasm/lib/manifoldCAD.ts
```diff
@@ -0,0 +1,67 @@
...
+ * @module manifoldCAD
+ */
+
+import type * as encapsulatedTypes from '../manifold-encapsulated-types.d.ts';
+import type * as manifoldCADTypes from '../types/manifoldCAD.d.ts';
+
+import {initialize} from './worker.ts';
+
+const evaluator = initialize();
+const context = await evaluator.getFullContext();
```
elalish: This may be fine, but I don't completely understand. This appears to be setting up garbage collection, but any call to `evaluate` also sets up garbage collection?

bindings/wasm/lib/worker.ts
```diff
@@ -58,18 +59,29 @@ export namespace MessageToWorker {
...
    *
    * The worker will respond with `MessageFromWorker.Done`
    * or `MessageFromWorker.Error`.
+   *
+   * `filename` doesn't have much meaning at the moment, but it
+   * is useful for ensuring effective error messages later.
+   *
+   * If `bundle` is set to false, the worker will not bundle code,
+   * and assume that it has already been bundled.  An undefined
+   * value will be treated by default as true.
```
elalish: `bundle` is ambiguous - it could mean "please bundle" or "is already a bundle". Maybe rename to `doBundle` or `applyBundling`?

bindings/wasm/lib/worker.ts
```diff
@@ -189,13 +211,26 @@ export function cleanup(): void {
...
 }
 
+/**
+ * If `bundle` is set to false, the worker will not bundle code,
+ * and assume that it has already been bundled.  An undefined
+ * value will be treated by default as true.
+ */
+export interface evaluateOptions {
+  filename?: string;
+  bundle?: boolean;
```
elalish: same comment as above.

bindings/wasm/lib/worker.ts
```diff
@@ -206,13 +241,18 @@ export async function evaluate(code: string): Promise<Document> {
...
+  const {bundle, ...bundleOpt} = options;
+  const bundled = bundle === false ? code : await bundleCode(code, bundleOpt);
   const t1 = performance.now();
+  if (bundle !== false) {
+    log(`Bundling code took ${((t1 - t0) / 1000).toFixed(2)} seconds`);
+  }
 
-  log(`Manifold took ${
-      (Math.round((t1 - t0) / 10) / 100).toLocaleString()} seconds`);
+  const manifold = await evaluator!.evaluate(bundled);
```
elalish: something about calling evaluator.evaluate inside of a function called evaluate makes me want to reevaluate our naming... Can we be any more specific?

bindings/wasm/lib/worker.ts
```diff
@@ -337,3 +381,5 @@ const initializeWebWorker = (): void => {
   }
 };
 if (isWebWorker()) initializeWebWorker();
+
+/* v8 ignore end -- @preserve */
```
elalish: What's this?

bindings/wasm/test/fixtures/unitSphere.mjs
```diff
@@ -0,0 +1,3 @@
+import {Manifold} from '../../lib/manifoldCAD.ts';
+
+const result = Manifold.sphere(1.0);
```
elalish, line 3: The only difference between these files is the `const` - what subtlety are we checking here?

bindings/wasm/test/cdn.test.ts
```diff
@@ -0,0 +1,48 @@
...
+let evaluator = worker.getEvaluator() || worker.initialize();
+beforeEach(() => worker.cleanup());
+
+suite('Import remote modules from', () => {
+  test('esm.sh', async () => {
+    const entrypoint =
+        resolve(import.meta.dirname, './fixtures/importFromCDN.mjs');
+    const bundle = await bundleFile(entrypoint, {jsCDN: 'esm.sh'});
+    const result = await evaluator.evaluate(bundle);
+    expect(result.volume()).toBeGreaterThan(0);
```
elalish: Nice end-to-end testing! I'm slightly concerned about our CI being dependent on external websites, but then again, I think the CDNs have much more stringent uptime requirements than we do, so I think this is good.

bindings/wasm/test/import.test.ts
```diff
@@ -0,0 +1,69 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+import {resolve} from 'node:path';
+import {beforeEach, expect, suite, test} from 'vitest';
+
+import {bundleFile} from '../lib/bundle.ts';
+import * as worker from '../lib/worker.ts';
+
+let evaluator = worker.getEvaluator() || worker.initialize();
```
elalish: I tend to use tests as usage examples when I'm starting to use a new library. This could be a good place for a comment about what's happening with these two ways to get an evaluator.

bindings/wasm/test/stacktrace.test.ts
```diff
@@ -0,0 +1,51 @@
...
+    const filepath =
+        resolve(import.meta.dirname, './fixtures/generateReferenceError.mjs');
+    let code = await bundleFile(filepath);
+
+    const ev = async () => await evaluator.evaluate(code);
+    await expect(ev()).rejects.toThrowError()
+  });
+});
+
+suite('Build model without the evaluator', () => {
```
elalish: Maybe also some comments about why or in what situations one would choose to use the evaluator or not?

bindings/wasm/lib/bundle.ts
tonious: `export default` works as well.

elalish: What do you mean? Can you give me an example?

tonious: Treat the main script as an ES module:
```javascript
const result = 3;
export default result;
```

Getting `result` in a node/esm environment is cleaner.  We don't even really need to know the name, just that it's the default export:
```javascript
import model from `thescript`;
```

In the evaluator, first we run it through esbuild, and get some big long transpiled string.  Syntactically, we need destructure the export, but we're looking for `default`.
```javascript
const bundled = 'code_here';
const evalFn = new AsyncFunction(bundled + "\n" + "return module.exports;");
const { default } = await evalFn();
```
And now we can check for `undefined` and all that jazz outside of the constructed function.  Even better, the user can still export objects beyond the default without breaking any of this.  ManifoldCAD won't see them, but it could be useful for libraries.

bindings/wasm/lib/worker.ts
tonious: I didn't want the big mess of worker specific code messing up my test coverage before I figured out how to actually test it. :sob:

I'm happy to remove this.

bindings/wasm/test/cdn.test.ts
tonious: I'm a little worried about that too, to be honest.  If I can figure out how to make these skip unless referenced specifically, I'll do that.

bindings/wasm/test/fixtures/unitSphere.mjs
tonious: One is `export`ed, while the other is not.

bindings/wasm/lib/worker.ts
tonious: Good call.

bindings/wasm/lib/manifoldCAD.ts
tonious: Ah... I'll rework this a little.  It should be something more like:
```javascript
const evaluator = worker.getEvaluator() || worker.initialize();
const context = await evaluator.getFullContext();
```
This should make it clear that line 34 is instantiating the worker (and implicitly the evaluator) while line 35 is getting the execution context.

bindings/wasm/lib/bundler.ts
tonious: In my working copy I have:
```javascript
const cdnUrl = (specifier: string, jsCDN?: string) => {
  if (!jsCDN) return specifier;
  const helper = cdnUrlHelpers[jsCDN];
  return helper ? helper(specifier) : `${jsCDN}${specifier}`;
};
```
So, if no CDN, return unchanged.  And if it's unknown CDN, us it as a prefix by default?  Or is that too much magic?

bindings/wasm/lib/worker.ts
tonious: I might have gotten carried away.  Maybe `worker.evaluate()` calls `evaluator.run(code)`?

bindings/wasm/test/cdn.test.ts
elalish: Honestly, I'd rather err on the side of more test coverage in the CI than less. If we start seeing flakes, let's reevaluate then.

bindings/wasm/lib/worker.ts
elalish: Yeah, that's fine. But even better would be to describe the difference, which I'm honestly still a touch unsure of. The worker is doing more, right? So what is the "more"?

bindings/wasm/lib/manifoldCAD.ts
elalish: I must still be missing something. Somewhere in manifoldCAD.ts it's calling `evaluator.evaluate()`, right? And that calls `getFullContext()` and sets up garbage collection, right? So why is it also happening at the top of this file? Or am I misunderstanding the architecture? Honestly, a little flow chart of some kind showing some of the building blocks and how they're used for the node, command line, and manifoldCAD.org could be really nice documentation.

bindings/wasm/lib/bundler.ts
elalish: Yeah, I suppose this is fine. I was concerned about the `/+esm` syntax changing under us, but I suppose they probably have to support that now for as long as their site exists, or they'll break everyone anyway.

bindings/wasm/lib/manifoldCAD.ts
tonious: Documentation is definitely on my list.

I think the meaning of the evaluator has shifted... Evaluation is down to about ten lines of actual code, or even less.  I wonder if it should _become_ `manifoldCAD.ts`, and move the actual eval function to the worker?  It's also the one object oriented part of the overall refactor, and I think that's unnecessarily complicating things.

tonious:
As part of this latest refactor, I've changed how models get their data to manifoldCAD.  Instead of defining result, they must now `export default` something.  Now, that something could be a `Manifold` object.  But if it's a function, we'll call it.  It's like a manifoldCAD specific callback.

This gives us a way for libraries to give a result in manifoldCAD (web or CLI) while not burning those CPU cycles when imported. 

This only applies to top level models -- entry points directly run by the worker -- so we don't have to worry about messing up imports within a model.

pca006132:
Since there is quite a lot of work on the JS binding side and an important bug fix (#1383), we should probably release v3.3 soon.

Do you think this PR will be ready soon (in a few days)? Or should we just do a release first? @tonious @elalish

bindings/wasm/test/cdn.test.ts
```diff
@@ -18,31 +18,39 @@ import {beforeEach, expect, suite, test} from 'vitest';
...
 import * as worker from '../lib/worker.ts';
 
-let evaluator = worker.getEvaluator() || worker.initialize();
+import {getSceneVertexCount, VertexCountMethod} from '@gltf-transform/functions';
+import { Document } from '@gltf-transform/core';
+
+const countVertices = (doc:Document) => {
+  const scene = doc.getRoot().listScenes()[0];
+  if (!scene) return -1;
+  return getSceneVertexCount(scene, VertexCountMethod.UPLOAD_NAIVE);
```
elalish, line 26: 👍

bindings/wasm/test/examples/gyroid-module.mjs
```diff
@@ -0,0 +1,65 @@
+// Recreates Modular Gyroid Puzzle by Emmett Lalish:
+// https://www.thingiverse.com/thing:25477. This sample demonstrates the
+// use of a Signed Distance Function (SDF) to create smooth, complex
+// manifolds.
+import {vec3} from 'gl-matrix';
```
elalish, line 5: Now that you have imports working, can we make gl-matrix just an import, so manifold-cad doesn't need to know about it anymore? Would make a good demonstration of the work you've done.

bindings/wasm/test/examples/voronoi.mjs
```diff
@@ -0,0 +1,24 @@
+import {DVMesh} from '@thi.ng/geom-voronoi';
+import PoissonDiskSampling from 'poisson-disk-sampling';
```
elalish, line 9: Good example, but could use some comments explaining where these are imported from and what importing options you have.

bindings/wasm/test/fixtures/importFromCDN.mjs
```diff
@@ -0,0 +1,23 @@
+import {DVMesh} from '@thi.ng/geom-voronoi';
+import PoissonDiskSampling from 'poisson-disk-sampling';
```
elalish: Maybe not necessary anymore since it's tested in the main examples now?

bindings/wasm/test/examples/gyroid-module.mjs
tonious: This _is_ an import 😂 
But I'll go fix `euler2quat()` to make it a little more explicit.

tonious:
I've got some a little work to do.  But not a whole lot -- this is getting close.

tonious:
I'm pretty sure this just leaves `euler2quat`.

bindings/wasm/test/cdn.test.ts
```diff
@@ -28,7 +28,7 @@ const countVertices = (doc: Document) => {
 
 beforeEach(() => worker.cleanup());
 
-suite('Import remote modules from', () => {
+suite.skip('Import remote modules from', () => {
```
elalish: Are these tests actually failing on the CI? I'm happy to give them a try and wait to see if things get flaky.

bindings/wasm/test/fixtures/unitSphereGLTF.mjs
```diff
@@ -0,0 +1,10 @@
+import {GLTFNode, Manifold} from '../../lib/manifoldCAD';
+
+// Models that build GLTF objects need not export their results.  The
+// scene builder will track GLTF nodes as they are created.  Essentially,
+// it's possible to build a manifoldCAD model as a side-effect of an import.
```
elalish: Hmm, thanks for noting this. Is this desirable behavior? If I make a library, I expect to e.g. `export function gear(params...)` and `const result gear(example params)` to show an example. On import, it shouldn't generate any geometry, just make the function available, which I believe is how your PR works now, right? If I instead switch the example result to a `new GLTFNode().manifold = gear(example params)`, I would expect it to work the same way, not pollute my node structure in the place I'm importing it with an example I have no way to remove. The node should only appear if I run the library directly as the top-level file. Is that doable?

bindings/wasm/examples/editor.js
```diff
@@ -522,9 +530,11 @@ async function run() {
   enableCancel();
   clearConsole();
   console.log('Running...');
-  const output = await tsWorker.getEmitOutput(editor.getModel().uri.toString());
+  // FIXME pass other files in for import?
```
elalish, line 535: What files?

elalish (changes requested):
I just want to make sure we have the library-loading behavior working as desired before we merge this. Thanks for all the tests!

bindings/wasm/test/fixtures/unitSphereGLTF.mjs
tonious: I'm just as inclined to entirely remove the side-effect.  The scene builder will not automatically collect `GLTFNodes` at creation time.  It's up to the user to give them to us.

Maybe a pseudo-type something along the lines of:

```typescript
type modelResult = Manifold | GLTFNode | Array<GLTFNode|Manifold>;
type modelExportDefault = modelResult | () => modelResult | () => Promise<modelResult>;
```

The end user would have to explicitly export either geometry or a function that returns geometry to see an example.  Libraries could export as many additional functions or properties as they'd like, but only a top level default export would be rendered in manifoldCAD.

tonious:
> I just want to make sure we have the library-loading behavior working as desired before we merge this. Thanks for all the tests!

Thanks for your patience with this!  I was honestly not expecting to refactor the world, but here we are.  This is a lot of ground to cover and I'm much happier knowing that the changes are well understood and reasonably tested.

bindings/wasm/test/fixtures/unitSphereGLTF.mjs
hrgdavor: I would vote to remove the side-effects. That would be my choice from the start, but I did not want to nag you. Since all I do lately is only comment, and do not want to bother you, seeing how much work you are doing.

bindings/wasm/examples/editor.js
tonious: Oh, just the other models available in the editor.  I've just added a library example.

bindings/wasm/test/cdn.test.ts
tonious: I was having issues with skypack.  I'll skip that one in particular.

tonious:
Screenshot 2025-10-26 at 3 57 49 pm

bindings/wasm/test/fixtures/unitSphereGLTF.mjs
tonious: At least on my stuff, feel free to comment.  More brain cells are good.

elalish: I like the sound of this plan quite a bit. From the tense I'm having a little trouble determining how much of this is how it currently works in this PR and how much is a change to still be made?

tonious: It wasn't done at time of comment -- I wanted feedback first.   But it's done now :joy:

Between this and changing from `result` to `export default`, evaluating a model is suddenly much simpler.

bindings/wasm/test/examples/tetrahedron-puzzle.mjs
```diff
@@ -70,3 +70,6 @@ moving.rotation = (t) => [0, 0, 270 * 2 * (1 - t)];
 
 globalDefaults.animationLength = 10;  // seconds
 globalDefaults.animationMode = 'ping-pong';
+
+export const nodes = [layFlat, fixed, layFlat2, moving];
```
elalish: Not a critique, but for my own understanding - if this were just `const nodes = [fixed, moving]`, would it still work? Since `fixed` references `layFlat` as its parent? Or could you do `const nodes = [layFlat, layFlat2]` since they keep track of their children? Or do you need to specifically export every node in the tree?

bindings/wasm/test/examples/rounded-frame.mjs
```diff
@@ -34,4 +34,6 @@ outsideNode.manifold = outside;
 
 const insideNode = new GLTFNode();
 insideNode.manifold = inside;
-insideNode.material = {baseColorFactor: [0, 1, 1]};
\ No newline at end of file
+insideNode.material = {baseColorFactor: [0, 1, 1]};
+export const nodes = [outsideNode, insideNode];
+export default nodes;
```
elalish, line 41: Is `export const nodes` important? What would be the difference if these two lines were condensed to `export default nodes = [outsideNode, insideNode];`?

bindings/wasm/test/examples/intro.mjs
```diff
@@ -0,0 +1,32 @@
...
+// "CrossSection." or "Manifold." to list the 2D and 3D constructors,
+// respectively. Type "module." to see the static API - these functions
+// can also be used bare. Use console.log() to print output (lower-right).
+// This editor defines Z as up and units of mm.
+const {cube, sphere} = Manifold;
+const box = cube([100, 100, 100], true);
+const ball = sphere(60, 100);
+
+// You must export your model as default, or create at least one
+// GLTFNode - see Menger Sponge and Gyroid Module examples.
```
elalish: I believe this text needs an update?

elalish (approved):
This looks much better! The only thing I feel a little bad about is this is a breaking change for any ManifoldCAD.org users out there, since their scripts won't compile anymore. It's a quick and easy fix, but that may not be obvious. Then again, I don't think we actually have very many users on that website...

What does the error look like they'll receive if, say, they opened one of our current example scripts, without the updates you've made to it here?

bindings/wasm/test/examples/rounded-frame.mjs
tonious: `export default nodes = [...` won't work.  But `export const nodes` can be reduced to just `const nodes`, given that `export default nodes;` will mark it as an export anyhow.  I'll go do that.

tonious:
I don't feel great about it being a breaking change either.  But I do think it's now a much better interface, and it'll be easier to grow from here.

> What does the error look like they'll receive if, say, they opened one of our current example scripts, without the updates you've made to it here?

```
Error: No output as no model was exported.  Try 'export default result'?
```

Or... at least it should.  I was doing my checks out of order, so until my next push it says:

```
Error: Cannot convert model to GLTFNode!
```

bindings/wasm/test/examples/tetrahedron-puzzle.mjs
tonious: I didn't think of that to be honest.  It was failing without the parent nodes, but I can just recurse up the tree.

elalish: We're getting there, but I wonder if we can make this any more automatic? Consider a design that's a big animated train of gears - before, just creating the nodes was enough, but now you also need to track all the leaf nodes and export them or it will break. Is it possible to make the nodes work a bit more like they did before, where that node list is stored internally and doesn't need to be exported explicitly? But in the case of an import, that node list is ignored so as to avoid side effects?

tonious: I've pushed up something that meets in the middle.  For top level scripts, it tracks creation of GLTF nodes.  Then, the user can call `getGLTFNodes()` to get that list and export it.  That function will explicitly return an empty array when called by library, and any nodes created by libraries will not add to the list.

This way there is no geometry creation by side-effect at any level, yet the user doesn't have to do any accounting.

bindings/wasm/test/examples/gyroid-module.mjs
```diff
@@ -55,10 +54,15 @@ if (m > 1) {
...
       }
     }
   }
 }
 
+// Get a list of GLTF nodes that have been created in this model.  This
+// function only works at the top level; in a library it will always return
+// an empty array, and nodes created in libraries will not be included in
+// the result. This is intentional; libraries must not create geometry as
+// a side effect.
```
elalish, line 66: 👍

tonious:
> Okay, I like that solution - simple, but explicit. Shall we go ahead and merge this? Ideally we should all take a little time to play with ManifoldCAD.org once this is up to make sure things are behaving as expected. When should we cut a release (including npm)?

That makes sense.  I've been doing a lot of testing with my local copy, but that's not the same.  It also provides some time to address documentation and any small things that I might have missed.

I am vaguely worried that I've done _something_ that will break when it's time to build the site.  So, sorry in advance :sob:

> Perhaps we should check that your cookiecad update is still working?

I updated it about two thirds of the way through this, though I haven't pushed my changes yet.  I'll make sure it's still doing the thing.

> Any other breaking API changes we might need?

I can't think of any.  I was trying pretty hard to not even break this much.  From here it's pretty much enhancements.  My mental to-do list looks like:

- [Automatic type loading in manifoldCAD.org](https://github.com/elalish/manifold/pull/1387#issuecomment-3408555871) as per @Loosetooth
- Minor exporter refactoring:
  - They should not be object oriented any more.
  - They should be able to write files (in line with the way `gltf-transform` does it).
- Model imports
- More test coverage.
  - Test bench for web workers?  There's a vitest plugin.
- Export `CrossSection` objects to GLTF.

elalish:
Awesome, let's see how it goes! I'm really looking forward to model imports.

Issue #1395, opened 2025-10-28
ManifoldCAD.org updates


elalish:
@tonious ManifoldCAD.org is looking pretty good with your latest changes, but I figured it would be good for us to have a place to note and discuss any regressions/problems we find.

So far I've noticed these:
1. We used to have e.g. `const { vec3 } = glMatrix;`, which is now changed to `import { vec3 } from 'gl-matrix';`. However, for anyone with existing scripts, all you get as explanation is `ReferenceError: glMatrix is not defined`. Is there a way to give any more info?
1. For those without `export default` we get a better error: `Error: No output as no model was exported.  Try 'export default result'?`, but it might be nice to also mention `getGLTFNodes()`, especially if the file has nodes.
1. Involute Gear Library includes the line `import {CrossSection} from "manifold-3d/manifoldCAD";`, but none of the other examples do. Can this be removed, or is it doing something important? It also seems to kill our TS declarations.
1. I have a [script](https://tinyurl.com/yckfrh5h) that's giving an odd error now: `BundlerError: Expected ")" but found ":"`. Maybe has to do with using `Mesh`?

Has anyone else noticed anything?

tonious:
I get a wasm error on iOS.  Granted, I can't remember loading the site on my phone before, but I'm gonna assume I broke it.

tonious:
#1396 can get merged in at any time.  It should fix points 1, 2, and 4.

Point 3 is trickier.  The import needs to stay in order to for it to work as a library.  I think there's something funky happening with how monaco is handling declarations.

elalish:
Thanks! I almost wonder if *all* the scripts should have that import line, or maybe `import {Manifold} from "manifold-3d";` - it would be more explicit and consistent. Only worth it if it makes it easier to solve the Monaco thing though.

hrgdavor:
Looks like you are inching to making it a normal js file with imports and no hidden funky stuff :)

tonious:
@elalish said:

> I almost wonder if all the scripts should have that import line, or maybe import {Manifold} from "manifold-3d"; - it would be more explicit and consistent.

I'm not against that.  It wouldn't make a difference for manifoldCAD.org, but it would remove the triple-slash-comment requirement for the vscode extension.  I think.

@hrgdavor said:

> Looks like you are inching to making it a normal js file with imports and no hidden funky stuff :)

[...]

PR #1396, opened 2025-10-28
Add specific error messages in wasm bindings.

tonious:
As per #1395, add custom messages.
 
When glMatrix is mentioned in a ReferenceError: 

> ReferenceError: ManifoldCAD no longer includes gl-matrix directly.  Import it by adding `import * as glMatrix from 'glMatrix';` to the top of your model.

When GLTF nodes are created at the top level, yet no geometry is exported to manifoldCAD:

> Error: GLTF Nodes were created, but not exported.  Add `const nodes = getGLTFNodes();` and `export default nodes;` to the end of your model.

### DIFF
--- a/bindings/wasm/test/fixtures/typeScriptSphere.ts
+++ b/bindings/wasm/test/fixtures/typeScriptSphere.ts
@@ -1,8 +1,8 @@
 import {Manifold} from '../../lib/manifoldCAD';
 
-const makeSphere: ((radius?:number) => any) = (radius?:number) => {
-  return Manifold.sphere(radius ?? 1.0)
-}
+const makeSphere: ((radius?: number) => any) = (radius?: number) => {
+  return Manifold.sphere(radius ?? 1.0);
+};
 
 const result = makeSphere();
 export default result;
--- a/scripts/format.sh
+++ b/scripts/format.sh
@@ -15,7 +15,7 @@ $CLANG_FORMAT -i bindings/wasm/lib/*.ts &
 $CLANG_FORMAT -i bindings/wasm/examples/*.{js,ts,html} &
 $CLANG_FORMAT -i bindings/wasm/examples/public/*.{js,ts} &
 $CLANG_FORMAT -i bindings/wasm/test/*.ts &
-$CLANG_FORMAT -i bindings/wasm/test/fixtures/*.mjs &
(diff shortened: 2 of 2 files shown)


tonious:
This was a good catch.  There was a bug where importing `manifoldCAD` at the top level would result in the version that libraries get.  In effect, including that import disabled GLTF node tracking.
Fixed now, with test cases to cover.

bindings/wasm/lib/worker.ts
```diff
@@ -242,7 +243,17 @@ export async function evaluate(
...
     // https://github.com/nodejs/node/issues/43047#issuecomment-1564068099
     const stacktrace = getSourceMappedStackTrace(bundled, error, -2);
-    const newError = new RuntimeError(error);
+    let newError: RuntimeError|null = null;
+
+    if (error.message.match(/glMatrix/)) {
+      newError = new RuntimeError(
+          error,
+          'ManifoldCAD no longer includes gl-matrix directly.  ' +
+              'Import it by adding `import * as glMatrix from \'glMatrix\';` ' +
```
elalish: I think it's `from 'gl-matrix'`, isn't it?

bindings/wasm/lib/worker.ts
```diff
@@ -251,9 +262,15 @@ export async function evaluate(
...
+  if (!result || (Array.isArray(result) && !result.length)) {
+    if (scenebuilder.getGLTFNodes().length) {
+      throw new Error(
+          'GLTF Nodes were created, but not exported.  ' +
+          'Add `const nodes = getGLTFNodes();` and `export default nodes;` ' +
+          'to the end of your model.');
+    }
     throw new Error(
-        'No output as no model was exported.  Try \'export default result\'?');
+        'No output as no model was exported.  Try `export default result`?');
```
elalish: What's the ? for?

elalish (changes requested):
Thanks for the quick fix. Might be worth adding a test around that `Mesh` constructor, or whatever it was that was failing.

bindings/wasm/lib/worker.ts
tonious: Oh poop. That's correct.

bindings/wasm/lib/worker.ts
tonious: It's a Canadianism.  Phrasing a statement as a question makes it less assertive, eh?

I'll change it.

PR #1397, opened 2025-10-29
Automagically bundle manifoldCAD.d.ts

tonious:
This MR uses [api-extractor](https://api-extractor.com/) to bundle `.d.ts` files for downstream use.  The generated declarations no longer have any imports; all of the upstream declarations are rolled into the same file.  The upshot is that it's easier to roll `manifold-global-types.d.ts`, `manifold-encapsulated-types.d.ts` and `manifoldCAD.d.ts` into a single file for use within an editor.

Today I learned that associating a set of declarations with an import can be done with the `module` statement, and has no relationship to any path:

```typescript
  const manifoldCadDTS = await getManifoldCadDTS();
  monaco.languages.typescript.typescriptDefaults.addExtraLib(manifoldCadDTS); // Global namespace.
  const wrapped = `declare module 'manifold-3d/manifoldCAD' {\n${manifoldCadDTS}\n};`
  monaco.languages.typescript.typescriptDefaults.addExtraLib(wrapped); // manifold-3d/manifoldCAD namespace
```

This MR is related to #1395.

Marking this as draft: It works, but still requires some meditation.

### DIFF
--- a/bindings/wasm/examples/editor.js
+++ b/bindings/wasm/examples/editor.js
@@ -263,36 +263,10 @@ function initializeRun() {
 // Editor ------------------------------------------------------------
 let tsWorker = undefined;
 
-async function getManifoldDTS() {
-  const global = await fetch('/manifold-global-types.d.ts')
-                     .then(response => response.text());
-
-  const encapsulated = await fetch('/manifold-encapsulated-types.d.ts')
-                           .then(response => response.text());
-
-  return `
-${global.replaceAll('export', '')}
-${encapsulated.replace(/^import.*$/gm, '').replaceAll('export', 'declare')}
-declare interface ManifoldToplevel {
-  CrossSection: typeof T.CrossSection;
-  Manifold: typeof T.Manifold;
-  Mesh: typeof T.Mesh;
-  triangulate: typeof T.triangulate;
-  setMinCircularAngle: typeof T.setMinCircularAngle;
-  setMinCircularEdgeLength: typeof T.setMinCircularEdgeLength;
-  setCircularSegments: typeof T.setCircularSegments;
-  getCircularSegments: typeof T.getCircularSegments;
-  resetToCircularDefaults: typeof T.resetToCircularDefaults;
-  setup: () => void;
-}
-declare const module: ManifoldToplevel;
-`;
-}
-
 async function getManifoldCadDTS() {
   const global =
       await fetch('/manifoldCAD.d.ts').then(response => response.text());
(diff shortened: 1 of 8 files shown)


tonious:
I was briefly thinking about how to get intellisense to work across exports within manifoldCAD.org, but that's too much scope creep for now.

bindings/wasm/examples/editor.js
```diff
@@ -263,47 +263,23 @@ function initializeRun() {
...
 });
 require(['vs/editor/editor.main'], async function() {
-  monaco.languages.typescript.typescriptDefaults.addExtraLib(
-      await getManifoldDTS());
-  monaco.languages.typescript.typescriptDefaults.addExtraLib(
-      await getManifoldCadDTS());
+  const manifoldCadDTS = await getManifoldCadDTS();
+  monaco.languages.typescript.typescriptDefaults.addExtraLib(manifoldCadDTS);
+  const wrapped =
+      `declare module 'manifold-3d/manifoldCAD' {\n${manifoldCadDTS}\n};`
```
elalish, line 280: Ah, very nice. So am I correct that if we required all manifold scripts to include `import * from 'manifold-3d/manifoldCAD';` at the top, then we could remove the first `addExtraLib` command? That would also give an easy way to check if a random piece of TS/JS is a manifoldCAD script - simply look for that import. What do you see as the pros and cons?

tonious: That's absolutely correct.

Pros:

- As you said, we can tell if it's a manifoldCAD script prior to execution.
- Our monaco setup becomes slightly simpler.
  - The vscode extension as well.
- No implied imports.  Models will behave _almost_[^1] the same inside or outside of ManifoldCAD.

Cons:

- One more line of boilerplate in every model.

[^1]: Importing `manifold-3d/manifoldCAD` at the top level through the worker (or bundler) will result in slightly different behaviour than importing it at a deeper level, or outside of the worker: `globalDefaults` will not be available, `getGLTFNodes()` will return an empty array, and `isManifoldCAD()` may not be true.

tonious: I'm not sure that's even really a con; it means we're forcin.... er... encouraging our users to be explicit about their imports.

elalish: Okay, well it sounds like it's time to go ahead and make the change then. I believe this may be the last breaking change we're considering for the npm API? Let's get that in so we can spend a few days testing and then cut a release, eh?

PR #1398, opened 2025-10-30
Add missing declarations.

tonious:
As per #1395.

@pca006132 pointed out that a few manifold functions were not recognized.


pca006132:
Is this exhaustive? And if we do https://github.com/elalish/manifold/pull/1397#discussion_r2478220631, do we still need this change?

tonious:
As far as I know, this is now exhaustive.

This change is still required even with #1397 -- we won't dynamically load our own types.

PR #1399, opened 2025-10-30
Fetch remote types.

tonious:
Use [monaco-editor-auto-typings](https://github.com/lukasbach/monaco-editor-auto-typings) to fetch type definitions on manifoldCAD.org.

On the way by, updated monaco to the latest npm version and packaged it as well.

Relates to #1395.

### DIFF
--- a/bindings/wasm/examples/editor.js
+++ b/bindings/wasm/examples/editor.js
@@ -104,6 +104,27 @@ function nthKey(n) {
   }
 }
 
+function getAllScripts() {
+  const files = {};
+  for (const [name, contents] of exampleFunctions) {
+    files[name] = contents;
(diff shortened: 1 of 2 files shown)


pca006132:
Is it possible to load other local files as declarations as well? So autocomplete and error checking can wodk for the multifile examples.

tonious:
> Is it possible to load other local files as declarations as well?

Theoretically, yes.  I haven't got that far.

One thing I've noticed already is that `gl-matrix` has the the source path to their types published -- not the path in the final npm package.  I wonder if there's some kind of way to alias types.

elalish (approved):
Cool, I look forward to trying this out.

PR #1401, opened 2025-10-31
Require explicit imports in/of manifoldCAD

tonious:
One last part of #1395, as per discussion over in #1397.

* Require models to explicitly import `manifold-3d/manifoldCAD`, rather than implicitly including it.
* Added a custom message if a member of `manifoldCAD` (e.g.: `Manifold` or `setMorphStart`) is mentioned in a reference error.
* Update tests and examples to match.
* Broke out `globalDefaults` into a separate dts as it still needs to be global for top level scripts.  All top-level vs top-level-import vs library-import logic is now in `lib/worker.ts`.
* Tweaked exports in `package.json`, so hopefully manifoldCAD types will be discoverable.

### DIFF
--- a/bindings/wasm/types/globalDefaults.d.ts
+++ b/bindings/wasm/types/globalDefaults.d.ts
@@ -1,0 +1,35 @@
+// Copyright 2025 The Manifold Authors.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+/**
+ * The globalDefaults object is shared between a manifoldCAD script and the
+ * scene builder.
+ *
+ * It can be used to set parameters elsewhere in ManifoldCAD.  For example, the
+ * GLTF exporter will look for animation type and framerate.
+ *
+ * It is only accessable as a global object, by a top level script evaluated by
+ * the worker.  Libraries will not have access to it.
+ *
+ * See 'Tetrahedron Puzzle' for an example.
+ */
+export declare const globalDefaults: {
+  roughness: number,
+  metallic: number,
+  baseColorFactor: [number, number, number],
+  alpha: number,
+  unlit: boolean,
+  animationLength: number,
+  animationMode: 'loop'|'ping-pong';
+}
--- a/bindings/wasm/esbuild.mjs
+++ b/bindings/wasm/esbuild.mjs
@@ -7,7 +7,7 @@ await esbuild.build({
   format: 'esm',
   target: 'esnext',
   sourcemap: 'inline',
-  sourcesContent: true,
+  sourcesContent: false,
   external: [
     'node:path', 'node:url', 'module',
   ]
--- a/bindings/wasm/examples/editor.js
+++ b/bindings/wasm/examples/editor.js
@@ -293,10 +293,10 @@ function initializeRun() {
 
 // Editor ------------------------------------------------------------
 
-async function getManifoldCadDTS() {
-  const global =
-      await fetch('/manifoldCAD.d.ts').then(response => response.text());
-  return global.replace(/^export \{ }.*$/gm, '').replace(/^export /gm, '');
+async function getDTS(url) {
+  const response = await fetch(url);
+  const text = await response.text();
+  return text.replace(/^export \{ }.*$/gm, '').replace(/^export /gm, '');
 }
 
 async function createEditor() {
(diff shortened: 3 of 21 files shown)


bindings/wasm/types/globalDefaults.d.ts
```diff
@@ -0,0 +1,35 @@
...
+
+/**
+ * The globalDefaults object is shared between a manifoldCAD script and the
+ * scene builder.
+ *
+ * It can be used to set parameters elsewhere in ManifoldCAD.  For example, the
+ * GLTF exporter will look for animation type and framerate.
+ *
+ * It is only accessable as a global object, by a top level script evaluated by
+ * the worker.  Libraries will not have access to it.
```
elalish, line 23: I'm trying to understand what globalDefaults really is. The first 5 are material properties assigned to the invisible "root" scene node in GLTF, but are overridden by any child nodes that redefine them. The last two are for animation and really do apply globally and cannot be reset anywhere (actually they could in a glTF, but we've kept things intentionally simpler here). 

Not included are functions like `setMinCircularAngle` etc, which also control global default state, but for Manifold instead of glTF. Those are global, but can be changed during execution to affect different objects differently. Would changes to those in a library have no effect when imported? Or could they create confusing state changes? 

These aren't problems with this PR, but rather a chance to check on old design decisions of mine that weren't made with import in mind.

elalish (approved):
Looks great - I'm going to push this in so we can kick the tires on it more easily. We might want to consider more specific naming than globalDefaults, at least once we figure out what they really represent.

bindings/wasm/types/globalDefaults.d.ts
tonious: I feel like the default material stuff can just be excluded.  Maybe provide a setter on `material.ts` or the scene builder, but I don't think manifoldCAD scripts need that access.

As for animation, I think moving to something like `setAnimationLength()`, `getAnimationLength()`, `setAnimationMode()`, `getAnimationMode()` makes sense.  That way the state in `animation.ts` doesn't depend on passing an object around.  Even though the affects are global, my gut feeling is that libraries would need to be able to get at these functions.  My mental test case is a script that builds an multipart model calling a library that animates the model assembling itself.

The LoD functions like `setMinCircularAngle()`... Well, libraries really do need to be able to access all of the getters at least.  I can also picture cases where libraries may need to change it.

Given that we now have a distinction between manifoldCAD imported at the top level and manifoldCAD imported as a library, we can emit a warning when a library calls a setter with global effects?  This could be done by wrapping a list of functions (i.e.: like the garbage collector) in the worker, but that leaves node users on their own.  Which is a perfectly valid choice, and probably what I would do.

Another approach might be to define a `globalSettingsChanged` event/callback, and modify the each setter implementation accordingly.  Either way, I'm not quite sure how to log _where_ the setter was called, just _that_ it was called.

elalish: I think in an ideal world libraries can't return or affect nodes at all. Nodes are purely a top-level construct, meaning a library can only use them to create examples when it is being run at the top level itself. 

The LoD functions set global state (for better or worse - that API idea came from openscad). However, ManifoldCAD effectively scopes them by reseting their values each time a new script is run (otherwise the settings would persist for the whole session). I think ideally imports would work similarly, having their LoD values effectively scoped independently of the encompassing script. However, I'm not sure how doable that is. Perhaps as a minimum we could undo any LoD changes whenever a function returns? And somehow disallow or ignore a library setting the LoD values in its own top-level?

What do you think is a reasonable approach here?

tonious: I'm not sure how to scope LoD like that.  JavaScript generally works on the basis that if you can reference it, it's in scope.  A callee doesn't really have a way of identifying a caller, at least not in current JS.  I don't think it's enforceable on the called side.

It's much easier to do it by convention, I guess.  A library can wrap their own functions with a decorator, or call set up or tear down operations.  We could do that through the bundler, but it would break for node users.

Thinking out loud here:

We can remove the `setMinimum*` calls from `manifoldCAD.ts`.  We can put them back in the top level namespace as we do for `GlobalDefaults`.  I like this approach because because it's very clear that libraries can't adjust global LoD.  They can still ignore it, though :grin:

We could do as we do with `getGLTFNodes()`: We ship dummy functions in `manifoldCAD.ts`, and replace them with real functions when bundling.  This is a negative default -- node users won't be able to adjust LoD unless they import the real functions from the scene builder.

Or, we could go the other way, and replace the accessors with dummy functions for libraries when bundling.  This is pretty much how `isManifoldCAD()` works -- by default it's a function that returns false, and we replace it with another function that returns true.  This would be a positive default -- we're packaging the real functions in `manifoldCAD.ts`, and replacing them at bundle time.  The downside to this approach is again with node users -- libraries will be able to affect globals as we won't have control of the bundler.

Of those, I like the first option.  It's the most explicit, and users will run into reference errors when they first try to package a library.  The other two introduce behaviour differences between node and manifoldCAD that may fly under the radar.

elalish: Yes, I like your first option as well. Libraries really shouldn't be setting defaults at all anyway - better they be explicit in constructors if the number of facets is actually important. Sounds like this is one more breaking change to hold release for. Maybe even the last one...

PR #1403, opened 2025-11-02
Refactor stack trace handling.

tonious:
I wasn't entirely happy with how stack traces worked.  As it turns out, stack traces through function constructors haven't really been explored in depth by the JavaScript community.  JavaScriptCore in particular just ignores the issue and doesn't provide line or column numbers until the trace leaves the constructed function.

Here's some programmer therapy.

* Gold master data collected from three JavaScript interpreters.
* Refactored with specific parsers for V8 and SpiderMonkey.
* Unit tests.

And now my conscience is clear.

### DIFF
--- a/bindings/wasm/lib/error.ts
+++ b/bindings/wasm/lib/error.ts
@@ -17,11 +17,18 @@ import type {BuildFailure, Location, Message} from 'esbuild';
 export class BundlerError extends Error {
   location?: Location;
   error: Message;
+  manifoldStack?: string;
 
   constructor(failure: BuildFailure, options?: ErrorOptions) {
     super(undefined, options);
     this.cause = failure;
     this.error = failure.errors[0];
+
+    if (this.error.location) {
+      const {file, line, column} = this.error.location!;
+      this.manifoldStack =
+          `${this.toString()}\n    at ${file}:${line}:${column}`;
+    }
   }
 
   get name(): string {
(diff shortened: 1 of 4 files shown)


elalish (approved):
Nice, thanks!

hrgdavor:
not sure what other issues you had, but I was able to get correct line numbers without parsing stack traces, with direct eval https://github.com/hrgdavor/jscadui/blob/c129f593b45072ced6441e42da7153864e072d24/packages/require/src/require.js#L18-L23

```js
// initially new Function was used to pass parameters: require, exports, module
// new Functions screws with sourcemaps as it adds a prefix to the source
// we need eval to do the same without prefix
// https://esbuild.github.io/content-types/#direct-eval
// to be nice to bundlers we need indirect eval
export const runModule = globalThis.eval('(require, exports, module, source)=>eval(source)')
```

unlike me you wrote test, and looks like you also looked into it bit deeper. anyway, I am sharing this, it may be useful, or I can find out more by looking into your solution.

tonious:
Oh, that's awesome.  I hadn't realized indirect eval was even a thing.

That said, using function constructors makes it easy to support async, and that got me going at the start of this crazy refactor.  So call it sentimental value? :sweat_smile:

hrgdavor:
not sure this eval would have issue with async ... maybe I am missing something ?

I am not expecting you to refactor again, as my code does not even have tests, so who knows what issues may be hiding. 

Anyway, just took the opportunity to share this eval trick, for small chance sbdy else finds it useful.

Issue #1337, opened 2025-08-15
Python packaging metadata with CMake


bkmgit:
When Python bindings are built and installed using CMake a metadata directory such as manifold3d-3.2.1.dist-info is missing, but this is installed when creating and installing a Python wheel.  It would be helpful to have the Python metadata directory also built when using CMake with the option -DMANIFOLD_PYBIND=ON

elalish:
If you know how, a PR would be most welcome.

pca006132:
I'm wondering what other libraries also generate that directory when installed via CMake. It feels that maybe the best way to do it is to *not* install the Python binding with CMake, but ask the user to install it via `build`.

pca006132:
Can probably copy https://github.com/elliotcmorris/OpenAssetIO-MediaCreation/commit/1a13d84cb6b41c2eeeaf3d07b17bd4f30c46ac41

bkmgit:
Yes, though the specifications change:
https://packaging.python.org/en/latest/specifications/core-metadata/#core-metadata
so maintenance maybe burdensome.  The approach at:
https://github.com/maximiliank/cmake_python_r_example
seems cleaner, but may want to have a different python backend.

bkmgit:
Another approach is to build a wheel from cmake:
- https://github.com/cmake-wheel/cmeel
- https://github.com/Klebert-Engineering/python-cmake-wheel

pca006132:
Maybe use setuptools to install instead? Then there is no need to ask python to trigger cmake in the build, we do separate cmake build first, use cmake install for the main shared lib, and use python build to install the python binding shared lib. https://python-packaging.readthedocs.io/en/latest/non-code-files.html

May need to specify flags, so builds on pypi can use the static linking variant.

bkmgit:
@pca006132 setuptools seems to work. Is the approach in #1432 reasonable?  At the moment it links to the static library, and there is no install step, but these can be updated.

PR #1407, opened 2025-11-02
Allow using an installed manifold shared library

### LINKED ISSUE #1406: Enable use of shared library from Python package
The current python package statically links the library.  It can be helpful to be able to use a shared library, especially in distribution packages which will likely have the C++ library also installed.

bkmgit:
closes #1406

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -55,6 +55,11 @@ option(
 )
 option(MANIFOLD_TEST "Enable testing suite" ON)
 option(MANIFOLD_PYBIND "Build python bindings" OFF)
+option(
+  MANIFOLD_PYBIND_SHARED_MANIFOLD
+  "Link Python bindings to shared Mainfold library"
+  OFF
+)
 # MANIFOLD_CBIND is only available when MANIFOLD_CROSS_SECTION is enabled
 cmake_dependent_option(
   MANIFOLD_CBIND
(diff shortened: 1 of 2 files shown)


pca006132:
Yes, but I haven't yet have time to check if this actually fixes the problem.

pca006132:
One issue here is that the binary from pypi will still provide its own manifold lib, AFAIK only nix will try to link against its own library?

bkmgit:
The Fedora package would also link against its own manifold lib - currently under review:
https://bugzilla.redhat.com/show_bug.cgi?id=2386079

pca006132:
Looking at it, it seems that the current build already links to shared `libmanifold.so`. What you need is a way to properly install the python binding like what pip does.

Whether the cmake library setting is public or private does not affect linking.

bkmgit:
Will check again.

At the moment, the Python package generated using CMake does not use a pyproject.toml file and does not generate a wheel. As mentioned in https://github.com/elalish/manifold/issues/1337 because the project primarily uses CMake to build, it would be better to allow for CMake to also drive generation of a Python wheel which could be published to PyPI or installed directly, rather than having a separate Python build to generate a wheel.  This should be done in another pull request.  For the wheel on PyPI, it is probably helpful to embed the static library so it is self-contained.  For GNU/Linux distribution packages, it is helpful for the Python package to link to a shared version of the system library.

pca006132:
Sounds good to me. I think we can probably modify the python build routine to make it reuse the existing binary.

pca006132:
Closing as this is not the correct solution.

PR #1408, opened 2025-11-04
Remove GlobalDefaults object.

tonious:
As per [discussion on #1401](https://github.com/elalish/manifold/pull/1401#pullrequestreview-3406949517).

On the way, this involved:

  * Added `lib/level-of-detail.ts`
  * Refactored `lib/manifoldCAD.ts` and `types/manifoldCAD.d.ts` to use re-exports.
  * Improved documentation.
    * Extracted GLTFNode into a module.
    * Added `@internal` typedoc tags to hide functions that aren't really public facing.
  * Updated examples to match.

Part of #1395

### DIFF
--- a/bindings/wasm/lib/level-of-detail.ts
+++ b/bindings/wasm/lib/level-of-detail.ts
@@ -12,7 +12,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-import {getManifoldModuleSync} from './wasm';
+import {getManifoldModuleSync} from './wasm.ts';
 
 /**
  * Control detail level for the
@@ -29,8 +29,8 @@ import {getManifoldModuleSync} from './wasm';
  * @packageDocumentation
  */
 
-let minCircularAngle: number = 10;
-let minCircularEdgeLength: number = 1;
+let minCircularAngle: number = 10.0;
+let minCircularEdgeLength: number = 1.0;
 
 /**
  * Set an angle constraint when calculating the number of segments in a circle.
@@ -44,7 +44,7 @@ let minCircularEdgeLength: number = 1;
  */
 export const setMinCircularAngle = (angle: number) => {
   minCircularAngle = angle;
-  getManifoldModuleSync().setMinCircularAngle(angle);
+  getManifoldModuleSync()?.setMinCircularAngle(angle);
 };
 
 /**
@@ -58,7 +58,7 @@ export const setMinCircularAngle = (angle: number) => {
  */
 export const setMinCircularEdgeLength = (length: number) => {
   minCircularEdgeLength = length;
-  getManifoldModuleSync().setMinCircularEdgeLength(length);
+  getManifoldModuleSync()?.setMinCircularEdgeLength(length);
 };
 
 /**
@@ -71,7 +71,7 @@ export const setMinCircularEdgeLength = (length: number) => {
  * @group Global Settings
  */
 export const setCircularSegments = (segments: number) =>
-    getManifoldModuleSync().setCircularSegments(segments);
+    getManifoldModuleSync()?.setCircularSegments(segments);
 
 /**
  * Reset the circular construction parameters to their defaults if
@@ -80,9 +80,9 @@ export const setCircularSegments = (segments: number) =>
  * @group Global Settings
  */
 export const resetToCircularDefaults = () => {
+  getManifoldModuleSync()?.resetToCircularDefaults();
   minCircularAngle = 10;
   minCircularEdgeLength = 1;
-  getManifoldModuleSync().resetToCircularDefaults();
 };
 
 /**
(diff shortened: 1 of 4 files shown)


bindings/wasm/lib/level-of-detail.ts
```diff
@@ -0,0 +1,120 @@
...
+ *
+ * Libraries should not change these values, and if run through manifoldCAD or
+ * the manifoldCAD CLI, will not be able to.  Libraries may get values to
+ * determine their own level of detail.
+ *
+ * @packageDocumentation
+ */
+
+let minCircularAngle: number = 10.0;
+let minCircularEdgeLength: number = 1.0;
```
elalish, line 33: Why store these again here? Is it just to enable e.g. `getMinCircularAngle`? Can you give me a use case for that getter?

bindings/wasm/test/examples/auger.mjs
```diff
@@ -3,7 +3,7 @@ const beadRadius = 2;
 const height = 40;
 const twist = 90;
 
-import {CrossSection, Manifold, setMinCircularEdgeLength} from 'manifold-3d/manifoldCAD';
```
elalish, line 6: Why don't we need to import this anymore? Seems to go against the idea of being explicit.

bindings/wasm/test/examples/rounded-frame.mjs
```diff
@@ -3,8 +3,7 @@
 // parameters.
 
 import {
-  Manifold, GLTFNode, getGLTFNodes,
-  setMinCircularAngle, setMinCircularEdgeLength
+  Manifold, GLTFNode, getGLTFNodes
 } from 'manifold-3d/manifoldCAD';
```
elalish, line 7: Would it be more clear to split these imports? Something like 
```
import {Manifold, CrossSection} from 'manifold-3d';
import {GLTFNode, setMinCircularAngle} from 'manifold-3d/manifoldCAD';
```
Then it would perhaps be more clear that the first type of imports will work for both libraries and top-level scripts, while the second type are strictly for top-level scripts and will be no-ops for imported libraries.

hrgdavor: that would be a strange runner specific behavior, if import behaves differently.

if aim is to allow creating libraries, but also allow the library to show something, it should be a specific exported function.

again I am coming from jscad where scripts that want to generate models export `main` function, which is nothing new from jscad, the concept is in C,Java and other languages. 

... I love how much effort is done here and pushing JS manifold forward

bindings/wasm/lib/level-of-detail.ts
tonious: I've been playing with `refineToLength` and a cylindrical warping function:

Screenshot 2025-11-06 at 10 44 57 am

If I know what `minCircularEdgeLength` is, I can refine to that length, and the resulting geometry has a scale-appropriate level of detail.

That said, I might have overthought this one.  Give me a thumbs down, and I'll tear it out.

bindings/wasm/test/examples/auger.mjs
tonious: Yes and no.  If called in a library, there's no way to get it, and it will always throw a reference error.  It's explicit in the sense that it fails fast when a developer packages their code into modules.  It's also documented as only being available to top level scripts.

If was packaged in a module like `manifold-3d/manifoldCAD/settings`.
We could determine if it's imported from the top level in the bundler... make it throw if it's in manifoldCAD, but not top level.  But that behaviour wouldn't hold as a generic ES module.  A library developer working in node could import the function, and wouldn't see it throw until they tried it through the CLI or website.

bindings/wasm/test/examples/rounded-frame.mjs
tonious: `GLTFNode` and `setMinCircularAngle` are slightly different beasts.`setMinCircularAngle` directly sets a global in the manifold kernel.  For top level scripts, we subclass `GLTFNode` so we can track created objects for the user.
Through the worker, we track nodes that are created at the top level as a helper function...

Huh.

Maybe 'TrackedGLTFNode' should be opt-in:

```typescript
import {GLTFNode} from 'manifold-3d/manifoldCAD/trackedGLTFNode';
// or 
import {TrackedGLTFNode as GLTFNode} from 'manifold-3d/manifoldCAD';
```

tonious: Anyhow, the issue is that a library could change a globally shared value, like `setMinCircularAngle` or the list of nodes underlying `TrackedGLTFNode`.  It doesn't really matter if this happens via an exported function or not, it will lead to unexpected results.

bindings/wasm/lib/level-of-detail.ts
elalish: That's reasonable, I suppose it serves as a getter for the default value. I don't have a strong opinion.

bindings/wasm/test/examples/rounded-frame.mjs
elalish: Thanks for the explanations, both of you. I'm satisfied with that.

elalish (approved):
Thanks!

PR #1432, opened 2025-11-17
WIP: Enable building Python wheel using setuptools

### LINKED ISSUE #1337: Python packaging metadata with CMake

bkmgit:
Closes #1337


bindings/python/setup.py
```diff
@@ -0,0 +1,21 @@
+# SPDX: unlicense
```
pca006132, line 1: is there any reason this should be `.in`? I don't see anything variable substitution here.

bindings/python/CMakeLists.txt
```diff
@@ -77,10 +77,54 @@ install(
...
+  configure_file(getwheelname.py.in
+        ${CMAKE_CURRENT_BINARY_DIR}/getwheelname.py @ONLY
+        )
+  configure_file(setup.py.in
+        ${CMAKE_CURRENT_BINARY_DIR}/setup.py @ONLY
+        )
+  execute_process(COMMAND ${CMAKE_COMMAND} -E make_directory "${CMAKE_CURRENT_BINARY_DIR}/manifold3d/libs"
+        WORKING_DIRECTORY "${CMAKE_CURRENT_BINARY_DIR}"
+        )
+  execute_process(COMMAND "${Python_EXECUTABLE}" getwheelname.py OUTPUT_VARIABLE PY_WHEELFILE
```
pca006132: is this for debugging only?

pyproject.toml
```diff
@@ -45,7 +45,7 @@ sdist.exclude = [
     "oneTBB", # we may have this when we build with cibuildwheel
 ]
 wheel.packages = ["manifold3d"]
-cmake.args = ["-DMANIFOLD_PYBIND=ON", "-DMANIFOLD_CBIND=OFF", "-DMANIFOLD_PAR=ON", "-DMANIFOLD_TEST=OFF", "-DBUILD_SHARED_LIBS=OFF"]
+cmake.args = ["-DMANIFOLD_PYBIND=ON", "-DMANIFOLD_WHEEL=OFF", "-DMANIFOLD_CBIND=OFF", "-DMANIFOLD_PAR=ON", "-DMANIFOLD_TEST=OFF", "-DBUILD_SHARED_LIBS=OFF"]
```
pca006132: Maybe we can get rid of scikit-build-core instead.

pca006132 (commented):
Looks good to me. I think if we can use `setuptools`, maybe we can just remove scikit-build-core from our build pipeline.

Please test this with cibuildwheels and also test if the sdist can be built after this. We have a CI action for this that I can invoke for you if you want https://github.com/elalish/manifold/blob/master/.github/workflows/build_wheels.yml.

bindings/python/CMakeLists.txt
bkmgit: Building the wheel assumes a particular folder structure, so the library and pyi file are copied later in the build.

Wheel name is needed later in the build.

bindings/python/setup.py
bkmgit: MANIFOLD_VERSION is substituted.

bindings/python/CMakeLists.txt
pca006132: but I don't see the variable `PY_WHEELFILE` used anywhere else, except the message after this command.

bindings/python/setup.py
pca006132: I don't see that in this file?

bkmgit: Yes you are correct. Should just be setup.py

pyproject.toml
bkmgit: Ok. Need to modify how upload wheels to PyPI and how generate a source distribution.

bindings/python/CMakeLists.txt
bkmgit: It is used in the install step

pyproject.toml
elalish: I'm a little surprised scikit-build doesn't do the things we need. What is so unique about our requirements?

bkmgit:
Nix build of the wheel  has the problem described at https://discourse.nixos.org/t/building-derivation-results-in-requirement-setuptools-40-8-0/52755

Official Nix package uses buildPythonPackage https://github.com/NixOS/nixpkgs/blob/nixos-unstable/pkgs/development/python-modules/manifold3d/default.nix

Examining https://github.com/NixOS/nixpkgs/blob/master/pkgs/development/interpreters/python/mk-python-derivation.nix

Approach used for Botan3 seems like it might work:
https://github.com/NixOS/nixpkgs/blob/master/pkgs/development/python-modules/botan3/default.nix

bkmgit:
> Looks good to me. I think if we can use `setuptools`, maybe we can just remove scikit-build-core from our build pipeline.
> 
> Please test this with cibuildwheels and also test if the sdist can be built after this. We have a CI action for this that I can invoke for you if you want https://github.com/elalish/manifold/blob/master/.github/workflows/build_wheels.yml.

For sdist, should the bindings only be made available or also the library sources? At present library sources are made available, but the binary packages also have tbb among other build dependencies which are not in the src uploaded to PyPI.

pyproject.toml
pca006132: It is not about our requirements, but distros want to share the base library and binding library. Pip does things in its own way, which is fine for us, but distros want to avoid that.

From the documentations, it is not clear to me how to make scikit-build reuse an existing build, so it is probably easier to just use setuptools.

pca006132:
> For sdist, should the bindings only be made available or also the library sources? At present library sources are made available, but the binary packages also have tbb among other build dependencies which are not in the src uploaded to PyPI.

Our source should be available. I am not sure about the other sources. I think ideally we should use the system provided libraries when available, so I don't include them. That said, it kinds of make the dependencies required when building from sdist, as the user cannot specify a certain build variant.

pca006132:
Feel free to skip the nix build for now. If you fixed the pip build (sdist and bdist), I can fix the nix build. Just let me know if you need any help.

elalish:
@bkmgit where are we with this?

pca006132:
I think this is not getting enough interest. Closing for now until someone revives this.

bkmgit:
It is still of interest, but have a number of packages with the same problem.  Can reopen once have made progress.

PR #1434, opened 2025-11-18
Import glTF models into manifoldCAD

tonious:
What's done so far:

  * Ported and refactored import code from `make-manifold`.
  * Extend `GLTFNode` tree to also include `VisualizationGLTFNode`.
  * Refactored exporters to be non-object oriented and extensible (`export-model.ts`).
  * Did the same for importers, even if it's just glTF right now (`import-model.ts`).
  * Merged `import-gltf.ts` into `gltf-io.ts`.
  * Fixed a bunch of documentation on the way by.
  * Added `importModel` and `importManifold` functions.
  * Added 'Import Manifold' example.
  * Made `import.meta.url` work inside manifoldCAD.

What needs to be done:

  * [x] Imported materials are duplicated on export.  Really obvious with textures.
  * [x] Only tested with models loaded over HTTP.
  * [x] We can serve up models for examples, but how do we expect users to handle theirs?
      * 'Out of scope' is my default answer for this one.  But it needs some thought.
  * [x] Import models from blobs/arraybuffers/whatever. 
  * Tests. There are no tests yet and it makes me nervous.
     * [x] 'Import Manifold' example runs as a test. 
     * [x] For that matter, our existing tests can be refactored to use the importer.
  * ~~[ ] A bunch of code in the scene builder may actually live better in `export-model.ts`.~~
  * [x] File types are assigned by extension.  We should also accept mime types.
  * ~~Test with `.gltf` files that have external assets like `.bin` or textures.~~
  * ~~[ ] Test with `.gltf` files that have url encoded buffers.~~
  * ~~[ ] Better warnings when exporting files with non-manifold components.~~
  * [x] Add merge tolerance parameter to `importManifold()`.

### DIFF
--- a/bindings/wasm/lib/worker.ts
+++ b/bindings/wasm/lib/worker.ts
@@ -238,7 +238,7 @@ export async function evaluate(
     _manifold_cad_library: manifoldImport,
 
     // Bundled code may be referencing files by relative paths.
-    // Set runtime value of import.meta.main
+    // Set runtime value of import.meta.url
     _manifold_runtime_url: options.baseUrl ?? null,
 
     // While this project is built using ES modules, and we assume models and


bindings/wasm/lib/import-gltf.ts
```diff
@@ -0,0 +1,34 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+import {WebIO} from '@gltf-transform/core';
+
+import {setupIO} from './gltf-io.ts';
```
elalish: This is such a small file - would it make sense to combine it into gltf-io.ts?

bindings/wasm/lib/import-model.ts
```diff
@@ -0,0 +1,231 @@
...
+  const converted: Array<Manifold> = [];
+  node.traverse(child => {
+    const manifold = gltfMeshToManifold(document, child);
+    if (manifold) converted.push(manifold);
+  });
+  if (!converted.length) {
+    throw new Error(`Model contains no manifold geometry.`);
+  }
+
+  return getManifoldModuleSync()!.Manifold.union(converted);
```
elalish: I thought about this for awhile - it's a major restriction, but I think appropriate for our use case. If people really want to be editing individual parts of an imported glTF assembly, they should probably be using gltf-transform's API directly instead of manifoldCAD.

bindings/wasm/lib/import-model.ts
```diff
@@ -0,0 +1,231 @@
...
+ * and then union the results together.  If a child node has no mesh, the mesh
+ * has no geometry, or the mesh is not manifold, that child node will be
+ * silently excluded.
+ *
+ * Other errors will be re-thrown for the caller to handle.
+ *
+ * @returns A valid Manifold object.
+ */
+export function gltfNodeToManifold(
+    document: GLTFTransform.Document, node: GLTFTransform.Node): Manifold {
```
elalish: Can we add a time argument? An animated glTF will yield very different manifolds at different times.

tonious:
Formerly #1418

bindings/wasm/lib/import-model.ts
```diff
@@ -0,0 +1,231 @@
...
+      attributes: attributes.filter(x => x !== 'POSITION'),
+      alpha: material.getAlpha(),
+      baseColorFactor: material.getBaseColorFactor() as any as Vec3,
+      metallic: material.getMetallicFactor(),
+      roughness: material.getRoughnessFactor(),
+      name: material.getName(),
+    });
+  }
+
+  const manifoldMesh = new Mesh(mesh);
```
elalish, line 477: Could we take everything before this and encapsulate it in another public function, `gltfNodeToMesh`? Then a user can run e.g. `mesh.merge()` with different tolerance to try to get something that `new Manifold(mesh)` will succeed with. They might also just take the mesh points and filter them or hull them instead of making a manifold directly. 

Is suppose you'll also need `gltfNodeToMeshes` returning an array of meshes based on traversing the children. Then `gltfNodeToManifold` can just be a high-level function for when the input is expected to be already manifold.

bindings/wasm/lib/gltf-node.ts
```diff
@@ -29,39 +43,104 @@ export class GLTFMaterial {
...
+   * The original imported model may consist of an entire tree of nodes, each of
+   * which may or may not be manifold.  This method will convert each child
+   * node, and then union the results together.  If a node has no mesh, the mesh
+   * has no geometry, or the mesh is not manifold, that node will be silently
+   * excluded.  Other errors will be re-thrown for the caller to handle.
+   *
+   * @returns A valid Manifold object.
+   */
+  makeManifold(): Manifold {
+    return gltfNodeToManifold(this.document, this.node);
```
elalish: Is this wrapper function necessary?

bindings/wasm/lib/import-model.ts
elalish: Do we need the separate node argument here? Can we simply start at the document root?

elalish (changes requested):
>We can serve up models for examples, but how do we expect users to handle theirs?
'Out of scope' is my default answer for this one. But it needs some thought.

I figure they can either spin up a web server, use someone else's, or use this through node / vscode extension where they have access to a local file system. 

>Test with .gltf files that have external assets like .bin or textures.

If this works easily, great, but don't spend a lot of time on it. 99% of glTF usage is GLBs because single file is so much easier to work with. If that's all we support, I doubt we'll get any complaints.

bindings/wasm/lib/gltf-node.ts
tonious: I think so.  It makes sense to have `gltfNodeToManifold()` as a general function, even if it's not directly accessible in manifoldCAD.  This isn't just for exporting -- This allows a user to import a model and then later convert it to a manifold.

I want users to be able to check if a model contains non-manifold elements, and maybe convert it depending on that.  That might entail adding a `strict` parameter to this call, or more likely a different method on this class.

bindings/wasm/lib/import-gltf.ts
tonious: I'm inclined to go the other way, and split `gltf-io.ts` across `import-gltf.ts` and `export-gltf.ts`.

bindings/wasm/lib/import-model.ts
tonious: Oh, that makes sense.

bindings/wasm/lib/import-model.ts
tonious: That's a good idea.

bindings/wasm/lib/import-model.ts
tonious: I was thinking of the case (not that I've seen it in person) where a glTF document contains an assembly with named nodes in it (or similar).  This would allow a manifold (not manifoldCAD) user to select a node.

I could make it optional?

bindings/wasm/lib/import-model.ts
tonious: It also makes the logic easier as we don't have to worry about order of operations.  Unions are commutative.[^1]

I think allowing non-manifold objects into manifoldCAD in the first place is a pretty major concession.

[^1]: I went to university to learn this term.  I'm gonna use it at least once.

bindings/wasm/lib/import-model.ts
elalish: Only if gltf-transform does the work for you though. No point writing a bunch of code for this small feature.

elalish: Optional sounds good.

bindings/wasm/lib/gltf-node.ts
elalish: I was thinking makeManifold is the one we could remove.

bindings/wasm/lib/import-model.ts
tonious: There isn't really a single command way of doing this.  It's not too bad, but I think it's out of scope for this MR.

tonious: Done.

bindings/wasm/lib/import-model.ts
tonious: Done.

bindings/wasm/lib/gltf-node.ts
```diff
@@ -140,7 +160,27 @@ export class NonManifoldGLTFNode extends BaseGLTFNode {
    */
   makeManifold(): Manifold {
     return gltfNodeToManifold(this.document, this.node);
-  };
+  }
+
+  /**
+   * Does this node contain non-manifold geometry?
+   */
+  isNonManifold() {
```
elalish: This is `NonManifoldGLTFNode.isNonManifold()`, which is confusing. This is kind of an expensive check with all the conversions and you don't even get to keep the results. I think we can remove this function - if they care, they can do it themselves. 

Let's also rename `NonManifoldGLTFNode` to `VisualizationGLTFNode`, since we're not sure yet if it's manifold.

bindings/wasm/lib/import-model.ts
```diff
@@ -137,28 +156,130 @@ export async function importManifold(uri: string): Promise<Manifold> {
...
+}
+
+/**
+ * Extract meshes from a gltf-transform node (and its children), or from all
+ * nodes in a document, and convert them to Mesh objects.  Meshless nodes will
+ * be silently skipped.
+ *
+ * @group Utility Functions
+ * @param document
+ * @param node Optionally, traverse the children of this node.
```
elalish: Do you mean children of this node or descendants of this node?

bindings/wasm/lib/import-model.ts
```diff
@@ -137,28 +156,130 @@ export async function importManifold(uri: string): Promise<Manifold> {
...
+ * @group Utility Functions
+ * @param meshes
+ * @returns
+ */
+export function meshesToManifold(meshes: Array<Mesh>): Manifold {
+  const {Manifold} = getManifoldModuleSync()!;
+
+  const manifolds = meshes.map(mesh => tryToMakeManifold(mesh))
+                        .filter(manifold => !!manifold)
+                        .filter(manifold => !manifold.isEmpty());
```
elalish: Do you need these checks, since they're done inside tryToMakeManifold too?

bindings/wasm/lib/import-model.ts
```diff
@@ -137,28 +156,130 @@ export async function importManifold(uri: string): Promise<Manifold> {
...
+/**
+ * Given a list of Mesh objects, attempt to convert them individually into
+ * Manifold objects, and return the union.  Non-manifold Meshes will be silently
+ * skipped.
+ *
+ * @group Utility Functions
+ * @param meshes
+ * @returns
+ */
+export function meshesToManifold(meshes: Array<Mesh>): Manifold {
```
elalish: I don't think this needs to be public, since we already have importManifold.

bindings/wasm/lib/import-model.ts
```diff
@@ -137,28 +156,130 @@ export async function importManifold(uri: string): Promise<Manifold> {
...
 export function gltfNodeToManifold(
-    document: GLTFTransform.Document, node: GLTFTransform.Node): Manifold {
-  const converted: Array<Manifold> = [];
-  node.traverse(child => {
-    const manifold = gltfMeshToManifold(document, child);
-    if (manifold) converted.push(manifold);
-  });
-  if (!converted.length) {
-    throw new Error(`Model contains no manifold geometry.`);
+    document: GLTFTransform.Document, node?: GLTFTransform.Node): Manifold {
```
elalish: I don't think this needs to be public either - `importManifold` is a nice high-level function; if that's not adequate, I think they should probably go all the way down to dealing with each mesh individually.

bindings/wasm/lib/import-model.ts
```diff
@@ -105,10 +116,18 @@ export async function importModel(uri: string): Promise<NonManifoldGLTFNode> {
...
 }
 
 /**
+ * Import a model, and convert it to a Manifold object for manipulation.
+ *
+ * The original imported model may consist of an entire tree of nodes, each of
+ * which may or may not be manifold.  This method will convert each child node,
+ * and then union the results together.  If a child node has no mesh, the mesh
+ * has no geometry, or the mesh is not manifold, that child node will be
+ * silently excluded.
```
elalish, line 185: I've suggested narrowing the scope of the public API here quite a bit, but one ease of use addition we could make is to add an optional tolerance parameter to importManifold and use `merge()` if it's specified. 

Except! I just realized that apparently I forgot to expose the mesh tolerance value in our JS API. That's a significant oversight on my part. If you feel comfortable patching that into our bindings, feel free - otherwise let me know and I'll put up a PR for that.

tonious: I can give that a shot.  Stretch my C++ legs.  It'll be a separate PR, though.

elalish:
Looking really good - what's left before you're ready to unmark it as a draft? Does your checklist at the top need any updating?

tonious:
Checklist up top is up to date, unfortunately.

I've been testing against a side project (3d printed parts for a bench power supply), which has helped.  But I haven't written any real tests for this yet.

I might leave off moving code from `gltf-io` -> `[import|export]-gltf` and from `scene-builder` -> `export-model`.  That shouldn't really be publicly visible if done right.

bindings/wasm/lib/import-model.ts
```diff
@@ -235,11 +239,12 @@ const tryToMakeManifold = (mesh: Mesh) => {
  * @param meshes
  * @returns
  */
-function meshesToManifold(meshes: Array<Mesh>): Manifold {
+function meshesToManifold(meshes: Array<Mesh>, tolerance?: number): Manifold {
   const {Manifold} = getManifoldModuleSync()!;
 
   const manifolds = [];
   for (const mesh of meshes) {
+    if (tolerance) mesh.tolerance = tolerance;
```
elalish: By itself this won't have any effect. You'll need to add a condition if `tryToMakeManifold` fails, you'll need to run `mesh.merge()` (where tolerance is used) and then rerun `tryToMakeManifold`. I think it's best to only use `merge` if necessary, since it can make topological changes.

tonious:
Screenshot 2025-11-27 at 1 06 22 pm

Left to right: `importModel()`, `importManifold()`, and `importManifold()` with a tolerance of 0.5.

bindings/wasm/lib/import-model.ts
tonious: Right.  Done.

tonious:
Steady progress.

The motherboard example was giving me fits because it's scaled wrong.  It's 145 meters wide, which is about the size of a destroyer.  The navy kind, not the star kind.  I've applied transforms from each glTF node to its manifold counterpart, and all three motherboards are now properly stadium sized.  The moon/space example files are 120 and 100 meters across, respectively.  It might be fun to scale it up to be actually moon sized :grin:

Next, I can import models and make them manifold, but I somehow broke it in a subtle way.  [The example](https://github.com/KhronosGroup/glTF/tree/main/extensions/2.0/Vendor/EXT_mesh_manifold/samples/manifold) from the `EXT_mesh_manifold` is not converting cleanly on this branch.  It works just fine through `make-manifold.html`, so it's definitely my fault.

I'm a little happier with import/export now.  It's intended to be easily extensible; glTF-transform being the baseline implementation, naturally.  This is still a work in progress, so it might shift a little.

There are a set of interfaces that define importers and exporters.  Any implementation can declare which extensions and their associated mime-types it will handle.  The glTF and 3MF versions are modules, but a class or object could implement these.

```typescript
export interface Format {
  extension: string;
  mimetype: string;
}

export interface Importer {
  importFormats: Array<Format>;
  fromBlob: (blob: Blob) => Promise<GLTFTransform.Document>;
  fromArrayBuffer:
      (buffer: Uint8Array<ArrayBufferLike>) => Promise<GLTFTransform.Document>;

  fetchModel?: (uri: string) => Promise<GLTFTransform.Document>;
  readFile?: (uri: string) => Promise<GLTFTransform.Document>;
}

export interface Exporter {
  exportFormats: Array<Format>;
  toBlob: (doc: GLTFTransform.Document) => Promise<Blob>;
  toArrayBuffer:
      (doc: GLTFTransform.Document) => Promise<Uint8Array<ArrayBufferLike>>;

  writeFile?: (filename: string, doc: GLTFTransform.Document) => Promise<void>;
}
```

The optional methods are all regarding file system access.  I have two competing constraints in mind:

  * I want to leave some room so that I can supply a file system mock or object later.  This will be useful in browser based environments, whether entirely virtual or uploading folders (like jscad does).  It will also make it easier to translate calls for the vscode extension.
  * I'd also like to retain support for multi-file models.  That said, it might be simpler to just implement a `PlatformIO` class for glTF-transform.  There's not much to [the node implementation](https://github.com/donmccurdy/glTF-Transform/blob/main/packages/core/src/io/node-io.ts) anyhow.

This is what I get for (over)thinking and typing at the same time.  I'll rip the optional methods out.  I'm probably also going to raise the `toBlob()` method out and make the implementation level as thin as possible, given that it still has to speak glTF-transform.

In other tangentially-related news, I've used [opencascade-tools](https://github.com/ghackenberg/opencascade-tools) to translate a few files.  It's a wrapper around fifty megs of webassembly, so it's not fast.  But it is an effective local converter for step files.

elalish:
Great progress, thanks! Sing out if you have any architectural / API questions, but it seems like you have it well in hand. 

I had no idea opencascade had WASM for tessellating STEP files - that's cool! And I'm not surprised it's huge; that's no easy feat, especially if they can generate a reasonably manifold mesh.

tonious:
Does [this look reasonable](https://github.com/elalish/manifold/pull/1434/files#diff-5de67c57bd5839b084a470558700a7c8e1251fc0d5bfebef1c8852bfca76eb75R21-R37)?

```typescript
  // Models can be imported by URL on manifoldCAD.org, or by path with the CLI.
  // `new URL(relativeURL, import.meta.url)` is a way of resolving relative
  // paths that works inside manifoldCAD.  It also works in node.js (after v21)
  // and most modern browsers, albeit with slightly different results.
  //
  // On disk, `import.meta.url` is the absolute url of the source file. Included
  // files can reference assets by relative paths, and we'll still be able to
  // find them.
  //
  // In a browser as there is no filesystem.  By default, `import.meta.main`
  // will be set to the location of the browser.  On manifoldCAD.org, this means
  // that relative URLs are resolved using `https://manifoldcad.org/` as the
  // base URL.
  const moon =
      await importManifold(new URL('./models/moon.glb', import.meta.url));
  const space =
      await importManifold(new URL('./models/space.glb', import.meta.url));
```

The worker can't even see the window location, the editor has to pass that forwards.  One can tell the worker to use a different base URL.

What I'm trying to wrap my scruffy head around is "How could a module package a `.glb` file?".  This approach is about 2/3 of the way there.  It works on a filesystem.  It works in a browser _if_ the developer has taken the time to put resources in a known place.

hrgdavor:
editor should pass baseurl to worker, and url of the script so you know the starting point. 
Depending how you execute the script ... I had to change the script to convert import.meta to require.meta because eval was not working

bindings/wasm/test/examples/import-manifold.ts
```diff
@@ -0,0 +1,68 @@
...
+// or combined like any other manifoldCAD object.
+
+import {importManifold, Manifold} from 'manifold-3d/manifoldCAD';
+const {compose} = Manifold;
+
+const csg = async () => {
+  // glTF (.glb or .gltf) models have a defined scale of 1 unit to 1 metre.  On
+  // the other hand, manifoldCAD has a defined scale of 1 unit to 1 millimetre.
+  // These models are 100m and 120m wide, respectively.  Smaller than you'd
+  // expect for a moon, but larger than you'd expect for a model.
```
elalish, line 18: Maybe we should just resave them at 0.1m scale so we can skip the note...

bindings/wasm/test/examples/import-manifold.ts
```diff
@@ -0,0 +1,68 @@
...
+  // Models can be imported by URL on manifoldCAD.org, or by path with the CLI.
+  // `new URL(relativeURL, import.meta.url)` is a way of resolving relative
+  // paths that works inside manifoldCAD.  It also works in node.js (after v21)
+  // and most modern browsers, albeit with slightly different results.
+  //
+  // On disk, `import.meta.url` is the absolute url of the source file. Included
+  // files can reference assets by relative paths, and we'll still be able to
+  // find them.
+  //
+  // In a browser as there is no filesystem.  By default, `import.meta.main`
```
elalish: Do you mean `import.meta.url`?

bindings/wasm/test/examples/import-manifold.ts
```diff
@@ -0,0 +1,68 @@
...
+  // These models are 100m and 120m wide, respectively.  Smaller than you'd
+  // expect for a moon, but larger than you'd expect for a model.
+  const metres = 1000;
+
+  // Models can be imported by URL on manifoldCAD.org, or by path with the CLI.
+  // `new URL(relativeURL, import.meta.url)` is a way of resolving relative
+  // paths that works inside manifoldCAD.  It also works in node.js (after v21)
+  // and most modern browsers, albeit with slightly different results.
+  //
+  // On disk, `import.meta.url` is the absolute url of the source file. Included
```
elalish: The including source file or the included source file?

bindings/wasm/test/examples/import-manifold.ts
```diff
@@ -0,0 +1,68 @@
...
+  // files can reference assets by relative paths, and we'll still be able to
+  // find them.
+  //
+  // In a browser as there is no filesystem.  By default, `import.meta.main`
+  // will be set to the location of the browser.  On manifoldCAD.org, this means
```
elalish: Perhaps tab URL instead of location of the browser? It's an app on my filesystem after all.

bindings/wasm/test/examples/import-manifold.ts
```diff
@@ -0,0 +1,68 @@
...
+    space.add(moon),
+    space.subtract(moon),
+    space.intersect(moon),
+  ];
+
+  // Space out the results.
+  const arranged = geometry.reduce((acc: Manifold, cur: Manifold) => {
+    // Place accumulated product to the left (x-), with the right side at x=0.
+    // The left side of the current object will be on the right, after a gap.
+    // They will not overlap, so `compose()` is preferable over `union()`.
```
elalish: actually, even if you union it'll just compose under the hood anyway due to optimizing the CSG tree internally.

hrgdavor:
>  convert import.meta to require.meta

to be celar script was written normally with import.meta, I just had to change it before evaulation in my worker.

tonious:
> convert import.meta to require.meta

That's actually something that esbuild does.  I haven't written a test up yet, but this should work the same way both inside and outside of manifoldCAD.  Outside of manifoldCAD it will use the native value.

bindings/wasm/test/examples/import-manifold.ts
tonious: I think it's worth keeping it in.   Mis-scaled models are a thing, and implicit scaling can make it more confusing.

bindings/wasm/test/examples/import-manifold.ts
tonious: Good point.

tonious:
> Anyway, this PR has been open for some time - maybe it would be best to get it merged even if it's not perfect, so we can try some things like this and kick the tires.

There are a few niggling details, but it feels reasonably stable.

bindings/wasm/test/examples/import-manifold.ts
elalish: Right, definitely avoid implicit scaling - I just figured the example might be simpler if the GLBs are already at a reasonable scale. Either way is fine though.

bindings/wasm/lib/bundler.ts
```diff
@@ -78,7 +93,7 @@ export const esbuildManifoldPlugin = (options: BundlerOptions = {}):
     let manifoldCADExportPath: string|null = null;
     const manifoldCADExportSpecifier = 'manifold-3d/manifoldCAD'
     const ManifoldCADExportMatch = /^manifold-3d\/manifoldCAD(.ts|.js)?$/
-    const manifoldCADExportNames = [
+    const manifoldCADExportNames = [ // FIXME update this list.
```
elalish: Any more updates needed?

bindings/wasm/lib/bundler.ts
```diff
@@ -59,6 +61,19 @@ function log(...args: any[]) {
   }
 }
 
+const insertMetaData =
+    (text: string, sourceUrl?: string) => {
+      if (sourceUrl) {
+        const st = new MagicString(text);
+        st.prepend(`const _import_meta_url=_manifold_runtime_url ?? '${
```
elalish, line 68: What does this do?

bindings/wasm/lib/gltf-io.ts
```diff
@@ -423,3 +434,26 @@ function readPrimitive(
...
 }
+
+let _io: GLTFTransform.PlatformIO|null = null;
+
+/**
+ * Return an appropriate PlatformIO object.
+ */
+const getIO = (): GLTFTransform.PlatformIO => {
+  if (!_io) {
+    _io = new GLTFTransform.WebIO();
```
elalish, line 445: Should this ever be NodeIO instead?

bindings/wasm/lib/gltf-node.ts
```diff
@@ -27,41 +52,98 @@ export class GLTFMaterial {
...
+  clone(newParent?: BaseGLTFNode) {
+    const copy = new GLTFNodeTracked(newParent ?? this.parent);
+    Object.assign(copy, this);
+    return copy;
+  }
+}
+
+/**
+ * Position an imported model for later export.
+ * _Non-manifold_ is over stating the case; _not-necessarily-manifold_ may be a
```
elalish: Update?

bindings/wasm/lib/import-model.ts
```diff
@@ -0,0 +1,480 @@
...
+ */
+export interface ImportOptions {
+  /**
+   * Use `mimetype` to determine the format of the imported model, rather than
+   * inferring it.
+   */
+  mimetype?: string;
+  /**
+   * Tolerance for mesh simplification.  Any edge smaller than tolerance may be
+   * collapsed.
```
elalish: Really this is even more about merging - maybe "will attempt to close gaps smaller than tolerance in an effort to make the mesh manifold."

bindings/wasm/lib/import-model.ts
```diff
@@ -0,0 +1,480 @@
...
+
+  throw new Error(`Could not import model \`${source}\`.`);
+}
+
+/**
+ * Fetch a model over HTTP/HTTPS.
+ *
+ * @group Low Level Functions
+ **/
+export async function fetchModel(
```
elalish, line 257: Do these lower functions need to be exported, or can we just expose `readModel` - it seems to handle what each of them do.

bindings/wasm/lib/worker.test.ts
```diff
@@ -104,8 +97,8 @@ suite('Examples', () => {
   test('Heart', async () => {
     const result = await runExample('Heart');
     expect(result?.genus).to.equal(0, 'Genus');
-    expect(result?.volume).to.be.closeTo(3.342, 0.001, 'Volume');
-    expect(result?.surfaceArea).to.be.closeTo(11.51, 0.01, 'Surface Area');
+    expect(result?.volume).to.be.closeTo(282742, 10, 'Volume');
+    expect(result?.surfaceArea).to.be.closeTo(22186, 1, 'Surface Area');
```
elalish, line 101: Why did these change? I don't see any changes to the heart example...

bindings/wasm/lib/gltf-io.ts
tonious: Actually no.  It's not doing any direct IO, that's one abstraction level up in import- or export-model.  Here, it just needs to be able to serialize into an ArrayBuffer.

bindings/wasm/lib/bundler.ts
tonious: This sets `import.meta.url`.  At run time, if `_manifold_runtime_url` has been set (by the worker) it will use that.  If not, it will use `sourceUrl`, which is the filesystem url of the file now being bundled.

bindings/wasm/lib/bundler.ts
tonious: Just did this, and added a test for it.

bindings/wasm/lib/worker.test.ts
tonious: The [glTF node gets scaled](https://github.com/elalish/manifold/blob/master/bindings/wasm/test/examples/heart.mjs#L44-L51).:

```typescript
const heart = ball.warp(func);
const box = heart.boundingBox();
const scale = 100 / (box.max[0] - box.min[0]);

setMorphStart(ball, func);
const node = new GLTFNode();
node.manifold = ball;
node.scale = [scale, scale, scale];
```

Before this PR, transforms on glTF nodes were ignored.  The previous values are in local space; the new values are in global space.

bindings/wasm/lib/bundler.ts
tonious: There's an [esbuild config change](https://github.com/elalish/manifold/pull/1434/files#diff-5f3c6c0b93bbad2d87d129ae8e536d1bc66911a1896b6742848f6d9750855e63R312-R314) that goes with this:

```javascript
    // Define some paths so we can find resources by relative URL.
    define: {
      'import.meta.url': '_import_meta_url',
    }
```

This pattern is pretty much straight from [this esbuild issue](https://github.com/evanw/esbuild/issues/1492#issuecomment-893144483), modified for use without a filesystem.

bindings/wasm/lib/import-model.ts
tonious: They're not exposed within manifoldCAD, and I don't really expect these to be useful to a manifoldCAD user.  I think they would be useful when writing utilities using the wasm bindings in general.

At any rate, I want to write unit tests for some of these, eventually.  For that to work, they have to be exported.  Gimme a thumbs down, and I'll mark them as `@internal`, or a thumbs up and I'll leave 'em.

elalish: Okay, I thought anything exported here was public API; if they're not exposed, that's fine.

PR #1436, opened 2025-11-20
Reorder package.json exports

tonious:
This should allow typescript to find definition files, when installed via npm.

Tested locally on a fresh project.


elalish:
Thanks!

PR #1439, opened 2025-11-22
Ensure that Manifold timing message includes deferred operations.

tonious:
@elalish [said](https://github.com/elalish/manifold/discussions/1437)

> Think we can just combine the manifold and glTF timers? I don't think gltf-transform is doing much work except at export anyway.

Sure 😀


elalish (approved):
Thanks!

PR #1442, opened 2025-11-28
Stop using deprecated macos-13 images in CI

coryrc:
I made the smallest bump possible. Could also jump to macos-15-intel at the appropriate places, but I figured you may be trying to keep this running for (barely-)older OS versions.

https://github.com/actions/runner-images/issues/13046

### DIFF
--- a/.github/workflows/manifold.yml
+++ b/.github/workflows/manifold.yml
@@ -280,9 +280,9 @@ jobs:
       matrix:
         parallelization: [OFF, ON]
         shared: [OFF, ON]
-        # macos-14-intel: intel
+        # macos-15-intel: intel
         # macos-latest: arm
-        os: [macos-14-intel, macos-latest]
+        os: [macos-15-intel, macos-latest]
     runs-on: ${{matrix.os}}
     steps:
     - name: Install common dependencies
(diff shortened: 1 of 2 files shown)


coryrc:
Also, from that link, looks like they're disabling it intermittently soon and completely in a little more than a week.

pca006132:
I think macos-15-intel is fine

coryrc:
> I think macos-15-intel is fine

Done.

.github/workflows/manifold.yml
```diff
@@ -280,9 +280,9 @@ jobs:
       matrix:
         parallelization: [OFF, ON]
         shared: [OFF, ON]
-        # macos-13: intel
+        # macos-15-intel: intel
         # macos-latest: arm
-        os: [macos-13, macos-latest]
+        os: [macos-15-intel, macos-latest]
```
pca006132, line 285: thinking about it, should we use `macos-15` instead of `macos-latest`?

pca006132:
And I think this PR requires @elalish to change the required CI checks setting.

.github/workflows/manifold.yml
coryrc: macos-15-intel will be the last supported macos on intel; otherwise I'd suggest the opposite. `macos-latest` means you don't have to make these updates in the future.

Issue #1430, opened 2025-11-15
Symbolic perturbation occasionally fails to handle perfectly coincident faces


luisfonsivevo:
My application of Manifold makes very heavy use of coincident faces, and seemingly stress tests symbolic perturbation past its breaking point even on simple meshes. I've produced several meshes with flat shards that are obviously visually incorrect. Manually setting the tolerance, even to something crazy like 1, has no effect. Notice how each sample does not perform any rotation that would introduce floating point rounding error.  It's possible that my use case doesn't fit what Manifold offers, but this first one in particular I believe may be a bug, because there are no self-overlaps and the coincident faces that left behind shards are axis-aligned.

```
import {Manifold} from 'manifold-3d/manifoldCAD';

const octagonCCW =
[
  [0, 2],
  [1, 1],
  [2, 0],
  [3, 0],
  [4, 1],
  [5, 2],
  [5, 3],
  [4, 4],
  [3, 5],
  [2, 5],
  [1, 4],
  [0, 3],
];

const leftHoleCW =
[
  [1, 2],
  [1, 3],
  [2, 4],
  [2, 1],
];

const rightHoleCW =
[
  [4, 2],
  [3, 1],
  [3, 4],
  [4, 3],
];

const octagonWithDividingBar = Manifold.extrude([octagonCCW, leftHoleCW, rightHoleCW], 5)
const bar = Manifold.cube([1, 3, 1], false);
const topBar = bar.translate([2, 1, 4]);
const bottomBar = bar.translate([2, 1, 0]);
export default octagonWithDividingBar.subtract(topBar).subtract(bottomBar);
```

`topBar` is successfully subtracted, and `bottomBar` is not.
<img width="546" height="392" alt="Image" src="https://github.com/user-attachments/assets/e9df4179-44d8-4866-a4b2-55fa66877c44" />

The next one may arguably constitute as self-overlapping. It consists of 2 boxes that meet at one edge, which Manifold is usually able to handle. The coordinates are dyadic/integers, so I don't think floating point rounding error is the culprit here. Shifting the big and little boxes away from each other "fixes" the shard. Depending on which `rot` is uncommented, the result appears incorrect in different ways.

```
import {Manifold} from 'manifold-3d/manifoldCAD';

const rot = [1, 0, 0, 0, 0, 1, 0, 0, 0, 0, 1, 0, 0, 0, 0, 1]; const axis = 2; //no rotation - shard is on the perimeter of the little box
[...]

PR #1443, opened 2025-11-30
Cleanup of Boolean3

elalish:
Not fixing #1430 here yet, but I've done a pass on refactoring to make the code a little easier to follow hopefully and ideally make it easier to improve the symbolic perturbation. I'll work on that next, but figured I may as well get this in first. 

@pca006132 This is the first time I've really delved into the code since your massive refactor to speed up our broad phase. It's so much nicer now! Thanks for making that so much cleaner at the same time as you made it so much faster. 

Now P and Q really mean the inputs inP and inQ, while A and B get swapped based on `forward`. I also tried making the perturbation a function, but it basically cut the speed in half, so I'm skipping that idea.

### DIFF
--- a/src/boolean3.cpp
+++ b/src/boolean3.cpp
@@ -33,40 +33,40 @@ namespace {
 // are carefully designed to minimize rounding error and to eliminate it at edge
 // cases to ensure consistency.
 
-vec2 Interpolate(vec3 pL, vec3 pR, double x) {
-  const double dxL = x - pL.x;
-  const double dxR = x - pR.x;
+vec2 Interpolate(vec3 aL, vec3 aR, double x) {
+  const double dxL = x - aL.x;
+  const double dxR = x - aR.x;
   DEBUG_ASSERT(dxL * dxR <= 0, logicErr,
                "Boolean manifold error: not in domain");
   const bool useL = fabs(dxL) < fabs(dxR);
-  const vec3 dLR = pR - pL;
+  const vec3 dLR = aR - aL;
   const double lambda = (useL ? dxL : dxR) / dLR.x;
   if (!std::isfinite(lambda) || !std::isfinite(dLR.y) || !std::isfinite(dLR.z))
-    return vec2(pL.y, pL.z);
+    return vec2(aL.y, aL.z);
   vec2 yz;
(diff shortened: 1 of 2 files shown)


src/boolean3.cpp
```diff
@@ -33,85 +33,86 @@ namespace {
...
-  const double q1ex = vertPosQ[q1e].x;
-  int s01 = reverse ? Shadows(q1sx, p0x, expandP * normal[q1s].x) -
-                          Shadows(q1ex, p0x, expandP * normal[q1e].x)
-                    : Shadows(p0x, q1ex, expandP * normal[p0].x) -
-                          Shadows(p0x, q1sx, expandP * normal[p0].x);
+inline std::pair<int, vec2> Shadow01(const int a0, const int b1,
+                                     const Manifold::Impl& inA,
+                                     const Manifold::Impl& inB,
+                                     const double expandP,
+                                     VecView<const vec3> normalP,
```
pca006132, line 81: I guess this `normalP` will be removed later when we change the perturbation logic? Or should we get this from `inA`/`inB` and `forward`?

src/boolean3.cpp
```diff
@@ -179,14 +181,13 @@ struct Kernel11 {
 };
 
 struct Kernel02 {
-  VecView<const vec3> vertPosP;
-  VecView<const Halfedge> halfedgeQ;
-  VecView<const vec3> vertPosQ;
+  const Manifold::Impl& inA;
+  const Manifold::Impl& inB;
   const double expandP;
   VecView<const vec3> vertNormalP;
```
pca006132, line 187: we can get this from `inA`/`inB` and `forward`?

src/boolean3.cpp
elalish: Yeah, this is going to change, so leaving it very explicit for the moment.

elalish: Yeah, this is going to change, so leaving it very explicit for the moment.

src/boolean3.cpp
elalish: Same as above.

elalish: Same as above.

PR #1446, opened 2025-12-04
export internal type modules for TypeScript portability

Loosetooth:
When users write ManifoldCAD scripts and try to transpile them with TypeScript tools (tsc, tsdown, rolldown), they encounter error TS2742:

```typescript
// script.ts
import { CrossSection } from 'manifold-3d/manifoldCAD';

export default CrossSection.square([10, 10]).extrude(5);
```

```
error TS2742: The inferred type of 'default' cannot be named without a 
reference to './node_modules/manifold-3d/manifold-encapsulated-types'. 
This is likely not portable. A type annotation is necessary.
```

**Root cause**: The `lib/*.d.ts` files (and `manifold.d.ts` itself) import from `./manifold-encapsulated-types` and `./manifold-global-types`, but these modules aren't declared in the package.json `exports` map. TypeScript sees these as internal, non-portable paths.

**Fix**: Add the missing exports:
```json
"./manifold-encapsulated-types": "./manifold-encapsulated-types.d.ts",
"./manifold-global-types": "./manifold-global-types.d.ts"
```

These files are already shipped in the package and imported by `manifold.d.ts`. They just weren't explicitly exported.

### DIFF
--- a/bindings/wasm/package.json
+++ b/bindings/wasm/package.json
@@ -46,6 +46,7 @@
     },
     "./manifold-encapsulated-types.d.ts": "./manifold-encapsulated-types.d.ts",
     "./manifold-global-types.d.ts": "./manifold-global-types.d.ts",
+    "./manifoldCAD.js": "./lib/manifoldCAD.js",
     "./manifoldCAD": {
       "import": {
         "types": "./dist/manifoldCAD.d.ts",


tonious:
This might already be fixed in #1436, just not published to npm.

elalish:
Interesting - anyone care to describe the difference between those two solutions?

We should probably put out another release once import gets merged and we've kicked the tires a bit.

Loosetooth:
@tonious You're right, the current `master` branch already works just fine. I didn't notice before.

> Interesting - anyone care to describe the difference between those two solutions?

If I understand correctly, there are two `manifoldCAD.d.ts` files in the manifold-3d package:
1. `manifold-3d/dist/manifoldCAD.d.ts` This one is correctly bundled/built by rollup, doesn't have any strange "internal" type imports.
2. `manifold-3d/lib/manifoldCAD.d.ts` This one is generated by `tsc` and **has** references to other type files that are not exported by the package.

Because of the order of the `exports` in package.json, when running `tsc` on a script that imports `manifold-3d/manifoldCAD`, number 2 was being hit before number 1.
`tsc` then encounters the "internal" (not exported) type files such as `manifold-encapsulated-types.d.ts`, and fails to generate types correctly.

@tonious If there is no other good reason to export these `manifold-encapsulated-types.d.ts` and `manifold-global-types.d.ts` files from the package, then I am happy to close this PR in favor of [your other changes](https://github.com/elalish/manifold/pull/1436/commits/5cadf2637e06a3069acbf1fa939784036d3d94e2).

tonious:
[API-Extractor](https://api-extractor.com/) rolls up the type declarations.  I've disabled a bunch of the functionality, as this is pretty much all we need it to do.  There are separate roll-ups for overall manifoldCAD functions, and the few functions like animation and level-of-detail that are only changeable at the top level.



@Loosetooth said:

> @tonious If there is no other good reason to export these manifold-encapsulated-types.d.ts and manifold-global-types.d.ts files from the package, then I am happy to close this PR in favor of [your other changes](https://github.com/elalish/manifold/pull/1436/commits/5cadf2637e06a3069acbf1fa939784036d3d94e2).

Now that I've thought about it a little, this PR makes sense. I didn't feel comfortable entirely omitting the not-rolled-up-version of `./lib/manifoldCAD.d.ts`, but hadn't really thought about _why_.  Not everyone is going to use the entire manifoldCAD part of manifold.  Still, one should be able to reasonably use _part_ of it, which means that the varying `./lib/*.d.ts` files should load cleanly.  They can't all be rolled up as that would result in duplicate type definitions.

Can I request a few small changes while you're cleaning up my mess?

Complete extensions for consistency:

```json
  "./manifold-encapsulated-types.d.ts": "./manifold-encapsulated-types.d.ts",
  "./manifold-global-types.d.ts": "./manifold-global-types.d.ts"
```

Also, we should probably explicitly export `manifold.(js|d.ts)`, even though it's under `main` and `types` in `package.json` already.

```json
  "./manifold.js": "./manifold.js",
  "./manifold.d.ts": "./manifold.d.ts",
  "./manifold": {
    "import": {
      "types": "./manifold.d.ts",
      "default": "./manifold.js"
    }
  },
```

@elalish said:

> We should probably put out another release once import gets merged and we've kicked the tires a bit.

I'm working on it! :joy:

Loosetooth:
@tonious I added your proposed changes.

I also added and explicit export for `manifoldCAD.js`.

tonious:
Looks good!  Thanks, @Loosetooth :)

PR #1454, opened 2025-12-15
Use currentName over scriptName.

tonious:
... or the 'format' button throws an exception.

### DIFF
--- a/bindings/wasm/examples/editor.js
+++ b/bindings/wasm/examples/editor.js
@@ -398,7 +398,7 @@ async function createEditor() {
 
     // The user edited an example.
     // Copy it into a new script.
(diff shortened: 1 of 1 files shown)


elalish (approved):
👍

elalish:
Not sure if this PR is the culprit, but we have a major regression on ManifoldCAD - when you edit an example it no longer turns into a new script, and when switching scripts your changes are lost.

tonious:
> Not sure if this PR is the culprit, but we have a major regression on ManifoldCAD - when you edit an example it no longer turns into a new script, and when switching scripts your changes are lost.

I haven't been able to replicate that.  I've tested in firefox and chrome, and it seems to work for me.

elalish:
Weird, I can't repro now either. Maybe the PWA got into a weird state. False alarm.

PR #1455, opened 2025-12-15
manifoldc minor API fixes

geoffder:
On @pca006132 's friendly prodding, I'm working to bring OManifold up to date and getting it to compile with build system changes etc. This prunes a declaration to nowhere and makes a few functions more consistent.

### DIFF
--- a/bindings/c/manifoldc.cpp
+++ b/bindings/c/manifoldc.cpp
@@ -481,9 +481,15 @@ ManifoldMeshGL64* manifold_meshgl64_merge(void* mem, ManifoldMeshGL64* m) {
   return m;
 }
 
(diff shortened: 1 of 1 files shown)


bindings/c/include/manifold/manifoldc.h
```diff
@@ -437,7 +437,6 @@ size_t manifold_meshgl_size();
 size_t manifold_meshgl64_size();
 size_t manifold_box_size();
 size_t manifold_rect_size();
-size_t manifold_curvature_size();
```
pca006132, line 440: interesting, so we never had this before?

geoffder: There used to be a curvature type (one of the things that I've had to prune from OManifold).

PR #1456, opened 2025-12-16
fix SplitPinchedVerts parallelization

pca006132:
This turns out to be rather simple. I probably did not think about determinism when writing the parallelized version.

Basically, the older behavior is to pick an arbitrary loop to keep and duplicate the pinched vertex for other loops. With this change, it always keep the loop associated with the halfedge with the smallest id.

I checked with #1445, this seems to work well.

### DIFF
--- a/src/edge_op.cpp
+++ b/src/edge_op.cpp
@@ -13,6 +13,7 @@
 // limitations under the License.
 
 #include <unordered_map>
+#include <unordered_set>
 
 #include "impl.h"
 #include "parallel.h"
@@ -735,8 +736,7 @@ void Manifold::Impl::SplitPinchedVerts() {
   ZoneScoped;
 
   auto nbEdges = halfedge_.size();
-  // FIXME: non-deterministic
-#if FALSE  // MANIFOLD_PAR == 1
+#if MANIFOLD_PAR == 1
   if (nbEdges > 1e4) {
     std::mutex mutex;
     std::vector<size_t> pinched;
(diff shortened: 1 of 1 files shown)


src/edge_op.cpp
```diff
@@ -13,6 +13,7 @@
 // limitations under the License.
 
 #include <unordered_map>
+#include <unordered_set>
```
elalish, line 16: necessary?

src/edge_op.cpp
```diff
@@ -792,13 +794,17 @@ void Manifold::Impl::SplitPinchedVerts() {
         });
 
     manifold::stable_sort(pinched.begin(), pinched.end());
-    std::vector<bool> halfedgeProcessed(nbEdges, false);
+    pinched.resize(std::distance(
+        pinched.begin(), manifold::unique(pinched.begin(), pinched.end())));
```
elalish, line 798: When is the distance between pinched.begin() and unique pinched.begin() not zero?

elalish (approved):
Thanks!

src/edge_op.cpp
pca006132: `unique` returns an iterator to the end of the deduplicated list, so the answer is usually :)

src/edge_op.cpp
pca006132: needed for the unordered set.

src/edge_op.cpp
elalish: Oh, right, that makes sense.

PR #1457, opened 2025-12-16
boolean3 performance optimization

pca006132:
Just make `expandP` a bool, and make this with `forward` a compile time constant via templates.

This can speed up boolean3 quite significantly. For examples with many intersections, e.g. ember test between 472042 and 99865, this can reduce the single threaded runtime from 1620ms to 1200ms, and multithreaded runtime from 140ms to 130ms.

### DIFF
--- a/include/manifold/linalg.h
+++ b/include/manifold/linalg.h
@@ -908,10 +908,8 @@ struct vec<T, 3> {
   LINALG_CONSTEXPR14 T& operator[](int i) {
     return i == 0 ? x : i == 1 ? y : z;
   }
-  constexpr const vec<T, 2>& xy() const {
-    return *reinterpret_cast<const vec<T, 2>*>(this);
-  }
-  vec<T, 2>& xy() { return *reinterpret_cast<vec<T, 2>*>(this); }
+  constexpr vec<T, 2> xy() const { return vec<T, 2>(x, y); }
+  constexpr vec<T, 2> yz() const { return vec<T, 2>(y, z); }
 
   template <class U, class = detail::conv_t<vec, U>>
   constexpr vec(const U& u) : vec(converter<vec, U>{}(u)) {}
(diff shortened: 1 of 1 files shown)


src/boolean3.cpp
```diff
@@ -389,9 +392,18 @@ Intersections Intersect12(const Manifold::Impl& inP, const Manifold::Impl& inQ,
...
 };
 
-Vec<int> Winding03(const Manifold::Impl& inP, const Manifold::Impl& inQ,
-                   const VecView<std::array<int, 2>> p1q2, double expandP,
-                   bool forward) {
+template <bool forward>
+Intersections Intersect12(const Manifold::Impl& inP, const Manifold::Impl& inQ,
+                          bool expandP) {
+  if (expandP)
+    return Intersect12_<true, forward>(inP, inQ);
```
elalish, line 399: These wrappers make me wonder if it would be better to put the expandP template on the Boolean class itself? Not sure if that would make the code simpler or more complicated though...

include/manifold/linalg.h
```diff
@@ -908,10 +908,8 @@ struct vec<T, 3> {
   LINALG_CONSTEXPR14 T& operator[](int i) {
     return i == 0 ? x : i == 1 ? y : z;
   }
-  constexpr const vec<T, 2>& xy() const {
-    return *reinterpret_cast<const vec<T, 2>*>(this);
-  }
-  vec<T, 2>& xy() { return *reinterpret_cast<vec<T, 2>*>(this); }
+  constexpr vec<T, 2> xy() const { return vec<T, 2>(x, y); }
+  constexpr vec<T, 2> yz() const { return vec<T, 2>(y, z); }
```
elalish, line 912: Is this related to the rest of this change? I'm a little surprised we don't want both of these: the const and the non-const.

src/boolean3.cpp
pca006132: At least it will force you do write all methods in the header file. IMO the current way is a bit ugly but simple enough.

include/manifold/linalg.h
pca006132: Oh, this is unrelated to other changes here but related to UB. The `reintetpret_cast` is illegal, and there is no way to expose the mutable method without changing the representation (e.g. make the struct contains an array instead).

PR #1461, opened 2025-12-17
Import model tests.

tonious:
Nothing sophisticated, just some basic smoke tests to exercise `import-models.ts`.


bindings/wasm/lib/import-model.test.ts
```diff
@@ -0,0 +1,95 @@
...
+suite('importModel()', () => {
+  test('import a model without EXT_mesh_manifold', async () => {
+    const model = await importer.importManifold(
+        new URL('../test/fixtures/models/box.glb', import.meta.url));
+    expect(model.volume()).to.be.closeTo(100 * 100 * 100, 1);
+  });
+
+  test('import a model by file URL', async () => {
+    const model = await importer.importManifold(
+        new URL('../test/fixtures/models/box.glb', import.meta.url));
```
elalish, line 83: What's the difference between this test and the one above?

bindings/wasm/lib/import-model.ts
```diff
@@ -298,7 +305,26 @@ export async function readFile(filename: string, options: ImportOptions = {}) {
...
+ *
+ * glTF has a defined scale of 1:1 metre.
+ * manifoldCAD has a defined scale of 1:1 mm.
+ *
+ * @internal
+ */
+function importTransform(doc: GLTFTransform.Document): GLTFTransform.Document {
+  // Find top level nodes and correct their scale.
+  for (const sourceNode of doc.getRoot().listNodes()) {
+    if (sourceNode.getParentNode()) continue;
```
elalish, line 322: Does gltf-transform really not have a simpler way to find top-level nodes? I thought there'd be a root.children() or something.

elalish (approved):
Thanks!

bindings/wasm/lib/import-model.test.ts
tonious: Semantic only.  Ultimately I want three samples to test: with `EXT_mesh_manifold`, without it, and without it, but with a known tolerance value.

At the same time, I don't want to lose the file url vs path distinction, because that's exactly the sort of thing future me is likely to mess up.

bindings/wasm/lib/import-model.ts
tonious: [It doesn't seem to](https://gltf-transform.dev/modules/core/classes/Root).  It's pretty light weight.

tonious: Now that I'm reading the docs a little closer, [scene.listChildren()](https://gltf-transform.dev/modules/core/classes/Scene#listChildren) might work.  I'll try it.

tonious: No dice.  Root nodes and scene nodes are different things, apparently.

bindings/wasm/test/fixtures/models/boxNotManifold.obj
```diff
@@ -0,0 +1,25 @@
...
+f 7 8 4
+f 5 7 4
+f 7 5 6
+f 7 6 8
+f 6 3 8
+f 5 1 6
+# Pull one corner out of position:
+#f 6 1 3
+v 0.045 0.045 0.05
+f 9 1 3
```
tonious, line 25: This corner is [-0.005,-0.005, 0] out of position, which has a magnitude of 0.00625.  However, it merges with a tolerance of 0.005.

:shrug: Probably not important, but thought I'd mention it.

Issue #1462, opened 2025-12-17
Crash regression (missing props)


stephomi:
I got a regression since 54baa62b4ad49a8a4b82ebe25785cce14d622d40
Can't share the model but it has a bunch of different objects and some of them have UVs and others don't.
I don't have the full context but I think it should be the fix (it fixes my issue at least):

```diff
diff --git a/src/csg_tree.cpp b/src/csg_tree.cpp
index 35585a61..c8998c0d 100644
--- a/src/csg_tree.cpp
+++ b/src/csg_tree.cpp
@@ -231,14 +231,15 @@ std::shared_ptr<CsgLeafNode> CsgLeafNode::Compose(
         const int nextVert = vertIndices[i];
         const int nextEdge = edgeIndices[i];
         const int nextProp = propVertIndices[i];
+        const bool hasProp = node->pImpl_->NumProp() > 0;
         transform(node->pImpl_->halfedge_.begin(),
                   node->pImpl_->halfedge_.end(),
                   combined.halfedge_.begin() + edgeIndices[i],
-                  [nextVert, nextEdge, nextProp](Halfedge edge) {
+                  [hasProp, nextVert, nextEdge, nextProp](Halfedge edge) {
                     edge.startVert += nextVert;
                     edge.endVert += nextVert;
                     edge.pairedHalfedge += nextEdge;
-                    edge.propVert += nextProp;
+                    if (hasProp) edge.propVert += nextProp;
                     return edge;
                   });
```

stephomi:
I still have a prop issue with another model though, so this change isn't enough.

Probably an ugly workaround but it seems to work:

```diff
diff --git a/src/sort.cpp b/src/sort.cpp
index a05933ca..65c73814 100644
--- a/src/sort.cpp
+++ b/src/sort.cpp
@@ -321,6 +321,14 @@ void Manifold::Impl::CompactProps() {
   ZoneScoped;
   if (numProp_ == 0) return;
 
+  int pmax = 0;
+  for (auto &k : halfedge_) {
+    if (k.propVert > pmax) pmax = k.propVert;
+  }
+  if ((pmax * NumProp()) >= properties_.size()) {
+    properties_.resize((pmax + 1) * NumProp());
+  }
+
   const int numProp = NumProp();
   const auto numVerts = properties_.size() / numProp;
   Vec<int> keep(numVerts, 0);
```

stephomi:
Test file (one mesh with uvs and one without):
[...]

PR #1465, opened 2025-12-19
Add test with props

stephomi:
Crashing test (https://github.com/elalish/manifold/issues/1462)
I think it happens when you have two objects with different numProp and the one with props has fewer vertices.
(test object are far away in this example, they don't collide)

If it doesn't crash you can always check/confirm the issue there: 
```diff
diff --git a/src/sort.cpp b/src/sort.cpp
index a05933ca..11aeb5f6 100644
--- a/src/sort.cpp
+++ b/src/sort.cpp
@@ -321,6 +321,15 @@ void Manifold::Impl::CompactProps() {
   ZoneScoped;
   if (numProp_ == 0) return;
 
+  int pmax = 0;
+  for (auto &k : halfedge_) {
+    if (k.propVert > pmax) pmax = k.propVert;
+  }
+  if ((pmax * NumProp()) >= properties_.size()) {
+    printf("ERROR PROPS %d %d\n", properties_.size(), pmax);
+    properties_.resize((pmax + 1) * NumProp());
+  }
+
```


elalish:
Good catch, you were right about where the regression was introduced. Thanks for test - I further simplified it. And you almost had the fix.

PR #1467, opened 2025-12-21
Add perturbation test

### LINKED ISSUE #1430: Symbolic perturbation occasionally fails to handle perfectly coincident faces

elalish:
Fixes #1430 

I played around with the symbolic perturbation - couldn't fix the example completely. But then I noticed something about this example: it was rotating a cube about its center through 360 degrees, so it was effectively unioning cubes exactly identical, along with all the rotations thereof. Once I removed that (rotating only through 90 degrees), I can no longer reproduce any problems with the symbolic perturbation, even though I'm not bothering to `AsOriginal` or `Simplify`. I even tried it doing rotations around the other axes to check the other parts of the perturbation and it's all working as expected. 

So, on the one hand, I could try to figure out why enough identical cubes eventually create some shards, but it doesn't seem a great use of time. Let's hold off until we find a more realistic test case that gives problems.

### DIFF
--- a/test/boolean_test.cpp
+++ b/test/boolean_test.cpp
@@ -13,7 +13,6 @@
 // limitations under the License.
 
 #include "../src/utils.h"
-#include "manifold/cross_section.h"
 #include "manifold/manifold.h"
 #include "test.h"
 
@@ -378,12 +377,12 @@ TEST(Boolean, Perturb3) {
   Manifold gear = Manifold::BatchBoolean(outerCubes, OpType::Add);
   Manifold outerGear = gear.Scale({2, 2, 1});
 
-  const float topArea = CrossSection(gear.Project()).Area();
-  const float sideArea = gear.SurfaceArea() - 2 * topArea;
-
   // Subtract inner from outer to create the nasty gear with slivers
   Manifold nastyGear = outerGear - gear;
-  const float expectedArea = 3 * sideArea + 6 * topArea;
+
+  // const float topArea = CrossSection(gear.Project()).Area();
+  // const float sideArea = gear.SurfaceArea() - 2 * topArea;
+  const float expectedArea = 26.972;  // 3 * sideArea + 6 * topArea;
   const float expectedVolume = outerGear.Volume() - gear.Volume();
 
   // The gear should be valid and manifold


elalish:
@zalo, does that change fix your upstream (python or scad) test as well?

zalo:
Huh!  That’s a great insight; I did notice similar behavior in a multiplayer WebAssembly app I toyed with a while ago ( https://zalo.github.io/WorldParty/ ) and didn’t think too hard about it… ended up adding a slight random judder when putting cubes down to eliminate the slivers 💀

Agreed there is perhaps something fundamental to dig into here…

Issue #1468, opened 2025-12-24
Crash in Manifold::Impl::SplitPinchedVerts()


kintel:
I'm reliably reproducing a crash in Manifold::Impl::SplitPinchedVerts(). Currently only reproducible using OpenSCAD, but I'll try to package this as a Manifold test case if necessary.

Corresponding OpenSCAD issue: https://github.com/openscad/openscad/issues/6235

I've tried both Manifold 3.3.2 and latest master.
TBB is onetbb 2022.2.0 or 2022.3.0

ASAN complains about a TBB worker thread:
```
==98363==ERROR: AddressSanitizer: unknown-crash on address 0x200062200003f0f8 at pc 0x00010284882c bp 0x00016e15dd80 sp 0x00016e15dd78
READ of size 8 at 0x200062200003f0f8 thread T3
    #0 0x000102848828 in std::__1::__bit_reference<std::__1::vector<bool, std::__1::allocator<bool>>, true>::operator=[abi:ne200100](bool) __bit_reference:85
    #1 0x00010731f88c in _ZZZN8manifold8Manifold4Impl17SplitPinchedVertsEvENK3$_2clIN3tbb6detail2d113blocked_rangeImEEEEDaRKT_ENKUliE_clEi edge_op.cpp:776
    #2 0x00010731f1ec in _ZN8manifold8Manifold4Impl7ForVertIZZNS1_17SplitPinchedVertsEvENK3$_2clIN3tbb6detail2d113blocked_rangeImEEEEDaRKT_EUliE_EEviSB_ impl.h:275
    #3 0x00010731e828 in _ZZN8manifold8Manifold4Impl17SplitPinchedVertsEvENK3$_2clIN3tbb6detail2d113blocked_rangeImEEEEDaRKT_ edge_op.cpp:775
    #4 0x00010731e138 in decltype(std::declval<manifold::Manifold::Impl::SplitPinchedVerts()::$_2 const&>()(std::declval<tbb::detail::d1::blocked_range<unsigned long>&>())) std::__1::__invoke[abi:ne200100]<manifold::Manifold::Impl::SplitPinchedVerts()::$_2 const&, tbb::detail::d1::blocked_range<unsigned long>&>(manifold::Manifold::Impl::SplitPinchedVerts()::$_2 const&, tbb::detail::d1::blocked_range<unsigned long>&) invoke.h:179
[...]

PR #1469, opened 2025-12-24
add reproducer for issue #1468

pca006132:
For this issue, the assimp importer will do something with the mesh, and the imported mesh becomes empty. I do the IO using our debug mode IO, and is able to reproduce the issue.

It seems that the issue is caused by the triangulation not producing a manifold result. And when intermediate checks are suppressed, we have a corrupted halfedge data structure.

### DIFF
--- a/test/manifold_test.cpp
+++ b/test/manifold_test.cpp
@@ -1138,7 +1138,7 @@ TEST(Manifold, MergeRefine) {
 #endif
 
 #ifdef MANIFOLD_DEBUG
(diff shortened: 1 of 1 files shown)


test/test.h
```diff
@@ -37,6 +37,12 @@ struct MeshSize {
   int numPropVert = numVert;
 };
 
+struct ManifoldParamGuard {
```
elalish, line 40: 💯

PR #1473, opened 2025-12-26
Lazy collider

pca006132:
Can probably be regarded as a late Christmas present or an early New Year's present :)

This implements part of point 3 in #1458. It makes the collider construction lazy. ~~After implementing it, I found that this approach is probably still not the best, because we still need to sort the faces with morton code. I tried avoiding such a sort, but it brings more issues. I haven't yet investigated to see if this indicates a bug in our code or something.
The gain is not too big, but at least the changes are quite simple.~~

Anyway, some data about performance:

For `perfTest`, it is around 5% improvement:
```
original:
nTri = 512, time = 0.00114211 sec
nTri = 2048, time = 0.00212679 sec
nTri = 8192, time = 0.00550277 sec
nTri = 32768, time = 0.0116046 sec
nTri = 131072, time = 0.0276725 sec
nTri = 524288, time = 0.0944312 sec
nTri = 2097152, time = 0.377017 sec
nTri = 8388608, time = 1.50151 sec

now:
nTri = 512, time = 0.00120504 sec
nTri = 2048, time = 0.00206288 sec
nTri = 8192, time = 0.00519169 sec
nTri = 32768, time = 0.0106602 sec
nTri = 131072, time = 0.024389 sec
nTri = 524288, time = 0.0789487 sec
nTri = 2097152, time = 0.338706 sec
nTri = 8388608, time = 1.43133 sec
```

Test statistics:
```
BooleanComplex: 921 -> 903
Samples: 11244 -> 10804
Hull: 3497 -> 3473
Smooth: 978 -> 912
SDF: 1258 -> 1229
Boolean: 228 -> 205
Manifold: 28 -> 28
Properties: 5389 -> 5316
Overall: 25287 -> 24726
```

For the ember test, I observed 5% performance improvement.

---

Btw, just a side note. I actually tried to use AI to do this, but it seems that it is quite difficult to give detailed enough prompts (maybe my skill issue). At least I ended up just rewriting almost everything that AI gave me :(

### DIFF
--- a/src/lazy_collider.h
+++ b/src/lazy_collider.h
@@ -54,27 +54,11 @@ class LazyCollider {
 
   bool IsBuilt() const;
 
-  template <const bool selfCollision, typename Recorder>
-  struct Adapter {
-    using Local = typename Recorder::Local;
-    Recorder* recorder;
-
-    Local& local() { return recorder->local(); }
-
-    void record(int queryIdx, int leafIdx, Local& local) {
-      if constexpr (selfCollision) {
-        if (leafIdx == queryIdx) return;
-      }
-      recorder->record(queryIdx, leafIdx, local);
-    }
-  };
-
   template <const bool selfCollision = false, typename Recorder,
             typename... Args>
   void Collisions(Recorder& recorder, Args... args) const {
     const Built& built = EnsureBuilt();
-    Adapter<selfCollision, Recorder> adapter{&recorder};
-    built.collider->Collisions<false>(adapter, built.transform, args...);
+    built.collider->Collisions<false>(recorder, built.transform, args...);
   }
 
   static bool IsAxisAligned(const mat3x4& transform);


src/collider.h
```diff
@@ -320,29 +302,50 @@ class Collider {
...
+                  std::optional<mat3x4> transform, bool parallel = true) const {
     ZoneScoped;
     using collider_internal::FindCollision;
     if (internalChildren_.empty()) return;
-    for_each_n(parallel ? autoPolicy(n, collider_internal::kSequentialThreshold)
-                        : ExecutionPolicy::Seq,
-               countAt(0), n,
-               FindCollision<decltype(f), selfCollision, Recorder>{
-                   f, nodeBBox_, internalChildren_, recorder});
+    if (transform) {
```
elalish, line 300: Seems like a lot of duplicated code between these two Collisions functions - can one call the other instead?

src/impl.cpp
```diff
@@ -645,15 +633,28 @@ Manifold::Impl Manifold::Impl::Transform(const mat3x4& transform_) const {
...
   result.SetEpsilon(result.epsilon_);
+
+  const bool canDeferCollider = LazyCollider::IsAxisAligned(transform_);
+  if (LazyCollider::IsAxisAligned(transform_)) {
+    result.collider_ = std::make_shared<LazyCollider>(collider_, transform_);
+  } else if (!result.IsEmpty()) {
+    Vec<Box> faceBox;
+    Vec<uint32_t> faceMorton;
+    result.GetFaceBoxMorton(faceBox, faceMorton);
+    result.SortFaces(faceBox, faceMorton);
```
elalish: I don't think this should be necessary - the old logic was to call Transform for axis-aligned or Update for non. Neither resorted - It's just that Update has to walk the tree to update the parent boxes to enclose their children since applying the transform doesn't suffice.

pca006132: true, will change that later and see how it goes

src/collider.h
pca006132: It should be possible. This is just mirroring the underlying collider API. We can rewrite the collider to eliminate one.

Forgot why exactly I did this, need to see if it makes any difference performance-wise.

pca006132:
After addressing the comments, there is no performance regression, which is nice

elalish (approved):
Nice, thanks!

PR #1475, opened 2025-12-28
Fix face normals

### LINKED ISSUE #1262: Cracks after simplification

elalish:
Fixes #1262 

This turned out to be an interesting one - SwapDegenerates was identifying the wrong faces as degenerate because it projects along their face normal and it was getting the wrong normal occasionally. The wrong normals were coming from CollapseColinearEdges. The problem here turned out to be from my change to MarkCoplanar - in this example with a sharp corner and very high tessellation, there were some triangles past the edge that get added to the other face since they are narrower than tolerance. This is fine, but they have a different face normal. Then when edge collapse happens, it was assuming all the coplanar faces had matching face normals, so it doesn't update them, but that can turn these degenerate triangles into big triangles, which now have clearly wrong normals.

The fix is to assign the same face normal to all triangles that are coplanar. Simple enough, but in so doing I realized that we weren't calculating normals in a place that made sense, so ended up doing a bunch of refactoring to clean that up. We should now be calculating vertex normals less often.

### DIFF
--- a/src/smoothing.cpp
+++ b/src/smoothing.cpp
@@ -825,7 +825,7 @@ void Manifold::Impl::CreateTangents(int normalIdx) {
         }
       });
 
-  halfedgeTangent_.swap(tangent);
+  halfedgeTangent_ = std::move(tangent);
   DistributeTangents(fixedHalfedge);
 }
 
@@ -864,7 +864,7 @@ void Manifold::Impl::CreateTangents(std::vector<Smoothness> sharpenedEdges) {
                              vertNormal[halfedge_[edgeIdx].startVert], edgeIdx);
              });
 
-  halfedgeTangent_.swap(tangent);
+  halfedgeTangent_ = std::move(tangent);
 
   // Add sharpened edges around faces, just on the face side.
   for (size_t tri = 0; tri < NumTri(); ++tri) {
@@ -985,6 +985,8 @@ void Manifold::Impl::Refine(std::function<int(vec3, vec4, vec4)> edgeDivisions,
   SortGeometry();
   if (old.halfedgeTangent_.size() == old.halfedge_.size()) {
     SetNormalsAndCoplanar();
+  } else {
+    CalculateVertNormals();
   }
   meshRelation_.originalID = -1;
 }
--- a/src/subdivision.cpp
+++ b/src/subdivision.cpp
@@ -556,7 +556,7 @@ Vec<Barycentric> Manifold::Impl::Subdivide(
 
           tmp[i] += la::max(Added(hIdx), Added(halfedge_[hIdx].pairedHalfedge));
         });
-    edgeAdded.swap(tmp);
+    edgeAdded = std::move(tmp);
   }
 
   Vec<int> edgeOffset(numEdge);
--- a/src/manifold.cpp
+++ b/src/manifold.cpp
@@ -811,8 +811,8 @@ Manifold Manifold::SmoothOut(double minSharpAngle, double minSmoothness) const {
       pImpl->CreateTangents(0);
       // Reset the properties to the original values, removing temporary normals
       pImpl->numProp_ = numProp;
-      pImpl->properties_.swap(properties);
-      pImpl->halfedge_.swap(halfedge);
+      pImpl->properties_ = std::move(properties);
+      pImpl->halfedge_ = std::move(halfedge);
     } else {
       pImpl->CreateTangents(pImpl->SharpenEdges(minSharpAngle, minSmoothness));
     }


.vscode/launch.json
```diff
@@ -12,7 +12,7 @@
       "program": "${workspaceFolder}/build/test/manifold_test",
       "args": [
         "--gtest_catch_exceptions=0",
-        "--gtest_filter=Manifold.Simplify"
+        "--gtest_filter=SDF.SineSurface"
```
elalish, line 15: This test is passing, but I think that's just because it's not tight enough; it looks wrong when I export it. I need to check on that before merging this.

src/impl.h
```diff
@@ -294,7 +293,7 @@ struct Manifold::Impl {
   void MarkCoplanar();
   void DedupePropVerts();
   void RemoveUnreferencedVerts();
-  void InitializeOriginal(bool keepFaceID = false);
+  void InitializeOriginal(bool keepFaceID = true);
```
elalish: I think I can remove this bool param entirely now.

src/impl.cpp
```diff
@@ -209,8 +209,10 @@ Manifold::Impl::Impl(Shape shape, const mat3x4 m) {
   vertPos_ = vertPos;
   for (auto& v : vertPos_) v = m * vec4(v, 1.0);
   CreateHalfedges(triVerts);
-  Finish();
   InitializeOriginal();
+  CalculateBBox();
+  SetEpsilon();
```
elalish, line 214: I was adding these in so I could remove them from Finish and not duplicate work, but I don't think I've actually removed them yet.

src/impl.cpp
```diff
@@ -292,7 +296,9 @@ void Manifold::Impl::MarkCoplanar() {
 
       const vec3 v = vertPos_[halfedge_[h].endVert];
       if (std::abs(dot(v - base, normal)) < tolerance_) {
-        meshRelation_.triRef[h / 3].coplanarID = tp.tri;
+        const size_t tri = h / 3;
+        meshRelation_.triRef[tri].coplanarID = tp.tri;
+        faceNormal_[tri] = normal;
```
elalish, line 300: This was the actual bug fix.

src/smoothing.cpp
```diff
@@ -115,7 +115,7 @@ struct InterpTri {
       cosTheta = -cosTheta;
     }
 
-    if (cosTheta > 1.0 - std::numeric_limits<double>::epsilon()) {
+    if (std::abs(cosTheta) > 1.0 - std::numeric_limits<double>::epsilon()) {
```
elalish, line 118: This bug has been around for quite some time.

src/impl.cpp
```diff
@@ -305,6 +311,7 @@ void Manifold::Impl::MarkCoplanar() {
       }
     }
   }
+  CalculateNormals();
```
elalish: I need to rename this, since it's only calculating vertNormals now. I should rename Finish too, since that's my worst-named function ever.

src/subdivision.cpp
```diff
@@ -663,7 +666,8 @@ Vec<Barycentric> Manifold::Impl::Subdivide(
                                          bary[rIdx[2]], bary[rIdx[3]]}});
                   });
       });
-  meshRelation_.triRef = triRef;
+  std::swap(meshRelation_.triRef, triRef);
```
pca006132: why swap?

src/subdivision.cpp
```diff
@@ -642,8 +643,10 @@ Vec<Barycentric> Manifold::Impl::Subdivide(
         Vec<ivec3> newTris = subTris[tri].Reindex(tri3, edgeOffsets, edgeFwd,
                                                   interiorOffset[tri]);
         copy(newTris.begin(), newTris.end(), triVerts.begin() + triOffset[tri]);
-        auto start = triRef.begin() + triOffset[tri];
-        fill(start, start + newTris.size(), meshRelation_.triRef[tri]);
+        auto startRef = triRef.begin() + triOffset[tri];
+        fill(startRef, startRef + newTris.size(), meshRelation_.triRef[tri]);
+        auto startN = faceNormal.begin() + triOffset[tri];
+        fill(startN, startN + newTris.size(), faceNormal_[tri]);
```
pca006132, line 649: Should this be filled or computed? iirc the vertex positions are interpolated with the tangents, so should the face normal change?

src/impl.cpp
```diff
@@ -250,24 +252,26 @@ void Manifold::Impl::InitializeOriginal(bool keepFaceID) {
 void Manifold::Impl::MarkCoplanar() {
   ZoneScoped;
   const int numTri = NumTri();
+  faceNormal_.resize(numTri);
```
pca006132, line 254: ok so we are now always computing face normal here?

.vscode/launch.json
pca006132: I'm thinking does this mean that we should have some image based check in our CI (also available locally), similar to what openscad does (https://github.com/openscad/openscad/blob/master/tests/image_compare.py).

src/subdivision.cpp
elalish: Better than a copy. But I suppose I should just move it instead.

src/impl.cpp
elalish: Yeah, we were doing the work anyway, so now it saves the results and we don't duplicate it elsewhere.

src/subdivision.cpp
pca006132: yeah, a move makes more sense

.vscode/launch.json
elalish: I think our standard volume and surface area checks will suffice, I just didn't have them on this one.

src/subdivision.cpp
elalish: The refine function runs MarkCoplanar to recalculate them if tangents are supplied.

test/sdf_test.cpp
```diff
@@ -191,6 +191,8 @@ TEST(SDF, SineSurface) {
 
   EXPECT_EQ(smoothed.Status(), Manifold::Error::NoError);
   EXPECT_EQ(smoothed.Genus(), 38);
+  EXPECT_NEAR(smoothed.Volume(), 107.4, 0.1);
+  EXPECT_NEAR(smoothed.SurfaceArea(), 394.7, 0.1);
```
elalish, line 195: I pulled these values from master so that the test properly fails now. Interestingly if I remove `.Simplify()` I get a good result - doesn't quite match master, but might be better. However, that means some combination of Simplify and SmoothOut now has a regression. I'm also getting a crash on smooth_out in the Python bindings, so maybe related. Must be something subtle, because I didn't change those functions much.

PR #1478, opened 2025-12-31
shared halfedge

pca006132:
Make halfedges shared by default. Most of the code is around changing `Vec` to make it shareable (using template to avoid overhead for non-shareable vectors), and adding code to make the vector unique for mutation. The current code just throws an error if you try to structurally modify the vector (e.g., `push_back` or `pop_back`), ideally we should assert this when users try to mutate the elements, but this is not doable without making things more complicated, e.g., return a cell object that checks the reference count. So the current version is probably good enough...

Performance-wise it is a bit faster, probably around 5%. Need to check that later.

And regarding code bloat in `Vec`: The major issue is that copy constructor and copy assignment operator cannot have templates, so I have to manually instantiate them. See https://en.cppreference.com/w/cpp/language/as_operator.html and https://en.cppreference.com/w/cpp/language/member_template.html#Member_function_templates

### DIFF
--- a/src/vec.h
+++ b/src/vec.h
@@ -321,6 +321,8 @@ class Vec : public VecView<T> {
   }
 
   static_assert(std::is_trivially_destructible<T>::value);
+  // This is required so we can access private fields between shared and
+  // non-shared vectors.
   friend Vec<T, !shared>;
 
   static void free_async(T* ptr, size_t size) {


src/boolean_result.cpp
```diff
@@ -317,9 +317,9 @@ void AppendPartialEdges(Manifold::Impl& outR, Vec<char>& wholeHalfedgeP,
   // while remapping them to the output using vP2R. Use the verts position
   // projected along the edge vector to pair them up, then distribute these
   // edges to their faces.
-  Vec<Halfedge>& halfedgeR = outR.halfedge_;
+  auto& halfedgeR = outR.halfedge_;
```
pca006132, line 320: And the annoying thing is that the compiler will not throw an error if you keep `Vec<Halfedge>&`: It will automatically copy construct a `Vec<Halfedge>` aka `Vec<Halfedge, false>` from the `Vec<Halfedge, true>` aka `SharedVec<Halfedge>`, and then pass it around as a reference...

src/quickhull.cpp
```diff
@@ -223,9 +223,9 @@ HalfEdgeMesh::HalfEdgeMesh(const MeshBuilder& builderObject,
 /*
  * Implementation of the algorithm
  */
-std::pair<Vec<Halfedge>, Vec<vec3>> QuickHull::buildMesh(double epsilon) {
+std::pair<SharedVec<Halfedge>, Vec<vec3>> QuickHull::buildMesh(double epsilon) {
```
pca006132, line 226: it works even if we do not change this, but we need to do a copy in that case

src/vec.h
```diff
@@ -41,36 +46,54 @@ class Vec;
  * implementation that did not consider things like non-trivial
  * constructor/destructor, please keep T trivial.
  */
-template <typename T>
+template <typename T, bool shared>
 class Vec : public VecView<T> {
  public:
-  Vec() {}
+  Vec() : VecView<T>() {
+    if constexpr (shared) count_ = new std::atomic<int>(1);
```
pca006132, line 59: need `if constexpr` to avoid type errors, because `count_` is of type `Empty` when `shared = false`.

src/vec.h
```diff
@@ -41,36 +46,54 @@ class Vec;
...
   }
 
-  Vec(size_t size, T val) : VecView<T>() { resize(size, val); }
+  inline Vec(size_t size, T val) : VecView<T>() {
+    if constexpr (shared) count_ = new std::atomic<int>(1);
+    resize(size, val);
+  }
 
-  Vec(const Vec<T>& vec) : VecView<T>() { *this = Vec(vec.view()); }
+  inline Vec(const Vec<T, true>& vec) : VecView<T>() {
```
pca006132, line 78: these are where manual specialization is done

src/vec.h
```diff
@@ -82,34 +105,27 @@ class Vec : public VecView<T> {
...
-      free_async(this->ptr_, capacity_);
-    }
-    this->ptr_ = nullptr;
-    this->size_ = 0;
-    capacity_ = 0;
-  }
+  ~Vec() { dealloc(); }
 
-  Vec<T>& operator=(const Vec<T>& other) {
+  Vec& operator=(const Vec& other) {
```
pca006132, line 124: The copy constructor for `SharedVec`, i.e., `shared = true`, just copies the reference and increments the reference count.

src/vec.h
```diff
@@ -214,12 +259,56 @@ class Vec : public VecView<T> {
...
+        *this = Vec<T, true>(this->view());
+      }
+    }
+  }
 
  private:
   size_t capacity_ = 0;
 
+#if __cplusplus >= 202002L
+  [[no_unique_address]]
```
pca006132, line 289: Normally empty structs have size 1 (1 byte) to ensure every field has a unique address, but in c++20 you can use `no_unique_address` to avoid this and the field takes no space. Not really required but I found it nice.

src/vec.h
```diff
@@ -214,12 +259,56 @@ class Vec : public VecView<T> {
...
+    this->ptr_ = other.ptr_;
+    this->size_ = other.size_;
+    this->capacity_ = other.capacity_;
+    other.ptr_ = nullptr;
+    other.size_ = 0;
+    other.capacity_ = 0;
+  }
+
   static_assert(std::is_trivially_destructible<T>::value);
+  friend Vec<T, !shared>;
```
pca006132, line 326: This is required so we can access private fields between shared and non-shared vectors.

src/vec.h
```diff
@@ -214,12 +259,56 @@ class Vec : public VecView<T> {
     capacity_ = this->size_;
   }
 
-  size_t capacity() const { return capacity_; }
+  inline size_t capacity() const { return capacity_; }
+
+  void MakeUnique() {
```
elalish, line 274: What does this do? Does it affect other shared copies?

elalish (approved):
This is getting a little fancy for me, but it is pretty cool. My recommendation is to turn all of your PR comments into code comments.

src/vec.h
pca006132: It does not affect the other shared copies. It copies the result and makes itself a unique copy.

PR #1483, opened 2026-01-02
Add Otterplans to users in README

mathieucivel:
I've added my side project ([otterplans](https://otterplans.com)) to the table. It's an app that can generate woodworking plans and instructions based on the users' dimensions. Your library made it possible ! If I ever get dollars with this project, I'll forward some to you for support and gratitude.


hrgdavor:
Users of manifold is a very nice source to find cool projects. just quick questcio @mathieucivel : What do you use to generate PDF,  I somehow expected preview to be enough to generate preview via browser print to pdf (I tend to do that cuz I am lazy)

mathieucivel:
> Users of manifold is a very nice source to find cool projects. just quick questcio @mathieucivel : What do you use to generate PDF, I somehow expected preview to be enough to generate preview via browser print to pdf (I tend to do that cuz I am lazy)

Thank you for your interest! I use `domtoimage` and `jspdf` for the PDF generation. I've explored the browser print option before but it seemed hard to consistently force page breaks in order to prevent page truncations while printing. Do you know if it possible ?

I'm not sure this the best place to discuss this though, I don't want to create to much noise.

PR #1484, opened 2026-01-04
TypeScript Documentation Overhaul

tonious:
Mostly documentation changes:

  * Reorganized typedocs into 'manifold', 'manifoldCAD', and 'manifoldCAD runtime' sections.
  * Added documentation headers to most library files.
  * Simplified TypeScript readme file
    * Link to other documentation where possible.  This includes absolute links to pages that are not published yet, but exist in branch.
    * Removed some redundant language copied over from the top level file.
  * Lots of small changes to make typedoc happier.
  * Made low level `import-model` and `export-model` functions (like `toBlob` and `fromBlob`) a little more analogous.  (Which is what I wanted to do in the first place)

Still to do:

  * [x] Fix build issue.
  * [ ] Make sure that all `manifoldCAD Runtime` docs explicitly call out what functions are available within manifoldCAD and which are not.  This will likely be by group.  Functions referenced under 'manifoldCAD' are _assumed_ to be available -- that's the end user API reference.
  * [ ] Standardize capitalization for manifold/Manifold and manifoldCAD/ManifoldCAD

### DIFF
--- a/bindings/wasm/examples/make-manifold.ts
+++ b/bindings/wasm/examples/make-manifold.ts
@@ -18,8 +18,7 @@ import {prune} from '@gltf-transform/functions';
 import {SimpleDropzone} from 'simple-dropzone';
 
 import {disposeMesh, Properties, readMesh, setupIO, writeMesh} from '../lib/gltf-io';
-
-import Module from './built/manifold.js';
+import Module from '../manifold.js';
 
 // Set up gltf-transform
 const io = setupIO(new WebIO());
--- a/bindings/wasm/examples/model-viewer-script.ts
+++ b/bindings/wasm/examples/model-viewer-script.ts
@@ -16,8 +16,8 @@ import {Document, WebIO} from '@gltf-transform/core';
 import {clearNodeTransform, flatten, mergeDocuments, prune} from '@gltf-transform/functions';
 
 import {disposeMesh, Properties, readMesh, setupIO, writeMesh} from '../lib/gltf-io';
-
-import Module, {Manifold, Mesh} from './built/manifold';
+import type {Manifold as ManifoldType} from '../manifold';
+import Module from '../manifold';
 
 // Set up gltf-transform
 const io = setupIO(new WebIO());
@@ -37,7 +37,7 @@ const id2properties = new Map<number, Properties>();
 // for reading single-object glTFs, as it simply unions any extra meshes
 // together rather than returning a scene hierarchy.
 async function readGLB(url: string) {
-  const manifolds = Array<Manifold>();
+  const manifolds = Array<ManifoldType>();
   const docIn = await io.read(url);
   await docIn.transform(flatten());
   const nodes = docIn.getRoot().listNodes();
@@ -110,7 +110,7 @@ const mv = document.querySelector('model-viewer');
 
 // Use gltf-io and gltf-transform to convert the resulting Manifold to a glTF
 // and display it with <model-viewer>.
-async function push2MV(manifold: Manifold) {
+async function push2MV(manifold: ManifoldType) {
   disposeMesh(node.getMesh()!);
   const mesh = writeMesh(doc, manifold.getMesh(), id2properties);
   node.setMesh(mesh);
--- a/bindings/wasm/examples/package.json
+++ b/bindings/wasm/examples/package.json
@@ -7,6 +7,7 @@
     "dev": "vite",
     "build": "tsc --noEmit && vite build",
     "build:watch": "vite build --watch",
+    "clean": "rm -fr ./dist",
     "prebuild": "cd .. && npm run build",
     "preinstall": "cd .. && npm ci",
     "preview": "vite preview"
--- a/bindings/wasm/examples/three.ts
+++ b/bindings/wasm/examples/three.ts
@@ -14,7 +14,8 @@
 
 import {BoxGeometry, BufferAttribute, BufferGeometry, IcosahedronGeometry, Mesh as ThreeMesh, MeshLambertMaterial, MeshNormalMaterial, PerspectiveCamera, PointLight, Scene, WebGLRenderer} from 'three';
 
-import Module, {Mesh} from './built/manifold';
+import type {Mesh as MeshType} from '../manifold';
+import Module from '../manifold';
 
 // Load Manifold WASM library
 const wasm = await Module();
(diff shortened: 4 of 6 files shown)


bindings/wasm/lib/export-model.ts
```diff
@@ -109,28 +129,36 @@ export function register(exporter: Exporter) {
...
 export async function toArrayBuffer(
-    doc: GLTFTransform.Document, identifier: string): Promise<ArrayBuffer> {
-  return await getExporter(identifier).toArrayBuffer(doc);
+    doc: GLTFTransform.Document,
+    options: ExportOptions = {}): Promise<ArrayBuffer> {
+  if (!(options.mimetype || options.extension)) {
+    throw new Error(
+        'Must specify a mimetype or extension when exporting to a Buffer.');
+  }
+  const format = getFormat((options.mimetype || options.extension)!);
```
elalish, line 160: What happens if they're unknown values? Seems like we could combine that error state with this missing value one.

bindings/wasm/types/manifoldCAD.d.ts
```diff
@@ -13,20 +13,38 @@
...
+ * module, and export it along with everything else listed here.
  * This allows models to behave identically when running on manifoldCAD.org,
  * through the CLI, or through nodejs.
  *
  * It can be imported as `manifold-3d/manifoldCAD`.
  *
  * @packageDocumentation
- * @module manifold-3d/manifoldCAD
+ * @module manifoldCAD
+ * @group manifoldCAD
```
elalish, line 33: Does this end up building a docs website somehow? Or how are these directives used?

bindings/wasm/lib/export-model.ts
tonious: `getFormat` will throw an `UnsupportedFormatError` in that case.  I can probably make that work here.

bindings/wasm/types/manifoldCAD.d.ts
tonious: Yep!  So, `@group` and `@category` provide semantic organization, while `@module` provides logical organization.  Here's how the docs look locally:

Screenshot 2026-01-07 at 10 39 27 am

tonious:
> would you be so kind as to define "runtime" for me? As many years as I've been a software engineer now, I still have no idea what it means, beyond "thing that does a thing".

For me, it's usually "things that must be in memory to make another thing work".  Often implies the presence of an interpreter.

That said, you've already proven that this is not the right nomenclature.  Other thoughts:

  - 'manifoldCAD Libraries' would be too easily confused with 'manifold-the-library'. 
  - 'manifoldCAD Internals', except I don't want them to be purely internal.  Also confusing when talking about what scripts can do inside manifoldCAD.
  - 'manifoldCAD Support'?

And while I've got you, should I capitalize 'Manifold'/'ManifoldCAD'?  My writing is all over the place right now.

JustinTulloss:
Couple more suggestions:

- engine
- just CAD ( ie the layers are "manifold", "cad", and "ui")
- cadAPI

In any event, a layer diagram will probably be needed to show how the dependency DAG is expected to work and what kinds of functionality lives in each layer

elalish:
>Make sure that all manifoldCAD Runtime docs explicitly call out what functions are available within manifoldCAD and which are not. This will likely be by group. Functions referenced under 'manifoldCAD' are assumed to be available -- that's the end user API reference.

Let's try to make this clearer with naming. I think we have `manifold` the C++ library/project, `manifold-3d` the npm package of JS/TS/WASM bindings, and ManifoldCAD.org the GUI application. Then we have `manifold-cad` the command-line application. Finally we have functions one can call within a manifold-cad/ManifoldCAD script. I like the idea of calling these docs the Manifold CAD API. 

However, I think that API should only document public functions, which are the manifold-3d TS API itself plus our various additions mostly around glTF. 

Then I believe you have also documented the lower-level (runtime?) API, which is not exposed in manifold-cad, but can be used by developers who want to make something more advanced on top of `manifold-3d`. I think it's fine call this Manifold runtime docs (or even helpers), but I think they should be buried a little - under a developer section or something. Most users who come to this documentation site will just be trying to write a script. 

Anyway, it'll be easier to review in website form. If that means we should merge this and then talk more, that's fine.

tonious:
I'm okay with merging and refining.

elalish:
Okay, having poked around in this documentation now, it's pretty cool. I would still ideally like to reorganize it as described above, but I'm not sure how feasible that is with tsdoc, and I don't think it's worth spending a pile of time/code on. What is here now is certainly usable and way better than what we had before. 

The main problem now is that all the ManifoldCAD docs are links due to being forwarded exports, but those links lose the useful context of the file-level doc page - plus it makes things a bit harder to navigate. Can we put the functions of a file into separate groups/categories? I'll have to play around a bit more.

tonious:
> The main problem now is that all the ManifoldCAD docs are links due to being forwarded exports, but those links lose the useful context of the file-level doc page - plus it makes things a bit harder to navigate.

I totally agree.  I've been fighting with TypeDoc to make that work, but haven't had any luck yet.

> Can we put the functions of a file into separate groups/categories?

Within a module, use the `@group` and `@category` pages.  Take a look at [import-model](https://manifoldcad.org/jsdocs/modules/lib_import-model.html) for a good example of groups within a module.

At the top level, the groups are `manifold`, `manifoldCAD`, `manifold-gltf`, and `manifoldCAD Runtime`.  The latter has categories underneath it for `Core`, `Input/Output`, and `Modelling`.  This builds the left hand navigation tree.

Maybe split it into two documents?  Generate the 'user' documentation from the rolled up declaration files?  We publish _that_ under a different url, like `/userdocs` or `/caddocs` or something.  Then here, we drop the `manifoldCAD` group, rename `manifoldCAD Runtime` to just `manifoldCAD`, and make it clear that _this_ is developer level documentation for contributing or embedding.

Huh.  C++ docs are at `/docs/html`, but there's nothing at `/docs`.  Maybe refactor to:

  * `/docs/jsuser/`
  * `/docs/jsapi/`
  * `/docs/cppapi/`

elalish:
> Maybe split it into two documents? Generate the 'user' documentation from the rolled up declaration files? We publish _that_ under a different url, like `/userdocs` or `/caddocs` or something. Then here, we drop the `manifoldCAD` group, rename `manifoldCAD Runtime` to just `manifoldCAD`, and make it clear that _this_ is developer level documentation for contributing or embedding.

Yes, this sounds good to me. I wonder if https://typedoc.org/documents/Tags._primaryExport.html would help with getting the user docs to generate without all those reference links?
> 
> Huh. C++ docs are at `/docs/html`, but there's nothing at `/docs`. Maybe refactor to:
> 
> * `/docs/jsuser/`
> * `/docs/jsapi/`
> * `/docs/cppapi/`

Yeah, this seems like a good idea. Let's make a forwarding link so links to the existing docs don't break.

tonious:
I've tried and `@primaryExport` didn't seem to fix the issue.  Splitting the documents seems to work locally.

PR #1489, opened 2026-01-14
Deprecate compose

### LINKED ISSUE #1464: Deprecating `Compose`
1. Check if all existing usages of compose can be replaced with union.
2. Mark the public API as deprecated.
3. Optional: Change the public compose API to use union, so if it causes problem in user code, they will report.
4. Decide when to remove it from the public API. Probably until a major version change.

pca006132:
Closes #1464.

### DIFF
--- a/src/manifold.cpp
+++ b/src/manifold.cpp
@@ -1068,7 +1068,18 @@ Manifold Manifold::Hull() const {
  * @param manifolds A vector of manifolds over which to compute a convex hull.
  */
 Manifold Manifold::Hull(const std::vector<Manifold>& manifolds) {
-  return Compose(manifolds).Hull();
+  std::vector<vec3> vertPos;
+  size_t size = 0;
+  for (const auto& man : manifolds) size += man.NumVert();
+  if (size == 0) return Manifold();
+  vertPos.reserve(size);
+  for (const auto& man : manifolds) {
+    const auto& impl = man.pNode_->ToLeafNode()->GetImpl();
+    vertPos.insert(vertPos.end(), impl->vertPos_.begin(), impl->vertPos_.end());
+  }
+  std::shared_ptr<Impl> impl = std::make_shared<Impl>();
+  impl->Hull(VecView<const vec3>(vertPos));
+  return Manifold(std::make_shared<CsgLeafNode>(impl));
 }
 
 /**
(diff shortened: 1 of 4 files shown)


test/manifold_test.cpp
```diff
@@ -829,15 +832,6 @@ TEST(Manifold, Invalid) {
 }
 #endif
 
-TEST(Manifold, MultiCompose) {
```
pca006132, line 832: It seems that this test is only concerned with compose, so I am removing it.

test/manifold_test.cpp
```diff
@@ -227,9 +227,10 @@ TEST(Manifold, OppositeFace) {
 TEST(Manifold, Decompose) {
   std::vector<Manifold> manifoldList;
   manifoldList.emplace_back(Manifold::Tetrahedron());
-  manifoldList.emplace_back(Manifold::Cube());
-  manifoldList.emplace_back(Manifold::Sphere(1, 4));
-  Manifold manifolds = Manifold::Compose(manifoldList);
+  manifoldList.emplace_back(Manifold::Cube().Translate({2, 0, 0}).AsOriginal());
+  manifoldList.emplace_back(
```
pca006132, line 231: Need to separate the meshes to avoid combining them into one. The `AsOriginal` is needed to pass the `originalID` check.

bindings/c/manifoldc.cpp
```diff
@@ -436,7 +436,7 @@ ManifoldManifold* manifold_revolve(void* mem, ManifoldPolygons* cs,
 }
 
 ManifoldManifold* manifold_compose(void* mem, ManifoldManifoldVec* ms) {
-  auto composed = Manifold::Compose(*from_c(ms));
+  auto composed = Manifold::BatchBoolean(*from_c(ms), OpType::Add);
```
pca006132, line 439: These bindings are manually changed to use `BatchBoolean` to avoid triggering deprecation warning.

src/manifold.cpp
```diff
@@ -1068,7 +1068,18 @@ Manifold Manifold::Hull() const {
  * @param manifolds A vector of manifolds over which to compute a convex hull.
  */
 Manifold Manifold::Hull(const std::vector<Manifold>& manifolds) {
-  return Compose(manifolds).Hull();
+  std::vector<vec3> vertPos;
```
pca006132, line 1071: The direct translation is to use `BatchBoolean`, but there is no point in doing union when we can just directly pass the vertices to quickhull.

PR #1492, opened 2026-01-16
Add user guide link to the editor

tonious:
Like so:

Screenshot 2026-01-16 at 9 32 07 am


bindings/wasm/examples/editor.css
```diff
@@ -219,6 +219,16 @@ model-viewer {
...
 }
 
+.docs {
+  background-image: url(/icons/docs.png);
+  width: 36px;
+  height: 36px;
+
+  cursor: pointer;
+  border: none;
+  border-radius: 18px;
```
elalish, line 229: Why does it need all these attributes that .share and such don't?

bindings/wasm/examples/index.html
```diff
@@ -44,7 +44,8 @@
           <button type="button" id="undo" class="top-icon undo squish"></button>
           <button type="button" id="redo" class="top-icon redo squish"></button>
           <button type="button" id="format" class="top-icon star squish"></button>
-          <button type="button" id="share" class="top-icon share"></button>
+          <button type="button" id="share" class="top-icon share squish"></button>
+          <a target="_blank" href="https://manifoldcad.org/docs/jsuser/" class="top-icon docs"></a>
```
elalish, line 48: This will open the docs in a new tab, right? I think that's what we want.

elalish (approved):
Good idea!

bindings/wasm/examples/editor.css
tonious: The other icons inherit from `button`.  This is on a link element, so it doesn't have these styles by default.

PR #1494, opened 2026-01-19
Add C-binding functions to access the normalIdx parameter of GetMeshGL/GetMeshGL64

WNP78:
This functionality was not previously accessible via the C API and can be exposed via two new functions.  Only new functions are added and no existing functions were changed, to not break FFI compatibility. Names chosen were `manifold_get_meshgl_w_normals` and `manifold_get_meshgl64_w_normals`, to match the naming established by `manifold_meshgl64_w_tangents`. I am unaware of any documentation for the C api but if this exists I can update it too.

### DIFF
--- a/bindings/c/manifoldc.cpp
+++ b/bindings/c/manifoldc.cpp
@@ -436,11 +436,9 @@ ManifoldMeshGL64* manifold_meshgl64_w_tangents(void* mem, double* vert_props,
   return to_c(mesh);
 }
 
-ManifoldMeshGL64* manifold_meshgl64_w_options(void* mem, double* vert_props,
-                                              size_t n_verts, size_t n_props,
-                                              uint64_t* tri_verts,
-                                              size_t n_tris,
-                                              ManifoldMeshGL64Options* options) {
+ManifoldMeshGL64* manifold_meshgl64_w_options(
+    void* mem, double* vert_props, size_t n_verts, size_t n_props,
+    uint64_t* tri_verts, size_t n_tris, ManifoldMeshGL64Options* options) {
   auto mesh = new (mem) MeshGL64();
   mesh->numProp = n_props;
   mesh->vertProperties = vector_of_array(vert_props, n_verts * n_props);


WNP78:
I'm also looking at a way to write to `MeshGLP.runIndex` and `MeshGLP.runOriginalID`, as well as `mergeFromVert` and `mergeToVert` through the C API - they are read only currently. Not sure whether to add it as an option on construction, or make functions to mutate them on an existing instance. Other similar properties (e.g. `halfedgeTangent`) seem to be set in constructor variants, but with tangents plus both optional "run" collections plus the merge collections (which should be provided both or none), plus the MeshGL/MeshGL64 split, it would turn into a bit of a combinatorial mess with 16 or 32 constructor options to cover, depending on if you consider run index + original ID as a unit. 

So I wonder if it would be best to just add e.g:
```c
void manifold_meshgl_set_run_index(ManifoldMeshGL* m, uint32_t* runIndex, size_t length);
void manifold_meshgl_set_run_original_id(ManifoldMeshGL* m, uint32_t* runOriginalID, size_t length);
void manifold_meshgl_set_merge_verts(ManifoldMeshGL* m, uint32_t* mergeFromVert, uint32_t* mergeToVert, size_t length);
```
(plus accompanying meshgl64 versions, which would be identical)

I don't know how strict the semantics on the C api are so I'd like your thoughts on this one

pca006132:
I don't have much opinion about this, it sounds ok to me. @geoffder what do you think?

elalish:
Could we make a `manifold_meshgl` constructor that takes a struct instead to effectively get default arguments?

WNP78:
> Could we make a `manifold_meshgl` constructor that takes a struct instead to effectively get default arguments?

In this case I assume non-set optional fields would just be left as null pointers, so it's not super different from just adding a big constructor function with all the options in it and allowing nullptr arguments, other than the conciseness of the API although I don't know how many people are actualy using the C API directly without some sort of binding.

I can implement that either way on this PR or open a separate one if you want

WNP78:
@elalish I did an implementation with a struct [here](https://github.com/WNP78/manifold/tree/meshgl-options), let me know if you want me to merge it into this PR or make a separate one

elalish:
Yeah, I think your struct version looks good. I like this because it's both more convenient to use, and it's easy for us to extend without breaking the API. Let's merge it in here.

WNP78:
> Yeah, I think your struct version looks good. I like this because it's both more convenient to use, and it's easy for us to extend without breaking the API. Let's merge it in here.

I'll merge it in here, although I would note that if the struct was extended in the future this could be potentially breaking if a newer dll is used with a program that was compiled for the pre-extension struct, since the API consumer is going to be allocating the options struct at the old smaller size, and the C-bindings would then read past it into undefined memory.

elalish:
Ugh, I hate C. I really hope that's not a use case - trying to upgrade a dependency without recompiling? What if we changed the order of an enum or something? Feels very fragile...

WNP78:
> Ugh, I hate C. I really hope that's not a use case - trying to upgrade a dependency without recompiling? What if we changed the order of an enum or something? Feels very fragile...

To be honest, I don't think that ABI compatibility is essential if you ship the bindings with headers, unless it's something that you guarantee. If you want to be extra sure I could add a function to get the size of the options struct, so that programs can armour against it growing in the future.

bindings/c/include/manifold/types.h
```diff
@@ -70,6 +70,28 @@ typedef struct ManifoldProperties {
...
 
+typedef struct ManifoldMeshGLOptions {
+  uint32_t* run_indices;
+  size_t run_indices_length;
+  uint32_t* run_original_ids;
+  size_t run_original_ids_length;
+  uint32_t* merge_from_vert;
+  uint32_t* merge_to_vert;
+  size_t merge_verts_length;
+  float* halfedge_tangents;
```
elalish, line 81: Do these need default values so that one doesn't have to specify all of them?

bindings/c/include/manifold/manifoldc.h
```diff
@@ -49,7 +49,13 @@ ManifoldMeshGL* manifold_meshgl_w_tangents(void* mem, float* vert_props,
                                            size_t n_verts, size_t n_props,
                                            uint32_t* tri_verts, size_t n_tris,
                                            float* halfedge_tangent);
+ManifoldMeshGL* manifold_meshgl_w_options(void* mem, float* vert_props,
```
elalish, line 52: with this function there's no longer a need for `manifold_meshgl_w_tangents`. I'm tempted to just remove it, but I suppose deprecating would be better.

bindings/c/include/manifold/types.h
pca006132: IIRC C doesn't have default values. One way is to provide a function that returns the default value of the struct.

PR #1495, opened 2026-01-19
Fix excessive memory usage in batch union operations

### LINKED ISSUE #1493: batch union memory usage issue
It seems that batch unioning thousands of meshes is using more memory than needed, should investigate why and maybe perform automated batching.

See https://github.com/elalish/manifold/pull/666#discussion_r2701717436

zalo:
## Summary
- Fix excessive memory usage during batch union operations by avoiding eager transform application during heap sorting and bounding box collection
- Add `NumVert()` and `GetBoundingBox()` accessor methods to `CsgLeafNode` that don't trigger transform application
- Add regression test for batch union with many transformed meshes

## Problem
The `MeshCompare` comparator used in `BatchUnion` was calling `GetImpl()->NumVert()`, and the bounding box collection was calling `GetImpl()->bBox_`, both of which trigger transform application. This creates full transformed mesh copies for ALL meshes during the initial sorting/partitioning phase, causing excessive memory usage when processing many transformed meshes that share a common base.

## Solution
1. Added a `NumVert()` method to `CsgLeafNode` that returns `pImpl_->NumVert()` without triggering transform application (vertex count doesn't change with transforms)
2. Added `GetBoundingBox()` method that computes the transformed bounding box by transforming the 8 corners, without applying the transform to the entire mesh

## Benchmark Results

Testing with 10,000 transformed instances of a shared high-detail sphere (1,026 vertices each):

| Metric | Master (simulated) | PR Branch | Improvement |
|--------|-------------------|-----------|-------------|
| Memory for hulls | ~1,299 MB | ~8 MB | **162x reduction** |
| Peak virtual memory | ~1,534 MB | ~243 MB | **6.3x reduction** |
| GetImpl-triggered allocations | 10,000 copies | 0 copies | **Eliminated** |

### How the benchmark works
The benchmark creates 10,000 `Manifold` instances that are transforms of a single shared sphere:
```cpp
Manifold baseSphere = Manifold::Sphere(0.4, 64);  // 1,026 vertices
for (...) {
    hulls.push_back(baseSphere.Translate({px, py, pz}));
}
```

On **master**, when `BatchUnion` calls `GetImpl()` on each mesh (for bounding box or vertex count), it triggers the lazy transform evaluation, creating 10,000 separate transformed mesh copies simultaneously (~129 KB each = ~1.26 GB total).

On the **PR branch**, `NumVert()` and `GetBoundingBox()` access the data without triggering transform application. The transforms are only applied on-the-fly during the final `Compose` operation, avoiding the memory spike.

## Test plan
- [x] All 46 Boolean tests pass
- [x] Added new test `BatchUnionManyTransformed` that creates 100 transformed cubes and unions them
- [x] Memory benchmark with 10,000 transformed meshes shows 162x memory reduction

Fixes #1493

🤖 Generated with [Claude Code](https://claude.ai/claude-code)

### DIFF
--- a/src/csg_tree.cpp
+++ b/src/csg_tree.cpp
@@ -181,8 +181,7 @@ std::shared_ptr<CsgLeafNode> SimpleBoolean(const Manifold::Impl& a,
     throw err;
   }
 #else
-  auto impl = Boolean3(a, b, op).Result(op);
-  return ImplToLeaf(std::move(impl));
+  return ImplToLeaf(Boolean3(a, b, op).Result(op));
 #endif
 }
 


test/boolean_test.cpp
```diff
@@ -757,3 +757,31 @@ TEST(Boolean, BatchBoolean) {
   EXPECT_FLOAT_EQ(subtract.Volume(), 7226.043);
   EXPECT_FLOAT_EQ(subtract.SurfaceArea(), 14904.597);
 }
+
+// Test batch union with many transformed meshes.
+// This exercises the optimization in MeshCompare that avoids triggering
+// transform application during heap operations (see issue #1493).
+TEST(Boolean, BatchUnionManyTransformed) {
```
pca006132: Is this increasing test coverage or does this fail without the optimization? I think we have other tests that do batch union with transformed meshes.

src/csg_tree.cpp
```diff
@@ -323,7 +354,14 @@ std::shared_ptr<CsgLeafNode> CsgLeafNode::Compose(
     const int offset = i * Manifold::Impl::meshIDCounter_;
 
     for (const auto& pair : node->pImpl_->meshRelation_.meshIDtransform) {
-      combined.meshRelation_.meshIDtransform[pair.first + offset] = pair.second;
+      Manifold::Impl::Relation rel = pair.second;
+      // Apply the node's transform to the mesh relation if not identity.
+      // This is necessary because we may not have called GetImpl() which would
+      // have applied the transform to the mesh relations.
+      if (node->transform_ != mat3x4(la::identity)) {
+        rel.transform = node->transform_ * Mat4(rel.transform);
```
pca006132, line 361: Good catch

pca006132:
In your table, what does `Master (simulated)` mean?

zalo:
Master (simulated) is basically that it ran the super memory test (union 10,000 hulls) just with the original GetImpl lines rather than the new GetImpl avoiding functions.

Would it make sense to ramp up that test to crashing levels?  … wouldn’t want it to take too long 😅 (and it might just hang instead of crashing 💀)

I can remove it otherwise

pca006132:
> Master (simulated) is basically that it ran the super memory test (union 10,000 hulls) just with the original GetImpl lines rather than the new GetImpl avoiding functions.

I see. The word `simulated` makes me a bit worried :P

> Would it make sense to ramp up that test to crashing levels? … wouldn’t want it to take too long 😅 (and it might just hang instead of crashing 💀)

No, but I think that if the test is not failing before and after, and it doesn't exercise more code paths, it is probably not worth having as a test.

I think in general we should have long(er) running performance tests, similar to `perfTest`, that is run in the CI with its own dedicated CI task that can be used to track timing and memory usages. And this test can be part of that. That said, I think that is probably best left as another PR, and I think for this PR you can remove the test for now.

src/csg_tree.cpp
```diff
@@ -104,6 +106,34 @@ std::shared_ptr<CsgNode> CsgLeafNode::Transform(const mat3x4& m) const {
...
+  }
+
+  // Transform all 8 corners and compute the axis-aligned bounding box
+  vec3 corners[8] = {
+      {box.min.x, box.min.y, box.min.z}, {box.max.x, box.min.y, box.min.z},
+      {box.min.x, box.max.y, box.min.z}, {box.max.x, box.max.y, box.min.z},
+      {box.min.x, box.min.y, box.max.z}, {box.max.x, box.min.y, box.max.z},
+      {box.min.x, box.max.y, box.max.z}, {box.max.x, box.max.y, box.max.z}};
+
+  vec3 newMin = transform_ * vec4(corners[0], 1.0);
```
pca006132: can probably use `std::minmax`

zalo: Claude doesn’t seem to think std::minmax can be used on vec3’s for element-wise minmaxing… there’s a method for computing new min/max bounds directly from the transformation matrix which might be a little faster, but the math looks scary and potentially unmaintainable…

zalo: I had it add Arvo’s algorithm for faster transformed AABB’s; it tried using std::minmax, but ran into some issue that broke the tests, so it’s what you see. ¯\\\_(ツ)_/¯

zalo:
I was thinking about a CI Benchmark that would run at regular intervals or on PR commits that would accumulate perf data to a folder in a branch published to Github Pages (potentially in another repo, so as to avoid polluting this one).  That would make it easier to see perf regressions over time…

src/csg_tree.cpp
```diff
@@ -151,7 +181,8 @@ std::shared_ptr<CsgLeafNode> SimpleBoolean(const Manifold::Impl& a,
     throw err;
   }
 #else
-  return ImplToLeaf(Boolean3(a, b, op).Result(op));
+  auto impl = Boolean3(a, b, op).Result(op);
```
pca006132: btw is this change still needed?

pca006132 (approved):
Thanks a lot!

pca006132:
> I was thinking about a CI Benchmark that would run at regular intervals or on PR commits that would accumulate perf data to a folder in a branch published to Github Pages (potentially in another repo, so as to avoid polluting this one). That would make it easier to see perf regressions over time…

Yeah this can be in its own issue.

src/csg_tree.cpp
zalo: (Good catch; this change is a remnant from some debug logic that was added; reverting soon.)

zalo:
Should be all ready.

elalish:
Nice work!

pca006132:
Btw, thinking about it, does this means we can drop the batching code for Minkowski? Need to test with openscad to see how it goes.

zalo:
Not sure yet; could be additional regressions 😅
I’m interested in doing a timing comparison between batching and not once it’s folly harrowed…

PR #1488, opened 2026-01-13
TypeScript Documentation Overhaul, Part II

tonious:
Continued from #1484.

To do: 

- ~~Standardize capitalization for manifold/Manifold and manifoldCAD/ManifoldCAD.~~ Documented here, moved to the [villianious tasklist](https://github.com/elalish/manifold/issues/1460).
- [x] Move ManifoldCAD user documentation to `/docs/jsuser`.
- [x] Move Manifold WASM and ManifoldCAD developer documentation to `/docs/jsapi`.
- [x] Make developer and user docs immediately distinguishable.
- ~~Move C++ documentation to `/docs/api`.  It gets the root api namespace, as it is ground truth~~
  - There are like 500 files in there.... 
- [x] Add redirects for current well known urls (e.g: `/docs/html` -> `/docs/api` and `/jsdocs`/ -> `/docs/jsuser`)
- ~~Add a 404 page.~~ Moved to the [villianious tasklist](https://github.com/elalish/manifold/issues/1460).
- [x] Set favicon in typedoc

Issue #1460, opened 2025-12-17
Villianous Task List to Conquer WASM


tonious:
This is my list of actionable javascript/typescript issues that I have noticed (or caused) while refactoring.  This particular issue has been split from #1459 -- high level strategic discussions should take place over there.

This is not an official list, not a complete list, not of all this needs to be implemented, and some of these ideas are kind of fuzzy.  That said, please feel free to make suggestions, and even more free to make pull requests.  After all, @elalish reviews those, not me :grin:.

Immediate next thoughts:
  - `show()` and `only()` should apply to all objects, including `GLTFNodes` and descendants.
  -  Revamp vscode extension.

#### User Documentation.

 - [x] 'manifold' / 'Manifold' / 'manifoldCAD' disambiguation.  I.e.: manifoldCAD is manifold + typescript + glTF. (#1484, #1488, #1497)
 - [x] Call out specific deviations from convention.  I.e.: glTF is 1m per unit, Y axis up.  manifoldCAD is 1mm per unit, Z axis up. (#1505)
 - [x] Add a direct link to TS documentation from the editor. (#1492)
 - [x] Handling precision issues (From #1135) (#1564, thanks @AnshulPatil2005!)
 - [x] [Degrees vs Radians](https://github.com/elalish/manifold/issues/1262#issuecomment-2913924196) (#1565, thanks @AnshulPatil2005!)
 - [ ] `smoothOut` -> `refine` -> `warp` example to demonstrate remeshing and warping (From #409 and #277)
 - [ ] `importModel` example. (As opposed to `importManifold`).
 - [ ] Example external library.
 - [ ] Example stand-alone model.  I.e.: checkout, `npm install`, `npm build:model` -> `helloworld.glb`.
 - [x] Standardize capitalization for 'manifold' and 'ManifoldCAD'. (From #1488)
 - [ ] Add a 404 page. (From #1488)

#### Developer documentation. 
Now that the API surface is both larger, and actually defined, it needs to be clear what documentation is applicable for a user of manifoldCAD (i.e.: modelling functions, anything in `manifold-3d/manifoldCAD.d.ts`) and what lower level functions can be called from outside the sandbox.
   - [ ] Explain bundling process, adding functions to manifoldCAD, &c.
   - [ ] Using `manifold-3d/manifoldCAD` without the worker.
   - [ ] Instantiating manifold in node and browser environments.
   - [ ] Add architecture diagram (see below)
 
### Refactoring
   - [ ] Centralize logging.

[...]

PR #1497, opened 2026-01-20
TypeScript Documentation Overhaul, Part III

tonious:
Following on #1484, #1488 and comments on #1460:

- [x] For top-level categories: Global State can be removed (seems to be a duplicate). Functions and Modeling functions can be combined and moved to right below Classes. Type Aliases should go at the bottom.
- [x] Actually, that was feedback about the right side panel ("on this page"), but there's also a left side panel that's almost but not quite the same. Should we pick one? Or make them sufficiently different to be useful?
- [x]  Is there a way to hide members? For instance _parent and the VisualizationGLTFNode constructor. Also sourceMaterial and sourceRunID in GLTFMaterial, which should also probably start with _.
  - Flagging a reflection (comment) as `@internal` flag will hide it from the User Guide but not the developer guide.
  - Flagging a reflection as `@hidden` will hide it from both.
- [x] Can we organize the functions under the Manifold class into the same categories as the C++ docs do?
- [x] Could use a top-level description of CrossSection...
- [x] ...and the various members of GLTFMaterial.
- [x] The top of Module ManifoldCAD reads like API documentation - would be better to have an intro to ManifoldCAD.org.
- [x] The top is ManifoldCAD User Guide, and under it are Using ManifoldCAD and manifoldCAD. I think these three could be combined into a single top-level header.

### DIFF
--- a/bindings/wasm/typedoc.api.json
+++ b/bindings/wasm/typedoc.api.json
@@ -8,6 +8,11 @@
     "./manifold.d.ts",
     "./lib/*.ts"
   ],
+  "exclude": [
+    "./lib/scene-builder.ts",
+    "./**/*.test.ts",
+    "./lib/*.d.ts"
+  ],
   "sortEntryPoints": false,
   "projectDocuments": [
     "./documents/embedding-manifoldcad.md",
--- a/scripts/format.sh
+++ b/scripts/format.sh
@@ -13,7 +13,7 @@ $CLANG_FORMAT -i bindings/c/include/manifold/*.h &
 $CLANG_FORMAT -i bindings/wasm/*.{js,ts} &
 $CLANG_FORMAT -i bindings/wasm/lib/*.ts &
 $CLANG_FORMAT -i bindings/wasm/examples/*.{js,ts,html} &
-$CLANG_FORMAT -i bindings/wasm/examples/public/*.{js,ts} &
+$CLANG_FORMAT -i bindings/wasm/examples/public/*.js &
 $CLANG_FORMAT -i bindings/wasm/test/*.ts &
 $CLANG_FORMAT -i bindings/wasm/test/fixtures/*.{ts,mjs} &
 $CLANG_FORMAT -i bindings/wasm/types/*.ts &


bindings/wasm/lib/gltf-node.ts
```diff
@@ -41,19 +41,84 @@ import type {Manifold, Vec3} from '../manifold';
...
+ * or set as a {@link GLTFNode.material | GLTFNode property}.
+ * 
+ * @see {@link https://registry.khronos.org/glTF/specs/2.0/glTF-2.0.html#materials | glTF 2.0 Specification: Materials}
+ * @see {@link https://physicallybased.info/ | Physically Based - The PBR values database}
+ * @sortStrategy source-order
+ */
 export interface GLTFMaterial {
+  /**
+   * Every vertex in a glTF Mesh has a set of attributes.
+   * `POSITION` cannot be specified -- ManifoldCAD will set it internally.
```
elalish, line 66: We should probably mention that these are positional, so e.g. ['TEXCOORD_0', 'NORMAL', 'SKIP_2', 'COLOR_0'] would end up using property channels 3-4 for texture, 5-7 for normal, ignore 8-9, and use 10-12 for color.

bindings/wasm/lib/gltf-node.ts
```diff
@@ -41,19 +41,84 @@ import type {Manifold, Vec3} from '../manifold';
...
+
+  /**
+   * Roughness of the material. 
+   * Ranges from 0 (smooth) to 1.0 (rough).
+   */
   roughness?: number;
+
+  /**
+   * Metallic property of the material.
+   * Ranges from 0 (more diffuse reflection) to 1.0 (more specular reflection).
```
elalish: Nope, diffuse and specular are roughness. Metallic makes it either a conductor (1) or dielectric (0), e.g. gold vs yellow plastic. Generally speaking materials are either one or the other and intermediate values are just for blending.

bindings/wasm/lib/gltf-node.ts
```diff
@@ -41,19 +41,84 @@ import type {Manifold, Vec3} from '../manifold';
...
+   */
   metallic?: number;
+
+  /**
+   * Base colour of the material.
+   * RGB values, ranging from 0 to 1.0.
```
elalish, line 105: Defaults to yellow: [1, 1, 0] - this multiplies `COLOR_0`, so if that vertex property is specified you'll likely want to set the baseColorFactor to white [1, 1, 1].

tonious:
> Are you planning to do the last two checkboxes in this PR?

Yep!  I'm fighting with TypeDoc, but I'm taking the approach of user documentation is more important than developer documentation.

tonious:
Screenshot 2026-01-21 at 11 58 14 am

elalish:
Thanks, that looks like a huge improvement! Looks totally useable, so these are just nits:

1. The right sidebar looks way more useful than the left. Can they switch places? Or hide the left one? The only thing the left one seems to be is a flat index, but it's not alphabetized and I'm pretty sure the search feature will serve that purpose better anyway. 
2. Could we concatenate the contents of the README and the User Guide in place of "These are the objects and functions available within ManifoldCAD itself", and then remove those separate documents?

tonious:
Screenshot 2026-01-21 at 4 58 00 pm

tonious:
Screenshot 2026-01-21 at 5 17 50 pm

bindings/wasm/lib/animation.ts
```diff
@@ -39,49 +40,49 @@ let animationFPS: number = 30;
...
+}
 
 /**
  * Set the animation frame rate.
  *
  * @param fps in frames per second.
- * @group Global Settings
  */
-export const setAnimationFPS = (fps: number): void => {
+export function setAnimationFPS(fps: number): void {
```
elalish, line 76: should setAnimationDuration get the same treatment?

bindings/wasm/types/manifoldCAD.d.ts
```diff
@@ -12,53 +12,92 @@
...
- * @category none
- * @module manifoldCAD
- * @primaryExport
- * @see {@link "Using ManifoldCAD" | Using ManifoldCAD}
+ * {@include ../documents/using-manifoldcad.md}
  *
- * @groupDescription Global State
- * These objects and functions are specific to top-level scripts
- * running within manifoldCAD.
+ * {@include ../README.md#NoStlPlease}
```
elalish: I'd put this section in dev docs instead.

elalish (approved):

bindings/wasm/types/manifoldCAD.d.ts
tonious: 👍 The dev docs have the full readme in them.

tonious:
Writing this documentation up has made me rethink the whole global settings thing :sweat:

But that'd be another PR.

PR #1502, opened 2026-01-23
Add transformation methods to GLTFNodes.

tonious:
This PR is intended to give `GLTFNode` type objects the same transformation API as `Manifold` objects.

This involves a breaking change: The current `scale` property has been renamed `scaling` in order to accommodate a `scale` method.

- [x] Add transformation methods.
- [ ] Should throw if an animation function has already been applied.
- [ ] Documentation.
    - Note that these transformations are applied to the GLTFNode immediately, but to the model at render time.  There's no CSG tree waiting in memory.
- [ ] Call signatures must match `Manifold` objects.
- [ ] Tests.

Part of the [Villianous Task List](https://github.com/elalish/manifold/issues/1460).

### DIFF
--- a/bindings/wasm/lib/gltf-node.ts
+++ b/bindings/wasm/lib/gltf-node.ts
@@ -36,8 +36,12 @@
  */
 
 import type * as GLTFTransform from '@gltf-transform/core';
+import {MathUtils} from '@gltf-transform/core';
 
-import type {Manifold, Vec3} from '../manifold.d.ts';
+import type {Manifold, Mat4, Vec3} from '../manifold.d.ts';
+
+import type {Quat} from './math.ts';
+import {euler2quat, identityMat4, mirrorMat4, multiplyMat4, quat2euler, rotateMat4, scaleMat4, translateMat4} from './math.ts';
 
 const nodes = new Array<BaseGLTFNode>();
 
@@ -154,18 +158,25 @@ export interface GLTFMaterial {
  * The abstract class from which other classes inherit.  Common methods and
  * properties live here.
  * @group Scene Graph
+ *
+ * @privateRemarks
+ * Internally, gltf-transform stores transformations as separate translation,
+ * rotation and scale vectors.  It can convert those vectors to and from a
+ * transformation matrix as needed.  Translation, rotation and scale will be
+ * applied in that order.
  */
 export abstract class BaseGLTFNode {
-  /** @internal */
-  _parent?: BaseGLTFNode;
-  name?: string;
+  private _parent?: BaseGLTFNode;
 
-  // Internally, gltf-transform stores transformations as separate translation,
-  // rotation and scale vectors.  It can convert those vectors to and from a
-  // transformation matrix as needed.
-  translation?: Vec3|((t: number) => Vec3);
-  rotation?: Vec3|((t: number) => Vec3);
-  scale?: Vec3|((t: number) => Vec3);
+  private _scaling: Vec3 = [1.0, 1.0, 1.0];
+  private _rotation: Vec3 = [0.0, 0.0, 0.0];
+  private _translation: Vec3 = [0.0, 0.0, 0.0];
+
+  private _scalingFn: ((t: number) => Vec3)|null = null;
+  private _rotationFn: ((t: number) => Vec3)|null = null;
+  private _translationFn: ((t: number) => Vec3)|null = null;
+
+  name?: string;
 
   constructor(parent?: BaseGLTFNode) {
     this._parent = parent;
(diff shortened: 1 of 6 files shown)


bindings/wasm/lib/gltf-node.ts
```diff
@@ -174,6 +185,81 @@ export abstract class BaseGLTFNode {
...
+    return this;
+  }
+
+  rotate(v: readonly[number, number, number]) {
+    this.transform(rotateMat4(v[0], v[1], v[2]));
+    return this;
+  }
+
+  scale(v: readonly[number, number, number]) {
+    this.transform(scaleMat4(v[0], v[1], v[2]));
```
elalish, line 255: Okay, so I'm remembering some of the gotchas that makes this all difficult. The glTF spec (for better or worse) does not allow general transforms, only TRS, which means no shear. If you concatenate TRS matrices, you don't in general still have a TRS. 

However, you can concatenate all the TRs you want and it's still a TR, and I think that's the primary use case anyway. 

How about this: scale is special - remove this function and say scale can only be set as a singular property as before. It gets applied first, so no problem.

bindings/wasm/lib/gltf-node.ts
```diff
@@ -174,6 +185,81 @@ export abstract class BaseGLTFNode {
...
+      this._translation = v;
+    } else if (typeof v === 'function') {
+      this._translationFn = v;
+    } else {
+      throw new TypeError('Must be a Vec3 or a function that returns a Vec3');
+    }
+  }
+
+  get rotation() {
+    return this._rotationFn ?? this._rotation;
```
elalish, line 204: This makes me sad because the animation API doesn't get to benefit from your extension here. What if instead the `rotate()` and `translate()` functions still take either values or a function? Then we just keep a single list of these operations (you could compose adjacent constant ones, but that's just an optimization). 

Then you'd just need to refactor the `addMotion` function. Scaling could stay the same, but then you'd run through the list of TR ops, evaluate at each time step and compose, then decompose back into a single TR to save in the glTF.

bindings/wasm/lib/gltf-node.ts
```diff
@@ -174,6 +185,81 @@ export abstract class BaseGLTFNode {
...
+        oldmatrix);
+    const newmatrix = multiplyMat4(transform, oldmatrix);
+
+    const quat: Quat = [0, 0, 0, 0];
+    decompose(newmatrix, this._translation, quat, this._scaling);
+    this.rotation = quat2euler(quat);
+  }
+
+  mirror(normal: readonly[number, number, number]) {
+    this.transform(mirrorMat4(normal as Vec3));
```
elalish, line 245: We'll have to drop this function too, for the same reason as scale.

bindings/wasm/lib/gltf-node.ts
```diff
@@ -174,6 +185,81 @@ export abstract class BaseGLTFNode {
...
+    if (Array.isArray(v)) {
+      this._scaling = v;
+    } else if (typeof v === 'function') {
+      this._scalingFn = v;
+    } else {
+      throw new TypeError('Must be a Vec3 or a function that returns a Vec3');
+    }
+  }
+
+  transform(transform: Mat4) {
```
elalish, line 231: This can't be a public function, since glTF doesn't allow general 4x4 transforms.

bindings/wasm/lib/math.ts
```diff
@@ -33,11 +34,155 @@ export function euler2quat(rotation: Vec3): Vec4 {
...
+  return result;
+}
+
+/**
+ * Create a rotation matrix, in ZYX order.
+ * @group Transformation Matrix
+ */
+export function rotateMat4(x: number, y: number, z: number): Mat4 {
+  const result = identityMat4();
+  const sx = sin(x * TAU / 360);
```
elalish, line 126: nit - can we please just use PI? That particular MIT professor kind of annoys me with his TAU manifesto... Plus, do I see 0.5 * TAU up in this file??

tonious: Fair.  I like it because I can think in whole revolutions, but it's not important at all.

bindings/wasm/lib/gltf-node.ts
tonious: > How about this: scale is special - remove this function and say scale can only be set as a singular property as before. It gets applied first, so no problem.

That's a reasonable compromise.

tonious:
> This makes me sad because the animation API doesn't get to benefit from your extension here. What if instead the rotate() and translate() functions still take either values or a function? Then we just keep a single list of these operations (you could compose adjacent constant ones, but that's just an optimization).

> Then you'd just need to refactor the addMotion function. Scaling could stay the same, but then you'd run through the list of TR ops, evaluate at each time step and compose, then decompose back into a single TR to save in the glTF.

Yeah, this is the right way to do it.  I think I need to meditate upon it for a bit.  I'm going to withdraw this PR, and re-open it later.

PR #1505, opened 2026-01-24
Rotate glTF models on import.

tonious:
GLTF defines 'up' as +Y, while ManifoldCAD defines it as '+Z'.

This PR rotates models on import so that up is up.  It also documents that behaviour.

Part of the [Villianous Task List](#1460).

(Smaller than usual 😅 )

### DIFF
--- a/bindings/wasm/lib/gltf-node.ts
+++ b/bindings/wasm/lib/gltf-node.ts
@@ -164,6 +164,12 @@ export abstract class BaseGLTFNode {
   // rotation and scale vectors.  It can convert those vectors to and from a
   // transformation matrix as needed.
   translation?: Vec3|((t: number) => Vec3);
+
+  /**
+   * Rotations are applied in `XYZ` order.
+   * That is roll first, then pitch and finally yaw.
+   * This matches the behaviour of `Manifold.rotate()`.
+   */
   rotation?: Vec3|((t: number) => Vec3);
   scale?: Vec3|((t: number) => Vec3);
 
--- a/bindings/wasm/lib/math.ts
+++ b/bindings/wasm/lib/math.ts
@@ -21,32 +21,48 @@ import type {Vec3} from '../manifold.d.ts';
 
 const {cos, sin, PI} = Math;
 
+/**
+ * A quaternion in `XYZW` order.
+ *
+ * ManifoldCAD, like OpenGL and many other libraries represent quaternions in
+ * `XYZW` order, with the scalar last. This differs from most math textbooks
+ * (and just as many other libraries) that use `WXYZ` order.
+ */
 export type Quat = [number, number, number, number];
 export type Vec4 = Quat;
 
+/**
+ * Convert Euler angles to a quaternion.
+ *
+ * Rotations are applied in `XYZ` order.
+ * That is roll first, then pitch and finally yaw.
+ * This matches the behaviour of `Manifold.rotate()`.
+ *
+ * @param rotation Degrees, in `XYZ` format.
+ */
 export function euler2quat(rotation: Vec3): Quat {
-  // https://en.wikipedia.org/wiki/Conversion_between_quaternions_and_Euler_angles#Source_code
-  const cr = cos(rotation[2] * PI / 360);
-  const sr = sin(rotation[2] * PI / 360);
-  const cp = cos(rotation[1] * PI / 360);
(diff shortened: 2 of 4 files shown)


bindings/wasm/lib/import-model.ts
```diff
@@ -329,16 +330,23 @@ export async function readFile(filename: string, options: ImportOptions = {}) {
...
+  // Find top level nodes and correct their scale and orientation.
   for (const sourceNode of doc.getRoot().listNodes()) {
     if (sourceNode.getParentNode()) continue;
 
     const scale = sourceNode.getScale();
     sourceNode.setScale([scale[0] * 1000, scale[1] * 1000, scale[2] * 1000]);
+
+    const halfRoot2 = Math.sqrt(2) / 2;
+    const original = sourceNode.getRotation();
+    const rotated = multiplyQuat(original, [-halfRoot2, 0, 0, halfRoot2])
```
elalish: I might just use euler2quat here for readability.

bindings/wasm/lib/import-model.test.ts
```diff
@@ -118,4 +119,28 @@ suite('importModel()', () => {
...
+    const model = await importer.importManifold(buffer, options);
+
+    // Count the number of vertices above the Z=0 plane.
+    let above = 0;
+    const mesh = model.getMesh();
+    for (let i = 0; i < mesh.numVert; i++) {
+      if (mesh.position(i)[2] > 0) above++;
+    }
+
+    expect(above).toEqual(1);
```
elalish: The only trouble with this test is it doesn't check the X and Y axes. We should document which axis we rotate about, and ensure it comes back to the original after an export/import cycle.

tonious: That's fair.

I've called out the rotation and scaling specifically at both import and export.
Now the test uses an [object](https://manifoldcad.org/#Test%20Object%3Ccode%3Eimport%20%7BManifold%7D%20from%20'manifold-3d%2FmanifoldCAD'%3B%0Aexport%20default%20Manifold.hull(%5B%5B0%2C0%2C5%5D%2C%5B-1%2C0%2C0%5D%2C%5B4%2C0%2C0%5D%2C%5B0%2C-2%2C0%5D%2C%5B0%2C3%2C0%5D%5D)%3B) that has no rotational or mirrored symmetry.

bindings/wasm/lib/import-model.ts
```diff
@@ -329,16 +330,22 @@ export async function readFile(filename: string, options: ImportOptions = {}) {
...
-  // Find top level nodes and correct their scale.
+  // Find top level nodes and correct their scale and orientation.
   for (const sourceNode of doc.getRoot().listNodes()) {
     if (sourceNode.getParentNode()) continue;
 
     const scale = sourceNode.getScale();
     sourceNode.setScale([scale[0] * 1000, scale[1] * 1000, scale[2] * 1000]);
+
+    const original = sourceNode.getRotation();
+    const rotated = multiplyQuat(original, euler2quat([90, 0, 0]));
```
elalish, line 347: Can you explain why this round-trip works when it looks like the same rotation is applied on import and export, rather than opposites? Your test seems to imply it's all working, but I just want to double check.

tonious: Y'know, I thought it was because multiplying quaternions is not transitive.  But flipping the sign and reversing didn't work, so I've got some deeper flaw here.

It should be the inverse of `exportTransform` just to adhere to the principle of least astonishment.

tonious: It's a quaternion `WXYZ` vs `XYZW` thing.  Time for more documentation.  (Ouch! That hurt me as a developer! 😁 )

But while I'm at it, `euler2quat` currently works in `ZYX` order, while `Manifold.rotate()` works in `XYZ` order.  Should I change this to match?  I'll have to mess with the Tetrahedron Puzzle.

elalish: Ouch, yeah euler2quat should definitely match Manifold.rotate. Why will that change tet puzzle? Were we actually doing the rotations wrong compared to our documentation? EDIT: ah, indeed. You know, I think that actually confused me when I was writing that example, but I just messed around until it worked instead of noticing the actual problem.

PR #1507, opened 2026-01-29
Make CrossSections visible.

tonious:
Does what it says on the tin.

- [x] Add a `CrossSectionGLTFNode` class.
- [x] Create glTF-transform meshes from cross sections.
  - Using `triangulate` with double sided materials.
  - Exported cross sections are not manifold and can be easily filtered out.
- [x] Scripts can return one or more `CrossSection` objects the same way they return `Manifold` objects.
- [x] Apply materials to cross section nodes.
- [x] Add an example (still in progress)
- [x] Tests

Part of the [Villianous Task List](#1460), first mentioned in #1392.

### DIFF
--- a/bindings/wasm/test/examples/godel-escher-bach.ts
+++ b/bindings/wasm/test/examples/godel-escher-bach.ts
@@ -7,35 +7,37 @@
 // manifold objects. When exporting to glTF, CrossSections will be visible, but
 // will not have `EXT_mesh_manifold` applied.
 
-import {Font, getGlyphPath, pathToSVG} from 'text-shaper';
+import {CrossSection, CrossSectionGLTFNode, GLTFNode, Manifold, Polygons, Vec3} from 'manifold-3d/manifoldCAD';
 import {pointsOnPath} from 'points-on-path';
-import {CrossSection, Manifold, CrossSectionGLTFNode, GLTFNode, Polygons, Vec3} from 'manifold-3d/manifoldCAD';
+import {Font, getGlyphPath, pathToSVG} from 'text-shaper';
 
-const fonturl = 'https://github.com/theleagueof/orbitron/raw/refs/heads/master/webfonts/orbitron-black-webfont.ttf';
+const fonturl =
+    'https://github.com/theleagueof/orbitron/raw/refs/heads/master/webfonts/orbitron-black-webfont.ttf';
 
 // Return a function that generates a CrossSection for a given letter.
 // Getting the function is asynchronous as it depends on a fetch, but
 // the function itself is synchronous.
-const characterGenerator = async (fonturl:string, height=100) => {
+const characterGenerator =
+    async (fonturl: string, height = 100) => {
   const font = await Font.fromURL(fonturl);
 
-  // Helper function to center and scale a CrossSection
-  // This particular demo looks best if characters are square, so ignore aspect ratio.
-  const resize = (cs:CrossSection) => {
+  // Helper function to center and scale a CrossSection.
+  // This particular demo looks best if characters are square, so ignore aspect
+  // ratio.
+  const resize = (cs: CrossSection) => {
     const {min, max} = cs.bounds();
-    return cs
-        .translate([-(min[0]+max[0])/2,-(min[1]+max[1])/2])
-        .scale([height/(max[0]-min[0]),height/(max[1]-min[1])]);
+    return cs.translate([-(min[0] + max[0]) / 2, -(min[1] + max[1]) / 2])
+        .scale([height / (max[0] - min[0]), height / (max[1] - min[1])]);
   };
 
-  return (char:string): CrossSection => {
+  return (char: string): CrossSection => {
     const glyphpath = getGlyphPath(font, font.glyphIdForChar(char));
     const svgpath = pathToSVG(glyphpath, {flipY: true, scale: 1});
     const cs = new CrossSection(pointsOnPath(svgpath) as Polygons);
-    return resize(cs).mirror([0,1]);
+    return resize(cs).mirror([0, 1]);
   }
 }
- 
+
 export default async () => {
   const height = 100;
   const offset = height / 2;
@@ -82,13 +84,13 @@ export default async () => {
   // Time for some Constructive Solid Geometry.
   // Extrude each CrossSection and apply the same rotation and translation.
   // Then compute the intersection.
-  const extrude = (node:CrossSectionGLTFNode) => {
+  const extrude = (node: CrossSectionGLTFNode) => {
     let extrusion = node.crossSection!.extrude(4 * height);
     extrusion = extrusion.rotate(node.rotation as Vec3);
     extrusion = extrusion.translate(node.translation as Vec3);
     return extrusion;
-  }
-  const intersect = ((acc:Manifold, cur:Manifold) => acc.intersect(cur))
+  };
+  const intersect = ((acc: Manifold, cur: Manifold) => acc.intersect(cur));
   const intersection = csNodes.map(extrude).reduce(intersect);
 
   // Put the result into a node so it can be oriented


elalish:
I wouldn't worry too much about materials yet - we don't even have properties support in Manifold CrossSections yet. That will be a bigger change. Let me know when you're ready for review.

elalish:
Also, it looks like npm is rate limiting our CI: https://github.com/elalish/manifold/actions/runs/21511230999/job/61978502259

Are we really hitting it that much? I'm guessing this must be a fairly common issue for Github Actions, right? Ideas?

pca006132:
I think it may not be related to us, but the action runner, because they may be doing this by IP.

It seems that if we use https://github.com/actions/setup-node, we get caching by default, which may avoid this rate limit?

tonious:
> Also, it looks like npm is rate limiting our CI

I've seen this locally, too.  It's not an install issue but a worker test that is triggering it.  I wonder if `esm.sh` is getting rate limited by NPM, or if we're getting rate limited by `esm.sh`.  Either way, I'll open another PR to skip that test.  If it's flakey, it's no good and jsDelivr (our default) seems fine.

> I wouldn't worry too much about materials yet - we don't even have properties support in Manifold CrossSections yet.

Oh, I wasn't going to get that fancy.  The material set on `CrossSectionGLTFNode` isn't wired up yet.  It'll apply to the whole cross section.

In other news, I might have messed up rotation stuff from #1505.  Here's [an example of differing rotation](https://manifoldcad.org/#Rotation%20Test%3Ccode%3Eimport%20%7BManifold%2C%20GLTFNode%7D%20from%20'manifold-3d%2FmanifoldCAD'%3B%0A%0Aconst%20shape%20%3D%20Manifold.hull(%5B%5B0%2C0%2C5%5D%2C%5B-1%2C0%2C0%5D%2C%5B4%2C0%2C0%5D%2C%5B0%2C-2%2C0%5D%2C%5B0%2C3%2C0%5D%5D)%3B%0A%0Aconst%20node%20%3D%20new%20GLTFNode()%3B%0Anode.manifold%20%3D%20shape%3B%0Anode.material%20%3D%20%7B%20baseColorFactor%3A%20%5B0%2C1%2C1%5D%7D%0Anode.rotation%20%3D%20%5B90%2C0%2C90%5D%3B%0A%0Aconst%20manifold%20%3D%20shape.rotate(node.rotation)%3B%0A%0Aexport%20default%20%5Bmanifold%2C%20node%5D%3B).  I'll have to fix that first to make my example work here :grin:

elalish:
Is this PR ready for review?

tonious:
Alas, my day job has gotten in the way this past week.  It's pretty close, though.

tonious:
Screenshot 2026-02-11 at 2 21 23 pm

Unlit materials work nicely here!

bindings/wasm/lib/gltf-node.ts
```diff
@@ -243,16 +264,48 @@ export class VisualizationGLTFNode extends BaseGLTFNode {
   uri?: string;
 
   constructor(
-      document: GLTFTransform.Document, node?: GLTFTransform.Node,
+      document: GLTFTransform.Document, node?: GLTFTransform.Node|null,
```
elalish: Is there a reason we need to accept `null` instead of just `undefined`? I tend to prefer just using undefined for simplicity.

bindings/wasm/lib/gltf-node.ts
```diff
@@ -243,16 +264,48 @@ export class VisualizationGLTFNode extends BaseGLTFNode {
...
+/**
+ * Display a CrossSection in 3D space.
+ *
+ * A CrossSection object is two dimensional.  Attaching it as a node
+ * allows it to be included in the final exported file, complete with
+ * transformations.
+ *
+ * > [!NOTICE]
+ * >
+ * > CrossSections are not -- and can never be -- manifold.  That means
```
elalish, line 291: We could make them manifold by doubling them up. Not sure that would be useful though...

bindings/wasm/lib/scene-builder.ts
```diff
@@ -256,6 +257,62 @@ function exportTransform(doc: Document) {
...
+  const {triangulate} = getManifoldModuleSync()!;
+  const polygons = cs.toPolygons();
+  const positions = new Float32Array(
+      polygons.flat().map((v: Vec2): Vec3 => ([...v, 0])).flat());
+  const triangles = triangulate(polygons);
+  const indices = new Uint32Array(triangles.flat());
+
+  log(`Exporting CrossSection ${
+      nodeDef.name ? `"${nodeDef.name}"` : 'object'} as mesh:`);
+  log(`  Triangles: ${triangles.length.toLocaleString()}`);
```
elalish: Is it possible that from about here down this code could be a shared function with createNodeFromManifold?

bindings/wasm/lib/scene-builder.ts
```diff
@@ -277,38 +334,56 @@ export async function manifoldToGLTFDoc(manifold: Manifold) {
...
-
   const doc = new Document();
   const root = exportTransform(doc);
 
   addAnimationToDoc(doc);
 
+  const uniqueNodes: Set<BaseGLTFNode> = new Set();
+  const recurse = (node: BaseGLTFNode) => {
+    uniqueNodes.add(node);
+    for (const child of node.listChildren()) {
```
elalish: Some bigger changes here - probably fine, but can you give me a little explanation? We used to not need the sets of children or to search the tree in this direction before.

bindings/wasm/test/examples/godel-escher-bach.ts
```diff
@@ -0,0 +1,122 @@
...
+// manifold objects. When exporting to glTF, CrossSections will be visible, but
+// will not have `EXT_mesh_manifold` applied.
+
+import {CrossSection, CrossSectionGLTFNode, GLTFNode, Vec3} from 'manifold-3d/manifoldCAD';
+
+const {circle, square, union, hull} = CrossSection;
+
+/**
+ * There are lots of ways to represent letters.
+ * This is probably not the best way, but it only really needs to do three
```
elalish: Impressive, but might this be a nice chance to show off your import tools? There's got to be a JS font library somewhere, right?

bindings/wasm/test/examples/godel-escher-bach.ts
```diff
@@ -0,0 +1,122 @@
+// This example recreats the cover of 'Gödel, Escher, Bach: an Eternal Golden
+// Braid' by Douglas Hofstadter.  It demonstrates the use of CrossSections and
+// CrossSectionGLTFNodes to incorporate two dimensional drawings in a model.
```
elalish, line 3: 💯

bindings/wasm/lib/gltf-node.ts
tonious: I guess that would be technically manifold, but zero volume.

IMHO, if a user wants a thin manifold object from a cross section, they should have to explicitly make it so.

bindings/wasm/lib/scene-builder.ts
tonious: Right.  In order to make the GEB demo work nicely, I brought in a few more tree functions -- `addChild()`, `listChildren()` -- and now can recurse down the scene graph.  This can lead to duplication, in cases like:

```typescript
const parent = new GLTFNode();
const child = new GLTFNode(parent);

export default [parent, child];
```

This would result in a node list of `[parent, child, child]`.   Throw in `GLTFNodeTracked`, and there are potentially more duplicates.

It seemed trivial to handle this silently.

bindings/wasm/test/examples/godel-escher-bach.ts
tonious: I've got some font code around.  Just need to find an available font that is chunky enough to look good in this demo.

bindings/wasm/lib/gltf-node.ts
tonious: This is an artifact of having `GLTFNode` subclasses with different constructor arguments.  I think I need to refactor that for consistency.

bindings/wasm/test/examples/godel-escher-bach.ts
```diff
@@ -0,0 +1,122 @@
...
+
+  // Time for some Constructive Solid Geometry.
+  // Extrude each CrossSection, and orient it in 3D space,
+  // then compute the intersection.
+  const intersection = nodes
+                           .map((node: CrossSectionGLTFNode) => {
+                             const extrusion =
+                                 node.crossSection!.extrude(4 * height);
+                             return extrusion.rotate(node.rotation as Vec3)
+                                 .translate(node.translation as Vec3);
```
tonious: I prototyped adding helpers for `extrude()` and `revolve()` on `CrossSectionGLTFNode()` that would perform the operation and then transform the result into the same space.  Then promptly banged my nose on two things:

* The exact same animation constraints discussed in #1502.
* If I apply transformations from this node, am I in the reference frame the user expects?  What if this frame has a parent?  Or a grandparent?

I'll add this thought to the Villianous Task List later, but thought I should mention it.

bindings/wasm/lib/scene-builder.ts
tonious: This is why I like working on this project.  When I was first hacking this up, I thought about that and "Naaah, too much special purpose code, too tricky".  And then during code review, I can suddenly see it :grin:

bindings/wasm/test/examples/godel-escher-bach.ts
elalish: I was actually going to ask about #1502 - I still really like that one. I was hoping my suggestion wasn't too much work, but it seems that it derailed the PR. Anything I can do to help?

Is there a reason we want to perform operations on a `CrossSectionGLTFNode`? Why not just use the actual `CrossSection`? I think that'll be less confusing and keep our API surface area down a bit.

bindings/wasm/lib/scene-builder.ts
elalish: I'm struggling to follow this explanation for some reason - maybe it's just late. Anyway, I trust you. Still, it can be helpful in situations like this to file an issue describing the problem and then mark this as fixing it, since it appears to be an existing problem beyond the immediate scope.

tonious: That's fair.  I'll pull this stuff out as it's tangential.

bindings/wasm/test/examples/godel-escher-bach.ts
tonious: I think you're right about keeping the API surface down, and ultimately I threw out the prototype.

It'd just be nice if there's a really simple way to take the transforms on the `CrossSectionGLTFNode` and apply them to the manifolds created by extruding or revolving its `CrossSection`.

```typescript
// Right now it's kinda verbose, and depends on applying transforms in the correct order.
let manifold = (csNode.crossSection!.extrude(10))
    .scale(csNode.scale)
    .rotate(csNode.rotation)
    .translate(csNode.translation);

// It could be more terse with matrix functions.
// This is essentially #1502.
manifold = (csNode.crossSection!.extrude(10))
    .transform(csNode.getMatrix());

// Even more terse if we have a helper that applies those transforms internally.
manifold = csNode.extrude(10);
```

I don't like that the last approach hides the transform.  IMHO this is a place where the user should understand what transformations are in play.

Anyhow, I'm still thinking about #1502.  I want to retain callbacks, but I also want the function signatures to match `Manifold`.  Leaving that thought aside, a method like `getMatrix(time:number = 0)` would make a natural starting place.

elalish:
Is this ready for review? I notice it's still marked as draft.

tonious:
It could probably stand another review.  I haven't redone the demo yet.

elalish:
Hmm, @tonious I think there's some kind of CORS issue with the new example now that it's deployed? 
```
Access to fetch at 'https://github.com/theleagueof/orbitron/raw/refs/heads/master/webfonts/orbitron-black-webfont.ttf' 
from origin 'https://manifoldcad.org' has been blocked by CORS policy: The 'Access-Control-Allow-Origin' header 
contains the invalid value ''. Have the server send the header with a valid value.Understand this error

worker.bundled-CZLECyfV.js:770 TypeError: Failed to fetch
    at zm (worker.bundled-CZLECyfV.js:770:16067)
    at async s (worker.bundled-CZLECyfV.js:770:17838)Caused by: TypeError: Failed to fetch
    at _Hy.fromURL (eval at zm (worker.bundled-CZLECyfV.js:770:15495), <anonymous>:13995:20)
    at characterGenerator (eval at zm (worker.bundled-CZLECyfV.js:770:15495), <anonymous>:19452:25)
    at G_del_Escher_Bach_default (eval at zm (worker.bundled-CZLECyfV.js:770:15495), <anonymous>:19467:22)
    at zm (worker.bundled-CZLECyfV.js:770:15604)
    at async s (worker.bundled-CZLECyfV.js:770:17838)
t @ worker.bundled-CZLECyfV.js:770Understand this error
worker.bundled-CZLECyfV.js:770 Caused by: TypeError: Failed to fetch
    at _Hy.fromURL (eval at zm (worker.bundled-CZLECyfV.js:770:15495), <anonymous>:13995:20)
    at characterGenerator (eval at zm (worker.bundled-CZLECyfV.js:770:15495), <anonymous>:19452:25)
    at G_del_Escher_Bach_default (eval at zm (worker.bundled-CZLECyfV.js:770:15495), <anonymous>:19467:22)
    at zm (worker.bundled-CZLECyfV.js:770:15604)
    at async s (worker.bundled-CZLECyfV.js:770:17838)
t @ worker.bundled-CZLECyfV.js:770Understand this error
worker.bundled-CZLECyfV.js:770 manifoldStack: TypeError: Failed to fetch
    at _Hy.fromURL (http-url:https:/cdn.jsdelivr.net/npm/text-shaper/+esm:7:337768)
    at characterGenerator (:22:27)
    at G_del_Escher_Bach_default (:44:22)
```

elalish:
This is great, but what we really need now is to hook up `show()` and `only()` so they work with CrossSections too, or maybe even plain polygons.

elalish:
@tonious Not sure if it was this PR or not, but our favorite old bug is back: once the imported modules load and the red squiggles disappear, it auto-saves and the example becomes "New Script". I wonder if there's a way to make a test for this...

tonious:
:poop:!
Unfortunately, this happens in the browser and depends on timing between Monaco, monaco-auto-typings and jsDelivr or local storage.  I'll have to try some defensive coding.

PR #1521, opened 2026-02-05
Doc improve extrude

coryrc:
Choose some self-documenting variable names. Don't mutate. Add a few
comments about the rest.

Includes my other PR because I was using it to test.

### DIFF
--- a/bindings/wasm/lib/worker.test.ts
+++ b/bindings/wasm/lib/worker.test.ts
@@ -84,7 +84,7 @@ suite('Examples', () => {
     const result = await runExample('Tetrahedron Puzzle');
     expect(result?.genus).to.equal(0, 'Genus');
     expect(result?.volume).to.be.closeTo(7223, 1, 'Volume');
-    expect(result?.surfaceArea).to.be.closeTo(3235, 1, 'Surface Area');
+    expect(result?.surfaceArea).to.be.closeTo(3233, 1, 'Surface Area');
   });
 
   test('Rounded Frame', async () => {


coryrc:
I want to do some visual comparisons of some shapes before I believe the comment.

src/constructors.cpp
```diff
@@ -270,18 +270,39 @@ Manifold Manifold::Extrude(const Polygons& crossSection, double height,
...
+            lastVertPos = vertPos[lastVert];
+          } else {
+            vec2 pos = transform * poly.back();
+            lastVertPos = {pos.x, pos.y, height * alpha};
+          }
+
+          // OpenSCAD chooses the shorter diagonal because it "works well in
+          // most cases".
+          if (la::distance2(vertPos[thisVert], vertPos[lastVertLower]) >=
+              la::distance2(lastVertPos, vertPos[thisVertLower])) {
```
elalish, line 299: Good call, but can we skip this check for the common case of zero rotation? Also, I believe this check should always yield the same result for the same polygon index of each slice, so perhaps it can be calculated just once for the polygon?

elalish (commented):
Good idea - out of curiosity, what does "Doc" refer to in your PR's name? I'll review in earnest when it's not marked draft.

elalish:
Are you still working on this?

coryrc:
Yes, sorry, I will get back to this but it'll be a while. I'm deep in trying to get extrusion to stay within a specified error tolerance. Some of that code might make parts of this redundant, well, not that there's a lot here. I still need to check visually whether this is the right thing.

PR #1528, opened 2026-02-12
Make keyholing more robust in triangulator

### LINKED ISSUE #1527: bug in keyhole algorithm
Downstream issue: https://github.com/openscad/openscad/issues/6607

Polygon:

```python
show(array([
  [10, 10],
  [-10, 10],
  [-10, -10],
  [10, -10]
]))
show(array([
  [-9, -9],
  [9, 0],
  [-9, 9],
  [9, 9],
  [9, -9]
]))
```

Image

Incorrect keyhole result:

```python
show(array([
  [10, 10],# 0, cost: 0.89553347118899
  [-10, 10],# 1, cost: 1
  [-10, -10],# 2, cost: 1
  [10, -10],# 3, cost: 1
  [10, 10],# 0, cost: -0.643564356445644
  [9, 0],# 5, cost: 1
  [-9, 9],# 6, cost: 1
  [9, 9],# 7, cost: 1
  [9, -9],# 8, cost: 1
  [-9, -9],# 4, cost: 1
  [9, 0],# 5, cost: 0
  [10, 10],# 0
]))
```

Image

Basically, when the polygon is self-intersecting (vertex touching an edge), the keyhole algorithm may poke through the intersected edge, causing issues.

elalish:
Fixes #1527 

I had a bunch of code to carefully determine the right point on the outer contour to keyhole to, but the logic for what point to start at on the hole was much too basic.

### DIFF
--- a/src/polygon.cpp
+++ b/src/polygon.cpp
@@ -14,7 +14,6 @@
 
 #include "manifold/polygon.h"
 
-#include <functional>
 #include <map>
 #include <set>
 
@@ -661,7 +660,9 @@ class EarClip {
   }
 
   // Find the actual rightmost starts after degenerate removal. Also calculate
-  // the polygon bounding boxes.
+  // the polygon bounding boxes. Only holes need rightmost starts, and the CCW
+  // check is to handle degenerate cases for them - the other contour starts are
+  // arbitrary.
   void FindStart(VertItr first) {
     const vec2 origin = first->pos;
 
@@ -680,7 +681,7 @@ class EarClip {
       areaCompensation += (area - t1) + area1;
       area = t1;
 
-      if (v->pos.x > maxX) {
+      if (v->pos.x > maxX && v->IsReflex(epsilon_)) {
         maxX = v->pos.x;
         start = v;
       }


src/polygon.cpp
```diff
@@ -680,7 +681,7 @@ class EarClip {
       areaCompensation += (area - t1) + area1;
       area = t1;
 
-      if (v->pos.x > maxX) {
+      if (v->pos.x > maxX && v->IsReflex(epsilon_)) {
```
pca006132, line 684: Is this sufficient? If we consider epsilon-valid polygons, the edge can go through another edge, do a bunch of complex stuff (within epsilon tolerance), and come back. This will not catch that.

elalish: It will actually - read the comment on the `IsReflex` method. Anything within epsilon doesn't count and it'll keep walking the polygon until a beyond-epsilon shape is found. Not saying this will be the last bug, but I think this'll make it significantly more robust.

pca006132: Ah I see. Makes sense.

Issue #1524, opened 2026-02-07
Remove MeshIO and update docs on what to do in its absence for C++


coryrc:
MeshIO should be removed from this repo because debugging output to OBJ is available.

Work items:

1. Update f3d_skeleton to use included OBJ export
2. Decide whether to preserve MeshIO. If so, move the code to a separate repo. If not, delete it.
3. Update [the docs for the debug OBJ export](https://github.com/elalish/manifold/blob/0bfc0d64f8b0ca51f76e48b0c4fc56d5cc3d22d2/include/manifold/manifold.h#L435) to make it clear what it is and is not good for
4. The [README suggests using MeshIO as a starting point](https://github.com/elalish/manifold/blob/0bfc0d64f8b0ca51f76e48b0c4fc56d5cc3d22d2/README.md?plain=1#L93). This will need to be updated. Would like it if has suggestions for how to export from C++ as a starting point, even if that's just "use the debug OBJ and use other utilities to convert if you need other formats"

~~I [got linker errors](https://github.com/elalish/manifold/pull/1523#issuecomment-3864365244) because `MANIFOLD_EXPORT` in cmake wasn't set. I should have got a compiler error.~~

elalish:
Do you have a proposal for how you'd like to change it?

coryrc:
#ifdef guard the contents of the header and cpp (then it doesn't need cmake to add or remove the file either)

elalish:
I think the idea was just to separate out MeshIO as much as possible, since it's not a core part of the library (and not very good), but since everyone wants file IO, it could easily be mistaken for "the right way", which it's not. I'm actually tempted to remove it entirely now that we have a basic OBJ I/O function for debug. @pca006132 what do you think?

pca006132:
I am fine with removing that. I think most of our users have other ways of doing IO.

starseeker:
Just my 2 cents - for our part I fully switched to using the built-in OBJ I/O for debugging, since we need its precision preservation in output to properly reproduce any problems.

coryrc:
I think I partially understand a comment you made on my f3d skeleton program: people should use https://assimp.org/ to export from C++; so MeshIO shouldn't exist?

coryrc:
Err, nope, MeshIO is just a wrapper on assimp. I guess I don't understand how you want people to export from C++ without it.

starseeker:
[...]

PR #1529, opened 2026-02-13
make MeshIO private

pca006132:
Part of #1524. The general OBJ file format can be quite annoying, so I switched to a regex-based implementation. This can at least parse output from assimp, and I tested that the output can be parsed by things like f3d.

There are some issues here:
1. With the current API, we cannot import non-manifold meshes as a `MeshGL`/`MeshGL64`, which is needed by `BooleanComplex.SimpleOffset`. It is easy to fix this by adding more APIs (one with just `MeshGL64`), but I am still thinking about whether that is a good way.
2. What should we do for C binding users? Expose an API that writes to a file? Or an API that takes a callback that accepts a char buffer? (the callback version is implemented)
3. The Ember benchmark uses stl import. We don't want to bundle the files ourselves, so we cannot convert them to obj files. I can write a simple stl parser for the Ember benchmark, but this feels a bit ad hoc. Or maybe we should require assimp for building the Ember benchmark, and this can also serve as an example for using assimp?

### DIFF
--- a/test/boolean_complex_test.cpp
+++ b/test/boolean_complex_test.cpp
@@ -1519,7 +1519,7 @@ TEST(BooleanComplex, HullMask) {
   MeshGL mesh = ret.GetMeshGL();
 }
 
-TEST(BooleanComplex, SimpleOffset) {
+TEST(BooleanComplex, DISABLED_SimpleOffset) {
   MeshGL64 seeds = ReadTestMeshGL64OBJ("Generic_Twin_91.1.t0.obj");
   EXPECT_TRUE(seeds.NumTri() > 10);
   EXPECT_TRUE(seeds.NumVert() > 10);


pca006132:
Thinking about it, removing `MeshIO` completely can be annoying:

1. We use it for fancy colored exports in some of our tests. Obj export will not have those properties. They are not crucial, but are nice to have.
2. We depend on it in `extras`, e.g., `convert_file`, `f3d_skeleton`, and `man_bench`.

For 1, we can make it optional for test exports. If assimp integration is enabled, it will output glb files in addition to obj files. Assimp is no longer a required dependency for any test. Import models will be obj files.
For 2, we can put the meshio code in extras.

Then we are just removing MeshIO from the public API, removing dependency on it for tests (we can run all tests without assimp). It is only needed for extras or when we want fancy output from tests.

include/manifold/manifold.h
```diff
@@ -478,10 +467,8 @@ class Manifold {
    * ofile.close();
    * @endcode
    */
-#ifdef MANIFOLD_DEBUG
```
elalish, line 481: How much does this bloat our WASM build? I think that was the primary reason we had this guarded by MANIFOLD_DEBUG.

src/impl.cpp
```diff
@@ -800,86 +826,57 @@ std::ostream& operator<<(std::ostream& stream, const Manifold::Impl& impl) {
...
+
+  constexpr size_t BUFFER_SIZE = 100;
+  std::array<char, BUFFER_SIZE> buffer;
+  std::cmatch m;
+
+  while (!stream.eof()) {
+    // extract line, and skip the line if it is longer than BUFFER_SIZE
+    // because the lines we care about should not be that long
+    // not using std::basic_istream<...>::getline because getline throws when
+    // the size exceeds the limit, but we don't want exception related code
```
elalish, line 847: 👍

elalish:
I think all I/O should ideally go through MeshGL, rather than directly to a Manifold. Then again, for debug, more direct is important. I guess adding a MeshGL64 import is the minimal solution. I'm not too worried about C bindings - no one is using C for its ease of use, and we only need import for repro, which is C++ only.

Yeah, okay, meshIO in extras does make sense. I don't think we need any fancy colored output from C++ tests - we have JS tests that do a much better job of that.

include/manifold/manifold.h
pca006132: I would expect emscripten to remove unused functions from the wasm build.

pca006132:
The issue here is that `MeshGL` does not have epsilon. The original goal of the function was to save things for debugging, where the epsilon value can be critical for some bugs. Not sure if adding the epsilon to `MeshGL` is a good idea, we don't want users to set it to some weird values...

pca006132:
The wasm code size is not increased if we are not building with `MANIFOLD_DEBUG=ON`.

Btw, I found that if we disable rtti (which we can), we get a 3% code size reduction:
```diff
diff --git i/CMakeLists.txt w/CMakeLists.txt
index bf81f04f..335619ec 100644
--- i/CMakeLists.txt
+++ w/CMakeLists.txt
@@ -136,6 +136,9 @@ if(EMSCRIPTEN)
     list(APPEND MANIFOLD_FLAGS -fexceptions)
     add_link_options(-fexceptions)
     add_link_options(-sDISABLE_EXCEPTION_CATCHING=0)
+  else()
+    add_compile_options(-fno-rtti -DEMSCRIPTEN_HAS_UNBOUND_TYPE_NAMES=0)
+    add_link_options(-fno-rtti)
   endif()
   set(MANIFOLD_PYBIND OFF)
   set(BUILD_SHARED_LIBS OFF)
```

bindings/c/include/manifold/manifoldc.h
```diff
@@ -501,31 +501,32 @@ void manifold_delete_triangulation(ManifoldTriangulation* m);
...
-void manifold_destruct_export_options(ManifoldExportOptions* options);
-
-void manifold_delete_material(ManifoldMaterial* m);
-void manifold_delete_export_options(ManifoldExportOptions* options);
-#endif
-
+// Import a manifold from a Wavefront obj file.
+// The obj_file parameter is the content of the obj file, not the filename,
+// and should be null-terminated.
+ManifoldManifold* manifold_read_obj(void* mem, char* obj_file);
```
elalish, line 507: Much simpler - I like this direction.

test/models/Cray_right.obj
```diff
@@ -0,0 +1,35 @@
...
+v 3.40282347e+38 -3.40282347e+38 -100
+
+# 0 UV coordinates
+
+# 1 vertex normals
+vn -nan -nan -nan
+
+# Mesh '' with 12 faces
+usemtl $Material_0
+f  1//1 2//1 3//1
```
elalish: Not a huge deal, but these OBJs are weird - what are these `//` doing here? And why does it have one random line on nan vertex normals? Would it be better to use our own WriteOBJ function to convert these GLBs?

test/models/Cray_right.mtl
elalish, line 1: And can we remove all these .mtl files? I don't think we need them for anything.

elalish (approved):
Looks great, thanks!

test/models/Cray_right.obj
pca006132: Yeah, these are from assimp. I can change that to using to our export function.

test/models/Cray_right.mtl
pca006132: Yes after changing the obj to using our export.

test/boolean_complex_test.cpp
```diff
@@ -1530,16 +1513,14 @@ TEST(BooleanComplex, CraycloudBool) {
...
+  Manifold mask = ReadTestOBJ("hull-mask.obj");
   Manifold ret = body - mask;
   MeshGL mesh = ret.GetMeshGL();
 }
 
-TEST(BooleanComplex, SimpleOffset) {
-  std::string file = __FILE__;
-  std::string dir = file.substr(0, file.rfind('/'));
-  MeshGL seeds = ImportMesh(dir + "/models/" + "Generic_Twin_91.1.t0.glb");
+TEST(BooleanComplex, DISABLED_SimpleOffset) {
```
pca006132, line 1522: This is causing triangulation failure for some reason, should be a bug.

pca006132: I don't know if this means that our export and import is inexact, cuz bitwise identical MeshGL64 should not cause this.

pca006132: Will investigate if this means the import/export is inexact.

pca006132: OK, this is not an issue about inexact output. We reorder faces in the output by sorting them, and this affects the order of Booleans in the test, and some order causes triangulation failure.

@elalish I think we can keep this for now, so we can fix the triangulation bug?

elalish: Yeah, sounds good.

PR #1532, opened 2026-02-16
fix test

elalish:
I don't love this test for several reasons, so I'm tempted to remove it entirely. I checked the output and it looks reasonable-ish despite the triangulation warnings. I also got this on importing the Generic_Twin_91.1.t0.obj to MeshLab:
```
Warning: Error encountered while loading file: "Generic_Twin_91.1.t0.obj" 
Error details: Identical vertex indices found in the same faces -- faces ignored
```
And sure enough:
```
f 1 2 3
f 1 3 4
f 1 4 143
f 1 8 2
f 1 142 141
f 1 143 142
f 4 3 5
f 4 4 5
```
@pca006132 does this mean our WriteOBJ function has an error? Or did we somehow get a topological edge degenerate in our data structure?


pca006132:
I am fine with removing this. The original mesh was topologically incorrect, so I think this is expected.

elalish:
Okay, SG. It still troubles me that we can get this kind of face in our data structures at all. Seems like we need to add another bit of sanitizing on our inputs.

elalish:
Oh wait, I think I see - the conversion probably never went through manifold, just meshGL. And Manifold is auto-removing these, hence not failing. Cool.

elalish:
Okay, removing since we now have a better version of this test anyway that doesn't involve import and is much more straight-forward to view and debug.

starseeker:
Yes, that's correct - the Generic_Twin_91.1.t0.obj was the raw input from the test model I was working with at the time that triggered the issue, so if the core bug it triggered is tested another way now there's no reason to hang on to this version.

PR #1537, opened 2026-02-19
fix: avoid reinterpret_cast to atomic, resolving undefined behavior (#1153)

### LINKED ISSUE #1153: avoid reinterpret_cast to atomic
This is technically undefined behavior but works in typical architecture.

aviralnull:
While going through the codebase, I noticed that we were using 
reinterpret_cast to cast plain memory buffers to std::atomic<T>* 
which is technically undefined behavior, even though it works fine 
on most architectures like x86 and ARM.

Fixed this by removing all reinterpret_cast to atomic throughout 
the src/ directory. No new methods were added and the existing 
structure of the code was kept intact. The change is minimal and 
focused only on eliminating the UB cast.

Verified the fix by scanning the entire src/ folder for any 
remaining reinterpret_cast combined with atomic and confirmed 
zero instances remain.

Closes #1153

### DIFF
--- a/src/collider.h
+++ b/src/collider.h
@@ -288,7 +288,7 @@ class Collider {
     // create global counters
     std::unique_ptr<std::atomic<int>[]> counter(
         new std::atomic<int>[NumInternal()]);
-    for (int i = 0; i < NumInternal(); ++i) {
+    for (size_t i = 0; i < NumInternal(); ++i) {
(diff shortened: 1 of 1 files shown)


pca006132:
The proper way is to upgrade to c++20 and use atomic ref, I don't think we are going to merge this, so closing.

PR #1538, opened 2026-02-22
Add ConvexDecomposition method

### LINKED ISSUE #415: Feature request: 3D convex decomposition operation
Convex decomposition is an operation that OpenSCAD relies on as part of its minkowski operations.
Currently the only implementation we have available is through [CGAL](https://doc.cgal.org/latest/Convex_decomposition_3/index.html), using its dreaded Nef Polyhedra which is slow and error prone.

zalo:
## Summary

Adds `Manifold::ConvexDecomposition(int maxClusterSize, int maxDepth)` that decomposes a manifold into convex pieces. Exposed to C++, Python, JS/WASM, and C bindings.

### Algorithm

1. **Coplanar reflex plane pre-pass** — scan all reflex edges for coplanar chains. When a plane contains > 50% of all reflex edges, split by it first, then recurse. This resolves sphere intersection seams without DT vertex poisoning (TwoSpheres: 271 → 2 pieces).
2. **Delaunay tetrahedralization** — unconstrained DT of mesh vertices with deterministic hash perturbation for cospheric robustness.
3. **Clip tets** — intersect each DT tet hull against the mesh (parallel via TBB). Interior tets keep the original hull to avoid introducing boolean vertices.
4. **Recover uncovered regions** — if clipped volume < 99.9% of total, compute difference and decompose the remainder.
5. **Cospheric merge** — group adjacent tets sharing a circumcenter (unperturbed positions) via union-find, merge by hull. Undoes arbitrary DT assignments for cospheric point sets (e.g. cube vertices).
6. **Greedy hull merge** — Phase A: pairwise merges via DT adjacency. Phase B: cluster merges (sizes 2 through maxClusterSize) with bbox overlap pruning, tightest-fit-first.
7. **Reflex edge splitting** — split remaining non-convex pieces by a plane through each reflex edge that bisects the dihedral angle.

### Minkowski upgrade

Adds `decompose` parameter to `MinkowskiSum`. When true, decomposes non-convex inputs into convex pieces for the fast vertex-addition path.

### Benchmarks

| Shape | Pieces | NC | Time | vs CGAL |
|-------|--------|-----|------|---------|
| LShape | 3 | 0 | 1ms | 11x faster |
| CubeSphere | 67 | 0 | 19ms | 25x faster |
| TwoSpheres | **2** | 0 | **3ms** | 800x faster |
| ThreeSpheres | 713 | 0 | 845ms | 9x faster |
| CubeCube | 5 | 0 | 3ms | 9x faster |
| FlatSlab | 59 | 0 | 317ms | CGAL crashes |

Always 0 non-convex pieces. 2-800x faster than CGAL. CGAL crashes on FlatSlab.

No new dependencies. Parallel tet clipping via TBB.

Gallery: https://convex-decomposition-gallery.pages.dev

Closes #415

## Test plan

- [x] 18 ConvexDecomposition tests (including ThreeSpheres, FlatSlab, thin geometry)
- [x] 9 MinkowskiSum tests (3 decompose=true variants)
- [x] All existing tests pass (no regressions)
- [x] clang-format passes
- [ ] CI passes


pca006132:
I think the best way to test this would be to test with openscad.

Btw, do you want to change the minkowski sum code to use this instead of the brute-force approach?

zalo:
Interestingly, it won’t always make sense to use this for the minkowski sum, because this volumetric approach won’t support erosion operations like the other surface approach does. (Though neither does OpenSCAD, so it’s fine.)

So I’ll probably leave it as an option, only for the sum for now…

Sorry this PR is failing on some OS’s, I’ll need a bit to troubleshoot…

Likewise; I haven’t done my full set of validity tests on it yet, but it seems more promising than my previous attempts…

It has the dubious advantage of being inelegant and much simpler 😅

zalo:
Bah sorry; premature, this is gonna need a little more time in the oven.

As mentioned in #415 , more cases than before work because of the simplicity of unconstrained tetrahedralization + not trying to make the violating parts delaunay… but the violating parts are still causing some problems when they still have reflex edges in them 😅

May have to add a tolerance where if a triangle is a small sliver enough next to a reflex edge, it just collapses and uses the orignal reflex edge vertices… and maybe I can “non-delaunay” tetrahedralize too…

zalo:
> Codecov

Just addressed this; inching towards viable

I’m a little miffed at how many parts the two spheres example outputs; kinda nuts.

I guess spheres are a pathological case for any delaunay adjacent algorithm; all points lie on the circumsphere, so most tetrahedra are delaunay degenerate in that sense…

src/convex_decomposition.h
elalish, line 1: What is the point of this file?

include/manifold/manifold.h
```diff
@@ -435,6 +435,15 @@ class Manifold {
   static Manifold Hull(const std::vector<vec3>& pts);
   ///@}
 
+  /** @name Convex Decomposition
+   */
+  ///@{
+  // mode: 0=DT+carve (default), 1=carve-only, 2=onion-peel
+  std::vector<Manifold> ConvexDecomposition(int maxClusterSize = 2,
```
elalish: Do we have other use cases in mind for this besides Minkowski? I think I'd prefer this stay out of the public API for now until we have a clear idea of how people want to use it. Also this method has too many options.

include/manifold/manifold.h
```diff
@@ -396,7 +396,7 @@ class Manifold {
   std::pair<Manifold, Manifold> SplitByPlane(vec3 normal,
                                              double originOffset) const;
   Manifold TrimByPlane(vec3 normal, double originOffset) const;
-  Manifold MinkowskiSum(const Manifold&) const;
+  Manifold MinkowskiSum(const Manifold&, bool decompose = false) const;
```
elalish, line 399: How would a user decide to set decompose to true or false? I think best to remove this option and figure out the right answer automatically.

src/manifold.cpp
```diff
@@ -958,6 +962,30 @@ Manifold Manifold::Hull(const std::vector<Manifold>& manifolds) {
...
+ *
+ * Uses unconstrained Delaunay tetrahedralization of mesh vertices, clips each
+ * tet against the mesh, then greedily merges adjacent convex pieces by hull.
+ * No new dependencies are required.
+ *
+ * @param maxClusterSize Maximum cluster size for greedy merge (2-4). Higher
+ * values produce fewer pieces but take longer. Default is 2.
+ * @param maxDepth Maximum recursion depth for decomposing non-convex pieces.
+ * Higher values produce more convex pieces at the cost of speed. Default is 1.
+ * @return Vector of approximately convex manifold pieces.
```
elalish, line 974: What about mode?

zalo:
@elalish I'm sorry; I didn't mean for you to spend time reviewing this yet! 🫠   It just doesn't run the CI if I have draft mode on 😅 

I agree wholeheartedly with all of your comments about hiding configurability; I'm still in the process of writing up visual tests (like I did with the Minkowski PR) for side-by-side comparisons and debugging and the relative pro's and con's of each method for each model case and how to detect each one's degenerate situations...

I've been testing out different methods of doing convex hull decomposition for years, and just "merging unconstrained delaunay tets + cleaning up the excess with splits" has gotten me more success than I've enjoyed in a while, so I was excited to get it down in PR form 😅 

The three methods:
- Unconstrained Delaunay -> Clip Tets that don't conform to the mesh (recursively decompose) -> Greedily Merge
- [Reddit Method](https://www.reddit.com/r/compsci/comments/17qusll/i_created_exact_convex_decomposition_algorithm/): Walk the Mesh and accumulate triangles into a convex hull (until that hull exits the mesh) -> Subtract and Repeat
- "Onion Peeling": Create a Convex Hull of the Reflex Edges -> Subtract, Decompose, and Repeat

Each generally seem to give reasonable results so far, but I'm gated on my test models right now, so I need to switch gears back to testing...

src/convex_decomposition.h
zalo: Oops; leftover

include/manifold/manifold.h
zalo: Agreed on parameters; I'll get rid of those once I know what works best.

ConvexDecomposition is a big deal to make objects collide in physics engines, and the current solutions for it (VHACD / CoACD) are both approximate and 10-100x slower than these ones...

A downstream dream of mine is to make a virtual woodworking/metalworking/etc. simulation, and in all my wanderings, I haven't found a good way out of convex decomposition (except doing SDF on SDF or particle based collisions?)

Here's a crappy example of particle based physics:
https://zalo.github.io/PhysicsWorkshop/

Also, convex decomposition is the foundational operation for minkowski sums, sweeps, and other fun solid geometry operations... I'm sure users will find _something_ to do with it 😅

include/manifold/manifold.h
zalo: I'll remove it once I'm done running robustness comparisons between all the modes and minkowski space (and hopefully it's shown minkowski sums are just better everywhere (except in the case of erosion operations).

src/manifold.cpp
zalo: Hoping to get rid of the mode selector, in a timsort-style sense...

Every shape has pathological/adversarial configurations; even CGAL crashes on one of my test pieces!
(I updated the PR description with short benchmarks; planning to make a comparison website soon...)

zalo:
Hmm, okay @elalish check this one out: https://manifold-convex-decomposition.pages.dev/

There are a few interesting comparisons; I'm not sure you've ever seen how CGAL's convex decomposition makes cuts in 2D and extrudes them along one axis: that's basically the core of its algorithm:
https://manifold-convex-decomposition.pages.dev/#tab=decomposition&shape=CubeSphere&explode=0

CGAL's output is extremely rotation dependent (this would have tied with the other algos if it were axis-aligned):
https://manifold-convex-decomposition.pages.dev/#tab=decomposition&shape=CubeSphereHole&explode=12

Likewise, here's an example that works better for some algorithms:
https://manifold-convex-decomposition.pages.dev/#tab=decomposition&shape=TwoSpheres&explode=15
and then those same ones fall apart with a slight variation:
https://manifold-convex-decomposition.pages.dev/#tab=decomposition&shape=ThreeSpheres&explode=15

There's even one that CGAL totally crashes on:
https://manifold-convex-decomposition.pages.dev/#tab=decomposition&shape=FlatSlab&explode=0

So, yeah, that's where it's at... going to need to figure out heuristics to tune the settings to get a reasonable case scenario in most instances...

elalish:
I'm a lot more interested in reliability than optimality. A big part of that is keeping code complexity down. Just make sure self-intersecting and degenerate geometry doesn't make for any big problems.

zalo:
I think the decomposition is pretty much correct to the best of my ability to test, but I think I'm going to be running into the old manifold union non-deterministic instability if I try to integrate it with the minkowksi sum ([which I think happened here](https://github.com/elalish/manifold/actions/runs/22340197247/job/64641592798)).

I think manifold's union reeeeally doesn't like to merge kissing faces 😅

include/manifold/manifold.h
elalish: Okay, if we can get it down to no parameters, I'm game for having it in the API.

elalish:
>I think manifold's union reeeeally doesn't like to merge kissing faces 😅

Can you show me an example where your Minkowski is generating kissing faces? Seems like if each chunk is offset, they'd tend to overlap quite a bit?

Anyway, your benchmarks are impressive and I love the little comparison website you spun up. Go ahead and request a review when you feel you're ready.

zalo:
Hmm, I have just learned of a better way of merging delaunay triangles, which I'm hoping will extend to delaunay tetrahedra...

Merged-CDT's are mentioned towards the end of this talk on state of the art pathfinding algorithms:
https://www.youtube.com/watch?v=7MOQOvGJyn4

And they have an MIT-Licensed repo for it: https://bitbucket.org/dharabor/pathfinding/src/master/anyangle/polyanya/
(Which includes a "throw-away script" for intelligently merging constrained delaunay tetrahedra: https://bitbucket.org/dharabor/pathfinding/src/master/anyangle/polyanya/utils/meshmerger.cpp )

So I've been porting it to typescript to play around with it: https://github.com/zalo/polyanya-ts

And... I think it's good... so that will be next... hopefully both much faster and much better...

zalo:
(In short: it was not good…)

Sebastien Loriot @sloriot was hinting that I was going to run into trouble with slivers doing this, and it seems like he was right 😅

Most of the common tetrahedralizers out there change the input (TetWild/fTetWild), and I guess TetGen, Marco Attene’s CDT, and CGAL’s new CDT are unique for preserving the input more exactly across a wider range of meshes.

But the incremental complexity of their implementations might be out of scope to re-implement in manifold (and their licenses are presently incompatible to import as well)… I’m not sure…

With a configurable precision threshold, I can broadly control the tradeoff of "number of parts + compute time <-> how exact the output is", where higher exactness causes increasing problems 🤔

elalish:
>With a configurable precision threshold, I can broadly control the tradeoff of "number of parts + compute time <-> how exact the output is", where higher exactness causes increasing problems 🤔

Honestly, precise convex decomposition is kind of annoying anyway. I'm definitely willing to consider a tolerance version if it's robust. Have some examples to show the tradeoffs / failure cases?

rujialiu:
> Most of the common tetrahedralizers out there change the input (TetWild/fTetWild), and I guess TetGen, Marco Attene’s CDT, and CGAL’s new CDT are unique for preserving the input more exactly across a wider range of meshes.
> 
> But the incremental complexity of their implementations might be out of scope to re-implement in manifold (and their licenses are presently incompatible to import as well)… I’m not sure…
> 
> With a configurable precision threshold, I can broadly control the tradeoff of "number of parts + compute time <-> how exact the output is", where higher exactness causes increasing problems 🤔

(Probably unrelated, but I couldn't help asking)
Currently I'm using fTetWild to preprocess non-manifold meshes to manifold (I also use OpenVDB's level set, but in many cases it's quality is lower than fTetWild), but I had to keep the preprocessor in a separate program because fTetWild is slow, RAM-hungry and depends on geogram (even after removing everything not needed, it's still 100k+ loc), so I wonder whether there will be some more lightweight ways of doing inexact tetrahedralization after this PR is merged.

BTW: I think fTetWild's license is the most permissive one (MPL) among all common tetrahedralizers, though it's still not as liberal as Apache-2/MIT/BSD

ali1234:
> I think the best way to test this would be to test with openscad.

ICYMI, OpenSCAD can now be compiled to use Manifold's Minkowski sum implementation. Just build it with `cmake -DUSE_MANIFOLD_MINKOWSKI ...`

In testing the original (non-decomposing) implementation I found a great deal of variance in performance. In some cases it is many times faster than CGAL, usually due increased parallelism - it still uses as much CPU time over all. In other cases it is much slower, especially when used on two concave objects.

fire:
There's a diff needed.

```patch
--- ./src/convex_decomposition.cpp	(original)
+++ ./src/convex_decomposition.cpp	(reformatted)
@@ -864,8 +864,7 @@
       Manifold hull = Manifold::Hull(combined);
       if (hull.IsEmpty()) return;
       double hullVol = hull.Volume();
-      if (hullVol <= sumVol * (1.0 + kMergeTol))
-        pq.push({hullVol, a, b});
+      if (hullVol <= sumVol * (1.0 + kMergeTol)) pq.push({hullVol, a, b});
     };
 
     // Seed PQ with all adjacent valid pairs.

```

zalo:
I think I'm going to have to close this for now; no time these days...

elalish:
I know the feeling - come back anytime!

PR #1541, opened 2026-02-26
 Fix Cylinder(h, 0, r) to support cone with apex at bottom

### LINKED ISSUE #1503: Can't directly use `cylinder` to make an ice cream cone
Briefly searched and didn't see any issues for this. I tried to use `cylinder(10, 0, 5)` to create an upside down cone and was surprised to find it doesn't work. I expected this to work similarly to `cylinder(10, 5, 0)`, and can obviously work around it with a rotation. Not sure if this is a binding issue or something deeper that causes it to fail like creating an inside out cylinder?
 
Image

AnshulPatil2005:
Fixes #1503.

Cylinder(h, 0, r) previously returned Invalid() because the implementation rejected radiusLow <= 0 unconditionally. This made it impossible to directly create a cone with the apex at the bottom (z=0) and base at the top (z=h), even though the symmetric case Cylinder(h, r, 0) worked fine.

##Fix:

When radiusLow == 0 and radiusHigh > 0, the function now creates the equivalent apex-at-top cone (Cylinder(h, radiusHigh, 0)) and mirrors it across the XY plane, then translates it up by h. This produces a valid manifold cone with apex at z=0 and base at z=h.

The following cases remain invalid and return InvalidConstruction:

Cylinder(h, 0) : both radii zero (default radiusHigh = -1)
Cylinder(h, 0, 0) : both radii explicitly zero
Cylinder(h, r < 0, ...) : negative radius

##Tests:

Added Manifold.CylinderZeroRadiusLow which verifies:

Cylinder(h, 0, r) produces a valid manifold with no error
Its volume equals Cylinder(h, r, 0) (same shape, mirrored)
Bounding box has apex at z=0 and base at z=h
Also added two new invalid-construction assertions for the radiusLow=0 edge cases.

### DIFF
--- a/src/constructors.cpp
+++ b/src/constructors.cpp
@@ -168,11 +168,9 @@ Manifold Manifold::Cylinder(double height, double radiusLow, double radiusHigh,
     }
     // Cone with apex at bottom: create the centered apex-at-top version and
     // mirror it
-    Manifold cone =
-        Cylinder(height, radiusHigh, 0.0, circularSegments, true);
+    Manifold cone = Cylinder(height, radiusHigh, 0.0, circularSegments, true);
     cone = cone.Mirror(vec3(0.0, 0.0, 1.0));
-    if (!center)
-      cone = cone.Translate(vec3(0.0, 0.0, height / 2.0));
+    if (!center) cone = cone.Translate(vec3(0.0, 0.0, height / 2.0));
     return cone.AsOriginal();
   }
   const double scale = radiusHigh >= 0.0 ? radiusHigh / radiusLow : 1.0;


src/constructors.cpp
```diff
@@ -158,9 +159,22 @@ Manifold Manifold::Cube(vec3 size, bool center) {
...
   }
+  if (radiusLow == 0.0) {
+    if (radiusHigh <= 0.0) {
+      return Invalid();
+    }
+    // Cone with apex at bottom: create the apex-at-top version and mirror it.
+    Manifold cone = Cylinder(height, radiusHigh, 0.0, circularSegments, false);
+    cone = cone.Mirror(vec3(0.0, 0.0, 1.0))
+               .Translate(vec3(0.0, 0.0, height))
+               .AsOriginal();
```
pca006132: I think you can just do `AsOriginal` when you return, so to not do this again when `center` is true.

Also, what if we just set `center = true` in l170, and only translate when `center` is false?

test/manifold_test.cpp
```diff
@@ -294,6 +294,22 @@ TEST(Manifold, Cylinder) {
   EXPECT_EQ(cylinder.NumTri(), 4 * n - 4);
 }
 
+TEST(Manifold, CylinderZeroRadiusLow) {
```
pca006132, line 297: It is good that you have a test, but the current test is not tight enough to my liking: it does not differentiate between the normal cone and the flipped cone. Maybe try to come up with some simple properties to check for? E.g., do some Boolean or take a slice from the cone.

PR #1547, opened 2026-02-27
Fix integer divide-by-zero in RefineToTolerance

### LINKED ISSUE #1545: WASM refineToTolerance not working
I was just making a ManifoldCAD [example](https://tinyurl.com/y8er2pd9) and found that `refine` and `refineToLength` both work, but `refineToTolerance` fails with
```
RuntimeError: divide by zero
    at Manifold.refineToTolerance (null:null:1)
    at :24:57
```
for any tolerance value. Probably something really simple, but it's late here, so I'll just file it for now.

AnshulPatil2005:
## Summary

Fixes #1545  refineToTolerance crashing with RuntimeError: divide by zero in WASM for any tolerance value
The bug was in the keepInterior block of Subdivide in src/subdivision.cpp, where an integer division / longest was reachable with longest == 0
Added a longest > 0 guard to the condition before the division

## Root Cause
In src/subdivision.cpp:555, the expression:

return extra > 0 ? (extra * (longest - thisAdded)) / longest : 0;
performs an integer divide by longest, which can be 0 when all edges in a triangle have zero required divisions (flat/uncurved regions where the tolerance lambda returns 0).

minExtra is always at least 1 (longest * 0.2 + 1), so extra can be positive even when longest == 0, making the division reachable.

## Fix

// before
return extra > 0 ? (extra * (longest - thisAdded)) / longest : 0;

// after
return (extra > 0 && longest > 0)
           ? (extra * (longest - thisAdded)) / longest
           : 0;
When longest == 0, no edge needs subdivision, so returning 0 is correct.

### DIFF
--- a/bindings/wasm/test/bindings.test.ts
+++ b/bindings/wasm/test/bindings.test.ts
@@ -40,13 +40,15 @@ suite('Manifold Bindings', () => {
   });
 
   test('refineToTolerance does not throw (issue #1545)', () => {
-    // Reproduces the original failing geometry from issue #1545: a flat-faced mesh
-    // with normals set via calculateNormals + smoothByNormals. On flat faces,
-    // tangents are parallel to the edges so d == 0 exactly, making edgeDivisions
-    // return 0 for *any* tolerance. This triggers longest == 0 in the keepInterior
-    // block — the integer divide-by-zero that traps in WASM.
-    const cube = manifoldModule.Manifold.cube([10, 10, 10]).calculateNormals(0, 30);
-    const smooth = manifoldModule.Manifold.ofMesh(cube.getMesh(0)).smoothByNormals(0);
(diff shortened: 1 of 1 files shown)


pca006132:
Thanks. Can you also add the failing example to the wasm tests?

For the CI failure, it seems to be something related to gh action setup. Don't need to worry about that, we can see if it is temporary or we can try to fix it later.

elalish:
Have you double-checked the original ManifoldCAD example I showed in the issue? That `null:null:` thing in the error made me think this was a binding problem rather than a C++ problem.

AnshulPatil2005:
> Have you double-checked the original ManifoldCAD example I showed in the issue? That `null:null:` thing in the error made me think this was a binding problem rather than a C++ problem.

Yes, The null:null:1 is just WASM's stack frame format when debug symbols are stripped not a binding issue. The binding is identical in form to refineToLength.

PR #1554, opened 2026-02-28
Fixes different verbose levels

### LINKED ISSUE #1107: Different verbose levels
https://github.com/elalish/manifold/pull/1104#issuecomment-2526974380

E.g. 0 as not verbose, 1 for dumping problematic meshes, 2 for printing statistics.

AnshulPatil2005:
fixes #1107
This separates the Boolean debug output levels so verbose=1 and verbose=2 no longer behave almost the same.
Changes
This PR moves routine Boolean stats output to verbose >= 2, including:

intersection timing
assembly / triangulation / simplification / sorting timing
vertex / triangle count summaries
edge cleanup statistics
It keeps failure-oriented debug output at verbose > 0, including:
operand dumps from SimpleBoolean()
self-intersection triangle dumps
It also updates the verbose-level documentation and test CLI help text to match the intended behavior.

Resulting behavior
verbose = 0: no verbose output
verbose = 1: Boolean debug dumps for failures / invalid intermediate meshes
verbose = 2: Boolean stats/timing output plus triangulator verbose output

### DIFF
--- a/include/manifold/common.h
+++ b/include/manifold/common.h
@@ -523,9 +523,8 @@ struct ExecutionParams {
   bool cleanupTriangles = true;
   /// Verbose level:
   /// - 0 for no verbose output
-  /// - 1 for verbose output for the Boolean, including timing info and vector
-  /// sizes.
-  /// - 2 for verbose output with triangulator action as well.
+  /// - 1 for Boolean debug dumps on failures and invalid intermediate meshes.
+  /// - 2 for Boolean timing and size statistics, plus triangulator action.
   int verbose = 0;
 };
 /** @} */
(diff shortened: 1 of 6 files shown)


pca006132:
Are the triangulator's verbose outputs now moved to `verbose >=  2`?

AnshulPatil2005:
> Are the triangulator's verbose outputs now moved to `verbose >= 2`?

it is at 2 only didn't change that, moved the Boolean timing/statistics output to level 2 as well.

elalish (approved):
I think this is fine - no strong opinion. I think this means if you run without `MANIFOLD_DEBUG` that you'll only get output in verbose = 2, since it'll only be stats at that point?

AnshulPatil2005:
> I think this is fine - no strong opinion. I think this means if you run without `MANIFOLD_DEBUG` that you'll only get output in verbose = 2, since it'll only be stats at that point?

I don't think so. Without MANIFOLD_DEBUG, these verbose paths are compiled out, so this only changes the split between verbose=1 and verbose=2 in debug builds.

PR #1556, opened 2026-03-02
Fix C include_pt bindings mutating temporary copies

### LINKED ISSUE #1544: C bindings: manifold_box_include_pt/manifold_rect_include_pt do not modify the passed object
### Summary
`manifold_box_include_pt()` and `manifold_rect_include_pt()` appear to operate on a local copy instead of the input pointer, so they have no observable effect.

### Where
- `bindings/c/box.cpp:57`
- `bindings/c/rect.cpp:62`

Current implementations do:
- `auto box = *from_c(b); box.Union(p);`
- `auto rect = *from_c(r); rect.Union(p);`

But `Union(const vec3)` / `Union(const vec2)` are in-place mutators:
- `include/manifold/common.h:211`
- `include/manifold/common.h:400`

### Expected
`manifold_*_include_pt` should expand the object referenced by the pointer argument.

### Actual
The pointer target is unchanged because only a temporary copy is mutated.

### Minimal repro idea (C API)
Create a small box/rect, call `*_include_pt` with an outside point, then verify with `*_contains_pt`; it still returns false.

### Suggested fix
Mutate through the pointer directly:
- `from_c(b)->Union(p);`
- `from_c(r)->Union(p);`

AnshulPatil2005:
Summary
Fixes manifold_box_include_pt() and manifold_rect_include_pt() so they mutate the passed Box and Rect in place instead of mutating a local copy.

Closes #1544.

Problem
Both C API functions copied the underlying bounds object before calling Union():

auto box = *from_c(b);
box.Union(p);
and similarly for Rect.

That updated only the temporary value, so the caller-visible object was unchanged. From the C API, include_pt effectively behaved like a no-op.

Changes
update manifold_box_include_pt() to mutate the actual Box
update manifold_rect_include_pt() to mutate the actual Rect
add a regression test covering both APIs


bindings/c/rect.cpp
```diff
@@ -60,7 +60,7 @@ int manifold_rect_contains_rect(ManifoldRect* a, ManifoldRect* b) {
 }
 
 void manifold_rect_include_pt(ManifoldRect* r, double x, double y) {
-  auto rect = *from_c(r);
+  auto& rect = *const_cast<Rect*>(from_c(r));
```
elalish: const_cast is an indication of other problems. It looks like our `from_c` functions are quite inconsistent about whether they return const or not. Let's at least remove the const from these ones. Bonus points if you take a look to understand why any of them return const in the first place. I have the feeling none of them should.

AnshulPatil2005: Implemented, I removed the const_cast and normalized the C bridge so from_c(...) no longer adds const to opaque handle conversions.

Updated in bindings/c/conv.h and bindings/c/conv.cpp for:

Manifold
CrossSection
SimplePolygon
Polygons
MeshGL
MeshGL64
Box
Rect
Triangulation

I checked the rest of the opaque-handle from_c(...) helpers and made them consistent as well, so they no longer return const pointers for those types either.the mutating C APIs now use from_c(...) directly and no longer need const_cast.

PR #1562, opened 2026-03-02
feat(manifoldcad): add tooltips, dynamic filenames, and type fetching

Divyansh200102:
This PR addresses several usability tasks for the ManifoldCAD editor.
### Changes
* **Tooltips:** Added standard HTML `title` attributes to the top-bar icons (Undo, Redo, Format, Share, Docs) to clarify their functions.
* **Dynamic Filenames:** The downloaded GLB and 3MF files now dynamically use the active script's name (e.g., `Intro.glb`) instead of hardcoding `manifold.glb`.
* **Fetching Types Indicator:** Added a "Fetching types..." text indicator to the top bar that triggers on `AutoTypings` updates, providing visual feedback when Monaco is resolving TypeScript definitions.
* **Editor Config Tightening:** Disabled the Monaco editor minimap to free up horizontal screen real estate.

### DIFF
--- a/bindings/wasm/examples/editor/editor.js
+++ b/bindings/wasm/examples/editor/editor.js
@@ -629,7 +629,8 @@ function clickSave(saveButton, extension, outputName) {
     }
     const link = document.createElement('a');
 
-    link.download = `${currentFileElement.textContent.trim() || 'manifold'}.${extension}`;
+    link.download =
+        `${currentFileElement.textContent.trim() || 'manifold'}.${extension}`;
 
     link.href = output[outputName];
     link.click();


bindings/wasm/examples/editor/editor.js
```diff
@@ -333,11 +334,23 @@ async function createEditor() {
...
     sourceCache: self.window.typecache,
     onError: e => {console.error(e)},
-    onUpdate: (update, text) => {console.debug(text)},
+    onUpdate: (update, text) => {
+      if (typeIndicator) {
+        typeIndicator.textContent = 'Fetching types...';
+        clearTimeout(typeTimeout);
+        typeTimeout = setTimeout(() => {
+          typeIndicator.textContent = '';
+        }, 1000);
```
elalish: Is there a way to make this disappear when the types get fetched, instead of after a timeout?

elalish:
@tonious I tried to request you as a reviewer, but Github won't let me - I think this is because you're not an official member here. You've certainly done plenty of work; would you like to be an official collaborator and in the author's list?

bindings/wasm/examples/editor/editor.js
Divyansh200102: @elalish I removed the 1000ms timeout. The indicator now follows the AutoTypings lifecycle directly — it shows in `onUpdate` when a fetch starts and clears in `onUpdateVersions` once the types are resolved and applied. Please have a look.

Divyansh200102:
> Thanks, that looks great! One thing: please don't force-push to your branches. If you add commits and merge instead of rebasing it makes it easier to do subsequent reviews. We squash on merge anyway, so no need to worry about a clean git history. 

Apologies.I'll keep that in mind moving forward.

tonious:
> Would you like to be an official collaborator and in the author's list?

Sure!  I'd be happy to do reviews.  I don't think I need to be able to merge to master though.

bindings/wasm/examples/editor/editor.js
```diff
@@ -619,14 +628,17 @@ function clickSave(saveButton, filename, outputName) {
       container.appendChild(saveArrow.parentElement);
     }
     const link = document.createElement('a');
-    link.download = filename;
+
+    const scriptName = currentFileElement.textContent.trim() || 'manifold';
```
tonious: `currentFileElement` is set from `currentName`, IIRC.  This could be simplified to:

```js
const scriptName = currentName.trim() || 'manifold';
link.download = `${scriptName}.${extension}`;
```

or even:

```js
link.download = `${currentName.trim() || 'manifold'}.${extension}`;
```

tonious (approved):

bindings/wasm/examples/editor/editor.js
Divyansh200102: @tonious shouldn’t we use `currentFileElement.textContent.trim()` instead of `currentName`? since `currentName` is scoped inside `createEditor()`, using it here would throw a reference error.

tonious: Good point.

bindings/wasm/examples/editor/editor.js
elalish: Something isn't working right - it seems to always say Fetching Types... now. Can you repro @Divyansh200102 ?

Divyansh200102: Apologies, @elalish. I wish I could take a look, but I’m currently dealing with a medical emergency involving a family member. If someone else can check what’s happening, that would be great.

elalish: Oh no! So sorry to hear - yeah, we can take care of it, no worries.

PR #1563, opened 2026-03-03
feat: integrate Prettier 3.8.1 for WASM editor and scripts

Divyansh200102:
### **Pull Request Description**

* **`bindings/wasm/examples/editor/package.json`**: Added `prettier ^3.8.1` to enable standalone browser and local formatting.
* **`bindings/wasm/examples/editor/editor.js`**: Implemented lazy-loaded Prettier 3.0+ formatting with undo-stack preservation and cursor/scroll state restoration.
* **`scripts/format.sh`**: Migrated WebAssembly file formatting (`.js`, `.ts`, `.html`) from `clang-format` to `npx prettier`.

Screenshot
image

Note: The diff is large because Prettier was run across all relevant files, and they were reformatted according to the new Prettier configuration.


pca006132:
Is it possible to split the commit? One commit before the formatting, one commit after that. Otherwise it is hard to read the diff.

Divyansh200102:
@pca006132 I’ve integrated Prettier 3.8.1 into the editor and updated `scripts/format.sh` and `check_format.yml` to use Prettier for JS/TS/HTML instead of clang-format. Because of this the CI is now failing due to these formatting changes as it uses clang-format for checking the format.

Would you prefer proceeding with this full migration to Prettier for consistency between the editor and CI, or should I limit the changes to just the editor button and keep the existing clang-format setup?

If we proceed with this then I'll modfy `.github/workflows/check_format.yml` with the relevant changes as well.

pca006132:
I am fine either way, @elalish and @tonious may have some opinion about which one to use.

I think you should have one commit before formatting, to make the diff readable. Also, it seems that you forgot to add the updated package lock to the commit, and the wasm CI build is failing due to that.

Divyansh200102:
@pca006132 I have made the commits please review them when you get the time.

bindings/wasm/examples/editor/editor.js
```diff
@@ -43,8 +43,62 @@ const shareButton = document.querySelector('#share');
...
+  if (!editor) return;
+
+  const model = editor.getModel();
+  if (!model) return;
+
+  const originalCode = model.getValue();
+  const language = (model.getModeId?.() || '').toLowerCase();
+  const tabSize = model.getOptions?.().tabSize ?? 2;
+
+  let parser = 'babel';
```
pca006132: what does it mean by this? when should the parser be `babel`?

bindings/wasm/examples/editor/editor.js
```diff
@@ -43,8 +43,62 @@ const shareButton = document.querySelector('#share');
...
+  if (language.includes('typescript')) parser = 'typescript';
```
pca006132: do we have cases in which the editor is not writing typescript/javascript?

bindings/wasm/examples/editor/editor.js
```diff
@@ -43,8 +43,62 @@ const shareButton = document.querySelector('#share');
...
+
+ 
+  const viewState = editor.saveViewState();
+  editor.pushUndoStop();
+
+  try {
+    const [{ format }, estree, babel, typescript] = await Promise.all([
+      import('prettier/standalone'),
+      import('prettier/plugins/estree'),
```
pca006132: same here, are the plugins for estree and babel needed?

bindings/wasm/examples/editor/editor.js
```diff
@@ -43,8 +43,62 @@ const shareButton = document.querySelector('#share');
...
+    }
+
+    editor.pushUndoStop();
+    if (viewState) editor.restoreViewState(viewState);
+
+    console.log('Formatted with Prettier 3.8.1');
+  } catch (err) {
+    console.error('Prettier formatting failed:', err);
+    consoleElement.textContent += `Prettier formatting failed: ${err.message}\r\n`;
+    consoleElement.scrollTop = consoleElement.scrollHeight;
```
elalish: I believe you can remove these lines if you use console.log instead of console.error.

bindings/wasm/examples/editor/editor.js
```diff
@@ -43,8 +43,62 @@ const shareButton = document.querySelector('#share');
...
+    if (formatted !== originalCode) {
+      editor.executeEdits('prettier-format', [{
+        range: model.getFullModelRange(),
+        text: formatted,
+        forceMoveMarkers: true,
+      }]);
+    }
+
+    editor.pushUndoStop();
+    if (viewState) editor.restoreViewState(viewState);
```
pca006132, line 94: Curious what does this do. I think it is related to undo, but why do we need to do `pushUndoStop` after applying the edit and `restoreViewState`? What happens if we remove the second run?

elalish (approved):
What's the line length limit now? That was a big one we were missing before with Monaco's auto-formatter. 

I'm fine with updating the format of all this as long as it's pretty easy to set up VSCode to use prettier to auto-format these files on save. I assume that's no big deal?

Also probably a good idea to add the format commit to git-blame-ignore-revs. @tonious do you have any major branches outstanding? I don't want to create a bunch of conflicts for you.

bindings/wasm/examples/editor/editor.js
Divyansh200102: Babel is not (and will be removed), but I believe estree is strictly required. According to the [Prettier 3.0 Release Notes](https://www.google.com/search?q=https://prettier.io/blog/2023/07/05/3.0.0.html%23npm-package-file-structures-changed):

> A new plugin has been added:
> 
> plugins/estree.mjs (ESM version)
> plugins/estree.js (UMD version)
> If you use standalone version, this plugin should be loaded when printing JavaScript, TypeScript, Flow, or JSON.
> 
> 

 Without estree, the standalone version cannot print the AST back to text.

bindings/wasm/examples/editor/editor.js
Divyansh200102: @pca006132 I originally included babel as a fallback for plain JS/JSX, as it is Prettier's default canonical parser for JavaScript. However, since we only use this editor for typescript in ManifoldCAD, you are right we don't need the JS fallback. I will remove the babel plugin .

bindings/wasm/examples/editor/editor.js
Divyansh200102: I called `pushUndoStop()` because according to Monaco docs 

> **"Execute edits on the editor. The edits will land on the undo-redo stack, but no 'undo stop' will be pushed."**

Since no undo boundary is created automatically, we must explicitly call `pushUndoStop()` to create one after formatting.
If we remove the second call, subsequent user edits would continue in the same undo group as the formatting edit instead of being separated.

Divyansh200102:
> What's the line length limit now? That was a big one we were missing before with Monaco's auto-formatter.
> 
> I'm fine with updating the format of all this as long as it's pretty easy to set up VSCode to use prettier to auto-format these files on save. I assume that's no big deal?
> 
> Also probably a good idea to add the format commit to git-blame-ignore-revs. @tonious do you have any major branches outstanding? I don't want to create a bunch of conflicts for you.

@elalish The line length limit should be 80 characters, which is Prettier’s default printWidth. We can adjust it if you prefer a different limit. As for using prettier in vs code easily we can add a `.prettierrc` to ensure consistent settings across different systems. I'll also add the  formatting commit to git-blame-ignore-revs as suggested.

tonious:
> Do you have any major branches outstanding? I don't want to create a bunch of conflicts for you.

I've got a branch that's fairly in-depth with `show`/`only`.  Believe it or not, I usually try to pare down my PRs before I post them -- I'll adjust to the new format at that point.

bindings/wasm/examples/editor/editor.js
tonious: If it's too much detail to print in the UI, it's reasonable to `console.log()` a short version, and then `console.debug()` or `console.error()` an in depth version.  It'll be accessible via the browser console.

bindings/wasm/examples/editor/editor.js
```diff
@@ -14,78 +14,133 @@
...
+  try {
+    const [{ format }, estree, babel, typescript] = await Promise.all([
+      import("prettier/standalone"),
+      import("prettier/plugins/estree"),
+      import("prettier/plugins/babel"),
+      import("prettier/plugins/typescript"),
+    ]);
+
+    const plugins =
+      parser === "typescript" ? [typescript, estree] : [babel, estree];
```
tonious: The lazy-loading approach here is interesting.  When built, does it end up in a separate bundle?

It looks like there's a bunch of initialization code here that will be rerun every time the user formats their code, including the dynamic imports.  Is it reasonable to split the initialization and store `format`, `parser` and `plugins` somewhere?

E.g.: At the top level, something like this:
```typescript
const prettierWrapper = () => {
    let prettierParts = null;
    const lazyLoad = async () => {
      const [{ format }, estree, babel, typescript] = await Promise.all([
        import("prettier/standalone"),
        import("prettier/plugins/estree"),
        import("prettier/plugins/babel"),
        import("prettier/plugins/typescript"),
      ]);
      return { format, estree, babel, typescript };
    };

    const format = async (originalCode, parser="typescript", options = {}) => {
      if (!prettierParts) {
        prettierParts = await lazyLoad();
      }
      const {format, estree, babel, typescript} = prettierParts;
  
      return await format(originalCode, {
        parser,
        plugins: [parser === "typescript" ? typescript : babel, estree],
        ...options
      });
    };
    return format;
};

const format = prettierWrapper();
```

Then, in this `onclick` handler, something like:
```typescript
  const formatted = await format(
    originalCode,
    language.includes("typescript") ? "typescript" : "babel",
    {
      tabWidth: tabSize,
      singleQuote: true,
      trailingComma: "all",
    }
);
```

Er, treat this as pseudocode -- it's completely untested :joy:

bindings/wasm/examples/editor/editor.js
Divyansh200102: I'll keep the console.error(..., err) and console.log(), so the clean string goes to the user's UI, while the full stack trace is still safely dumped into the browser DevTools console for debugging.

Divyansh200102:
Can you please review the pr again and see if the changes look good?

image
 
Also I formated every file in binding/wasm* so I am not sure if that's something I should do or just the format the editor.js file? Please let me if should revert this change or not?

elalish:
>Also I formated every file in binding/wasm* so I am not sure if that's something I should do or just the format the editor.js file? Please let me if should revert this change or not?

I think that's good - the examples are the main thing that need to be formatted, since we don't want them to change in ManifoldCAD.org when someone hits the format button. I agree that having all of our JS/TS files formatted consistently is also good, so I agree with formatting it all. It also looks like prettier is catching more things than clang-format was.

I'll let @tonious give the final approval here, since he's working on these files the most. I assume it's expected that you'll need to update a few line numbers of stack trace tests due to the formatting changes?

Divyansh200102:
> I assume it's expected that you'll need to update a few line numbers of stack trace tests due to the formatting changes?

yes I just saw the CI jobs and there seems to be some conflicts as well. will push the fix soon.

tonious (approved):
I'm good with this, providing that checks pass, and that this branch is updated to be conflict free.

elalish:
@Divyansh200102 shall I go ahead and take over this PR? You've done a nice job and I'd like to see it merged.

Divyansh200102:
> @Divyansh200102 shall I go ahead and take over this PR? You've done a nice job and I'd like to see it merged.

Yes @elalish please go ahead and thank you for stepping in and helping it get merged. Appreciate it.

elalish:
@israrkhan921 This might be up your alley - would you be interested in taking over this PR and pushing it across the finish line?

israrkhan921:
Yes, I can take it, I will check the PR and update you soon, and then start work on this.

PR #1564, opened 2026-03-03
doc update Handling Precision Issues

AnshulPatil2005:
Adds a "Handling Precision Issues" section to bindings/wasm/documents/bindings.md explaining that IEEE 754 floating point drift (e.g. 0.1 + 0.2 = 0.30000000000000004) can silently break operations like union() when adjacent manifolds fail to share a face.

Documents the recommended workaround: scale coordinates up to integer-friendly units, use Math.round() to eliminate drift, then scale back down. Includes a concrete code example and links to discussion.

reference:
https://github.com/elalish/manifold/issues/1460  ->(doc update part)
https://github.com/elalish/manifold/discussions/1135

### DIFF
--- a/bindings/wasm/documents/bindings.md
+++ b/bindings/wasm/documents/bindings.md
@@ -39,33 +39,6 @@ const result = box.subtract(ball);
 You must manually `delete()` each object constructed by your scripts (both `Manifold` and `CrossSection`).
 See [discussion](https://github.com/elalish/manifold/discussions/256#discussioncomment-3944287).
 
-## Handling Precision Issues
-
-> [!WARNING]
-> JavaScript uses IEEE 754 binary floating point, which means expressions like `0.1 + 0.2` evaluate to `0.30000000000000004`. Manifold intentionally follows standard floating point math and does not compensate for this in its transform pipeline.
-
-This becomes a problem when two manifolds are meant to share a face. For example, translating a cube by `[0.2, 0.2, 0.2]` with size `[0.1, 0.1, 0.1]` produces a bounding box maximum of `0.30000000000000004` instead of `0.3`. That tiny epsilon gap means an adjacent manifold at exactly `0.3` will not properly union with it.
-
-The recommended workaround is to **work in larger, integer-friendly units**:
-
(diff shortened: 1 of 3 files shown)


bindings/wasm/documents/bindings.md
```diff
@@ -39,6 +39,33 @@ const result = box.subtract(ball);
 You must manually `delete()` each object constructed by your scripts (both `Manifold` and `CrossSection`).
 See [discussion](https://github.com/elalish/manifold/discussions/256#discussioncomment-3944287).
 
+## Handling Precision Issues
```
tonious: Same comment as on #1565

https://github.com/elalish/manifold/pull/1565#discussion_r2879242651

bindings/wasm/documents/bindings.md
```diff
@@ -39,6 +39,33 @@ const result = box.subtract(ball);
...
+> JavaScript uses IEEE 754 binary floating point, which means expressions like `0.1 + 0.2` evaluate to `0.30000000000000004`. Manifold intentionally follows standard floating point math and does not compensate for this in its transform pipeline.
+
+This becomes a problem when two manifolds are meant to share a face. For example, translating a cube by `[0.2, 0.2, 0.2]` with size `[0.1, 0.1, 0.1]` produces a bounding box maximum of `0.30000000000000004` instead of `0.3`. That tiny epsilon gap means an adjacent manifold at exactly `0.3` will not properly union with it.
+
+The recommended workaround is to **work in larger, integer-friendly units**:
+
+1. Scale coordinates up by a round factor (e.g., 1000) so your values are whole numbers
+2. Use `Math.round()` after any arithmetic to eliminate accumulated drift
+3. Perform all Manifold operations
+4. Scale the result back down if needed
```
elalish: There are actually a variety of other difficulties you may run into when trying to rely on our symbolic perturbation - our general term for the type of geometry you're describing is "marginal" and should generally be avoided if possible, as multiple "correct" topological answers are within rounding error. 

For instance: even with floating-point identical coordinates, if the planar faces have different triangulations, that can also lead to gaps and slivers.

PR #1565, opened 2026-03-03
docs: document degrees vs. radians in JS/TS rotation API

AnshulPatil2005:
Adds a "Degrees vs. Radians" section to bindings/wasm/documents/bindings.md clarifying that Manifold's .rotate() takes degrees, not radians a deliberate design choice that eliminates floating point error entirely for multiples of 90° and enables faster code paths.

Shows a correct vs. wrong usage example (Math.PI / 2 vs. 90), and includes a helper snippet for converting from radians-based math (e.g. Math.atan2). Links to issue #1262 where passing -89.999999999999 instead of -90 caused visible mesh cracks.

Closes part of #1460.

### DIFF
--- a/bindings/wasm/documents/tips.md
+++ b/bindings/wasm/documents/tips.md
@@ -30,3 +30,24 @@ const box = cube([size, size, size]).translate([offset, offset, offset]);
 ```
 
 See [discussion #1135](https://github.com/elalish/manifold/discussions/1135) for more context.
+
+## Degrees vs. Radians
+
+Manifold's rotation API takes **degrees**, not radians. This is a deliberate design choice: Manifold can eliminate floating point error entirely for multiples of 90°, and uses more efficient code paths when it detects such rotations.
+
+```js
+// Pass degrees directly
+box.rotate([90, 0, 0]);
+box.rotate([0, 45, 0]);
+```
+
+If your own code works in radians (e.g. when using `Math.atan2`), convert to degrees before passing to Manifold:
+
+```js
+const toDeg = (rad) => rad * (180 / Math.PI);
+
+const angle = Math.atan2(y, x); // radians
+box.rotate([0, 0, toDeg(angle)]);
+```
+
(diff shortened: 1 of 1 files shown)


bindings/wasm/documents/bindings.md
```diff
@@ -39,6 +39,30 @@ const result = box.subtract(ball);
...
+
+Manifold's rotation API takes **degrees**, not radians. This is a deliberate design choice: Manifold can eliminate floating point error entirely for multiples of 90°, and uses more efficient code paths when it detects such rotations.
+
+```js
+// Correct — pass degrees directly
+box.rotate([90, 0, 0]);
+box.rotate([0, 45, 0]);
+
+// Wrong — Math.PI / 2 introduces floating point error
+box.rotate([Math.PI / 2, 0, 0]); // ≈ 1.5707963..., not 90
```
tonious: Er, this is wrong because `Math.PI/2` is in *radians*.   I'd just omit this entirely.

bindings/wasm/documents/bindings.md
```diff
@@ -39,6 +39,30 @@ const result = box.subtract(ball);
 You must manually `delete()` each object constructed by your scripts (both `Manifold` and `CrossSection`).
 See [discussion](https://github.com/elalish/manifold/discussions/256#discussioncomment-3944287).
 
+## Degrees vs. Radians
```
tonious: So, `bindings.md` is a bit lower level -- it's included in the developer guide, but not the user guide.  This is (I think) user level documentation.

Maybe put it in a new file like `documents/tips.md` or something.  Then include it into the user documentation by adding it to user-facing declaration file.  E.g.:

https://github.com/elalish/manifold/blob/9c4f4d75dfd213cc955adff4ee0a8c6fcb3f49b3/bindings/wasm/types/manifoldCAD.d.ts#L46

elalish:
I think your last merge may have gone awry: your changes are all gone.

AnshulPatil2005:
> I think your last merge may have gone awry: your changes are all gone.

it seems i messed something up while clearing the conflicts with tips.md, will fix it in a while

PR #1568, opened 2026-03-05
`show()` and `only()` now work for all object types, including `GLTFNode` and all descendants in its scene-graph subtree.

AnshulPatil2005:
Fixes part of #1460(show() and only() should apply to all objects, including GLTFNodes and descendants).

##Changes

- Extended debug API support from `Manifold` only to `Manifold | BaseGLTFNode`.
- Added node-level debug mode resolution using ancestor traversal.
- Applied node debug mode during export for:
  - `GLTFNode` manifold meshes
  - `CrossSectionGLTFNode` meshes
  - `VisualizationGLTFNode` (imported glTF content), recursively across children
- Updated scene-builder caching to include debug mode in cache keys so debug/non-debug variants are not mixed.
- Hardened garbage collection wrapper to only register WASM-deletable results (or arrays), avoiding tracking plain `GLTFNode` returns.
- Added regression tests covering:
  - `show(rootNode)` affects root + descendants, not unrelated siblings
  - `only(rootNode)` keeps root + descendants normal and ghosts unrelated siblings

### DIFF
--- a/bindings/wasm/lib/debug.ts
+++ b/bindings/wasm/lib/debug.ts
@@ -145,8 +145,8 @@ export function getNodeMaterialByID(
 /**
  * @internal
  */
-export function getNodeMaterialOverride(
-    debugMode: NodeDebugMode): GLTFMaterial|undefined {
+export function getNodeMaterialOverride(debugMode: NodeDebugMode): GLTFMaterial|
+    undefined {
   if (debugMode === 'show') {
     return SHOW;
   }
--- a/bindings/wasm/lib/scene-builder.ts
+++ b/bindings/wasm/lib/scene-builder.ts
@@ -276,7 +276,8 @@ function createNodeFromCache(
       writeManifold(doc, node, nodeDef as GLTFNode, material, nodeDebugMode);
     }
 
-    const nodeCache = cachedNodeDefs ?? new Map<NodeDebugMode, Map<GLTFMaterial, Node>>();
+    const nodeCache =
+        cachedNodeDefs ?? new Map<NodeDebugMode, Map<GLTFMaterial, Node>>();
     const materialCache = new Map<GLTFMaterial, Node>();
     materialCache.set(material, node);
     nodeCache.set(nodeDebugMode, materialCache);
--- a/bindings/wasm/test/import.test.ts
+++ b/bindings/wasm/test/import.test.ts
@@ -27,20 +27,22 @@ const countVertices = (doc: Document) => {
   return getSceneVertexCount(scene, VertexCountMethod.UPLOAD_NAIVE);
 };
 
-const getNodeColor = (doc: Document, name: string): [number, number, number, number] => {
-  const node = doc.getRoot().listNodes().find((node) => node.getName() === name);
-  expect(node).toBeTruthy();
+const getNodeColor =
+    (doc: Document, name: string): [number, number, number, number] => {
+      const node =
+          doc.getRoot().listNodes().find((node) => node.getName() === name);
+      expect(node).toBeTruthy();
 
-  const mesh = node!.getMesh();
-  expect(mesh).toBeTruthy();
+      const mesh = node!.getMesh();
+      expect(mesh).toBeTruthy();
 
-  const primitive = mesh!.listPrimitives()[0];
-  expect(primitive).toBeTruthy();
+      const primitive = mesh!.listPrimitives()[0];
+      expect(primitive).toBeTruthy();
 
-  const material = primitive.getMaterial();
-  expect(material).toBeTruthy();
-  return material!.getBaseColorFactor();
-};
+      const material = primitive.getMaterial();
+      expect(material).toBeTruthy();
+      return material!.getBaseColorFactor();
+    };
 
 beforeEach(() => worker.cleanup());
 


elalish:
@tonious were you working on this too? How does this compare to yours? 

We may need to start getting more serious about making separate issues an assigning them to ourselves...

AnshulPatil2005:
> @tonious were you working on this too? How does this compare to yours?
> 
> We may need to start getting more serious about making separate issues an assigning them to ourselves...

hey, my bad for starting on it without mentioning. i’ve just been picking up issues randomly while exploring. should i ask for assignment before working on one?

pca006132:
I think we just need a way to synchronize progress.

Maybe you and @tonious can just comment on the meta issue about what you are about to work on.

AnshulPatil2005:
> I think we just need a way to synchronize progress.
> 
> Maybe you and @tonious can just comment on the meta issue about what you are about to work on.

yes, that works too.

tonious:
This is a bit of a different approach.  I was refactoring so that debug used only nodes.  This is actually simpler.  I'll check this out and play with it a bit.

bindings/wasm/lib/garbage-collector.ts
```diff
@@ -101,7 +101,9 @@ export const garbageCollectFunction = (originalFn: any): any => {
   return (...args: any) => {
     //@ts-ignore
     const result = originalFn(...args);
-    memoryRegistry.push(result);
+    if (Array.isArray(result) || typeof result?.delete === 'function') {
+      memoryRegistry.push(result);
+    }
```
tonious, line 106: If this is getting triggered, then something is being garbage collected that should not be.  Can you provide an example?

bindings/wasm/lib/debug.ts
```diff
@@ -101,16 +106,71 @@ const debug = (manifold: Manifold, map: Map<number, Mesh>) => {
...
  * transparent red. This is particularly useful for debugging `subtract()` as it
  * will allow you find the object even if it doesn't currently intersect the
  * result.
  *
- * @param manifold The object to show - returned for chaining.
+ * @param object The object to show - returned for chaining.
  */
-export function show(manifold: Manifold) {
-  return debug(manifold, shown);
+export function show<T extends Manifold|BaseGLTFNode>(object: T): T {
```
tonious, line 167: Bonus points for generics.

bindings/wasm/lib/debug.ts
```diff
@@ -101,16 +106,71 @@ const debug = (manifold: Manifold, map: Map<number, Mesh>) => {
...
+ */
+export function getNodeMaterialOverride(debugMode: NodeDebugMode): GLTFMaterial|
+    undefined {
+  if (debugMode === 'show') {
+    return SHOW;
+  }
+  if (debugMode === 'single') {
+    return undefined;
+  }
+  return ghost ? GHOST : undefined;
```
tonious, line 156: `null` rather than `undefined`, please.

bindings/wasm/lib/debug.ts
```diff
@@ -101,16 +106,71 @@ const debug = (manifold: Manifold, map: Map<number, Mesh>) => {
...
+    node: BaseGLTFNode, debugRoots: Set<BaseGLTFNode>): boolean {
+  let current: BaseGLTFNode|undefined = node;
+  while (current) {
+    if (debugRoots.has(current)) return true;
+    current = current.parent;
+  }
+  return false;
+}
+
+export type NodeDebugMode = 'show'|'single'|undefined;
```
tonious, line 119: `null` rather than undefined as part of the type definition. 

Instead of `getNodeDebugMode()`, maybe try mode specific helpers, e.g.: `isNodeShown()`

bindings/wasm/lib/scene-builder.ts
```diff
@@ -298,6 +324,10 @@ function copyNodeToDocument(
     }
   }
   applyTransformation(doc, nodeDef, targetNode);
+  const nodeMaterial = getNodeMaterialOverride(nodeDebugMode);
+  if (nodeMaterial) {
+    setNodeMaterialRecursive(targetNode, getCachedMaterial(doc, nodeMaterial));
+  }
```
tonious, line 330: In this case, copy geometry without materials?  Otherwise, we could end up with unreferenced materials present in the target glTF document.  If those materials include textures (i.e.: part of an imported model) this could add considerable weight to the final result.

bindings/wasm/lib/scene-builder.ts
```diff
@@ -368,13 +400,13 @@ export async function GLTFNodesToGLTFDoc(nodes: Array<BaseGLTFNode>) {
...
       // Copy from another glTF document in memory.
-      node = copyNodeToDocument(doc, nodeDef);
+      node = copyNodeToDocument(doc, nodeDef, nodeDebugMode);
       ++visualizationNodes;
 
     } else {
       // Manifold or CrossSection Object.
       // Previous meshes and materials are cached in `source2node`.
-      node = createNodeFromCache(doc, nodeDef, source2node);
+      node = createNodeFromCache(doc, nodeDef, source2node, nodeDebugMode);
```
tonious, line 409: This isn't a node.  Either it doesn't need `nodeDebugMode` or that flag has a misleading name.

bindings/wasm/lib/scene-builder.ts
```diff
@@ -350,7 +380,8 @@ export async function GLTFNodesToGLTFDoc(nodes: Array<BaseGLTFNode>) {
   addAnimationToDoc(doc);
 
   const node2gltf = new Map<BaseGLTFNode, Node>();
-  const source2node = new Map<Cacheable, Map<GLTFMaterial, Node>>();
+  const source2node =
+      new Map<Cacheable, Map<NodeDebugMode, Map<GLTFMaterial, Node>>>();
```
tonious, line 384: This is really messy.  The outer level map is going to have at most three keys.

Given that `nodeDebugMode`  and `GLTFMaterial` are related, I think this can be done without the outer level map.

elalish:
>I do have some concerns, though. As is, this PR is very tightly coupled. It would be difficult to refactor or expand (e.g.: to add CrossSection support, or deduplicate when copying glTF assets).

So @tonious, the real question is, which is closer to where we want to be, this PR or yours? We've got plenty of things for @AnshulPatil2005 to work on, so just give your honest preference.

elalish:
> > I do have some concerns, though. As is, this PR is very tightly coupled. It would be difficult to refactor or expand (e.g.: to add CrossSection support, or deduplicate when copying glTF assets).
> 
> So @tonious, the real question is, which is closer to where we want to be, this PR or yours? We've got plenty of things for @AnshulPatil2005 to work on, so just give your honest preference.

@tonious Friendly ping on the above... I actually sent you an email about this, but it probably ended up in your spam.

tonious:
Sorry, day job got in the way again.

I think I'd prefer to clean up my own implementation.  Not that I consider this wasted effort -- it has absolutely shown me a few things about how to approach this issue.

elalish:
No prob, I'm just glad you weren't hit by a bus or something @tonious :)

Thanks for taking the initiative, @AnshulPatil2005!

PR #1573, opened 2026-03-06
Add importModel example

AnshulPatil2005:
Adds an example showing how to use importModel() to load a model for display only, as opposed to importManifold() which converts it for CSG use.

Closes part of #1460

### DIFF
--- a/bindings/wasm/lib/import-model.ts
+++ b/bindings/wasm/lib/import-model.ts
@@ -132,8 +132,8 @@ function getImporter(identifier: ImportFormat|string) {
   return importers.find(im => im.importFormats.includes(format))!;
 }
 
-function getSourceFilename(
-    source: string|Blob|URL|ArrayBuffer): string|undefined {
+function getSourceFilename(source: string|Blob|URL|ArrayBuffer): string|
+    undefined {
   let path: string|undefined;
   if (source instanceof URL) {
     if (source.protocol === 'blob:' || source.protocol === 'data:') return;
@@ -207,7 +207,8 @@ export async function importModel(
     const [sourceNode] = sourceNodes;
     targetNode.node = sourceNode;
     targetNode.name = sourceNode.getName() || getSourceFilename(source);
-  } else {
+  }
+  else {
     targetNode.name = getSourceFilename(source);
   }
   if (typeof source === 'string') targetNode.uri = source;


tonious:
Nice!  Could you add a test?

It doesn't need to do much, just ensure that this example runs cleanly.

tonious:
The other tests are in `lib/worker.test.ts`.

bindings/wasm/test/examples/import-model.ts
```diff
@@ -0,0 +1,46 @@
...
+//
+// Unlike `importManifold()`, which converts the source model into a fully
+// manifold object that can be modified with CSG operations, `importModel()`
+// preserves the original model exactly — including all materials, textures
+// and geometry — but the result cannot be used in boolean operations.
+//
+// Use `importModel()` when you want to:
+//   - Display a reference model alongside your design.
+//   - Include a non-manifold or complex mesh that would be lost in conversion.
+//   - Preserve original materials and textures from the source file.
```
elalish: I believe we already preserve these when a GLB is imported as a manifold too, right?

bindings/wasm/test/examples/import-model.ts
```diff
@@ -0,0 +1,46 @@
...
+
+import {importModel, Manifold} from 'manifold-3d/manifoldCAD';
+
+const csg = async () => {
+  // glTF (.glb or .gltf) models have a defined scale of 1 unit to 1 metre.  On
+  // the other hand, manifoldCAD has a defined scale of 1 unit to 1 millimetre.
+  const metres = 1000;
```
elalish: I think this is confusing - it makes it sounds like we don't properly convert the units, but we do. Is the moon at an annoying scale? If so, let's just fix the GLB itself instead.

bindings/wasm/test/examples/import-model.ts
```diff
@@ -0,0 +1,46 @@
...
+  // `VisualizationGLTFNode`.  The original materials, textures and geometry
+  // are preserved as-is.
+  //
+  // Models can be imported by URL on manifoldCAD.org, or by path with the CLI.
+  // `new URL(relativeURL, import.meta.url)` resolves paths relative to this
+  // file on disk, or relative to the tab URL in a browser.
+  const moon = await importModel(new URL('./models/moon.glb', import.meta.url));
+
+  // Give the imported model a name so it appears clearly in scene viewers.
+  moon.name = 'Moon (display only)';
```
elalish: Is this set to anything by default? Perhaps we should set it to the filename automatically.

bindings/wasm/test/examples/import-model.ts
```diff
@@ -0,0 +1,46 @@
...
+
+  // The imported model can be transformed but not used in CSG operations.
+  moon.translation = [0, 0, 50 * metres];
+
+  // A regular Manifold object can be returned alongside a VisualizationGLTFNode
+  // by placing both in an array.  ManifoldCAD will render them together.
+  const platform = Manifold.cylinder(5 * metres, 60 * metres);
+
+  // Return the display-only model and the manifold geometry together.
```
elalish: I appreciate the comments, but it's a bit wordy. Can you condense a bit and try to avoid repetition (e.g. this line is kind of implied by the comment above).

bindings/wasm/test/examples/import-model.ts
AnshulPatil2005: Good call. I removed the unit-conversion explanation and the metres helper from this example, and kept it focused on importModel usage with direct values.

bindings/wasm/test/examples/import-model.ts
AnshulPatil2005: Updated: the intro no longer uses materials/textures as a differentiator vs importManifold(), and instead only calls out the CSG/display-only distinction.

bindings/wasm/test/examples/import-model.ts
AnshulPatil2005: Condensed the example comments to remove repetition and keep only the essential behavior notes.

bindings/wasm/test/examples/import-model.ts
AnshulPatil2005: Implemented this in code: importModel() now falls back to the source filename when imported nodes are unnamed, and scene export now applies VisualizationGLTFNode.name overrides. Added tests in import-model.test.ts.

PR #1574, opened 2026-03-07
CI cache

pca006132:


.github/workflows/manifold.yml
```diff
@@ -38,67 +104,103 @@ jobs:
...
+          . -B build
+        cmake --build build
+    - name: Test
+      if: matrix.cross_section == 'ON'
       run: |
+        ./build/test/manifold_test
+        # python tests
         export PYTHONPATH=$PYTHONPATH:$(pwd)/build/bindings/python
         python3 bindings/python/examples/run_all.py -e
         python3 -m pytest
```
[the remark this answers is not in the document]
pca006132: We don't need to run tests with cross-section disabled, since it can only disable tests instead of testing more code path. It is a strictly compile-only check CI check.

scripts/test-cmake-subdir.sh
```diff
@@ -21,7 +21,7 @@ EOT
 cp -r ../manifold ./
 mkdir build
 cd build
-cmake ..
+cmake -DFETCHCONTENT_SOURCE_DIR_CLIPPER2=$CLIPPER2 -DFETCHCONTENT_SOURCE_DIR_TBB=$TBB ..
```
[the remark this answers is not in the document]
pca006132: only add these when the variables are non-empty.

.github/workflows/manifold.yml
```diff
@@ -15,6 +15,68 @@ concurrency:
...
+    - uses: actions/checkout@v4
+    - uses: lukka/get-cmake@v4.2.3
+      with:
+        cmakeVersion: 4.0.2
+    - &clone-tbb
+      uses: ./.github/actions/git-clone
+      with:
+        name: tbb
+        url: https://github.com/oneapi-src/oneTBB.git
+        rev: f1862f38f83568d96e814e469ab61f88336cc595
```
elalish: Can we use tagged releases for any of these?

.github/workflows/manifold.yml
```diff
@@ -109,25 +208,37 @@ jobs:
...
-      run: |
-        # setup emscripten
-        git clone https://github.com/emscripten-core/emsdk.git
-        cd emsdk
-        ./emsdk install 3.1.64
-        ./emsdk activate 3.1.64
-    - uses: jwlawson/actions-setup-cmake@v2
+    - uses: actions/setup-node@v6
+      with:
+        node-version: 24
```
elalish: I think I used 25 somewhere else. Should probably be consistent.

.github/workflows/manifold.yml
```diff
@@ -195,80 +306,42 @@ jobs:
...
-    - uses: actions/checkout@v4
-    - uses: jwlawson/actions-setup-cmake@v2
-    - name: Build
-      run: |
-        mkdir build
-        cd build
-        cmake -DCMAKE_BUILD_TYPE=RelWithDebInfo -DMANIFOLD_STRICT=ON -DMANIFOLD_DEBUG=ON -DMANIFOLD_FUZZ=ON -DMANIFOLD_PYBIND=OFF -DCMAKE_CXX_COMPILER=clang++ ..
-        make
-
-  build_fuzzer_mac:
```
elalish, line 213: Are there some CI runs that I need to remove from the github required list?

.github/workflows/manifold.yml
pca006132: I did not find a way to do a shallow pull with tagged releases previously. I can probably ask LLM to see if there is a way.

.github/workflows/manifold.yml
pca006132: Some of the names have been changed. E.g., `build_fuzzer_mac` and `build_fuzzer` become `build_fuzzer` with an os matrix.

pca006132:
> There's a lot here - do you have a general description of what's changing at a high level?

Will do later today, busy with other stuff for now.

pca006132:
The copilot summary above was quite good, but I can expand on that.

1. This now caches built-in dependencies like tbb, clipper2, googletest, and nanobind. So we are not pulling them from GitHub for every CI job. This one is a self-written GitHub action script. I haven't found a good way to do shallow clones with release tags, so sha256 ids are used for now.
2. This caches npm, pip, emsdk, and apt installs using GitHub Actions.
3. Uses YML anchors and references to deduplicate steps like pulling in tbb dependency.
4. Reorganize actions a bit.
  1. Use one job for fuzzer builds, and use matrix for testing different OSes.
  2. For Linux builds, skip tests for `MANIFOLD_CROSS_SECTION=OFF`, since we only need to check if it compiles. The option should not affect code paths.
  3. Rewrite stuff a bit to avoid `cd`. Not a big deal.


Not yet solved:
1. CMake install cache on Windows. Maybe we can use a Chocolatey action for that.
2. Homebrew install cache on macOS. There are GitHub Actions for this, but there are not that many stars, so should review before using that.
3. Should gather code coverage for parallel build, but it is too slow for now to be run in the CI.

.github/workflows/manifold.yml
pca006132: Just checked, you are using 25 for publish_npm, will change this to 25.

.github/workflows/manifold.yml
pca006132: You need to change the required list to merge this.

PR #1575, opened 2026-03-07
Add 3MF importer and test coverage

AnshulPatil2005:
closes part of #1460 
Summary
Adds lib/export-3mf.test.ts with 15 tests covering the previously untested 3MF export functionality
Tests are split into tests that run without WASM (format metadata, ZIP structure, header options) and tests that require WASM (real manifold model exports)
What is tested
exportFormats metadata -> correct extension (3mf) and mimetype (model/3mf)
export-model integration -> supports() correctly recognizes 3mf and model/3mf, rejects unknown formats
ZIP structure -> output is a valid ZIP, contains required 3D/3dmodel.model and [Content_Types].xml entries
Header options ->default header contains ManifoldCAD.org, custom title/unit appear in model XML, default unit is millimeter
Real model exports ->cube, sphere, GLTFNode scene all export successfully with valid geometry (<vertices>, <triangles>)
Parent-child node hierarchy ->nested GLTFNode scenes export correctly and produce <component> references in the XML
Dispatch through export-model.toArrayBuffer -> correctly routes to the 3MF exporter via both extension and mimetype options


bindings/wasm/lib/export-3mf.test.ts
```diff
@@ -0,0 +1,206 @@
...
+  });
+
+  test('cube model XML contains vertices and triangles', async () => {
+    const script = `import {Manifold} from 'manifold-3d/manifoldCAD';\n` +
+        `export default Manifold.cube([1,1,1]);`;
+    const doc = await worker.evaluate(script);
+    const result = await toArrayBuffer(doc);
+    const files = unzipSync(new Uint8Array(result));
+    const modelXml = new TextDecoder().decode(files['3D/3dmodel.model']);
+    expect(modelXml).toContain('<vertices>');
```
elalish: I think some of these tests can be combined, like this one and the above. Less code for more testing is ideal.

bindings/wasm/lib/export-3mf.test.ts
```diff
@@ -0,0 +1,206 @@
...
+
+  test('exports a sphere', async () => {
+    const script = `import {Manifold} from 'manifold-3d/manifoldCAD';\n` +
+        `export default Manifold.sphere(10, 64);`;
+    const doc = await worker.evaluate(script);
+    const result = await toArrayBuffer(doc);
+    const files = unzipSync(new Uint8Array(result));
+    const modelXml = new TextDecoder().decode(files['3D/3dmodel.model']);
+    expect(modelXml).toContain('<vertices>');
+    expect(modelXml).toContain('<triangles>');
```
elalish: I think these tests might be tighter and simpler by making them round-trips - reimport the exported 3MF and then you have the JS APIs to test that the result is actually identical.

AnshulPatil2005: The round-trip tests currently export to GLB and reimport via importManifold to verify geometry using the JS API, since there's no 3MF importer. Was the intent to add a 3MF importer to enable a true 3MF round-trip, or the GLB round-trip sufficient here?

bindings/wasm/lib/export-3mf.test.ts
AnshulPatil2005: yes, on it.

bindings/wasm/lib/export-3mf.test.ts
elalish: oh, I forgot we were missing a 3MF importer - yes adding that would be ideal. Care to give it a shot? Should be simpler than GLB at least, since no need to worry about materials and properties.

AnshulPatil2005: Added a minimal 3MF importer (import-3mf.ts) to enable true round-trip tests. (One thing to note - importManifold can't be used directly here because it applies importTransform (scale×1000, rotate+90°) which is designed to convert GLTF's metres/Y-up to mm/Z-up. Since 3MF already stores geometry in mm/Z-up, that transform would corrupt the coordinates. Instead, gltfDocToManifold is called directly after importing, which verifies the geometry via the JS API without the unwanted coordinate conversion(also added this in the importer as comments)). I can integrate this fully into the import system (registering the 3MF importer and skipping importTransform for it), that could also be done.

tonious: If you integrate this into the importer system, it should transform to glTF style Y-up, 1 metre units.  This is because anything on the manifold side of that API boundary is expecting a conforming glTF document.  I don't want two code paths for 'conforming' or 'almost conforming' implementations.

AnshulPatil2005: I was thinking, I’ll keep this PR as test-only (no importer registration), so we don’t introduce a non-conforming importer path. Then I’ll open a follow-up PR to normalize 3MF to glTF conventions (Y-up, meters) and integrate it into the main importer system. If you’d prefer that integration in this PR instead, I can do that.

bindings/wasm/lib/import-3mf.ts
```diff
@@ -0,0 +1,111 @@
...
+  // the primitives (it calls getMaterial() on each primitive).
+  const defaultMaterial = doc.createMaterial();
+
+  const scene = doc.createScene();
+
+  // Find every <object> that has a <mesh> child and create one
+  // gltf-transform node per mesh.  Component objects (those with only
+  // <components>) are skipped; their transforms are all identity for
+  // geometry exported by manifoldCAD, so no information is lost.
+  const objectRe = /<object\b[^>]*>([\s\S]*?)<\/object>/g;
```
elalish: Instead of all this custom regex parsing, can we just use something like `fast-xml-parser` since 3MF is just another XML format? I think that'll make it easier to maintain.

bindings/wasm/lib/import-3mf.ts
```diff
@@ -0,0 +1,111 @@
...
+    const vertRe = /<vertex\b[^>]*>/g;
+    let vMatch: RegExpExecArray|null;
+    while ((vMatch = vertRe.exec(meshXml)) !== null) {
+      const tag = vMatch[0];
+      const x = tag.match(/\bx="([^"]+)"/)?.[1];
+      const y = tag.match(/\by="([^"]+)"/)?.[1];
+      const z = tag.match(/\bz="([^"]+)"/)?.[1];
+      if (x && y && z) positions.push(+x, +y, +z);
+    }
+    if (!positions.length) continue;
```
elalish: I don't see any parsing logic in here about mesh instances and transformations, but I'm pretty sure those exist in 3MF, and I believe we export them. I think if you make a test that exports several nodes that have the same manifold, you'll see.

bindings/wasm/lib/export-3mf.test.ts
elalish: Yeah, let's do it all here. I agree it'll get very confusing to have inconsistent conventions, and since the transforms are just part of the data and not actually applied to the verts yet, they'll just cancel out anyway.

AnshulPatil2005: okay, will do it here in this pr it self then

bindings/wasm/lib/import-3mf.ts
AnshulPatil2005: done

bindings/wasm/lib/import-3mf.ts
AnshulPatil2005: added it

AnshulPatil2005:
note ->I updated lockfile root version (3.3.2 -> 3.4.0) to match package.json

bindings/wasm/lib/import-3mf.ts
```diff
@@ -0,0 +1,290 @@
...
+    const y = toFiniteNumber(vertex.y);
+    const z = toFiniteNumber(vertex.z);
+    if (x == null || y == null || z == null) continue;
+
+    // 3MF is mm/+Z-up. Convert to glTF m/+Y-up by applying:
+    // scale = unitToMeters and rotation = -90 degrees around X.
+    const xf = x * unitToMeters;
+    const yf = y * unitToMeters;
+    const zf = z * unitToMeters;
+    positions.push(xf, zf, -yf);
```
elalish: There's no need to do all this computation - the `GLTFTransform.Document` also supports transforms, so you can simply apply a top-level transform once to the whole tree. Conveniently, we'll apply the inverse transform when pulling it into Manifold's data structures, and if we're careful we should be able to see that the composed transform is then the identity and avoid ever transforming the verts.

bindings/wasm/lib/export-3mf.test.ts
```diff
@@ -0,0 +1,163 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+import * as GLTFTransform from '@gltf-transform/core';
+import {unzipSync} from 'fflate';
+import {afterEach, beforeAll, expect, suite, test} from 'vitest';
+
+import {exportFormats, toArrayBuffer} from './export-3mf.ts';
```
elalish: For readability below, can we not destructure these, or else rename them to somehow include `3MF`? It's unclear when reading the tests where 3MF is being specified.

bindings/wasm/lib/export-3mf.test.ts
```diff
@@ -0,0 +1,163 @@
...
+        const script = `import {Manifold} from 'manifold-3d/manifoldCAD';\n` +
+            `export default Manifold.cube([100,100,100]);`;
+        const doc = await worker.evaluate(script);
+        const result = await toArrayBuffer(doc);
+        expect(result).toBeInstanceOf(ArrayBuffer);
+        expect(unzipSync(new Uint8Array(result))['3D/3dmodel.model'])
+            .toBeDefined();
+
+        const model = await importManifold(result, {mimetype: 'model/3mf'});
+        worker.cleanup();
```
elalish: Is this what's causing your tests to fail? Is it necessary given this is included in `afterEach`?

bindings/wasm/lib/export-3mf.test.ts
```diff
@@ -0,0 +1,163 @@
...
+        expect(model.volume()).toBeCloseTo(100 * 100 * 100, 1);
```
elalish, line 104: I believe we can tighten this test - for export/import round-trip we shouldn't be changing anything, so let's verify that the triVerts arrays are identical before and after. We can also check each vert position - maybe not exactly, but with a very small tolerance, which will depend on how much precision we use in our 3MF export, since the order of everything should be identical.

bindings/wasm/lib/export-3mf.test.ts
```diff
@@ -0,0 +1,163 @@
...
+            `export default parent;`;
+        const doc = await worker.evaluate(script);
+        const result = await toArrayBuffer(doc);
+        const files = unzipSync(new Uint8Array(result));
+        const modelXml = new TextDecoder().decode(files['3D/3dmodel.model']);
+        expect(modelXml).toContain('<vertices>');
+        expect(modelXml).toContain('<triangles>');
+        expect(modelXml).toContain('<component');
+
+        const model = await importManifold(result, {mimetype: 'model/3mf'});
```
elalish: Good test - looks like it caught something. Does this have something to do with the `no <build>` logic I saw you adding? It's always possible there's an error in the 3MF exporter too - curious what you find.

AnshulPatil2005: I checked this and it’s not from the no build path for normal round-trips.

export-3mf is emitting build items as expected. The issue was importer-side: I was doing per-vertex normalization. I changed that to a single top-level node transform (units + up-axis), and kept no <build> only as a fallback.

bindings/wasm/lib/export-3mf.test.ts
AnshulPatil2005: yea,I removed the explicit worker.cleanup() from the test body and now rely only on afterEach cleanup. It doesn’t look like the root cause of the failures, but this avoids double-cleanup so makes it cleaner.

bindings/wasm/lib/import-model.ts
```diff
@@ -458,8 +464,7 @@ function meshesToManifold(meshes: Array<Mesh>, tolerance?: number): Manifold {
     // We have a manifold object, but it is in the local coordinate system of
     // the original glTF-transform node.  Find that node, and transform it back
     // if possible.
-    const sourceMesh = mesh2mesh.get(mesh);
-    const sourceNode = sourceMesh ? mesh2node.get(sourceMesh) : null;
+    const sourceNode = mesh2sourceNode.get(mesh) ?? null;
```
elalish: Why is this needed? It seems like an extra map that contains the same info as the existing ones, but maybe I'm missing something?

bindings/wasm/lib/export-3mf.test.ts
```diff
@@ -0,0 +1,192 @@
...
+        expect(Array.from(roundTripMesh.triVerts))
+            .toEqual(Array.from(sourceMesh.triVerts));
+
+        const sourcePositions = meshToVec3Array(sourceMesh);
+        const roundTripPositions = meshToVec3Array(roundTripMesh);
+        expect(roundTripPositions.length).toBe(sourcePositions.length);
+        for (let i = 0; i < sourcePositions.length; ++i) {
+          for (let j = 0; j < 3; ++j) {
+            expect(Math.abs(roundTripPositions[i][j] - sourcePositions[i][j]))
+                .toBeLessThan(1.0e-5);
```
elalish: To avoid repetition, how about a function called `expectMeshesMatch`? In fact, I wonder if we have or need something like that in the glTF tests? In any case it should be available to both.

bindings/wasm/lib/import-model.ts
AnshulPatil2005: The extra map was a workaround for instancing: mesh2node was Map<GLTFMesh, GLTFNode>, so shared meshes overwrote earlier node entries and we lost per-instance transforms but your point is correct, I removed mesh2sourceNode and instead changed mesh2node to Map<GLTFMesh, GLTFNode[]>, then consume those nodes in order in meshesToManifold(). That keeps instance transforms correct without extra map.

bindings/wasm/lib/export-3mf.test.ts
AnshulPatil2005: it is shared via test/util.ts; 3MF uses expectMeshesMatch and is available for dlTF tests too.

AnshulPatil2005:
@elalish how does it look now ?

bindings/wasm/lib/import-model.ts
elalish: Ah, okay. Honestly I think now your first way may have been clearer after all. Perhaps a comment is all that's needed. 

Anyway, this all looks good to me, though I'd like to know if @tonious has any feedback.

AnshulPatil2005: Yea, was confused on the decision but implemented it as it prevented extra map at the cost of a small confusion.

elalish:
ugh, sorry - mind updating package-lock to fix the conflict? Then we can merge this.

AnshulPatil2005:
> ugh, sorry - mind updating package-lock to fix the conflict? Then we can merge this.

done

PR #1580, opened 2026-03-13
Restore ManifoldCAD PWA paths and manifest metadata

israrkhan921:
This updates ManifoldCAD PWA configuration so it works reliably in local/dev and non-root hosting setups.

What changed
1. Switched PWA asset references from root-absolute paths to relative paths.
2. Updated service worker registration to use a URL resolved from the current module and a sane app scope.
3. Added screenshot form factors in the manifest so install UI checks pass on desktop.
4. Added optional manifest fields recommended by DevTools:
- protocol_handlers
- display_override

Files touched
- editor.js
- index.html
- manifest.json

Why
The previous setup assumed root deployment for manifest/service worker/icon URLs. That can break PWA behavior when served from subpaths. This change keeps behavior the same for root hosting and fixes path resolution for other environments.

Manual verification
1. Start editor locally and open http://localhost:5173/
2. DevTools → Application → Manifest
- manifest loads correctly
- screenshots show wide and narrow form factors
3. DevTools → Application → Service Workers
- service worker registers successfully
- update on reload works as expected
4. Reload once online, then offline reload confirms app shell caching behavior

PR #1583, opened 2026-03-14
Follow-up fix for deployed PWA service-worker path and type-fetch 404 noise

israrkhan921:
After #1580, deployment testing found issues not seen locally.
This PR focuses only on deployment reliability and 404-noise reduction.
It registers the service worker relative to the current page URL (with explicit registration error logging), keeps ManifoldCAD typings local while preventing unnecessary manifold-3d CDN probes, and hardens the service-worker fetch path by handling only GET requests, adding network try/catch with cache fallback, plus skipWaiting()/clients.claim() for faster updates.
 The expanded type-indicator timing/state-machine changes were removed from this PR to keep scope small and reviewable; indicator UX changes can be proposed separately if needed.

### DIFF
--- a/bindings/wasm/examples/editor/editor.js
+++ b/bindings/wasm/examples/editor/editor.js
@@ -344,91 +344,8 @@ async function createEditor() {
   }
 
   // Initialize auto typing on monaco editor.
-  let typeIndicator = document.getElementById('type-indicator');
-  if (!typeIndicator) {
-    console.warn(
-        '[AutoTypings] #type-indicator not found; creating fallback element.');
-    const headerTitle = document.querySelector('.header p');
-    if (headerTitle) {
-      const createdIndicator = document.createElement('span');
-      createdIndicator.id = 'type-indicator';
-      createdIndicator.style.fontSize = '14px';
-      createdIndicator.style.marginLeft = '12px';
-      createdIndicator.style.fontWeight = 'normal';
-      createdIndicator.style.opacity = '0.8';
-      headerTitle.appendChild(createdIndicator);
-      typeIndicator = createdIndicator;
-    }
-  }
+  const typeIndicator = document.getElementById('type-indicator');
   self.window.typecache = new LocalStorageCache();
-  let typeIndicatorFrame = 0;
-  let autoTypings;
-  let resolvingSince = 0;
-  let resolveEventUntil = 0;
-
-  const MIN_INDICATOR_MS = 350;
-
-  const setTypeIndicator = (resolving, reason = '') => {
-    if (!typeIndicator) return;
-    typeIndicator.textContent = resolving ? 'Fetching Types...' : '';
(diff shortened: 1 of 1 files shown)


bindings/wasm/examples/editor/editor.js
```diff
@@ -335,24 +344,123 @@ async function createEditor() {
   }
 
   // Initialize auto typing on monaco editor.
-  const typeIndicator = document.getElementById('type-indicator');
+  let typeIndicator = document.getElementById('type-indicator');
+  if (!typeIndicator) {
```
elalish: This block feels unnecessary - we defined this in our own HTML, so if it's not there, that's on us anyway. I think it's fine to make this a `const typeIndicator = document.getElementById('type-indicator')!;`

bindings/wasm/examples/editor/editor.js
```diff
@@ -335,24 +344,123 @@ async function createEditor() {
...
+            Boolean(autoTypings?.isResolving) || frameNow < resolveEventUntil;
+
+        if (stillResolving) {
+          typeIndicatorFrame = requestAnimationFrame(waitForResolve);
+          return;
+        }
+
+        const visibleFor = frameNow - resolvingSince;
+        if (visibleFor < MIN_INDICATOR_MS) {
+          typeIndicatorFrame = requestAnimationFrame(waitForResolve);
```
elalish: So, thank you for fixing this, but this little indicator is a very small feature. Is there any way to simplify this implementation? I'm tempted to remove it entirely if it really requires this much code to make function.

bindings/wasm/examples/editor/editor.js
```diff
@@ -335,24 +344,123 @@ async function createEditor() {
   }
 
   // Initialize auto typing on monaco editor.
-  const typeIndicator = document.getElementById('type-indicator');
+  let typeIndicator = document.getElementById('type-indicator');
```
elalish: This is actually a great example of where it's helpful to split PRs. If you wouldn't mind splitting the type-indicator change into it's own PR, I think the rest of this is ready to merge. Makes it easier to review too.

elalish (changes requested):
Thanks!

israrkhan921:
@elalish 
Ok Sir 
I will let you know

israrkhan921:
yes Sir
I made a follow-up simplification for the fetching types UI:
I removed the indicator flow entirely and kept only AutoTypings error logging. This avoids extra UI state logic while preserving the deployment and type-fetch reliability fixes from #1583.
I can open this as a small separate PR for review.

israrkhan921:
This PR is superseded by a clean follow-up PR opened from branch followup/remove-fetching-types-ui, which contains only the fetching-types UI simplification (single commit, minimal diff).
Closing this one to avoid carrying over the previous branch history.

Then close #1586.

elalish:
I think that you missed that this one already got merged...

israrkhan921:
Yes I have mentioned

PR #1587, opened 2026-03-15
refactor(editor): restore minimal fetching types hint

israrkhan921:
1.  Restores a minimal fetching-types hint in the editor header.
2.  Shows static "Fetching types..." while autoTypings is resolving imports.
3.  Clears the hint as soon as resolving finishes, without timers, fallback DOM creation, or extra state.
4.  Keeps the follow-up service-worker and type-fetch reliability fixes unchanged.

### DIFF
--- a/bindings/wasm/examples/editor/editor.js
+++ b/bindings/wasm/examples/editor/editor.js
@@ -350,7 +350,8 @@ async function createEditor() {
 
   const syncTypeIndicator = () => {
     if (!typeIndicator || !autoTypings) return;
-    typeIndicator.textContent = autoTypings.isResolving ? 'Fetching types...' : '';
+    typeIndicator.textContent =
+        autoTypings.isResolving ? 'Fetching types...' : '';
     typeIndicatorFrame =
         autoTypings.isResolving ? requestAnimationFrame(syncTypeIndicator) : 0;
   };


israrkhan921:
@elalish 
I went with the simplest version for now and removed the fetching-types indicator entirely.

If we want a tiny UX hint later, I’d keep it to one rule only:
show static “Fetching types…” while autoTypings.isResolving is true, and clear it when false.

No timers, no minimum display duration, no fallback DOM creation, and no state machine.

elalish:
>If we want a tiny UX hint later, I’d keep it to one rule only:
show static “Fetching types…” while autoTypings.isResolving is true, and clear it when false.

This sounds reasonable - let's try that instead.

israrkhan921:
Ok Sir

elalish:
Hmm, I think this needs more work. Did you test this manually? I'm seeing a lot of 404s in the terminal for voronoi and especiall godel, escher, bach examples. The fetching label does appear to go away eventually now, but then it also causes the example to become "new script" as though its been edited, when it hasn't. This is a bug we've had occasionally now, so it would be great if you could look into it.

I also feel like I couldn't get the new deployment of the PWA with a few regular refreshes, but I needed to do a hard refresh with cmd-r. I feel like that shouldn't be necessary? I didn't test enough to be 100% sure of the behavior, so would be great to know if you can repro.

israrkhan921:
> Hmm, I think this needs more work. Did you test this manually? I'm seeing a lot of 404s in the terminal for voronoi and especiall godel, escher, bach examples. The fetching label does appear to go away eventually now, but then it also causes the example to become "new script" as though its been edited, when it hasn't. This is a bug we've had occasionally now, so it would be great if you could look into it.
> 
> I also feel like I couldn't get the new deployment of the PWA with a few regular refreshes, but I needed to do a hard refresh with cmd-r. I feel like that shouldn't be necessary? I didn't test enough to be 100% sure of the behavior, so would be great to know if you can repro.

i am working now on it, to solve these errors.

PR #1593, opened 2026-03-16
fix(editor): stabilize refresh handling and offline behavior

israrkhan921:
Fix editor state transitions during auto-typing refreshes and reduce noisy logging from worker/bundler paths.

Improve PWA/dev offline reliability by updating service worker registration, cache versioning, app-shell precache, and local font asset usage for the Godel example.

Use same-origin manifest/favicon assets and move model-viewer import out of inline HTML to avoid Vite html-proxy issues.

### DIFF
--- a/bindings/wasm/examples/editor/editor.js
+++ b/bindings/wasm/examples/editor/editor.js
@@ -50,9 +50,8 @@ if (navigator.serviceWorker) {
   } else {
     // Resolve against the current page URL so production asset paths don't
     // redirect registration into /assets.
-    // v=4 is only a URL cache-bust for the worker script fetch path.
     const serviceWorkerUrl =
-        new URL('./service-worker.js?v=4', window.location.href);
+        new URL('./service-worker.js', window.location.href);
     navigator.serviceWorker
         .register(serviceWorkerUrl, {scope: './', updateViaCache: 'none'})
         .then(async registration => {
@@ -410,13 +409,12 @@ async function createEditor() {
   // broken declaration re-exports to non-existent source files. Avoid CDN
   // probes for those packages to keep refreshes quiet.
   // This skip list only affects Monaco auto-typing CDN lookups, not runtime
-  // imports; scoped @thi.ng entries are included to suppress noisy probes.
+  // imports.
   const jsDelivrResolver = new JsDelivrSourceResolver();
   const skippedTypingPackages =
       new Set(['manifold-3d', 'text-shaper', '@types/require']);
   const shouldSkipTypingPackage = packageName => {
-    return skippedTypingPackages.has(packageName) ||
-        packageName.startsWith('@thi.ng/');
+    return skippedTypingPackages.has(packageName);
   };
   const sourceResolver = {
     resolvePackageJson: async (packageName, version, subPath) => {


israrkhan921:
Sorry again, sir! 
I should have verified everything properly before adding changes.

The main mistake on my side was simple, I didn’t refresh the browser at right time, so I was looking at stale output and misunderstanding what was actually failing.
Because of that why, I treated noisy logs as runtime failures, even though the core geometry pipeline was still working.

Here is the full breakdown of what happened and what I have fixed.

I reproduced the issue carefully in the WASM editor examples and checked both paths:
1. Run flow: code execution and mesh export.
2. Refresh flow: editor/UI updates background type resolution, and cache behavior.

At first, the console looked like many failures. 
After tracing each message source I found three different causes mixed together:
1. Worker and bundler logging noise.
2. Monaco auto-typing network probes.
3. Browser policy and caching behavior.

So this was not one single bug. 
It was a signal-vs-noise problem across several layers.

1. Example state randomly became New Script  
What was happening:  
- The editor change handler could treat background typing refresh as user edits.
- In some paths it relied on stale captured script state.
- That combination could falsely convert examples into New Script.

What I changed:  
- Used the active script name at event time.
- Ignored flush/change events while type resolution is active.
- Prevented non-user typing refresh from triggering example conversion.

Why this works:  
- Background typing updates and real user edits are now separated.
- Decisions use live state instead of stale state.

Result:  
- Example switching/editing is stable.
- Random New Script conversion no longer occurs.

2. Fetching log spam in console  
What was happening:  
- Per-request fetch logs were emitted in the bundler path.
- Worker logs were mirrored in a way that duplicated output.
- Editor log UI showed repetitive fetch lines with low value.

What I changed:  
- Removed per-request fetch log emission in bundler.
- Stopped duplicate worker console mirroring while keeping useful forwarding.
- Filtered repetitive Fetching http lines in the editor log display path.

Why this works:  
- Important execution logs remain.
- Transport-level noise is reduced.

Result:  
- Console is much cleaner.
- Useful run/export logs are still visible.

3. Monaco auto-typing jsDelivr 404 noise  
What was happening:  
- Auto-typing aggressively probed dependency typing paths.
- Some dependencies expose typings in ways that cause repeated failed probes.

What I changed:  
- Added targeted resolver skips for:
  - manifold-3d
  - text-shaper
  - @types/require
  - @thi.ng/*
- Reduced recursion depth for package/file probing.

Why this works:  
- Known noisy paths are skipped.
- Probe volume is reduced without disabling useful typing support.

Result:  
- 404 noise dropped significantly.
- Editor typing support still works.

4. Favicon blocked by browser policy (COEP/CORP)  
What was happening:  
- Favicon came from cross-origin URL and got blocked under policy.
- This caused warnings and inconsistent icon behavior.

What I changed:  
- Added same-origin local favicon asset.
- Updated HTML to use local icon and manifest path.

Why this works:  
- Same-origin assets avoid cross-origin blocking.

Result:  
- Same look.
- Cross-origin favicon warnings removed.

5. Offline/cache behavior during local testing  
What was happening:  
- Local behavior could become stale depending on service worker/cache state.
- Offline testing in dev mode also showed expected Vite HMR websocket warnings, which looked scary but were not app failures.
- Gödel example had a real offline failure due to remote font fetching.

What I changed:  
- Reworked service worker flow for reliable offline behavior:
  - deterministic registration/update behavior,
  - cache versioning cleanup,
  - app-shell precache for key UI assets.
- Added local Orbitron font asset and switched Gödel example font loading to same-origin first (with safe fallback).

Why this works:  
- Required assets are locally cached and available offline.
- Example-specific remote dependency no longer blocks Gödel offline.
- Dev websocket warnings are understood as expected HMR behavior in offline mode, not runtime defects.

Result:  
- Offline mode works reliably for the editor and examples (including Gödel after font localization).
- Remaining Vite websocket warning in offline dev mode is expected and non-functional.

Important clarification  
- Lit dev-mode warnings from model-viewer are development warnings, not functional failures.
- Core runtime/export path remained healthy throughout.
- Most fixes were around state correctness, diagnostics quality, and offline reliability.

What I learned  
1. Always refresh and confirm runtime state before diagnosing.
2. Separate log sources before deciding severity.
3. Reduce noise carefully so real errors stay visible.
4. Treat service worker/cache behavior as a first-class part of local debugging.

Final apology  
Sorry again, sir!  
Next time I will confirm refresh state, verify the exact source of errors first, and only then add changes.
after reviewing it i will be waiting to your response and feedback Sir.

elalish:
Okay, so AI code gens tools are great and all, and you're welcome to use them, but a couple things: 

1. Using AI for your PR descriptions and replies is not a great idea - even if you're not confident with english, I will trust your own voice more than I will trust the generic AI descriptions. The one above is particularly bad - remember that brevity is the soul of wit. I'm looking for more information from fewer words - longer descriptions are not better.
2. AI can code decently these days, but testing is harder because the problem is much more open. I need confidence that you understand what we're trying to build here well enough to really test it - manually! So, before I review this, please post a "before" screen recording showing the problem we're solving (have the dev tools panel open too), and then an "after" showing how it's improved with this PR. 
3. Remember that the most important thing in contributing code, in open source or a company, is trust. I need to know that you understand the feedback I give and that you can identify and fix problems on your own. I can't catch every mistake in code review, so I need to know that I can trust you to find most of the problems yourself. AI just isn't that good yet.

israrkhan921:
> Okay, so AI code gens tools are great and all, and you're welcome to use them, but a couple things:
> 
> 1. Using AI for your PR descriptions and replies is not a great idea - even if you're not confident with english, I will trust your own voice more than I will trust the generic AI descriptions. The one above is particularly bad - remember that brevity is the soul of wit. I'm looking for more information from fewer words - longer descriptions are not better.
> 2. AI can code decently these days, but testing is harder because the problem is much more open. I need confidence that you understand what we're trying to build here well enough to really test it - manually! So, before I review this, please post a "before" screen recording showing the problem we're solving (have the dev tools panel open too), and then an "after" showing how it's improved with this PR.
> 3. Remember that the most important thing in contributing code, in open source or a company, is trust. I need to know that you understand the feedback I give and that you can identify and fix problems on your own. I can't catch every mistake in code review, so I need to know that I can trust you to find most of the problems yourself. AI just isn't that good yet.

i got it sir
i will follow rules and instructions and thanks for feedback

israrkhan921:
@elalish 
so what is my next task sir?

israrkhan921:
its a short video of new changes in PR#1593

https://drive.google.com/file/d/1Nz9RHtZx8c0ULW758LkJ8lb-b5fhQt8g/view?usp=drive_link

israrkhan921:
@elalish 
Sir i am waiting for your response and also for reviewing this PR#1593

pca006132:
> Sir i am waiting for your response and also for reviewing this PR#1593

@israrkhan921 We typically reply in 1-2 days, but it may take longer if we have something else to do. Please be patient about it.

israrkhan921:
ok Sir
i will wait

israrkhan921:
@elalish @pca006123
before changes in PR#1593:
 before PR#1593 changes.mp4
<https://drive.google.com/file/d/1Fs08m61ijQqAZzmbVKboeMjzL99OnHD-/view?usp=drive_web>
after changes in PR#1593:
 after PR#1593 changes.mp4
<https://drive.google.com/file/d/1aTi2OH7_7MYgLeKJF3QX2StDFesSiONX/view?usp=drive_web>

bindings/wasm/examples/editor/editor.js
```diff
@@ -28,13 +30,49 @@ const CODE_START = '<code>';
...
+  if (disableServiceWorker) {
+    // Explicit escape hatch for debugging cache-related issues.
+    navigator.serviceWorker.getRegistrations().then(registrations => {
+      registrations.forEach(registration => registration.unregister());
+    });
+  } else {
+    // Resolve against the current page URL so production asset paths don't
+    // redirect registration into /assets.
+    const serviceWorkerUrl =
+        new URL('./service-worker.js?v=4', window.location.href);
```
elalish: What does this `v=4` do? Does it need to be updated when the cache version changes? Should it be a named constant that's shared instead?

And please remember to tell me when I'm wrong! 😄

bindings/wasm/examples/editor/editor.js
```diff
@@ -28,13 +30,49 @@ const CODE_START = '<code>';
...
+    navigator.serviceWorker
+        .register(serviceWorkerUrl, {scope: './', updateViaCache: 'none'})
+        .then(async registration => {
+          await navigator.serviceWorker.ready;
+          if (!navigator.serviceWorker.controller) {
+            const key = 'manifoldcad-sw-controller-refresh';
+            if (!window.sessionStorage.getItem(key)) {
+              window.sessionStorage.setItem(key, '1');
```
elalish, line 63: What does this 1 mean?

bindings/wasm/examples/editor/editor.js
```diff
@@ -366,29 +404,41 @@ async function createEditor() {
...
-  // that package path. This prevents noisy 404s in production logs.
+  // We inject manifold-3d typings locally above, and text-shaper publishes
+  // broken declaration re-exports to non-existent source files. Avoid CDN
+  // probes for those packages to keep refreshes quiet.
   const jsDelivrResolver = new JsDelivrSourceResolver();
+  const skippedTypingPackages =
+      new Set(['manifold-3d', 'text-shaper', '@types/require']);
+  const shouldSkipTypingPackage = packageName => {
+    return skippedTypingPackages.has(packageName) ||
+        packageName.startsWith('@thi.ng/');
```
elalish: Do `@thi.ng` packages have broken types as well? Might be good to add a comment about these as well. I'm a little surprised the declarations are broken - any chance we missed something?

bindings/wasm/examples/editor/editor.js
```diff
@@ -366,29 +404,41 @@ async function createEditor() {
...
+      if (shouldSkipTypingPackage(packageName)) return '';
       return jsDelivrResolver.resolveSourceFile(packageName, version, path);
     }
   };
 
   autoTypings = await AutoTypings.create(editor, {
     sourceResolver,
     sourceCache: self.window.typecache,
+    packageRecursionDepth: 1,
+    fileRecursionDepth: 2,
```
elalish, line 436: These feel a little arbitrary - can  we at least have comments about how you chose these limits?

bindings/wasm/examples/editor/editor.js
```diff
@@ -593,8 +645,12 @@ function createWorker() {
       createWorker();
 
     } else if (message?.type === 'log') {
-      consoleElement.textContent += message.message + '\r\n';
-      consoleElement.scrollTop = consoleElement.scrollHeight;
+      const logMessage = String(message.message ?? '');
+      // Hide noisy per-module CDN fetch traces, keep other worker logs.
```
elalish, line 653: 👍

elalish (changes requested):
Looking good, thanks! And I appreciate the video you made. Just a few small things.

bindings/wasm/examples/editor/editor.js
israrkhan921: > What does this `v=4` do? Does it need to be updated when the cache version changes? Should it be a named constant that's shared instead?
> 
> And please remember to tell me when I'm wrong! 😄

v=4 is only for cache-bust in service worker URL.
Browser sees service-worker.js?v=4 as new URL, so it re-fetches worker script.
It is not automatically connected to cache version name.
So no, it does not always need update when cache version changes, unless we want both linked.
Also we already use updateViaCache: 'none' and registration.update(), so maybe this v=4 is extra.
I agree:
if we keep it, better use shared named constant with comment.
if not needed, better remove it.

bindings/wasm/examples/editor/editor.js
israrkhan921: > What does this 1 mean?

1 is just flag in sessionStorage.
Meaning is:refresh already tried one time in this tab.
I use this so page does not go infinite reload when service worker controller is still null.
So 1 has no special meaning, only marker value.

bindings/wasm/examples/editor/editor.js
israrkhan921: > Do `@thi.ng` packages have broken types as well? Might be good to add a comment about these as well. I'm a little surprised the declarations are broken - any chance we missed something?

I don’t think all @thi.ng packages have broken types.
This skip is mainly for Monaco auto-typing CDN fetch noise/fail, not runtime import issue.
For manifold-3d we already inject local d.ts.
For text-shaper we saw bad declaration re-export path.
For @thi.ng rule, yes maybe we missed better root fix.
This rule can be too broad, I will add comment to explain why this skip is here.
but in the future we will fix it in best way

bindings/wasm/examples/editor/editor.js
israrkhan921: > These feel a little arbitrary - can we at least have comments about how you chose these limits?

Yes, you are right
Without comment these values look random
I set these limits to reduce Monaco auto-typing fetch noise and keep editor stable + fast (also better for offline)
packageRecursionDepth: 1 = only shallow package typing lookup, no deep transitive crawl
fileRecursionDepth: 2 = allow small local import chain, but stop before big recursive file graph
So not random numbers, just conservative limits for stability/performance
I will add clear comments for this

bindings/wasm/examples/editor/editor.js
elalish: I'd like to have type info for @thi.ng packages - they seem like a likely dependency people may directly import. Let's remove this case so that we don't hide the problem. And I'm curious if your other fixes here will cover the problem anyway?

bindings/wasm/examples/editor/editor.js
elalish: When in doubt, remove it.

israrkhan921:
i am checking it now

israrkhan921:
@elalish 
i have removed both:

after removing that :    https://drive.google.com/file/d/1C3vjqrfH50fGx8AZsEpRjSBxLIxviX7X/view?usp=drive_link

hrgdavor:
I was in similar boat as @israrkhan921 when I started contributing to other projects, I would be eager to get feedback, so at first I was annoying ppl, and later I would take few days to respond to feedback(switch of roles and being the slow er one), so after few PR's I became less annoying :D

anyway, nice to see even more contributors

PR #1594, opened 2026-03-16
ci: add cross-platform determinism check (Part 1)

AnshulPatil2005:
https://github.com/opencax/GSoC/issues/114
Adds a new workflow (`.github/workflows/determinism.yml`) that verifies
exact mesh output consistency across Linux, macOS, and Windows.

**Three export jobs** (one per platform) each:
- Build `manifold_test` with `MANIFOLD_PAR=OFF` and a minimal config
- Run only `Boolean.Perturb3` and `Boolean.MultiCoplanar` with the `-e`
  flag to export canonical OBJ files via `GetMeshGL64()`
- Upload `nastyGear.obj` and `coplanar.obj` as artifacts
   -- Started with two floating-point-sensitive cases (`Perturb3` rotated 
  gear with many boolean subtractions, `MultiCoplanar` coplanar face 
  decisions) as a baseline; more cases can be added in follow-up PRs based 
  on feedback

**One compare job** then downloads all three artifact sets, SHA256 hashes
each file per platform, and fails with a human-readable diff if anything
diverges.

## Notes

- `MANIFOLD_PAR=OFF` is intentional removes threading as a source of
  ordering nondeterminism for this check
- Test cases chosen for floating-point stress: `Perturb3` (gear with
  rotated cubes) and `MultiCoplanar`
- Kept as a separate workflow so it can be tuned or disabled independently
  of the main CI


pca006132:
I would expect you to reuse our existing CIs, as they already compile for these targets.

You probably want to wait for #1574, since that changes the YML file a lot.

AnshulPatil2005:
integrated the check into existing CI (manifold.yml) and removed the standalone workflow.

For the required cross-platform compare, I switched from Perturb3 (nastyGear.obj) to Cubes + Coplanar (cubes.obj, coplanar.obj) because nastyGear was consistently producing platform-dependent mismatches(expected behaviour) and keeping the gate red. I expected cubes + coplanar to give green checks, but current runs still show cross-platform mismatches under exact byte/hash comparison:

cubes.obj: Linux vs macOS has real mesh/order differences,
coplanar.obj: Linux/macOS match, but Windows differs at very small floating-point text precision.
So at the moment we have mixed failures: some real cross-platform differences, and some compare-noise (line ending / tiny FP formatting sensitivity).

pca006132:
How did you merge the master branch into this branch? It is hard to read the diff. Can you do a rebase?

It is good that we catch differences. And the question here is why do we have these differences. We can start from simpler models with just 1 boolean operation first. But do record those with differing results, we want to eventually make them pass this test.

Things we can do here is:

1. Run a line ending replacement for windows.
2. Make a new version of the obj dump that dumps floating point as hex.
3. Dump the intermediate data of boolean3 and boolean_result.

AnshulPatil2005:
> How did you merge the master branch into this branch? It is hard to read the diff. Can you do a rebase?
> 
> It is good that we catch differences. And the question here is why do we have these differences. We can start from simpler models with just 1 boolean operation first. But do record those with differing results, we want to eventually make them pass this test.
> 

rebase done.
had a doubt, should I first add multiple small single-operation cases and mainly record which ones differ across platforms, or should I prioritize fixing current differences as soon as they appear case-by-case?

> Things we can do here is:
> 
> 1. Run a line ending replacement for windows.
> 2. Make a new version of the obj dump that dumps floating point as hex.
> 3. Dump the intermediate data of boolean3 and boolean_result.

I can continue all three in this PR as separate small commits ?

pca006132:
> I can continue all three in this PR as separate small commits ?

Yes, just ask me if you have questions about how to implement that.

>  had a doubt, should I first add multiple small single-operation cases and mainly record which ones differ across platforms, or should I prioritize fixing current differences as soon as they appear case-by-case?

How would you fix the differences without figuring the bug out? We need to try simpler cases, check the differences, figure out what is causing the platform inconsistencies.

AnshulPatil2005:
> How would you fix the differences without figuring the bug out? We need to try simpler cases, check the differences, figure out what is causing the platform inconsistencies.

yes, i am planning to,
1)Add a small set of simple single-op determinism cases (3–5) to get a minimal reproducible cross-platform mismatch.
2)Normalize line endings before compare to remove CRLF/LF false mismatches(should be relatively easy)
3)Add a hex-float OBJ dump mode so float comparisons are exact.
4)Add intermediate dumps for Perturb3 / boolean_result to locate the first stage where outputs diverge.
Then I’ll post a short results note (which cases differ and where divergence starts) before proposing/finalizing fixes.
does this sound good ?
> Yes, just ask me if you have questions about how to implement that.

Yes, I’ll ask as I go.

AnshulPatil2005:
simple-case set added-->

Boolean.DeterminismSimpleSubtract (det_simple_subtract.obj)
Boolean.DeterminismSimpleUnion (det_simple_union.obj)
Boolean.DeterminismSimpleIntersect (det_simple_intersect.obj),
line ending + hex-float canonicalization in compare+hex float OBJ,
intermediate debug dumps available via env flags,
current checks green on this set.

pca006132:
We need to decompose the failing cases into simple Boolean operations and find out where it is non-deterministic.

Green CI means that it does not catch the problem.

AnshulPatil2005:
> We need to decompose the failing cases into simple Boolean operations and find out where it is non-deterministic.
> 
> Green CI means that it does not catch the problem.

Yes, got it 
I was just testing with the previous set to validate the compare pipeline itself,It was initially failing due to cross-toolchain hex-float text representation differences, so I added canonicalization in compare ([8654539](https://github.com/elalish/manifold/pull/1594/commits/8654539714925b18f5d28f836b7d1d7bd12ea52b))
 I’ll decompose the known failing case and identify the first stage of output divergence.

pca006132:
Thanks, but I don't think the current way is correct. We should print hex from the cpp files.

pca006132:
And this is not dumping enough of the intermediate states. We should dump the array content, not just their lengths.

AnshulPatil2005:
> And this is not dumping enough of the intermediate states. We should dump the array content, not just their lengths.

> And this is not dumping enough of the intermediate states. We should dump the array content, not just their lengths.

Understood, I’ll update this
Then I’ll move on to Perturb3 decomposition + a discovery lane to locate the first divergence across platforms.
does this seem like a correct approach ? @pca006132

pca006132:
Yes

AnshulPatil2005:
first divergence is at perturb3_gear (single pair union is deterministic)
 
 will decompose it further to find the point of divergence
 
upload full intermediate dumps in discovery CI(compare)

Diff dumps stage-by-stage across OSes

Implement deterministic fix at that stage

AnshulPatil2005:
I checked the `Cross-platform determinism discovery` logs.

here is what i found 
- The `445` mismatches are mostly a chain reaction, 
- The first files that match on all OSes are:
  - `perturb3_pair_union.obj`
  - `perturb3_pair01.obj`
  - `perturb3_pair23.obj`
- The first file that does NOT match is:
  - `perturb3_gear3_seq.obj`
- The first internal mismatch starts here:
  - `boolean3_000004_post_winding_add_inP.obj`
- `boolean_result` mismatches happen later, so the root problem seems to start in `boolean3` first.

so i was thinking we should focus only on this first `boolean3_000004` mismatch, fix that, then rerun discovery. That should remove many downstream mismatches automatically.

does this direction looks right, I can next isolate it and figure out the bug better.

pca006132:
Yes

pca006132:
What happens if you change all the 4s here to 1?

https://github.com/elalish/manifold/blob/98c317fff84082db32fe4f6e7135a171720005b5/src/csg_tree.cpp#L397-L420

AnshulPatil2005:
> What happens if you change all the 4s here to 1?
> 
> https://github.com/elalish/manifold/blob/98c317fff84082db32fe4f6e7135a171720005b5/src/csg_tree.cpp#L397-L420

I’ll test the 4 -> 1
If mismatches drop, thats an order/grouping (reduction tree and possibly concurrency) effect.

AnshulPatil2005:
4->1 logs
- Mesh mismatches: `7` (same set as before)
- `perturb3_gear3_seq.obj`
- `perturb3_gear3_batch.obj`
- `perturb3_gear4_seq.obj`
- `perturb3_gear4_batch.obj`
- `perturb3_gear.obj`
- `perturb3_outerGear.obj`
- `perturb3_nastyGear.obj`
- Dump mismatches: `480` total
- `boolean3_*`: `211`
- `boolean_result_*`: `269`
- First dump mismatch is still `boolean3_000004`
- Last mismatch reaches `boolean_result_000055_*`

increased dump mismatches previous `445` by `+35` so 480 total.

AnshulPatil2005:
I'll isolate and figure out the divergence in  `boolean3_000004_post_winding_add_inP.obj` next

AnshulPatil2005:
Bug was **non-deterministic reduction order** in `BatchBoolean`.

`std::make_heap/pop_heap` was ordering only by `NumVert()`. For ties (same vertex count), heap behavior is not same across platforms/standard-library implementations, so different pairs were combined first. That changed the boolean reduction tree, so by `boolean3_000004_post_winding_add` the `inP` input already differed (linux/windows one path, mac another).

So the difference was caused by missing deterministic tie-break for equal-size operands.

The fix was to make `BatchBoolean` heap selection deterministic for ties.

In [src/csg_tree.cpp](manifold/src/csg_tree.cpp):
- I changed heap items from just `shared_ptr<CsgLeafNode>` to `(shared_ptr<CsgLeafNode>, serial_id)`.
- Comparator now sorts by:
 `NumVert()` then by`serial_id` (tie-break)
- New intermediate results also get a deterministic increasing `serial_id`.

Mismatches went from 480 to 94 due to this fix 

edit :
next first divergence in the new logs :boolean3_000015_post_winding_add_inQ.obj (line 965)

AnshulPatil2005:
Latest update: the determinism discovery run is now clean.

- Cross-platform determinism discovery (non-blocking): **0 mesh mismatches**
- Boolean debug dump comparison: **0 dump mismatches**
- Previous run was `3` mesh mismatches and `94` dump mismatches, so this is now fully resolved for the current Perturb3 discovery set.

Last two commits in this series:
- `c5ccf3d3` -> targeted the remaining first divergence at `boolean3_000015_post_winding_add_inQ.obj`
- `9b5dffef` -> added focused trig canonicalization at 45° (`±sqrt(1/2)`) to avoid global changes 

At this point, Linux/macOS/Windows hashes match for the tracked Perturb3 outputs and intermediate dumps.

pca006132:
So I guess part of the problem is due to platform inconsistencies with trigonometric functions that we use for mesh rotation. One way to solve this is to provide an implementation of that ourselves. @elalish what do you think?

@AnshulPatil2005 do we have more failures?

AnshulPatil2005:
I’ll also added MultiCoplanar into the blocking determinism compare set; next CI run will confirm whether it is deterministic cross-platform as well if not I'll solve it in the same way as pertub3.
If you prefer, I can follow up with a deterministic in-repo trig implementation (instead of targeted canonicalization) so this is solved in a more general way.

AnshulPatil2005:
> @AnshulPatil2005 do we have more failures?

For the current Perturb3 discovery corpus, I don’t see additional failures in the latest run: mesh mismatches are 0 and dump mismatches are 0 across Linux/macOS/Windows.
Will find out for multicoplanar

AnshulPatil2005:
@pca006132 @elalish

`MultiCoplanar` is now deterministic across Linux/macOS/Windows in CI.

Current setup:
- blocking determinism check for the fixed case set (now including `MultiCoplanar`)
- non-blocking discovery lane for `Perturb3`

Could you please review these changes? If this looks good, I’ll move on to Part 1 Task 2 (ASan+UBSan lane).

also, 
quick question on gating policy:

Right now we have:
- **blocking** cross-platform determinism check for `DeterminismSimple*` + `MultiCoplanar`
- **non-blocking discovery** lane for `Perturb3` (with debug dump comparison)

Would you like me to keep `Perturb3` as discovery-only, or move `Perturb3` into the blocking determinism gate as well?  
I can do either based on your preference for strictness vs. flake/noise risk.

AnshulPatil2005:
> So I guess part of the problem is due to platform inconsistencies with trigonometric functions that we use for mesh rotation. One way to solve this is to provide an implementation of that ourselves. @elalish what do you think?

For longer-term robustness in this, I’d be happy to help prototype a deterministic in-repo trig path (so we depend less on platform `libm` behavior) if you think that direction is worth pursuing.

elalish:
Yes, deterministic trig functions sounds like a good idea, and one I like better than snapping the inputs. It might be simpler to review this if you just make separate PRs for any actual fixes you make and this PR can just be for the actual CI changes. Does that sound reasonable?

AnshulPatil2005:
Yes, that sounds reasonable to me.

I’ll keep this PR focused on CI changes only, and I’ll move the trig-related fix into a separate PR so it can be reviewed independently.

pca006132:
For the trig functions, I think we should just grab one from libm or something, with compatible license.

AnshulPatil2005:
yep, that works too.

also, what should be my next steps now ?
> Could you please review these changes? If this looks good, I’ll move on to Part 1 Task 2 (ASan+UBSan lane).
> 
> also, quick question on gating policy:
> 
> Right now we have:
> 
> * **blocking** cross-platform determinism check for `DeterminismSimple*` + `MultiCoplanar`
> * **non-blocking discovery** lane for `Perturb3` (with debug dump comparison)
> 
> Would you like me to keep `Perturb3` as discovery-only, or move `Perturb3` into the blocking determinism gate as well? I can do either based on your preference for strictness vs. flake/noise risk.

pca006132:
Add the trigonometry function in another PR, change the rotation code to use those instead of `std` implementations. Tell us the source of the trigonometry functions (which libm for example, ideally provide the url) and license, so we can check if they are compatible.

After that PR is merged, merge the master branch into here and remove the existing fix, check if the result is still consistent across platforms. If everything is good, add a few more models for determinism check.

AnshulPatil2005:
Sounds good. I’ll do that.

AnshulPatil2005:
@elalish @pca006132 Could you suggest which additional determinism models that would be in blocking CI set?
Right now we already cover `DeterminismSimple*` + `MultiCoplanar` (blocking) and `Perturb3` (discovery).  
For expansion, I was thinking a few stable cases with broader geometry coverage like
- `Boolean.NearlyCoplanar`
- `Boolean.NonOverlap`
- one `CrossSection`-heavy case
- one transform/rotation-heavy case

AnshulPatil2005:
also I was thinking that I can start Part 1 Task 2 (ASan+UBSan lane) in parallel on a separate PR while we finalize this CI-model expansion PR.  
potential clash exists, but its manageable with clear scope boundaries.
should I proceed that way, or wait until this PR is fully merged first?

elalish:
I think your ideas for expanding the test sound good - use your judgement and don't go overboard. We can always add more later. Best to get this PR merged before it gets too stale. 

You're welcome to start the next task if you want, but hopefully we can get this one merged pretty quickly now.

pca006132:
I think we want to reduce the code needed to print intermediate states. We already have some code for dumping vectors and testing verbosity level, maybe we can use that.

AnshulPatil2005:
> I think your ideas for expanding the test sound good - use your judgement and don't go overboard. We can always add more later. Best to get this PR merged before it gets too stale.
> 
> You're welcome to start the next task if you want, but hopefully we can get this one merged pretty quickly now.

I added two additional blocking models:
Boolean.NonIntersecting (det_non_overlap.obj)
Boolean.AlmostCoplanar (det_nearly_coplanar.obj)
This should be ready to merge now. I’ll start the next task in some time, and I’ll address any follow-up review comments here as they come.


> I think we want to reduce the code needed to print intermediate states. We already have some code for dumping vectors and testing verbosity level, maybe we can use that.

makes sense, I’ll refactor it accordingly.

AnshulPatil2005:
> I think we want to reduce the code needed to print intermediate states. We already have some code for dumping vectors and testing verbosity level, maybe we can use that.

I refactored this to reuse existing verbosity plumbing (`ManifoldParams().verbose`, now triggered via `-v`) for enabling dumps.
For vector/object artifact output, I kept a shared helper (`boolean_dump.h`) because existing `Dump()` utilities are debug stdout-only, while this lane needs stable file artifacts (including hex-float formatting) for cross-platform comparison.

.github/workflows/manifold.yml
```diff
@@ -491,3 +602,167 @@ jobs:
...
+      uses: actions/download-artifact@v4
+      with:
+        name: meshes-windows
+        path: meshes/windows
+    - name: Compare meshes across platforms
+      run: |
+        set -euo pipefail
+        failed=0
+        normalized_dir="$(mktemp -d)"
+        for f in det_simple_subtract.obj det_simple_union.obj det_simple_intersect.obj det_multi_coplanar.obj det_non_overlap.obj det_nearly_coplanar.obj; do
```
elalish: There's a lot here - maybe bundle these into shell scripts instead?

scripts/compare_boolean_dumps.sh
elalish, line 1: Why are these dumps necessary? If the OBJs are identical, isn't that enough to validate determinism?

.github/workflows/manifold.yml
AnshulPatil2005: Yes, on it, that would be cleaner

scripts/compare_boolean_dumps.sh
AnshulPatil2005: right, the blocking one doesnt need the dumps, it was used in the initial stages before another discovery lane was made; so its safe to remove from it now, will do that.

discovery would need it so will keep it for discovery

.github/workflows/manifold.yml
```diff
@@ -491,3 +602,103 @@ jobs:
...
+          det_simple_union.obj
+          det_simple_intersect.obj
+          det_multi_coplanar.obj
+          det_non_overlap.obj
+          det_nearly_coplanar.obj
+      run: |
+        bash ./scripts/compare_artifact.sh meshes
+
+  determinism_discovery:
+    name: Cross-platform determinism discovery (non-blocking)
```
elalish, line 639: What is the purpose of this non-blocking check versus the blocking one above? Are these ones still not deterministic?

scripts/compare_boolean_dumps.sh
elalish: Thanks, but I think this discovery code should just live on one of your branches while you're working on determinism. It's a lot of code to add to the repo and I'm not sure everyone will debug the same way you do. Also, right now we are careful to not include any e.g. <string> or <iostream> headers unless we're in `MANIFOLD_DEBUG` mode, to keep our WASM build size down.

AnshulPatil2005: oh right, didnt thought of this actually; in that case i will only keep the blocking lane which could be updated as time goes according to the models to be used

.github/workflows/manifold.yml
AnshulPatil2005: my idea was Blocking =required gate. Discovery = broader (currently Perturb3 + dumps) to catch and debug then stable discovery cases can be promoted into blocking later.

src/boolean_dump.h
elalish, line 1: Can we remove this?

src/boolean3.cpp
```diff
@@ -28,6 +30,44 @@ using namespace manifold;
 
 namespace {
 
+std::atomic<uint64_t> gBoolean3DumpCounter{0};
+
+void DumpBoolean3State(const char* stage, const OpType op,
```
elalish: This also looks like debug code?

src/boolean_result.cpp
```diff
@@ -46,6 +47,47 @@ namespace {
 
 constexpr int kParallelThreshold = 128;
 
+std::atomic<uint64_t> gBooleanResultDumpCounter{0};
+
+void DumpBooleanResultState(const char* stage, const OpType op,
```
elalish: This too.

elalish (changes requested):

src/boolean_dump.h
AnshulPatil2005: yes definitely, its not in use now

elalish (approved):
This looks great, thank you!

PR #1597, opened 2026-03-17
fix: implement manifold_cross_section_warp_context C binding

### LINKED ISSUE #1531: Missing implementation for manifold_cross_section_warp_context in C bindings
### Missing C-binding implementation on obj-import branch

Hello 

I noticed that `manifold_cross_section_warp_context` is declared in `manifoldc.h`, but I can't find the implementation in `manifoldc.cpp`. 

It seems it got lost during the recent refactoring in the `obj-import` branch (I'm on commit 52c7fe0). Trying to use it leads to linker errors.

Could you guys take a look or let me know if this is still work-in-progress?

Thanks!

swaparup36:
resolves #1531 

Changes:
`bindings/c/cross.cpp`
Renamed the existing context-style implementation to manifold_cross_section_warp_context that matches the header, and added the correct non-context manifold_cross_section_warp overload that takes ManifoldVec2 (*fun)(double,double) and forwards via a small thunk.

### DIFF
--- a/bindings/c/cross.cpp
+++ b/bindings/c/cross.cpp
@@ -194,22 +194,6 @@ ManifoldCrossSection* manifold_cross_section_warp_context(
   return to_c(new (mem) CrossSection(warped));
 }
 
-ManifoldCrossSection* manifold_cross_section_warp(
-    void* mem, ManifoldCrossSection* cs,
-    ManifoldVec2 (*fun)(double, double)) {
-  struct Ctx {
-    ManifoldVec2 (*fun)(double, double);
-  };
-  Ctx ctx = {fun};
-
-  auto thunk = [](double x, double y, void* vctx) -> ManifoldVec2 {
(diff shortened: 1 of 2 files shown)


pca006132:
Do we need another version? I think it will be fine to just expose the old one in the header.

swaparup36:
> Do we need another version? I think it will be fine to just expose the old one in the header.

I see. I am keeping manifold_cross_section_warp_context and removing the manifold_cross_section_warp. In that case I have to remove the declarations for manifold_cross_section_warp in the manifoldc.h too. Will that be fine? I mean is that what you meant?

pca006132:
Yes, that sounds fine. Btw, you need to use clang-format to format the code, so the CI does not complain.

swaparup36:
> Yes, that sounds fine. Btw, you need to use clang-format to format the code, so the CI does not complain.

Got it, I will format the code using clang-format. Thanks!

PR #1599, opened 2026-03-18
Feature/editor-draggable-pane-dividers

israrkhan921:
Added draggable pane dividers for code editor, model viewer, and console and made monaco editor content wrap to its window and removed horizontal scrollbar.

### DIFF
--- a/bindings/wasm/examples/editor/editor.js
+++ b/bindings/wasm/examples/editor/editor.js
@@ -80,6 +80,11 @@ let editor = undefined;
 
 // Pane resizing - draggable pane dividers ---------------------
 
+const LEFT_PANE_MIN_PERCENT = 20;
+const LEFT_PANE_MAX_PERCENT = 80;
+const VIEWER_PANE_MIN_PERCENT = 35;
+const VIEWER_PANE_MAX_PERCENT = 90;
+
 // Keep percentages within practical bounds so panes stay usable.
 function clampToRange(value, minValue, maxValue) {
   return Math.min(maxValue, Math.max(minValue, value));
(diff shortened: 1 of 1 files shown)


hrgdavor:
Hi, even though I have decades in JS experince, I was using suboptimal event handling for drag and drop. At first I thought @israrkhan921 your version is less optimal, but since nowdays we chat a lot with AI models as kind of knowledgebase, I was informed that your version where you remove listeners is better than my old approach of keeping them and having a tracking variable. After further nagging gemini, it proposed an even more modern solution using abort signals, that looks interesting.

here is an example if you are curious, and would like to consider it.

```js
splitterElement.addEventListener('pointerdown', e => {
  if (window.matchMedia('(max-width: 820px)').matches) return;

  e.preventDefault();
  splitterElement.setPointerCapture(e.pointerId);

  const controller = new AbortController();
  const { signal } = controller;

  splitterElement.addEventListener('pointermove', handleDragMove, { signal });
  
  // Use a one-time "end" listener
    const onPointerEnd = (endEvent) => {
    splitterElement.releasePointerCapture(endEvent.pointerId);
    controller.abort(); // Kills move, up, and cancel listeners instantly
  };
  splitterElement.addEventListener('pointerup', onPointerEnd, { signal });
  splitterElement.addEventListener('pointercancel', onPointerEnd, { signal });
});
```

I looks legit to me, @tonious what are your thoughts on this one ?

bindings/wasm/examples/editor/editor.js
```diff
@@ -78,6 +78,95 @@ if (navigator.serviceWorker) {
...
+  const savedLeftPane = Number(window.localStorage.getItem(leftPaneStorageKey));
+  if (Number.isFinite(savedLeftPane)) {
+    const clampedLeftPanePercent = clampToRange(savedLeftPane, 20, 80);
+    pageElement.style.setProperty('--left-pane', `${clampedLeftPanePercent}%`);
+  }
+
+  const savedViewerPane =
+      Number(window.localStorage.getItem(viewerPaneStorageKey));
+  if (Number.isFinite(savedViewerPane)) {
+    const clampedViewerPanePercent = clampToRange(savedViewerPane, 35, 90);
```
elalish: Let's make these named constants so they're easier to find and change.

bindings/wasm/examples/editor/editor.js
```diff
@@ -78,6 +78,95 @@ if (navigator.serviceWorker) {
...
+        '--viewer-pane', `${clampedViewerPanePercent}%`);
+  }
+
+  attachSplitterDrag(horizontalSplitterElement, moveEvent => {
+    // Convert pointer X position to a percentage of the full workbench width.
+    const workbenchBounds = workbenchElement.getBoundingClientRect();
+    const leftPanePercent =
+        ((moveEvent.clientX - workbenchBounds.left) / workbenchBounds.width) *
+        100;
+    const clampedLeftPanePercent = clampToRange(leftPanePercent, 20, 80);
```
elalish: Same here.

israrkhan921:
I have added the named pane-bound constants and replaced all hardcoded clamp values

israrkhan921:
now its best , but in the future we will give it more optimization, smooth behavior, best performance and from all sides best feature.

full screen and zooming: https://drive.google.com/file/d/1M1iqq50vLhKa531b0jfqnYVOvis_nBUR/view?usp=drive_link

note: we will improve any feature of the web and full web, for its performance, speed, maintainability, accessibility, best practices and from all sides , we will improve it.

elalish (approved):
Looks great, thanks! I appreciate the video demo.

israrkhan921:
@elalish thanks Sir

PR #1602, opened 2026-03-19
consistent transformations of normals

### LINKED ISSUE #1601: Problem with transforming normals
Unfortunately I think I've found an API problem with how we handle vertex normal properties. I tried to make Manifold agnostic of what properties it's carrying, but we've had to break that a little by adding a `normalIdx` input to a few methods like `GetMeshGL`, `CalculateNormals`, `SmoothByNormals`, and such. `GetMeshGL` is the problem, as this is the function that applies transforms to the normals, if indicated. However, we currently assume that all transformations were applied *after* the normals were calculated, which *may* be true for input meshes (though probably not if `runTransform` is not identity), but is definitely not true for any mesh that gets transformed and then has `CalculateNormals` called. 

Here's a simple [demo](http://localhost:5173/#normal%20twist%3Ccode%3Eimport%20%7B%20Manifold%20%7D%20from%20'manifold-3d%2FmanifoldCAD'%3B%0A%0A%2F%2F%20putting%20the%20rotate%20after%20calculateNormals%20fixes%20the%20problem%0Aconst%20cube%20%3D%20Manifold.cube(1).rotate(30).calculateNormals(0)%3B%0A%0Aconst%20mesh%20%3D%20cube.getMesh(0)%3B%0A%0Aconst%20input%20%3D%20Manifold.ofMesh(mesh)%3B%0A%0A%2F%2F%20using%20cube%20instead%20of%20input%20works%20fine%0Aconst%20smooth%20%3D%20input.smoothByNormals(0).refine(10)%3B%0A%0Aexport%20default%20smooth%3B) of the problem.

Hmm, thinking about this further, perhaps this can be fixed in a non-breaking way after all. Perhaps we only need to change `CalculateNormals` to apply the reverse of `runTransform`, making it effectively put them in a consistent input reference frame. Still doesn't completely solve the round-trip problem, but perhaps we can handle that with documentation.

elalish:
Fixes #1601 

The changes the behavior of GetMeshGL a bit, but I think it's small enough we don't need to consider it a breaking change. Now when specifying a normalIdx, it clears the runTransform vector, since those transforms have now been applied to the normals. Of course those transforms might be useful for other things too, in which case you'll just have to use GetMeshGL without doing auto normals transforms. 

UPDATE:

Okay, the above is still true, but it was inadequate. I've expanded the MeshGL API to include Backside info, so that it now has all the info it needs to now also have an `UpdateNormals` method that can do what's described above, but after the transforms have been read out, so no need to do `GetMeshGL` two different ways now. The Flags struct also has space for some more data down the road if needed - possibly we can store the normalIdx here and remove those as method inputs.

### DIFF
--- a/include/manifold/manifold.h
+++ b/include/manifold/manifold.h
@@ -111,9 +111,6 @@ class CsgLeafNode;
  */
 template <typename Precision, typename I = uint32_t>
 struct MeshGLP {
-  struct Flags {
-    bool backside : 1;
-  };
   /// Number of property vertices
   I NumVert() const { return vertProperties.size() / numProp; };
   /// Number of triangles


include/manifold/manifold.h
```diff
@@ -111,10 +111,15 @@ class CsgLeafNode;
  */
 template <typename Precision, typename I = uint32_t>
 struct MeshGLP {
+  struct Flags {
+    bool backside : 1;
```
elalish: @pca006132 Does this look like a reasonable way to pack booleans into a byte? The idea is to give us some extensibility here, for instance to store the index of the normals channel.

In any case, probably best to push out 3.4.1 before we merge this.

pca006132: It will not work, c++ structs are still padded to bytes, so the size of this is still 1 byte.

elalish: Yeah, so it's one byte as you recommended, right? And we can add 7 more bits to it without breaking anything, right?

pca006132: Yes, but this is practically the same as storing `uint8_t`.

elalish: Yes, that was the goal. So, my only question is, do you like the struct like this, or is there a better way for it to look?

include/manifold/manifold.h
```diff
@@ -169,19 +178,31 @@ struct MeshGLP {
...
+  /**
+   * Updates the normals of the mesh based on the runTransform matrices and
+   * backside flags, which are then cleared to avoid double-applying them when
+   * round-tripping.
+   *
+   * @param normalIdx Specifies the first of the three consecutive property
+   * channels forming the (x, y, z) normals to update. NumProp must be at least
+   * normalIdx + 3 and normalIdx must be >= 3.
+   */
+  void UpdateNormals(int normalIdx);
```
elalish, line 187: @pca006132 How do you like this new API? I'd like to remove the normalIdx input to GetMeshGL and have this be the new way to update normals. Ideally I'd also like to remove the normalIdx input from here and save it in each triangle run instead. 

Curious what you think about how we should approach that with our release cycles and when we should update the various bindings...

include/manifold/manifold.h
pca006132: This looks fine to me.

Do you need review for the other parts? I can do it tonight.

elalish: That'd be great if you have time - this became a little larger than I bargained for at first.

pca006132: Ah, when I searched for bit field order, it seems that this is implementation defined and not specified by the standard. Using 0 and 1 is simpler for now, and we can say we use LSB order for bits later when we plan to support more flags (manually computing the bits instead of using bit field)

include/manifold/manifold.h
```diff
@@ -219,6 +240,30 @@ struct MeshGLP {
         halfedgeTangent[offset], halfedgeTangent[offset + 1],
         halfedgeTangent[offset + 2], halfedgeTangent[offset + 3]);
   }
+
+  /**
+   * Returns the transformation matrix for the specified run.
+   *
+   * @param run The index of the triangle run (0 <= run < runOriginalID.size()).
+   */
+  mat3x4 GetRunTransform(size_t run) const {
```
pca006132, line 246: `size_t` here but using `I` below?

src/shared.h
```diff
@@ -40,6 +40,10 @@ inline mat3 NormalTransform(const mat3x4& transform) {
   return la::inverse(la::transpose(mat3(transform)));
 }
 
+inline mat3 InverseNormalTransform(const mat3x4& transform) {
```
pca006132, line 52: The name sounds a bit like the inverse of `NormalTransform` above, but it is not?

src/smoothing.cpp
```diff
@@ -479,15 +484,24 @@ void Manifold::Impl::SetNormals(int normalIdx, double minSharpAngle) {
                halfedge_[i].propVert = -1;
              });
```
pca006132, line 486: Faster if we use `std::unordered_map`, but it probably doesn't matter too much here.

CMakeLists.txt
```diff
@@ -159,6 +159,7 @@ else()
     -Wno-unknown-warning-option
     -Wno-unused
     -Wno-shorten-64-to-32
+    -Wno-sign-compare
```
pca006132, line 162: TBH I think we should address sign compare, but that should be another issue.

src/sort.cpp
```diff
@@ -178,6 +178,36 @@ bool MergeMeshGLP(MeshGLP<Precision, I>& mesh) {
...
+      if (vertUpdated[vert]) continue;
+      vertUpdated[vert] = true;
+      vec3 normal;
+      const int start = vert * mesh.numProp + normalIdx;
+      for (int i : {0, 1, 2}) {
+        normal[i] = mesh.vertProperties[start + i];
+      }
+      normal = SafeNormalize(transform * normal);
+      for (int i : {0, 1, 2}) {
+        mesh.vertProperties[start + i] = normal[i];
```
pca006132, line 204: Is it possible that a vertex is shared between different runs with different transforms?

And it should be possible to parallelize this part with atomics.

include/manifold/manifold.h
pca006132: 1. What is the problem with the old API? Duplicated work for updating normals? Unable to use updated normals in `Manifold`?
2. Does saving `normalIdx` in triangle runs mean that we can have different `normalIdx` per run?
3. I think removing the `normalIdx` parameter from `GetMeshGL` is not a big enough breaking change to bump the major version. And if we don't want a breaking change, we can mark things as deprecated and remove them in the major version bump later (who knows when...). But that said, I am probably a bit too conservative in bumping major versions...
4. Bindings update can come later, I think as long as we update all bindings before a release, it should be fine. We probably want a ticket to track that though.

include/manifold/manifold.h
elalish: Good catch, thanks.

CMakeLists.txt
elalish: This one was ticking me off mostly because Clang wasn't complaining about this, so it was only hitting on the CI. Feels pretty pedantic anyway...

src/shared.h
elalish: I suppose normal transform of the inverse would be more accurate, but it's the only kind of inverse that makes any sense for us. I'll add a comment.

src/sort.cpp
elalish: No, because this is a MeshGL, so there are no shared verts - they are all duplicated at run boundaries.

elalish: parallelization sounds good, but I think I'll leave that for another day...

include/manifold/manifold.h
elalish: 1. If you update the normals, you need to remove the transforms, or they'll get updated again when you go through another Manifold round-trip. If you remove the transforms in GetMeshGL, then they aren't available for non-normals use. This way you can read them and then update and clear them, without having to remember to clear them yourself.
2. Yes, another reason it feels like a better way - plus less to remember.
3. Cool, I'm fine doing this for 3.5
4. Agreed, SG.

include/manifold/manifold.h
elalish: using size_t consistently now.

PR #1603, opened 2026-03-22
Fix handle empty splitbyplane

### LINKED ISSUE #1515: SplitByPlane() doesn't handle empty Manifold objects
If you call <del>`Manifold::TrimByPlane()` and</del> `Manifold::SplitByPlane()` on empty `Manifold` objects, rather than just producing empty results, it flags an error. In my code I am working around the issue by yielding an empty result manually if the input is empty.

My suspicion is that an empty `Manifold` has an invalid bounding box (typically with the minimum set to positive infinity and the maximum set to negative infinity, effectively inside-out with a maximally negative volume). If the min/max values from the invalid bounding box were used in arithmetic with finite vertex values, it would turn them invalid and possibly cause this issue.

Here is a gist to test the issue:

https://gist.github.com/fdarling/508e8b6b126ceca7b65e7e7191ec6423

And the output of that sample code:

```
$ ./manifold-splitbyplane-test 
emptyManifold.Status() = 0
Calling emptyManifold.TrimByPlane()...
trimmedManifold.Status() = 0
Calling emptyManifold.SplitByPlane()...
splitResult.first.Status() = 1
splitResult.second.Status() = 1
```

The `Error` code of `1` means `NonFiniteVertex`:

https://github.com/elalish/manifold/blob/93bea2834f681d7e7fb96dfd012db8a29ee3e1fb/include/manifold/manifold.h#L313-L327

swaparup36:
resolves #1515 

Changes:
- Added a simple early return in Manifold::SplitByPlane() for the IsEmpty() case so it doesn’t end up going through the non-finite BoundingBox() and Halfspace() path. 
- Added a small regression test for that.

### DIFF
--- a/src/manifold.cpp
+++ b/src/manifold.cpp
@@ -848,6 +848,7 @@ std::pair<Manifold, Manifold> Manifold::Split(const Manifold& cutter) const {
  */
 std::pair<Manifold, Manifold> Manifold::SplitByPlane(
     vec3 normal, double originOffset) const {
+  if (IsEmpty()) return {Manifold(), Manifold()};
   return Split(Halfspace(BoundingBox(), normal, originOffset));
 }
 
--- a/test/boolean_test.cpp
+++ b/test/boolean_test.cpp
@@ -477,6 +477,20 @@ TEST(Boolean, Split) {
                   cube.Volume());
 }
 
+TEST(Boolean, SplitByPlaneEmpty) {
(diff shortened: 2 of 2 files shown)


swaparup36:
I'm really sorry! I included commits for another fix on my branch by mistake. So, I had to close this PR. I will be more careful further.

PR #1606, opened 2026-03-23
Deterministic trig

AnshulPatil2005:
This PR moves the trig determinism fix to a licensed in-repo implementation (separate from CI-only changes).

- Replaced `std::sin` / `std::cos` usage in `sind()` with inlined reduced-range trig kernels adapted from musl (`__sin.c`, `__cos.c`), which originate from FreeBSD msun.
- Added attribution and license files:
  - `third_party/libm_trig/README.md`
  - `third_party/libm_trig/LICENSE.Sun.txt`

- musl sources:
  - https://git.musl-libc.org/cgit/musl/plain/src/math/__sin.c
  - https://git.musl-libc.org/cgit/musl/plain/src/math/__cos.c
- musl license (MIT):
  - https://git.musl-libc.org/cgit/musl/plain/COPYRIGHT
- Sun permissive notice included in the README 

ref : https://github.com/elalish/manifold/pull/1594#issuecomment-4109139273

### DIFF
--- a/include/manifold/linalg.h
+++ b/include/manifold/linalg.h
@@ -662,45 +662,21 @@ struct std_sqrt {
   }
 };
 struct std_sin {
-  template <class A>
-  auto operator()(A a) const {
-    return std::sin(a);
-  }
   double operator()(double a) const { return manifold::math::sin(a); }
 };
 struct std_cos {
-  template <class A>
-  auto operator()(A a) const {
-    return std::cos(a);
-  }
   double operator()(double a) const { return manifold::math::cos(a); }
 };
 struct std_tan {
-  template <class A>
-  auto operator()(A a) const {
-    return std::tan(a);
-  }
   double operator()(double a) const { return manifold::math::tan(a); }
 };
 struct std_asin {
-  template <class A>
-  auto operator()(A a) const {
-    return std::asin(a);
-  }
   double operator()(double a) const { return manifold::math::asin(a); }
 };
 struct std_acos {
-  template <class A>
-  auto operator()(A a) const {
-    return std::acos(a);
-  }
   double operator()(double a) const { return manifold::math::acos(a); }
 };
 struct std_atan {
-  template <class A>
-  auto operator()(A a) const {
-    return std::atan(a);
-  }
   double operator()(double a) const { return manifold::math::atan(a); }
 };
 struct std_sinh {
@@ -740,10 +716,6 @@ struct std_pow {
   }
 };
 struct std_atan2 {
-  template <class A, class B>
-  auto operator()(A a, B b) const {
-    return std::atan2(a, b);
-  }
   double operator()(double a, double b) const {
     return manifold::math::atan2(a, b);
   }


AnshulPatil2005:
was also considering CORE-MATH: modern correctly-rounded functions (MIT), but adoption is heavier for it.
from what i read CORE-MATH is strong for a long-term math-quality initiative, but for this determinism goal, musl kernel reuse is lower risk and faster to review.
CORE-MATH mission/integration notes: https://core-math.gitlabpages.inria.fr/

include/manifold/common.h
```diff
@@ -91,20 +91,57 @@ constexpr double smoothstep(double edge0, double edge1, double a) {
  *
  * @param x Angle in degrees.
  */
+// Deterministic reduced-range kernels adapted from musl libc (MIT):
```
pca006132: 1. I think it suffices to attach the license here, and we can remove the `third_party` files.
2. Naming-wise, it should be fine to just use `sin` and `cos`, maybe under a special namespace `manifold::math` or something. We should probably move `sun_acos` there and rename it as `acos`. @elalish what do you think?

We should double-check what other math functions we use are not precisely defined by IEEE 754, and fix them as well.

elalish: Agreed. I generally just use the manifold namespace directly, but it's not a strong opinion.

pca006132: The thing I am worried about is that we may call `sin` and somehow it resolves to the one from the standard library. Putting it inside another nested namespace can avoid this problem if we always use the qualified name `math::sin`.

I am not quite sure why we can currently access things like `remquo` without qualifying it with `std::` (a few lines below in `include/manifold/common.h`), so this name resolution problem is a potential issue.

AnshulPatil2005: Moved deterministic trig into manifold::math::{sin, cos, acos} in common.h and moved the notice too

AnshulPatil2005: > I am not quite sure why we can currently access things like `remquo` without qualifying it with `std::` (a few lines below in `include/manifold/common.h`), so this name resolution problem is a potential issue.

 I’ve switched those calls to names (std::remquo, std::fabs, etc.)

pca006132: > > I am not quite sure why we can currently access things like `remquo` without qualifying it with `std::` (a few lines below in `include/manifold/common.h`), so this name resolution problem is a potential issue.
> 
> I’ve switched those calls to names (std::remquo, std::fabs, etc.)

You don't have to do so, since the issue is not about these calls, but other stuff that makes these calls compile. (so `sin` may resolve to `std::sin` or `manifold::sin`, haven't checked but this is a possibility)

include/manifold/common.h
```diff
@@ -86,25 +89,126 @@ constexpr double smoothstep(double edge0, double edge1, double a) {
...
+ * - https://git.musl-libc.org/cgit/musl/plain/src/math/__cos.c
+ * - https://git.musl-libc.org/cgit/musl/plain/src/math/e_acos.c
+ *
+ * Copyright (C) 1993 by Sun Microsystems, Inc. All rights reserved.
+ * Developed at SunPro/SunSoft, a Sun Microsystems, Inc. business.
+ * Permission to use, copy, modify, and distribute this software is freely
+ * granted, provided that this notice is preserved.
+ */
+namespace math {
+inline double sin(double x) {
```
pca006132: Can the functions be `constexpr`?

include/manifold/common.h
```diff
@@ -86,25 +89,126 @@ constexpr double smoothstep(double edge0, double edge1, double a) {
...
+}  // namespace math
+
 /**
  * Sine function where multiples of 90 degrees come out exact.
  *
  * @param x Angle in degrees.
  */
 inline double sind(double x) {
-  if (!la::isfinite(x)) return sin(x);
+  if (!la::isfinite(x)) return std::sin(x);
```
pca006132: Maybe we should just return `NAN`.

include/manifold/common.h
AnshulPatil2005: Oh, I misunderstood the earlier comment.  

I could do a broader pass to qualify other math call sites in a separate follow-up PR ?

pca006132: > I could do a broader pass to qualify other math call sites in a separate follow-up PR ?

No, it is not needed. Again, the issue is why we can do unqualified calls, and changing the calls to qualified calls does not solve the problem. Anyway, this is not a major issue anyway.

pca006132:
We have a lot of uses of trig functions in places like `smoothing.cpp`, `cross_section.cpp`, `constructors.cpp`, `properties.cpp`. Maybe replace those as well.

include/manifold/common.h
AnshulPatil2005: Makes sense, thanks for clarifying

AnshulPatil2005:
> We have a lot of uses of trig functions in places like `smoothing.cpp`, `cross_section.cpp`, `constructors.cpp`, `properties.cpp`. Maybe replace those as well.

okay, will do that

AnshulPatil2005:
i replaced trig usage in `smoothing.cpp`, `cross_section.cpp`, `constructors.cpp`, and `properties.cpp` with deterministic `math::` functions.

there are still some trig call sites outside these files (`manifold.cpp` and generic paths in `linalg.h`), so please let me know if you want me to include those in this pr as well or keep this scope focused to the files you listed.

also reworked math::sin/cos to be full-range (range reduction + quadrant handling) and kept kernel math in SinKernel/CosKernel, so these new call sites are safe for general radian inputs.

AnshulPatil2005:
I traced the CI failures to the `constructors.cpp` sphere trig replacement (`la::cos(...) -> math::cos(...)`). With that change, `Boolean.NonConvexConvexMinkowskiDifference` fails consistently (surface area drift and genus 3 vs expected 5). Reverting that call makes the test pass again.

I think we need a higher-fidelity deterministic trig implementation for this (full libm-style argument reduction + kernel correction), not the current simplified wrapper.
@pca006132

pca006132:
1. Replace all usages. There is no reason why you shouldnt do that.
2. Why not just port the wrappers like https://git.musl-libc.org/cgit/musl/tree/src/math/sin.c?

AnshulPatil2005:
> 1. Replace all usages. There is no reason why you shouldnt do that.

yes yes was going to continue replacing others too.

> 2. Why not just port the wrappers like https://git.musl-libc.org/cgit/musl/tree/src/math/sin.c?

started with kernels first because it was the smallest way to validate the root cause and get an initial deterministic fix in place.
agreed though, for full replacement this should prolly use proper wrappers. i’ll port musl-style `sin/cos`

elalish:
>With that change, Boolean.NonConvexConvexMinkowskiDifference fails consistently (surface area drift and genus 3 vs expected 5). 

There's always a chance it was wrong before and right now. Can you export the before/after mesh and give us some screenshots? My rule of thumb is when in doubt, lower genus is usually better.

AnshulPatil2005:
> There's always a chance it was wrong before and right now. Can you export the before/after mesh and give us some screenshots? My rule of thumb is when in doubt, lower genus is usually better.

yea, here are the results

Comparison: old 503ca80 vs head 2b3a6575 (before vs after)
Test: Boolean.NonConvexConvexMinkowskiDifference

Execution result
- old 503ca80: FAILED
  - SurfaceArea = 17.029823126037613 (expected 16.70 +/- 1e-2)
  - Genus() = 3 (expected 5)
  
- head 2b3a6575: PASSED

Exported meshes
- old_503ca80.obj      (115956 bytes)
- head_2b3a6575.obj    (96248 bytes)

SHA256
- old_503ca80.obj      d5a84bd3a57ec19fbd06ddda3288ce3926746815c37e3d38df76eb657f5d2945
- head_2b3a6575.obj    094f459475ff3bc96eb7bc75fd829c3e2fe51abb3f71a23fc32e40747bab5443
- byte-identical: false

OBJ-derived geometry stats
- old_503ca80.obj
  - V=1182, E=3558, F=2372
  - components=3, chi=-4, genus_total=5.0
  - surface_area=17.029823126037666
- head_2b3a6575.obj
  - V=990, E=2994, F=1996
  - components=1, chi=-8, genus_total=5.0
  - surface_area=16.703740785913283
  
head_2b3a6575
old_503ca80

elalish:
Okay, thanks - yes old only has a lower genus because it has more components (3 vs 1) which is definitely a problem. 

Your "full trig coverage" commit is quite large - maybe break that down and see which piece is causing trouble? While you're at it, see if you can clean that up and simplify it - feels like it's more verbose than necessary, which is always fodder for bugs.

elalish:
>Okay, thanks - yes old only has a lower genus because it has more components (3 vs 1) which is definitely a problem.

Wait, do you have those names swapped? If old failed and head passed, wouldn't that mean the PR is working as it stands?

AnshulPatil2005:
> > Okay, thanks - yes old only has a lower genus because it has more components (3 vs 1) which is definitely a problem.
> 
> Wait, do you have those names swapped? If old failed and head passed, wouldn't that mean the PR is working as it stands?

my bad, I mixed up the labels earlier.
`503ca80` is the revision where I switched sphere trig (`la::cos(...) -> math::cos(...)`), and that revision fails (`Genus() = 3`, expected `5`).  
Current `HEAD` is after reverting that sphere change, and it passes again (`Genus() = 5`).

AnshulPatil2005:
> Your "full trig coverage" commit is quite large - maybe break that down and see which piece is causing trouble? While you're at it, see if you can clean that up and simplify it - feels like it's more verbose than necessary, which is always fodder for bugs.

yup, you're right doing it right npw

AnshulPatil2005:
quick note on the commit timeline since I force-pushed a few times:

from 86c85e5 onward, the current linear history is:

86c85e5 add deterministic trig implementation (musl-based)
0e43a74 switch source trig callsites
209a9c7 format
20b5122 switch linalg trig callsites
you may also see 75f025d, 2589ecc, and a13a571 in PR activity; those are superseded snapshots from earlier force-pushes and are not ancestors of current head.

include/manifold/common.h
```diff
@@ -86,25 +90,530 @@ constexpr double smoothstep(double edge0, double edge1, double a) {
   return x * x * (3 - 2 * x);
 }
 
+/**
+ * Deterministic trigonometric helpers.
```
elalish: Let's move this to its own file. As a general rule, if something has its own license header, it's clearer to make it a separate file. Plus this is quite a bit of related code.

include/manifold/linalg.h
```diff
@@ -659,40 +671,106 @@ struct std_sqrt {
     return std::sqrt(a);
   }
 };
+template <class A>
+constexpr auto deterministic_sin(A a) -> decltype(std::sin(a)) {
```
elalish: This seems like a crazy amount of boilerplate added to this file. And we have `manifold::math::sin`, `deterministic_sin`, `std::sin`, and `std_sin`. Can we please just have one? Or some really excellent reasons why multiple makes things simpler/better?

src/smoothing.cpp
```diff
@@ -40,7 +40,28 @@ double Wrap(double radians) {
...
   const double dot = la::dot(a, b);
-  return dot >= 1 ? 0 : (dot <= -1 ? kPi : la::acos(dot));
+  return dot >= 1 ? 0 : (dot <= -1 ? kPi : math::acos(dot));
+}
+
+quat RotationQuatAxisAngle(const vec3& axis, double angle) {
+  return {axis * math::sin(angle * 0.5), math::cos(angle * 0.5)};
+}
+
+quat RotationQuatBetween(const vec3& orig, const vec3& dest) {
```
elalish: Where did these functions come from? Don't they exist in linalg already? Or was this some kind of bug fix?

elalish:
Looks like you got the tests passing now - what was the problem?

include/manifold/common.h
AnshulPatil2005: done

include/manifold/linalg.h
AnshulPatil2005: Agreed. I removed the extra deterministic_* wrapper layer and kept a single dispatch path in linalg (std fallback + manifold::math overloads for floating-point trig).

src/smoothing.cpp
AnshulPatil2005: yup, good catch i missed this earlier these were temporary local duplicates while isolating deterministic trig behavior. I’ve now removed them and switched smoothing back to the existing linalg quaternion APIs (la::rotation_quat), so there’s no duplicate rotation implementation in this file.

AnshulPatil2005:
> Looks like you got the tests passing now - what was the problem?

the failure came from my earlier sphere-path trig replacement in `constructors.cpp` (`la::cos(...) -> math::cos(...)`), which changed geometry enough to break `Boolean.NonConvexConvexMinkowskiDifference` (area drift + genus mismatch).

that previous path was incomplete for full-range behavior. I switched to the full deterministic libm-style implementation (argument reduction + kernels), and with that change the failing test and CI checks pass.

include/manifold/linalg.h
```diff
@@ -664,36 +676,84 @@ struct std_sin {
   constexpr auto operator()(A a) const -> decltype(std::sin(a)) {
     return std::sin(a);
   }
+  double operator()(double a) const { return manifold::math::sin(a); }
```
elalish, line 665: I still don't understand what all this extra boilerplate is doing here. Also, above it's still calling `std::sin`, which is incorrect now, right? Can't we just update the original line and not add any extra cases?

include/manifold/linalg.h
```diff
@@ -41,6 +41,18 @@
...
 
+namespace manifold {
+namespace math {
+double sin(double);
+double cos(double);
+double tan(double);
+double asin(double);
+double acos(double);
+double atan(double);
+double atan2(double, double);
```
elalish: Are these necessary?

include/manifold/linalg.h
```diff
@@ -1900,8 +1971,7 @@ template <class T>
 vec<T, 4> qexp(const vec<T, 4>& q) {
   const auto v = q.xyz();
   const auto vv = length(v);
-  return std::exp(q.w) *
-         vec<T, 4>{v * (vv > 0 ? std::sin(vv) / vv : 0), std::cos(vv)};
+  return std::exp(q.w) * vec<T, 4>{v * (vv > 0 ? sin(vv) / vv : 0), cos(vv)};
```
elalish: Are `std::exp` and `std::log` deterministic already?

include/manifold/deterministic_trig.h
```diff
@@ -0,0 +1,527 @@
+// Copyright 2021 The Manifold Authors.
```
pca006132: 1. update the year
2. I think it is better to name this file as just `math.h` or something. The word deterministic can be confusing to some, since the functions are deterministic, just not producing bitwise-identical results across platforms.
3. Btw, did you port this manually or are you using LLM? If you are using LLM, maybe we need to check this a bit more carefully...

include/manifold/linalg.h
```diff
@@ -1801,7 +1872,7 @@ T distance(const vec<T, M>& a, const vec<T, M>& b) {
 template <class T, int M>
 T uangle(const vec<T, M>& a, const vec<T, M>& b) {
   T d = dot(a, b);
-  return d > 1 ? 0 : std::acos(d < -1 ? -1 : d);
+  return d > 1 ? 0 : acos(d < -1 ? -1 : d);
```
pca006132: `manifold::math::acos`? Similarly, for the function calls below.

include/manifold/linalg.h
pca006132: Probably not, these are recommended operations.

src/constructors.cpp
```diff
@@ -210,7 +210,9 @@ Manifold Manifold::Sphere(double radius, int circularSegments) {
   pImpl_->Subdivide([n](vec3, vec4, vec4) { return n - 1; });
   for_each_n(autoPolicy(pImpl_->NumVert(), 1e5), pImpl_->vertPos_.begin(),
              pImpl_->NumVert(), [radius](vec3& v) {
-               v = la::cos(kHalfPi * (1.0 - v));
```
pca006132, line 213: do you need to change this now, when `la::cos` implementation is changed?

include/manifold/linalg.h
AnshulPatil2005: my bad, removed the extra overload boilerplate

Also, std::sin there is only in decltype(...) for return-type deduction. Runtime now goes through manifold::math::sin(...) (same for cos/tan/asin/acos/atan/atan2), so no std::sin execution in that path anymore.

include/manifold/linalg.h
AnshulPatil2005: > Are `std::exp` and `std::log` deterministic already?

Not guaranteed, need to evaluate it separately.
I can do a separate follow-up after the trig pr is done

include/manifold/linalg.h
pca006132: Why do you need that for the return type declaration? Why not just replace that as well?

And I think we just need the double version. No need for single.

include/manifold/deterministic_trig.h
AnshulPatil2005: > 1. update the year

done

> 2. I think it is better to name this file as just `math.h` or something. The word deterministic can be confusing to some, since the functions are deterministic, just not producing bitwise-identical results across platforms.

yep, makes sense.

> 3. Btw, did you port this manually or are you using LLM? If you are using LLM, maybe we need to check this a bit more carefully...

Most of the port was done manually from the musl/FreeBSD msun sources. A coding agent helped me with some repetitive follow-up edits(added the notice header afterwards), which is why the year needed correction. I’ll be more careful next time wherever i use LLMs.

pca006132: It is normal to copy the copyright header without changing the year. I was just asking about the llm thing just in case, since we are trusting you to have port the thing correctly (and llm may hallucinate...).

include/manifold/linalg.h
AnshulPatil2005: not necessarily, I removed those forward declarations and now include math.h directly in linalg.h

src/constructors.cpp
AnshulPatil2005: good point, its not needed anymore. Since la::cos now routes to the deterministic trig path, I switched this sphere callsite back to la::cos for consistency.

include/manifold/deterministic_trig.h
AnshulPatil2005: > It is normal to copy the copyright header without changing the year. I was just asking about the llm thing just in case, since we are trusting you to have port the thing correctly (and llm may hallucinate...).

Makes sense, thanks for clarifying. I don’t rely on LLMs for writing core part of code. I only use it for small cleanups/edits. I’ll do one final line-by-line pass against upstream and test it thoroughly before merge.

include/manifold/common.h
```diff
@@ -21,6 +22,7 @@
 #endif
 
 #include "linalg.h"
+#include "math.h"
```
pca006132: Maybe use `./math.h`, since `math.h` is also a system header.

include/manifold/linalg.h
```diff
@@ -41,6 +41,8 @@
 #include <functional>   // For std::hash declaration
 #include <type_traits>  // For std::enable_if, std::is_same, std::declval
 
+#include "math.h"
```
pca006132: same here.

include/manifold/linalg.h
pca006132: I think you can simply remove this.

AnshulPatil2005: removed constexpr from those trig functor templates if that was the agenda.
that template overload is still needed right ? as a generic fallback for non-double types; the deterministic path is handled by the double overload (manifold::math::...)

pca006132: You can just remove the overload. I don't think there is any meaningful overload for this.

PR #1611, opened 2026-03-25
OIDC is the worst

elalish:
At least it's not just me: https://github.com/actions/setup-node/issues/1440?timeline_page=1

Just loving how they make the most opaque and hard to debug part of the pipeline crazy sensitive to useless crap. And apparently workflow_dispatch runs an out of date workflow now too. Argh!

### DIFF
--- a/.github/workflows/publish_npm.yml
+++ b/.github/workflows/publish_npm.yml
@@ -21,7 +21,9 @@ jobs:
         registry-url: 'https://registry.npmjs.org/'
 
     - name: Verify npm auth
-      run: npm whoami
+      run: |
+        unset NODE_AUTH_TOKEN
+        npm whoami
 
     - name: Update npm # Ensure npm 11.5.1 or later for trusted publishing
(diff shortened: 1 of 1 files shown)


pca006132:
Well wtf, I will have a look at their doc later this week I guess...

elalish:
Okay, I'm completely out of ideas: https://github.com/elalish/manifold/actions/runs/23536996243/job/68514624392

I thought I set up the whole OIDC thing properly without typos, but who knows?

pca006132:
> https://docs.npmjs.com/trusted-publishers#limitations-and-future-improvements
> OIDC authentication is currently limited to the publish operation. Other npm commands such as install, view, or access still require traditional authentication methods. The npm whoami command will not reflect OIDC authentication status since the authentication occurs only during the publish operation.

So `npm whoami` is expected to fail.

elalish:
Wow, good find, I never would have expected that - apparently gemini didn't either...

PR #1613, opened 2026-03-25
feat: exposed WarpBatch in WASM

swaparup36:
Exposed `WarpBatch()` in WASM - https://github.com/elalish/manifold/discussions/885#discussioncomment-10327590

Changes:
- Created a `WarpBatch` wrapper in `bindings/wasm/helpers.cpp` that calls `Manifold::WarpBatch()`.
- Exposed it in `bindings/wasm/bindings.cpp` as `._WarpBatch()` on the Manifold class.
- Also added `Manifold.prototype.warpBatch()` in `bindings/wasm/bindings.js`.
- Updated the TypeScript definitions in `bindings/wasm/manifold-encapsulated-types.d.ts`.
- Added a benchmark comparison between the warpBatch and warp function to show how much faster it is.

Here is the log of the test -

mesh: sphere refine(0), numVert=326
warp():      avg 1.18 ms over 80 runs
warpBatch(): avg 0.93 ms over 80 runs
speedup:     1.26x
mesh: sphere refine(1), numVert=326
warp():      avg 0.91 ms over 80 runs
warpBatch(): avg 0.83 ms over 80 runs
speedup:     1.09x
mesh: sphere refine(2), numVert=1298
warp():      avg 3.61 ms over 80 runs
warpBatch(): avg 3.24 ms over 80 runs
speedup:     1.11x
mesh: sphere refine(3), numVert=2918
warp():      avg 7.65 ms over 80 runs
warpBatch(): avg 6.93 ms over 80 runs
speedup:     1.10x
mesh: sphere refine(4), numVert=5186
warp():      avg 13.09 ms over 80 runs
warpBatch(): avg 12.38 ms over 80 runs
speedup:     1.06x

### DIFF
--- a/bindings/wasm/test/warpBatch.perf.test.ts
+++ b/bindings/wasm/test/warpBatch.perf.test.ts
@@ -12,6 +12,7 @@ beforeAll(async () => {
 function nowMs(): number {
   return Number(process.hrtime.bigint()) / 1e6;
 }
+
 suite('warpBatch tests', () => {
   test('actually changes geometry', async () => {
     const m = manifoldModule.Manifold.cube(2, true);
@@ -41,61 +42,5 @@ suite('warpBatch tests', () => {
     const origAfter = m.boundingBox();
     expect(origAfter.min[0]).toBe(before.min[0]);
     expect(origAfter.max[0]).toBe(before.max[0]);
-  });
-
-  test('warp vs warpBatch benchmark', async () => {
-    const refineLevels = [0, 1, 2, 3, 4];
-    const iters = 80;
-
-    for (const refineLevel of refineLevels) {
-      let m = manifoldModule.Manifold.sphere(10, 0);
-      m = m.refine(refineLevel);
-
-      m.warp((v: number[]) => {
-         v[0] += v[2] * v[2];
-       }).boundingBox();
-
-      m.warpBatch((verts: Float64Array, count: number) => {
-         for (let i = 0; i < count; i++) {
-           const z = verts[i * 3 + 2];
-           verts[i * 3 + 0] += z * z;
-         }
-       }).boundingBox();
-
-      let warpTotal = 0;
-      const warpT0 = nowMs();
-      for (let k = 0; k < iters; k++) {
-        const out = m.warp((v: number[]) => {
-          const z = v[2];
-          v[0] += z * z;
-        });
-      }
-      const warpT1 = nowMs();
-      warpTotal += (warpT1 - warpT0);
-      const warpAvg = warpTotal / iters;
-
-      let batchTotal = 0;
-      const warpBatchT0 = nowMs();
-      for (let k = 0; k < iters; k++) {
-        const out = m.warpBatch((verts: Float64Array, count: number) => {
-          for (let i = 0; i < count; i++) {
-            const z = verts[i * 3 + 2];
-            verts[i * 3 + 0] += z * z;
-          }
-        });
-      }
-      const warpBatchT1 = nowMs();
-      batchTotal += (warpBatchT1 - warpBatchT0);
-      const batchAvg = batchTotal / iters;
-
-      console.log(
-          `warp():      avg ${warpAvg.toFixed(2)} ms over ${iters} runs`);
-      console.log(
-          `warpBatch(): avg ${batchAvg.toFixed(2)} ms over ${iters} runs`);
-      console.log(`speedup:     ${(warpAvg / batchAvg).toFixed(2)}x`);
-
-      expect(warpAvg).toBeGreaterThan(0);
-      expect(batchAvg).toBeGreaterThan(0);
-    }
   });
 });


bindings/wasm/test/warpBatch.perf.test.ts
```diff
@@ -0,0 +1,82 @@
...
+import {expect, test} from 'vitest';
+
+import Module, {type ManifoldToplevel} from '../manifold';
+
+function nowMs(): number {
+  return Number(process.hrtime.bigint()) / 1e6;
+}
+
+test('warp vs warpBatch benchmark', async () => {
+  const wasmPath = join(process.cwd(), 'manifold.wasm');
```
pca006132: Do we really need this complicated initialization? I think we were just doing

```
beforeAll(async () => {
  manifoldModule = await Module();
  manifoldModule.setup();
});
```

Please have a look at the other code before you ask the llm to generate something...

bindings/wasm/test/warpBatch.perf.test.ts
```diff
@@ -0,0 +1,82 @@
...
+  const iters = 80;
+
+  for (const refineLevel of refineLevels) {
+    let m = Manifold.sphere(10, 0);
+    m = m.refine(refineLevel);
+
+    const n = m.numVert();
+    console.log(`mesh: sphere refine(${refineLevel}), numVert=${n}`);
+
+    m.warp((v: number[]) => {
```
pca006132: What are these? Are they related to the benchmark?

bindings/wasm/test/warpBatch.perf.test.ts
```diff
@@ -0,0 +1,82 @@
...
+
+    let warpTotal = 0;
+    for (let k = 0; k < iters; k++) {
+      const t0 = nowMs();
+      const out = m.warp((v: number[]) => {
+        const z = v[2];
+        v[0] += z * z;
+      });
+      const t1 = nowMs();
+      out.volume();
```
pca006132: Why do you call `volume`?

bindings/wasm/test/warpBatch.perf.test.ts
```diff
@@ -0,0 +1,82 @@
...
+      for (let i = 0; i < count; i++) {
+        const z = verts[i * 3 + 2];
+        verts[i * 3 + 0] += z * z;
+      }
+    });
+
+    let warpTotal = 0;
+    for (let k = 0; k < iters; k++) {
+      const t0 = nowMs();
+      const out = m.warp((v: number[]) => {
```
pca006132: Typically, if we expect the runtime of each iteration to be short, we just record the time for the whole loop and do division, instead of recording the time for each iteration, sum them up and do the division. That can avoid the overhead caused by recording the time.

bindings/wasm/bindings.js
```diff
@@ -230,6 +230,27 @@ Module.setup = function() {
     return out;
   };
 
+  Module.Manifold.prototype.warpBatch = function(func) {
+    const wasmFuncPtr = addFunction(function(ptr, count) {
+      const heapF64 = Module.HEAPF64 ?? HEAPF64;
+      if (!heapF64) {
+        throw new Error('WASM heap is not initialized (HEAPF64 unavailable)');
+      }
+      const verts = new Float64Array(heapF64.buffer, ptr, count * 3);
```
pca006132, line 239: Did you test if `warpBatch` actually did apply the warp? I am not sure if modifying `verts` will change the underlying heap buffer.

bindings/wasm/helpers.cpp
```diff
@@ -204,6 +204,15 @@ Manifold Warp(Manifold& manifold, uintptr_t funcPtr) {
   return manifold.Warp(f);
 }
 
+Manifold WarpBatch(Manifold& manifold, uintptr_t funcPtr) {
+  void (*f)(uintptr_t, size_t) =
+      reinterpret_cast<void (*)(uintptr_t, size_t)>(funcPtr);
+
+  return manifold.WarpBatch([&](manifold::VecView<manifold::vec3> vecs) {
+    f(reinterpret_cast<uintptr_t>(vecs.data()), vecs.size());
```
pca006132: Technically, there is nothing preventing the compiler from padding `vec3`, and that will break the assumption that a vector of `vec3` is the same as a vector of doubles with length `3n`.

I think a better approach is to do some copying to and from a flattened vector of doubles, or at least do some static assert to make sure the compiler is not padding stuff.

bindings/wasm/test/warpBatch.perf.test.ts
swaparup36: Sorry for that! I'll take care of these things now on.

bindings/wasm/test/warpBatch.perf.test.ts
swaparup36: These were some logs related to the benchmark - I removed these

bindings/wasm/test/warpBatch.perf.test.ts
swaparup36: out variable was unused that's why. I removed it also.

bindings/wasm/test/warpBatch.perf.test.ts
swaparup36: Yes, changed it.

bindings/wasm/bindings.js
swaparup36: Yes, I added a test regarding that to confirm.

bindings/wasm/helpers.cpp
swaparup36: I updated WarpBatch to implement this approach

elalish:
Honestly, I'm expecting a lot more than a 5-25% gain here. I was under the impression that the back-and-forth of calling piles of C++ functions from JS was a lot higher than this. For only this much improvement I don't think it's worth adding a whole extra API. Can you play around with your benchmark, perhaps at different scales or different function complexity, and see if you can find a situation where the difference is more like 10x?

swaparup36:
I added another benchmark that measures only the time spent inside the JS callback. In that warpBatch is roughly 7x faster on average.

However, when measuring the full end-to-end warp vs warpBatch call, the speedup is still only 1.0x–1.1x, because Manifold’s native post-warp processing dominates the total runtime. So the per-vertex callback overhead is real, but it’s mostly hidden by the cost of the rebuild step.

elalish:
That's an interesting result in itself. I'm guessing this is due to SetNormalsAndCoplanar being slow - it's not very parallelized. Okay, well 7x on the part you're working on is good, now that just raises the priority of some of this other performance work.

swaparup36:
I profiled the native Warp/WarpBatch path and found - the callback optimization is real. But end to end is dominated by native post-processing.

Profiling data:
CalculateVertNormals: about 25.9% self time
Collider construction: about 22.4% self time
SetNormalsAndCoplanar: about 11.9% self time

So, these are the things currently taking the most of the time.

elalish:
@pca006132 Look good to you?

pca006132:
LGTM. I am not concerned about performance here: I think the target use case in the old discussion is about using GPU to set the vertex positions in parallel.

I think maybe we can remove the benchmark code and just keep the test code, since the benchmark is not useful for now.

swaparup36:
> LGTM. I am not concerned about performance here: I think the target use case in the old discussion is about using GPU to set the vertex positions in parallel.
> 
> I think maybe we can remove the benchmark code and just keep the test code, since the benchmark is not useful for now.

Okay I will remove the benchmark code.

Issue #1474, opened 2025-12-28
misc issues


pca006132:
1. Properly propagating status. We have it for some functions, but it is not comprehensive. E.g., we should properly handle this in csg tree, methods like `SetProperties`, etc.
2. Functions like `Refine` are copying too many things. We only need to copy `halfedge`, `halfedgeTangent`, and `vertPos`. The copying should also be conditional on whether we have `halfedgeTangent`.
3. ~~It seems that compiling with `x86-64-v4` with clang will cause issues with the `Boolean.AlmostCoplanar`, need investigation.~~ fixed

Some by AI:

4. ~~`Warp` and `WarpBatch` leave `halfedgeTangent_` stale, and later smooth related operations may give incorrect results. Maybe we should just clear the data.~~
5. `MergeMeshGLP`'s `openEdges` processing can be quite slow because it is single-threaded and uses `std::multimap`, which is not hash-based. Maybe we should optimize that.
6. ~~It thinks this code to be quite confusing, which I agree~~: https://github.com/elalish/manifold/blob/b918447ad1c297febd00cc9e35bddc8993a07a23/src/manifold.cpp#L539-L541
7. `CurvatureAngles` calculation is not robust to degenerate faces, it may produce NAN when `edgeLength` is zero: https://github.com/elalish/manifold/blob/b918447ad1c297febd00cc9e35bddc8993a07a23/src/properties.cpp#L41-L53

SouravVerma-art:
I'd like to work on this issue. I've reviewed the problem and have a good understanding of what needs to be done. Could you please assign this to me?

pca006132:
You can try 3, 5, 7.

1 can be a bit confusing, 2 is too simple (and is already done in my end), 4 needs discussion.

SouravVerma-art:
Got it, I’ll work on 3, 5, and 7 and open PRs for them. We can discuss 4 separately when you’re ready.

elalish:
I noticed 6 as well and am fixing it in #1475. 

I bet 4 is correct - should verify first with a simple test. Should be an easy fix too.

pca006132:
3 is now fixed

elalish:
6 is fixed too now. I think you made a comment somewhere about making more functions lazy, which you already did to the collider. After #1475, I really agree. For instance, the vert normals - we should just mark them dirty and then recompute them when needed. Possibly SetNormalsAndCoplanar should work similarly.

zmerlynn:
[...]

PR #1620, opened 2026-03-31
Make CurvatureAngles more robust to degenerate faces

Renovic:
Fixes 7 in #1474. It implements a some checks for degenerate edges and faces within Curvature Angles, and creates a test case to ensure no NANs are produced when degenerate faces are inputted.

### DIFF
--- a/src/properties.cpp
+++ b/src/properties.cpp
@@ -37,12 +37,24 @@ struct CurvatureAngles {
   void operator()(size_t tri) {
     vec3 edge[3];
     vec3 edgeLength(0.0);
+
+    // Checking for Degenerate Edges
     for (int i : {0, 1, 2}) {
       const int startVert = halfedge[3 * tri + i].startVert;
       const int endVert = halfedge[3 * tri + i].endVert;
       edge[i] = vertPos[endVert] - vertPos[startVert];
       edgeLength[i] = la::length(edge[i]);
+
+      if (!std::isfinite(edgeLength[i]) || edgeLength[i] == 0.0) {
+        return;
+      }
+    }
+
+    for (int i : {0, 1, 2}) {
       edge[i] /= edgeLength[i];
+
+      const int startVert = halfedge[3 * tri + i].startVert;
+      const int endVert = halfedge[3 * tri + i].endVert;
       const int neighborTri = halfedge[3 * tri + i].pairedHalfedge / 3;
       const double dihedral =
           0.25 * edgeLength[i] *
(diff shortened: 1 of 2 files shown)


test/properties_test.cpp
```diff
@@ -121,6 +122,38 @@ TEST(Properties, CalculateCurvature) {
   }
 }
 
+// This test verifies that curvature calculation does not produce NaN values on
+// degenerate faces
+
+TEST(Properties, DegenerateCurvature) {
+  Manifold::Impl impl;
```
elalish, line 129: In general we test Manifold rather than Impl to stick with testing the public API surface as much as possible. It's also a good opportunity to demonstrate how to read properties from MeshGL.

test/properties_test.cpp
```diff
@@ -121,6 +122,38 @@ TEST(Properties, CalculateCurvature) {
...
+
+  impl.halfedge_[3] = Halfedge{0, 2, 2, 0};
+  impl.halfedge_[4] = Halfedge{2, 1, 1, 1};
+  impl.halfedge_[5] = Halfedge{1, 0, 0, 2};
+
+  impl.faceNormal_.resize(2);
+  impl.faceNormal_[0] = vec3(0.0, 0.0, 1.0);
+  impl.faceNormal_[1] = vec3(0.0, 0.0, 1.0);
+
+  impl.CalculateCurvature(-1, 0);
```
elalish, line 149: Let's test both curvatures as long as we're here.

src/properties.cpp
```diff
@@ -309,9 +321,13 @@ void Manifold::Impl::CalculateCurvature(int gaussianIdx, int meanIdx) {
   for_each_n(policy, countAt(0), NumVert(),
              [&vertMeanCurvature, &vertGaussianCurvature, &vertArea,
               &degree](const int vert) {
-               const double factor = degree[vert] / (6 * vertArea[vert]);
-               vertMeanCurvature[vert] *= factor;
-               vertGaussianCurvature[vert] *= factor;
+               // Degenerate vertices can have zero area, which can cause the
+               // curvature to become NaN
+               if (vertArea[vert] > 0.0) {
```
elalish, line 326: What will the curvature be if the area is 0? It might be good to explicitly set it to zero here for clarity.

test/properties_test.cpp
Renovic: I'm a little confused on what to do. When using the Manifold constructor it automatically cleans up the degenerate faces, but I need them to test the CurvatureAngles fixes.

elalish: Ah, fair point. I guess that's why we never saw this bug in practice. Okay, you can leave the Impl test then.

PR #1631, opened 2026-04-08
feat: add axes indicator to model-viewer

swaparup36:
resolves #1627 

image

### DIFF
--- a/bindings/wasm/examples/editor/editor.js
+++ b/bindings/wasm/examples/editor/editor.js
@@ -780,10 +780,13 @@ function updateOrientationGrid() {
   const extentZ = Math.max(Math.abs(bounds.min.z), Math.abs(bounds.max.z));
   const modelHalfExtent = Math.max(extentX, extentZ, 0.1);
   const paddedHalfExtent = modelHalfExtent * 1.2;
-  const majorStep = ceilNiceNumber(Math.max(paddedHalfExtent * 2, 0.1 * 2) / 8);
+  const majorStep = ceilNiceNumber(Math.max(paddedHalfExtent, 0.1) / 8);
   const minorStep = majorStep / 10;
   const gridHalfExtent = Math.ceil(paddedHalfExtent / majorStep) * majorStep;
   const maxGridDimension = gridHalfExtent * 2;
+  const minorLineCountPerAxis =
+      Math.floor((maxGridDimension / minorStep) + 1e-6) + 1;
+  const showMinorLines = minorLineCountPerAxis <= 120;
 
   const nearAxis = value => Math.abs(value) < minorStep * 0.01;
   const isMajorLine = value => {
@@ -791,15 +794,32 @@ function updateOrientationGrid() {
     return Math.abs(value - nearest) <= minorStep * 0.01;
   };
 
-  const minorGeometry = createGridLinesGeometry(
-      gridHalfExtent, minorStep,
-      coordinate => !nearAxis(coordinate) && !isMajorLine(coordinate));
+  const minorGeometry = showMinorLines ?
+      createGridLinesGeometry(
+          gridHalfExtent, minorStep,
+          coordinate => !nearAxis(coordinate) && !isMajorLine(coordinate)) :
+      null;
   const majorGeometry = createGridLinesGeometry(
       gridHalfExtent, majorStep, coordinate => !nearAxis(coordinate));
 
   const grid = new Group();
   grid.userData[ORIENTATION_GRID_FLAG] = true;
   grid.renderOrder = 2;
+
+  const gridPlane = new Mesh(
+      new PlaneGeometry(maxGridDimension, maxGridDimension),
+      new MeshBasicMaterial({
+        color: 0x5a6169,
+        transparent: true,
+        opacity: 0.12,
+        side: 2,
+        depthWrite: false,
+        toneMapped: false,
+      }));
+  gridPlane.rotation.x = -Math.PI / 2;
+  gridPlane.position.y = -0.0005;
+  gridPlane.userData[ORIENTATION_GRID_FLAG] = true;
+  grid.add(gridPlane);
 
   if (minorGeometry) {
     const minorLines = new LineSegments(minorGeometry, new LineBasicMaterial({
--- a/bindings/wasm/examples/editor/editor.js
+++ b/bindings/wasm/examples/editor/editor.js
@@ -664,18 +664,8 @@ const ORIENTATION_GRID_KEY = '__orientationGrid';
 
 function ceilNiceNumber(value) {
   const positiveValue = Math.max(value, 1e-9);
-  const exponent = Math.floor(Math.log10(positiveValue));
-  const magnitude = 10 ** exponent;
-  const fraction = positiveValue / magnitude;
-  let roundedFraction = 10;
-  if (fraction <= 1) {
-    roundedFraction = 1;
-  } else if (fraction <= 2) {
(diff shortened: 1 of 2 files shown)


bindings/wasm/examples/editor/editor.js
```diff
@@ -690,6 +659,210 @@ let paused = false;
...
+  root.add(grid);
+  root.userData[ORIENTATION_GRID_KEY] = grid;
+
+  scene.queueRender?.();
+  if (needsRenderSym) {
+    mv[needsRenderSym]?.();
+  }
+
+  const filteredLines = consoleElement.textContent.split(/\r?\n/).filter(
+      line => line && !line.startsWith('Max grid dimension: '));
```
elalish: I like your clever use of CanvasTexture - since that makes text so easy, let's just put the label right on the grid and skip the console part. So each axis label can be e.g. 
```
X
10cm
```

bindings/wasm/examples/editor/editor.js
```diff
@@ -690,6 +659,210 @@ let paused = false;
...
+      new PlaneGeometry(axisStripWidth, maxGridDimension),
+      axisMaterial.clone());
+  yAxis.rotation.x = -Math.PI / 2;
+  yAxis.position.y = 0.001;
+  yAxis.userData[ORIENTATION_GRID_FLAG] = true;
+  grid.add(yAxis);
+
+  const labelSize =
+      clampToRange(maxGridDimension * 0.045, majorStep * 0.9, minorStep * 0.9);
+  const labelOffset = majorStep * 0.25;
```
elalish: I think I might choose different values here, but you've made them easily accessible, so probably best I just do that as a follow-on.

bindings/wasm/examples/editor/editor.js
```diff
@@ -690,6 +659,210 @@ let paused = false;
...
+
+function updateOrientationGrid() {
+  const scene = sceneSym ? mv[sceneSym] : null;
+  if (!scene) return;
+
+  const root = scene.model ?? scene;
+  const previousGrid = root.userData[ORIENTATION_GRID_KEY];
+  if (previousGrid) {
+    previousGrid.removeFromParent();
+    delete root.userData[ORIENTATION_GRID_KEY];
```
elalish, line 773: Have you looked for memory leaks? I don't think this is enough to fully clean up - many of three's objects have `dispose()` methods that need to be called manually because the garbage collector won't find them.

bindings/wasm/examples/editor/editor.js
```diff
@@ -690,6 +659,210 @@ let paused = false;
...
+  const exponent = Math.floor(Math.log10(positiveValue));
+  const magnitude = 10 ** exponent;
+  const fraction = positiveValue / magnitude;
+  let roundedFraction = 10;
+  if (fraction <= 1) {
+    roundedFraction = 1;
+  } else if (fraction <= 2) {
+    roundedFraction = 2;
+  } else if (fraction <= 5) {
+    roundedFraction = 5;
```
elalish: Let's simplify this so that major/minor grid lines can only show up at order of magnitude spacings, e.g. every 1mm, 1cm, 10cm, 100cm, 1m, etc.

elalish (changes requested):
Looking really good! Would you mind sharing a screen recording, just to demonstrate how it looks when you switch between objects of different scale / position?

swaparup36:
> Looking really good! Would you mind sharing a screen recording, just to demonstrate how it looks when you switch between objects of different scale / position?

Sure!

bindings/wasm/examples/editor/editor.js
swaparup36: done!

bindings/wasm/examples/editor/editor.js
swaparup36: handled that.

bindings/wasm/examples/editor/editor.js
swaparup36: Yes, always. You can tell me either what you prefer.

swaparup36:
Here is the demo of how it looks when I switch between objects of different scale / position.

https://github.com/user-attachments/assets/da30b496-7ed4-4f43-a29b-d774ac11d270

bindings/wasm/examples/editor/editor.js
```diff
@@ -690,6 +659,223 @@ let paused = false;
 let showEdges = false;
 const EDGE_KEY = 'edgeLines';
 const EDGE_OVERLAY_FLAG = '__isEdgeOverlay';
+const ORIENTATION_GRID_FLAG = '__isOrientationGrid';
+const ORIENTATION_GRID_KEY = '__orientationGrid';
+
+function ceilNiceNumber(value) {
+  const positiveValue = Math.max(value, 1e-9);
+  const exponent = Math.ceil(Math.log10(positiveValue));
+  return 10 ** exponent;
```
elalish, line 668: This is good for the major grid lines, but the minors should always be 10x more, unless there are too many and we just skip minor grid lines entirely.

bindings/wasm/examples/editor/editor.js
```diff
@@ -690,6 +659,223 @@ let paused = false;
...
+    toneMapped: false,
+    depthWrite: false,
+  });
+  const mesh = new Mesh(new PlaneGeometry(size, size), material);
+  mesh.userData[ORIENTATION_GRID_FLAG] = true;
+  mesh.rotation.x = -Math.PI / 2;
+  return mesh;
+}
+
+function createGridLinesGeometry(gridHalfExtent, spacing, includeLine) {
```
elalish, line 725: The height of the grid plane isn't quite obvious enough yet - I think we should add a bit of transparent gray to the whole grid quad.

bindings/wasm/examples/editor/editor.js
```diff
@@ -690,6 +659,223 @@ let paused = false;
...
+  }
+
+  const bounds = new Box3().setFromObject(root);
+  if (bounds.isEmpty()) return;
+
+  const extentX = Math.max(Math.abs(bounds.min.x), Math.abs(bounds.max.x));
+  const extentZ = Math.max(Math.abs(bounds.min.z), Math.abs(bounds.max.z));
+  const modelHalfExtent = Math.max(extentX, extentZ, 0.1);
+  const paddedHalfExtent = modelHalfExtent * 1.2;
+  const majorStep = ceilNiceNumber(Math.max(paddedHalfExtent * 2, 0.1 * 2) / 8);
```
elalish: I think the grid is about twice the size it needs to be - maybe one of these *2 could be removed? Not sure which.

elalish (changes requested):
This is looking excellent, but I think we can still improve a few things.

bindings/wasm/examples/editor/editor.js
swaparup36: fixed! removed both.

bindings/wasm/examples/editor/editor.js
swaparup36: done!

bindings/wasm/examples/editor/editor.js
swaparup36: capped minor axis per major axis.

elalish (approved):
Thanks for doing the heavy lifting - I'll do a quick follow-on with my nit-picks; it's easier if I can see the results realtime.

tonious:
This is slick!  Nicely done!

PR #1632, opened 2026-04-09
Fix meshgl merge functions returning input pointer on no-op

zmerlynn:
## Summary

`manifold_meshgl_merge` and `manifold_meshgl64_merge` always return the caller-owned copy now, instead of returning the input pointer when `Merge()` finds nothing to do.

## Problem

When `Merge()` returns `false` (mesh is already manifold), both functions destruct the valid copy in `mem` and return the input pointer `m`. This violates the C API's ownership convention: every function taking `void* mem` returns a pointer into that buffer. Callers (including binding layers) that assume the returned pointer is the output allocation will double-free or leak `mem`.

This appears to have been an oversight in the original implementation (#820), which also used `delete` on a placement-new'd object. #1196 correctly fixed the UB by switching to an explicit destructor call, but the `return m` remained.

## Fix

`Merge()` mutates in place and leaves the mesh valid regardless of its return value — when it returns `false`, it simply means nothing needed to change. The copy is always a correct result, so we can return it unconditionally. This matches the pattern of `manifold_meshgl_copy` and every other `void* mem` function in the C API.

## Test plan

- [x] New `CBIND.meshgl_merge_returns_mem` and `CBIND.meshgl64_merge_returns_mem` tests: verify the returned pointer comes from `mem` and the result constructs a valid Manifold
- [x] Full test suite (321 tests) passes

### DIFF
--- a/bindings/c/manifoldc.cpp
+++ b/bindings/c/manifoldc.cpp
@@ -548,11 +548,8 @@ ManifoldMeshGL* manifold_meshgl_copy(void* mem, ManifoldMeshGL* m) {
 
 ManifoldMeshGL* manifold_meshgl_merge(void* mem, ManifoldMeshGL* m) {
   auto duplicate = new (mem) MeshGL(*from_c(m));
-  if (duplicate->Merge()) {
-    return to_c(duplicate);
-  }
-  duplicate->~MeshGL();
-  return m;
+  duplicate->Merge();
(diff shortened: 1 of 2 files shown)


elalish (approved):
Thank you! I'm no C expert - @pca006132 does this look good to you?

pca006132 (approved):

PR #1633, opened 2026-04-09
Add missing C API bindings

zmerlynn:
## Summary

- Add 16 new C API functions exposing `Manifold` and `MeshGL`/`MeshGL64`
  methods that had no C binding:
  - `manifold_get_tolerance`, `manifold_set_tolerance`, `manifold_simplify`,
    `manifold_num_prop_vert`
  - `manifold_meshgl{,64}_tolerance`, `_run_flags_length`, `_run_flags`,
    `_num_run`, `_update_normals`
- Add thread safety note to the C header: objects are not safe for concurrent
  access, even read-only, due to lazy evaluation with mutable internal state.
- Add tests for all new bindings.

## Motivation

These gaps were found while building [manifold-csg](https://github.com/zmerlynn/manifold-csg),
a new Rust FFI binding to the C API. Missing accessors for `runFlags`, `tolerance`,
`NumRun`, and `UpdateNormals` prevent lossless round-tripping of `MeshGL` data.
`GetTolerance`/`SetTolerance`/`Simplify` are useful public `Manifold` methods
with no C-side equivalent (`manifold_epsilon` exists but calls the `GetEpsilon`
testing hook rather than the public `GetTolerance`).

The thread safety note documents the `mutable shared_ptr` lazy-evaluation
pattern in `Manifold` and `CrossSection`, which is relevant for FFI consumers
that need to decide whether concurrent read access is safe (it isn't).

## Test plan

- [x] All 15 CBIND tests pass (4 new: `tolerance`, `num_prop_vert`,
  `meshgl_run_accessors`, `meshgl_update_normals`)
- [x] No changes to existing tests or behavior

### DIFF
--- a/bindings/c/include/manifold/manifoldc.h
+++ b/bindings/c/include/manifold/manifoldc.h
@@ -173,6 +173,10 @@ ManifoldManifold* manifold_refine_to_length(void* mem, ManifoldManifold* m,
                                             double length);
 ManifoldManifold* manifold_refine_to_tolerance(void* mem, ManifoldManifold* m,
                                                double tolerance);
+ManifoldManifold* manifold_set_tolerance(void* mem, ManifoldManifold* m,
+                                         double tolerance);
+ManifoldManifold* manifold_simplify(void* mem, ManifoldManifold* m,
+                                    double tolerance);
 
 // Manifold Shapes / Constructors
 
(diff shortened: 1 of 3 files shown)


pca006132:
IIRC the usage of the shared ptrs should be guarded by a mutex, so operations should be thread safe. I haven't really looked into that (I should have), but if it is not thread safe we should fix that.

zmerlynn:
Good point — I took a look and it seems like `Manifold::pNode_` itself isn't guarded, even though `CsgOpNode::impl_` uses `ConcurrentSharedPtr` with a mutex. The race is in `GetCsgLeafNode()` which reassigns `pNode_` from `const` methods without synchronization.

That said, this is orthogonal to the bindings PR, so I'll remove the thread safety comment from this PR and follow up with a separate issue to discuss the fix. (A `std::call_once` approach looks promising — low overhead on the fast path and naturally shareable across copies.)

elalish (approved):
Thank you! Our original C API contributor is long gone, so it's great to see this getting some love.

I'm also excited to see the Rust bindings you're working on - we've had several users interested in having that. It'll be great to add to the list when it's ready.

PR #1636, opened 2026-04-10
Make Manifold and CrossSection safe for concurrent const access

### LINKED ISSUE #1635: Manifold::pNode_ is not thread-safe for concurrent reads
## Problem

`Manifold::pNode_` is a `mutable std::shared_ptr<CsgNode>` that gets reassigned without synchronization in `GetCsgLeafNode()`:

https://github.com/elalish/manifold/blob/d88e2950/include/manifold/manifold.h#L533 https://github.com/elalish/manifold/blob/d88e2950/src/manifold.cpp#L145-L150

Every `const` method that triggers lazy CSG evaluation goes through this path, so two threads calling e.g. `NumVert()` on the same `Manifold` object can race on the `shared_ptr` assignment. `CrossSection` has the same pattern with `mutable shared_ptr<const PathImpl> paths_` and `mutable mat2x3 transform_`.

This came up in review of #1633 — see https://github.com/elalish/manifold/pull/1633#issuecomment-4220178271

Note that `CsgOpNode::impl_` already uses `ConcurrentSharedPtr` with a `recursive_mutex` for exactly this kind of guarded access, but `Manifold::pNode_` does not.

## Impact

FFI binding authors (Rust, Python, etc.) need to know whether concurrent read access is safe. Currently it is not, which means Rust bindings must mark `Manifold` and `CrossSection` as `!Sync`.

## Possible fix

`GetCsgLeafNode()` is a lazy one-shot evaluation — evaluate the CSG tree once, then it stays as a leaf. `std::call_once` fits naturally:

```cpp
mutable std::shared_ptr<std::once_flag> pNodeOnce_ =
    std::make_shared<std::once_flag>();
mutable std::shared_ptr<CsgNode> pNode_;

CsgLeafNode& Manifold::GetCsgLeafNode() const {
  std::call_once(*pNodeOnce_, [this] {
    if (pNode_->GetNodeType() != CsgNodeType::Leaf) {
      pNode_ = pNode_->ToLeafNode();
    }
  });
  return *std::static_pointer_cast<CsgLeafNode>(pNode_);
}
```

Cost is one extra `shared_ptr` (16 bytes) per `Manifold` — the `once_flag` is naturally shared across copies since they share the same `pNode_`. After first evaluation, `call_once` is just an atomic flag check on the fast path.

zmerlynn:
## Summary

Add a per-instance `shared_ptr<mutex>` to both `Manifold` and `CrossSection`, protecting the mutable `pNode_`/`paths_` fields from data races when multiple threads call const methods on the same object.

All reads of `pNode_` in const methods go through either `LoadPNode()` (builder methods like `Translate`, `Boolean`) or `GetCsgLeafNode()` (query methods like `NumVert`, `GetMeshGL`), both of which lock the mutex. `CrossSection::GetPaths()` and `Transform()` similarly lock `pathsMutex_`.

The mutex is wrapped in `shared_ptr` so it works with defaulted move constructors. Copy constructor and `operator=` lock the source's mutex to read its fields, and `operator=` creates a fresh mutex for the destination.

## Design

Cost is one `shared_ptr` (16 bytes) per object plus an uncontended mutex lock (~20ns) per method call — negligible compared to actual mesh operations. We considered `std::call_once` (protects queries but not builders) and C++17 `std::atomic_load`/`std::atomic_store` on `shared_ptr`, but the per-object mutex is clearer — all `pNode_` access goes through two accessors (`LoadPNode` and `GetCsgLeafNode`), making it hard to accidentally add an unprotected read.

When the project moves to C++20, this can be simplified to `std::atomic<std::shared_ptr<CsgNode>>` with no mutex needed (see #1153).

Fixes #1635

## Test plan

- [x] All 326 existing tests pass
- [x] No behavioral changes for single-threaded use

### DIFF
--- a/src/cross_section/cross_section.cpp
+++ b/src/cross_section/cross_section.cpp
@@ -228,12 +228,14 @@ CrossSection& CrossSection::operator=(CrossSection&&) noexcept = default;
  * const methods.
  */
 CrossSection::CrossSection(const CrossSection& other) {
+  std::lock_guard<std::mutex> lock(*other.pathsMutex_);
   paths_ = other.paths_;
   transform_ = other.transform_;
 }
 
 CrossSection& CrossSection::operator=(const CrossSection& other) {
   if (this != &other) {
+    std::scoped_lock lock(*pathsMutex_, *other.pathsMutex_);
     paths_ = other.paths_;
     transform_ = other.transform_;
   }
@@ -295,6 +297,7 @@ CrossSection::CrossSection(const Rect& rect) {
 // All access to paths_ should be done through the GetPaths() method, which
 // applies the accumulated transform_
 std::shared_ptr<const PathImpl> CrossSection::GetPaths() const {
+  std::lock_guard<std::mutex> lock(*pathsMutex_);
   if (transform_ == mat2x3(la::identity)) {
     return paths_;
   }
@@ -556,6 +559,7 @@ CrossSection CrossSection::Mirror(const vec2 ax) const {
  * @param m The affine transform matrix to apply to all the vertices.
  */
 CrossSection CrossSection::Transform(const mat2x3& m) const {
+  std::lock_guard<std::mutex> lock(*pathsMutex_);
   auto transformed = CrossSection();
   transformed.transform_ = m * Mat3(transform_);
   transformed.paths_ = paths_;
--- a/src/manifold.cpp
+++ b/src/manifold.cpp
@@ -122,7 +122,10 @@ Manifold::~Manifold() = default;
 Manifold::Manifold(Manifold&&) noexcept = default;
 Manifold& Manifold::operator=(Manifold&&) noexcept = default;
 
-Manifold::Manifold(const Manifold& other) : pNode_(other.pNode_) {}
+Manifold::Manifold(const Manifold& other) {
+  std::lock_guard<std::mutex> lock(*other.pNodeMutex_);
+  pNode_ = other.pNode_;
+}
 
 Manifold::Manifold(std::shared_ptr<CsgNode> pNode) : pNode_(pNode) {}
 
@@ -143,12 +146,19 @@ Manifold Manifold::PropagateStatus(Error status) {
 
 Manifold& Manifold::operator=(const Manifold& other) {
   if (this != &other) {
+    std::scoped_lock lock(*pNodeMutex_, *other.pNodeMutex_);
     pNode_ = other.pNode_;
   }
   return *this;
 }
 
+std::shared_ptr<CsgNode> Manifold::LoadPNode() const {
+  std::lock_guard<std::mutex> lock(*pNodeMutex_);
+  return pNode_;
+}
+
 CsgLeafNode& Manifold::GetCsgLeafNode() const {
+  std::lock_guard<std::mutex> lock(*pNodeMutex_);
   if (pNode_->GetNodeType() != CsgNodeType::Leaf) {
     pNode_ = pNode_->ToLeafNode();
   }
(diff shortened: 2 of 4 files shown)


pca006132:
I'm thinking about whether to use this or mutex, since I doubt if we can move to c++20 in a short period of time, since we probably want to wait until the major users move to c++20.

I am not worried about the overhead introduced by mutex (probably a `std::shared_ptr<std::mutex>` so it is cloneable), since the overhead is minuscule compared to constructing a MeshGL object.

zmerlynn:
Good point about the C++20 timeline. I'm happy to switch to a `shared_ptr<mutex>` — same size cost as the `once_flag` approach and it closes the builder method gap entirely.

One question on the semantics: should copies share the mutex or get independent ones? Currently the copy constructor copies `pNode_` (so both copies start with the same `shared_ptr<CsgNode>` value) but each has its own member that gets independently written during evaluation. So I think each copy needs its own mutex — sharing would be wrong since the mutex protects the per-instance `pNode_` member, not the underlying `CsgNode`. Does that match your thinking?

Concretely, the approach would be:
- `mutable shared_ptr<recursive_mutex> pNodeMutex_` (fresh per instance, same pattern as `ConcurrentSharedPtr` in `utils.h`)
- All `pNode_` reads go through a `GetPNode()` accessor that locks + returns a copy of the shared_ptr
- `GetCsgLeafNode()` locks for the read-check-write
- Copy constructor / `operator=` get a fresh mutex (and lock the source's mutex to read `pNode_`)
- Move stays defaulted

zmerlynn:
Updated the PR to use `shared_ptr<mutex>` as you suggested. All `pNode_` reads now go through two locked accessors (`LoadPNode` for builder methods, `GetCsgLeafNode` for query methods), so both query+query and query+builder concurrent access is safe.

Approach matches what I outlined in the earlier comment — per-instance mutex (non-recursive is sufficient), fresh mutex on copy/assign, source locked during copy. `CrossSection` gets the same treatment for `paths_`/`transform_`.

src/cross_section/cross_section.cpp
```diff
@@ -228,12 +228,15 @@ CrossSection& CrossSection::operator=(CrossSection&&) noexcept = default;
  * const methods.
  */
 CrossSection::CrossSection(const CrossSection& other) {
+  std::lock_guard<std::mutex> lock(*other.pathsMutex_);
   paths_ = other.paths_;
   transform_ = other.transform_;
 }
 
 CrossSection& CrossSection::operator=(const CrossSection& other) {
```
pca006132, line 236: I think for this to be thread safe, we need another mutex for the `CrossSection` object itself?

src/manifold.cpp
```diff
@@ -137,12 +140,20 @@ Manifold Manifold::Invalid() {
 
 Manifold& Manifold::operator=(const Manifold& other) {
   if (this != &other) {
+    std::lock_guard<std::mutex> lock(*other.pNodeMutex_);
```
pca006132: Same issue as the `CrossSection` copy constructor.

src/cross_section/cross_section.cpp
zmerlynn: Good catch — updated `operator=` for both classes to lock `this` as well, using `std::scoped_lock` for deadlock avoidance. Left the copy constructors locking only the source since `this` is under construction and can't be accessed concurrently.

src/manifold.cpp
zmerlynn: Same fix applied here — `std::scoped_lock` on both mutexes to avoid deadlock from concurrent `a = b` / `b = a`.

pca006132 (approved):

PR #1637, opened 2026-04-11
Propagate error status through Manifold operations

zmerlynn:
## Summary

Error status is supposed to propagate through operations "like NaN" — once a Manifold has an error, it should stick. Several code paths silently dropped it or did expensive work on errored inputs.

**Status was silently lost:**
- `Decompose`: created fresh `Impl` objects for each component without copying status
- `Hull`: created fresh `Impl` without copying status
- `Hull(vector<Manifold>)`: didn't check input statuses at all

**Expensive work on errored inputs:**
- `SetProperties`, `CalculateCurvature`, `CalculateNormals`, `SmoothByNormals`, `SmoothOut`, `Refine`, `RefineToLength`, `RefineToTolerance`, `SetTolerance`, `Simplify`: these copy the `Impl` (which preserves status) but proceed to do expensive mesh operations on the errored copy

**Fix:** Add a private `PropagateStatus(Error)` helper (alongside the existing `Invalid()`) that creates an empty Manifold with a specific error code. Add early-return guards to all 13 affected methods, and refactor the 3 existing guards in `AsOriginal`, `Warp`, `WarpBatch` to use the new helper.

Partial fix for #1474

## Test plan

- [x] 12 new tests (`ErrorPropagation*`) covering every guarded method
- [x] All 338 tests pass (326 existing + 12 new)
- [x] No behavioral changes for valid inputs

### DIFF
--- a/test/manifold_test.cpp
+++ b/test/manifold_test.cpp
@@ -171,6 +171,126 @@ TEST(Manifold, InvalidInput7) {
   EXPECT_EQ(tet.Status(), Manifold::Error::RunIndexWrongLength);
 }
 
+TEST(Manifold, ErrorPropagationDecompose) {
+  MeshGL in = TetGL();
+  in.vertProperties[2 * 3 + 1] = NAN;
+  Manifold errored(in);
+  ASSERT_EQ(errored.Status(), Manifold::Error::NonFiniteVertex);
+  auto parts = errored.Decompose();
+  ASSERT_EQ(parts.size(), 1);
+  EXPECT_EQ(parts[0].Status(), Manifold::Error::NonFiniteVertex);
+}
+
+TEST(Manifold, ErrorPropagationHull) {
+  MeshGL in = TetGL();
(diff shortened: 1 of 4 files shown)


zmerlynn:
cc @pca006132

test/manifold_test.cpp
```diff
@@ -183,6 +303,7 @@ TEST(Manifold, ObjRoundTrip) {
 
 TEST(Manifold, OppositeFace) {
   MeshGL gl;
+  // clang-format off
```
elalish: What happened here? Are on we using different versions of clang format?

elalish (approved):
Thanks, I like the consistency. I'll defer to @pca006132 for the specifics.

test/manifold_test.cpp
zmerlynn: Yeah, I was on clang-format 22 which reformats those vertex arrays. Switched to 20 to match CI, dropped the guards.

pca006132 (approved):

PR #1639, opened 2026-04-11
Fix README "bare" typos

klausw:
"bare" implies a lack of clothing, I assume you meant "bear a resemblance".


elalish (approved):
Ha, thanks Klaus! Good to hear from you again, it's been ages. Are you still at Google? Is this a pure drive-by, or are you thinking of using Manifold for something?

Issue #1640, opened 2026-04-11
Add ray casting via existing BVH


zmerlynn:
The internal `Collider` (radix tree over triangle AABBs) seems like a natural fit for ray casting queries. I'd like to propose adding two methods to `Manifold`:

```cpp
struct RayHit {
  int triIdx;
  double distance;
  vec3 barycentric;
  vec3 normal;  // geometric face normal (free lookup from faceNormal_[triIdx])
};

std::optional<RayHit> RayCastFirst(vec3 origin, vec3 direction,
    double maxDist = inf) const;
std::vector<RayHit> RayCastAll(vec3 origin, vec3 direction,
    double maxDist = inf) const;
```

`RayCastFirst` returns the nearest hit; `RayCastAll` returns all hits sorted by distance. Both would use the existing BVH with a ray-AABB slab test for traversal (paralleling `FindCollision` but with ray-specific culling), and Moller-Trumbore for the narrow phase. Ray primitives would live in a new `src/ray.h` alongside `tri_dist.h`.

My use case is mold design — I need to cast rays against manifolds for undercut detection. Currently doing this with O(n) brute force outside the library. This may also be useful for the mesh offset work in #192 (@zalo noted "Manifold would just need to be able to do raycasts with its BVH"), and I imagine rendering/simulation users could benefit as well.

I'd be happy to start implementing this if the API direction looks right. I'd stage it as:

1. Ray primitives + `Collider::RayCast` traversal (internal, no public API)
2. Public C++ API (`RayHit`, `Manifold::RayCastFirst`/`RayCastAll`)
3. C API bindings

elalish:
Given that I now work in path-trace rendering, I definitely see the utility of a ray caster, and you're right that it fits pretty well here. 

However I'm concerned about the numerical stability of Moller-Trumbore. I think it would be better if we reuse the internals of boolean3 - after all, edge-tri intersection is basically the same as ray-tri intersection. The difference is that ours uses no espilon values and will always give consistent results, e.g. if you hit an edge exactly between two front-facing triangles it will always return a hit with exactly one of the triangles - never neither or both, which I think Moller-Trumbore will have trouble with. 

Your API looks good, though I'm tempted to change the input to `vec3 origin, vec3 endpoint` in order to not hide any internal rounding errors. It might be less ergonomic though - I'm interested in your feedback here as a user.

zalo:
For a winding number calculation via raycasting, it’d be nice to have a convenient infinite length ray…

[...]

PR #1642, opened 2026-04-12
Add RayCast, WindingNumber, and NearestPoint query APIs

zalo:
Ref: #1640

## Summary

Implements spatial query APIs for Manifold surfaces, reusing the robust edge-crossing math from `boolean3.cpp` (Shadow01, Kernel02, Interpolate, Shadows) as discussed in the issue thread.

### New APIs

- **`RayCast(origin, endpoint)`** — Segment intersection returning nearest hit. Uses ray-plane intersection + Shadow01-style 2D containment with `GetAxisAlignedProjection`-based coordinate permutation and `vertNormal_` symbolic-perturbation tie-breaking. Reuses `Shadows()` and `Interpolate()` from boolean3.
- **`RayCast(origin, direction, maxDist)`** — Direction + distance form. When `maxDist` is infinity, the ray is clipped to the mesh bounding box for precision (as discussed in the issue).
- **`WindingNumber(point)`** — Directly calls `Kernel02<false, true>` via a minimal single-vertex `Impl`, reusing the exact same BVH Z-projection query pattern as `Winding03_`. No new intersection math.
- **`NearestPoint(point)`** — Closest point on mesh surface via BVH-accelerated expanding-radius search.

### Design: reuse of boolean3 internals

Per elalish's feedback, `WindingNumber` and `RayCast` are implemented in `boolean3.cpp` (same translation unit as the anonymous-namespace kernels) so they can call the boolean engine's mathematical primitives directly:

- **WindingNumber**: Creates a one-vertex `Impl` with zero perturbation normal, passes it to `Kernel02<false, true>` alongside the mesh. The BVH query uses `DoesOverlap(vec3)` for XY containment (Z-axis ray filter). This is the same pattern as `Winding03_`.
- **RayCast**: After ray-plane intersection, tests 2D containment using the same `Shadows()` and `Interpolate()` functions that `Shadow01` calls, with the same `faceNormal_` edge-direction tie-breaking and `vertNormal_` symbolic perturbation.
- **NearestPoint** lives in `queries.cpp` — it's a distance-based query, not an intersection query, so it doesn't have a boolean3 equivalent to reuse.

### Bindings

All exposed through C++, C, Python, and JavaScript/WASM:
- **C**: `manifold_ray_cast()`, `manifold_ray_cast_direction()`, `manifold_winding_number()`, `manifold_nearest_point()`
- **Python**: `ray_cast(origin, endpoint)`, `ray_cast(origin, direction, max_dist)`, `winding_number(point)`, `nearest_point(point)`
- **JS/WASM**: `rayCast(origin, endpoint)`, `rayCast(origin, direction, maxDist)`, `windingNumber(point)`, `nearestPoint(point)`

## Test plan

- [x] 21 C++ GoogleTest cases:
  - RayCast: hit, miss, sphere, internal origin, empty, normal direction, direction form, finite distance
  - WindingNumber: inside/outside cube/sphere, empty, boundary
  - NearestPoint: face, interior, sphere, corner, empty
  - Remeshing: sphere via LevelSet using WindingNumber+NearestPoint as SDF
  - MeshSDFSignCorrectness: validates correct sign at inside/outside/near-surface points
- [x] 3 C binding tests (CBIND.ray_cast, CBIND.winding_number, CBIND.nearest_point)
- [x] All 353 tests pass (including all existing tests)
- [x] clang-format applied to all changed files


zalo:
Closing and reopening to clear stale review.

PR #1643, opened 2026-04-12
Add RayCast, WindingNumber, and NearestPoint query APIs

zalo:
Ref: #1640

## Summary

Implements spatial query APIs for Manifold surfaces, reusing the robust edge-crossing math from `boolean3.cpp` (Shadow01, Kernel02, Interpolate, Shadows) as discussed in the issue thread.

### New APIs

- **`RayCast(origin, endpoint)`** — Segment intersection returning nearest hit. Uses ray-plane intersection + Shadow01-style 2D containment with `GetAxisAlignedProjection`-based coordinate permutation and `vertNormal_` symbolic-perturbation tie-breaking. Reuses `Shadows()` and `Interpolate()` from boolean3.
- **`RayCast(origin, direction, maxDist)`** — Direction + distance form. When `maxDist` is infinity, the ray is clipped to the mesh bounding box for precision (as discussed in the issue). Early-returns on empty/non-finite bbox to prevent NaN propagation.
- **`WindingNumber(point)`** — Directly calls `Kernel02<false, true>` via a minimal single-vertex `Impl`, reusing the exact same BVH Z-projection query pattern as `Winding03_`. No new intersection math.
- **`NearestPoint(point)`** — Closest point on mesh surface via BVH-accelerated expanding-radius search. Uses squared-distance comparisons during traversal (sqrt only at the end).

### Design: reuse of boolean3 internals

Per elalish's [feedback](https://github.com/elalish/manifold/issues/1640#issuecomment-2792116519), `WindingNumber` and `RayCast` are implemented in `boolean3.cpp` (same translation unit as the anonymous-namespace kernels) so they can call the boolean engine's mathematical primitives directly:

- **WindingNumber**: Creates a one-vertex `Impl` with zero perturbation normal, passes it to `Kernel02<false, true>` alongside the mesh. The BVH query uses `DoesOverlap(vec3)` for XY containment (Z-axis ray filter). This is the same pattern as `Winding03_`.
- **RayCast**: After ray-plane intersection, tests 2D containment using the same `Shadows()` and `Interpolate()` functions that `Shadow01` calls, with the same `faceNormal_` edge-direction tie-breaking and `vertNormal_` symbolic perturbation.
- **NearestPoint** lives in `queries.cpp` — it's a distance-based query with no boolean3 equivalent. Uses the existing `Collider` BVH with an expanding-then-refining two-pass search.

### Bindings

All exposed through C++, C, Python, and JavaScript/WASM:

| Language | Segment cast | Direction cast | Winding number | Nearest point |
|----------|-------------|---------------|---------------|--------------|
| C++ | `RayCast(origin, endpoint)` | `RayCast(origin, dir, maxDist)` | `WindingNumber(point)` | `NearestPoint(point)` |
| C | `manifold_ray_cast()` | `manifold_ray_cast_direction()` | `manifold_winding_number()` | `manifold_nearest_point()` |
| Python | `ray_cast(origin, endpoint)` | `ray_cast(origin, dir, max_dist)` | `winding_number(point)` | `nearest_point(point)` |
| JS/WASM | `rayCast(origin, endpoint)` | `rayCast(origin, dir, maxDist)` | `windingNumber(point)` | `nearestPoint(point)` |

JS `windingNumber` includes a `vararg2vec3` wrapper so `[x,y,z]` tuple input works consistently.

## Test plan

- [x] 35 C++ GoogleTest cases:
  - RayCast: hit, miss, sphere, internal origin, empty, normal direction, direction form (infinity and finite), along edge, through vertex, parallel to face, diagonal through sphere, translated mesh
  - WindingNumber: inside/outside cube/sphere, empty, boundary, below edge, below vertex, exhaustive grid, translated mesh, tiny mesh (1e-6 scale)
  - NearestPoint: face, interior, sphere, corner, empty, equidistant, far away (distance 1000), translated mesh, thin slab
  - Remeshing: sphere via LevelSet using WindingNumber+NearestPoint as SDF
  - MeshSDFSignCorrectness: validates correct sign at inside/outside/near-surface points
- [x] 3 C binding tests
- [x] All 367 tests pass (including all existing tests)
- [x] clang-format applied to all changed files


zalo:
Closing for now

PR #1645, opened 2026-04-12
Add ray casting via Kernel12 edge-triangle intersection

zmerlynn:
## Summary

Implement `Manifold::RayCast(origin, endpoint)` and `Manifold::RayCastAll(origin, endpoint)` by constructing a minimal single-edge `Impl` for the ray segment and calling `Kernel12` from boolean3 directly. This reuses the full symbolic perturbation machinery (Shadow01, Kernel02, Kernel11, Kernel12) for watertight results at shared edges and vertices, with zero new intersection math.

The ray `Impl` has two vertices (origin, endpoint), two halfedges forming one edge, and zero normals — so the ray contributes nothing to perturbation tiebreakers, and consistency at shared edges/vertices depends entirely on the mesh's own normals. Candidate triangles are filtered via the BVH using the ray's AABB.

- **`RayCast`** returns the nearest hit as a `RayHit` (triangle index, parametric distance `[0, 1]`, 3D position, face normal). Returns `faceID == -1` on miss.
- **`RayCastAll`** returns all hits sorted by distance.

Both share a `RayCastImpl` helper that builds the ray Impl and runs the BVH query.

## Follow-up work

- **Bindings**: C API (`ManifoldRayHit`, `manifold_ray_cast`), Python (`.ray_cast()`), WASM (`.rayCast()`)
- **Convenience overload**: `RayCast(origin, direction, maxDist)` for the direction + distance form

Partial implementation for #1640

## Test plan

- [x] 19 new tests: RayCast (14) + RayCastAll (5) covering all three axes, diagonal rays, sphere, nearest-hit, two-object traversal, empty manifold, zero-length ray, watertight edge/vertex, normals, hit positions
- [x] All 358 tests pass (326 existing + 19 new + 13 from other merged PRs)

### DIFF
--- a/include/manifold/common.h
+++ b/include/manifold/common.h
@@ -155,6 +155,22 @@ struct Smoothness {
 };
 
 /**
+ * @brief Result of a ray cast query against a Manifold.
+ */
+struct RayHit {
+  /// The triangle index that was hit.
+  uint64_t faceID = 0;
+  /// The parametric distance along the ray segment in the closed interval
+  /// [0, 1], where 0 is the origin and 1 is the endpoint. Hits exactly at
+  /// the origin or endpoint are included.
+  double distance = 0;
+  /// The 3D position of the hit point.
+  vec3 position = vec3(0.0);
+  /// The geometric face normal at the hit.
+  vec3 normal = vec3(0.0);
+};
+
+/**
  * @brief Axis-aligned 3D box, primarily for bounding.
  */
 struct Box {
--- a/include/manifold/manifold.h
+++ b/include/manifold/manifold.h
@@ -397,6 +397,12 @@ class Manifold {
   double MinGap(const Manifold& other, double searchLength) const;
   ///@}
 
+  /** @name Spatial Queries
+   */
+  ///@{
+  std::vector<RayHit> RayCast(vec3 origin, vec3 endpoint) const;
+  ///@}
+
   /** @name Mesh ID
    *  Details of the manifold's relation to its input meshes, for the purposes
    * of reapplying mesh properties.
--- a/src/boolean3.cpp
+++ b/src/boolean3.cpp
@@ -528,4 +528,66 @@ Boolean3::Boolean3(const Manifold::Impl& inP, const Manifold::Impl& inQ,
   }
 #endif
 }
+std::vector<RayHit> Manifold::Impl::RayCast(vec3 origin, vec3 endpoint) const {
+  ZoneScoped;
+  if (IsEmpty()) return {};
+  const vec3 dir = endpoint - origin;
+  if (la::dot(dir, dir) == 0.0) return {};
+
+  // Build a minimal single-edge Impl representing the ray segment.
+  // Kernel12 treats inA as an edge mesh, so we need two vertices, two
+  // halfedges (forward + backward), and one face normal. Zero vertex normals
+  // and face normal mean the ray contributes nothing to perturbation
+  // tiebreakers — consistency at shared edges/vertices depends entirely on
+  // the mesh's own normals.
+  Impl rayImpl;
+  rayImpl.vertPos_.resize(2);
+  rayImpl.vertPos_[0] = origin;
+  rayImpl.vertPos_[1] = endpoint;
+  rayImpl.vertNormal_.resize(2);
+  rayImpl.vertNormal_[0] = vec3(0.0);
+  rayImpl.vertNormal_[1] = vec3(0.0);
+  rayImpl.halfedge_.resize(2);
+  rayImpl.halfedge_[0] = {0, 1, 1, 0};  // forward: vert 0 → 1
+  rayImpl.halfedge_[1] = {1, 0, 0, 0};  // backward: vert 1 → 0
+  rayImpl.faceNormal_.resize(1);
+  rayImpl.faceNormal_[0] = vec3(0.0);
+
+  // expandP=false with zero vertNormal means the ray-side perturbation is
+  // zero. forward=true means we project along +Z for the lower-dimensional
+  // kernel cascade (Shadow01 → Kernel02 → Kernel11 → Kernel12).
+  Kernel02<false, true> k02{rayImpl, *this};
+  Kernel11<false> k11{rayImpl, *this};
+  Kernel12<false, true> k12{rayImpl, *this, k02, k11};
+
+  // Use the component with largest magnitude for stable t computation.
+  const vec3 absDir = la::abs(dir);
+  const int tAxis = absDir.x > absDir.y && absDir.x > absDir.z ? 0
+                    : absDir.y > absDir.z                      ? 1
+                                                               : 2;
+
+  std::vector<RayHit> hits;
+  // Query the BVH with the ray's AABB.
+  const Box rayBox(la::min(origin, endpoint), la::max(origin, endpoint));
+  auto recorderf = [&](int /*queryIdx*/, int tri) {
+    const auto [s, v] = k12(0, tri);  // halfedge 0 vs triangle tri
+    if (s != 0 && std::isfinite(v.x)) {
+      // v is the 3D intersection point computed by Kernel12.
+      // Compute parametric t ∈ [0,1] along the ray segment.
+      const double t = (v[tAxis] - origin[tAxis]) / dir[tAxis];
+      if (t >= 0.0 && t <= 1.0) {
+        hits.push_back({static_cast<uint64_t>(tri), t, v, faceNormal_[tri]});
+      }
+    }
+  };
+  auto recorder = MakeSimpleRecorder(recorderf);
+  auto f = [&rayBox](int) { return rayBox; };
+  collider_.Collisions<false>(recorder, f, 1, false);
+
+  std::sort(hits.begin(), hits.end(), [](const RayHit& a, const RayHit& b) {
+    return a.distance < b.distance;
+  });
+  return hits;
+}
+
 }  // namespace manifold
--- a/test/manifold_test.cpp
+++ b/test/manifold_test.cpp
@@ -1273,3 +1273,103 @@ TEST(Manifold, OpenscadCrash) {
   EXPECT_EQ(m2.IsEmpty(), false);
 }
 #endif
+
+TEST(Manifold, RayCastHitCube) {
+  // Ray through center along Z — also tests watertight shared edge at (0,0)
+  Manifold cube = Manifold::Cube(vec3(1), true);
+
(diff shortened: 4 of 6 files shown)


zmerlynn:
Added `RayCastAll` since implementing it required refactoring `RayCast` into a shared `RayCastImpl` helper anyway — the only difference is accumulate-all vs track-best, so it made more sense to include both than to churn the same code twice.

include/manifold/common.h
```diff
@@ -154,6 +154,22 @@ struct Smoothness {
   double smoothness;
 };
 
+/**
+ * @brief Result of a ray cast query against a Manifold.
+ */
+struct RayHit {
+  /// The triangle index that was hit, or -1 for no hit.
+  int faceID = -1;
```
elalish: I think this should be a `uint64_t` so as to ensure it can handle anything `MeshGL64` throws at it.

include/manifold/manifold.h
```diff
@@ -395,6 +395,13 @@ class Manifold {
   double MinGap(const Manifold& other, double searchLength) const;
   ///@}
 
+  /** @name Spatial Queries
+   */
+  ///@{
+  RayHit RayCast(vec3 origin, vec3 endpoint) const;
```
elalish: I think I prefer your original that returned `std::optional<RayHit>` - any reason you changed it?

src/boolean3.cpp
```diff
@@ -528,4 +528,107 @@ Boolean3::Boolean3(const Manifold::Impl& inP, const Manifold::Impl& inQ,
   }
 #endif
 }
+namespace {
```
elalish: Don't we already have an anonymous namespace in this file somewhere? In any case this namespace is generally at the top of the file.

src/boolean3.cpp
```diff
@@ -528,4 +528,107 @@ Boolean3::Boolean3(const Manifold::Impl& inP, const Manifold::Impl& inQ,
...
+template <typename RecordHit>
+void RayCastImpl(const Manifold::Impl& mesh, vec3 origin, vec3 endpoint,
+                 RecordHit& recordHit) {
+  using Impl = Manifold::Impl;
+  // Build a minimal single-edge Impl representing the ray segment.
+  // Kernel12 treats inA as an edge mesh, so we need two vertices, two
+  // halfedges (forward + backward), and one face normal. Zero vertex normals
+  // and face normal mean the ray contributes nothing to perturbation
+  // tiebreakers — consistency at shared edges/vertices depends entirely on
+  // the mesh's own normals.
```
elalish, line 542: Sounds good, but make sure you've checked the tests with ASAN, just in case we didn't notice something being accessed out of bounds.

src/boolean3.cpp
```diff
@@ -528,4 +528,107 @@ Boolean3::Boolean3(const Manifold::Impl& inP, const Manifold::Impl& inQ,
...
+      bestTri = tri;
+    }
+  };
+  RayCastImpl(*this, origin, endpoint, recordHit);
+
+  if (bestTri < 0) return {};
+  RayHit hit;
+  hit.faceID = bestTri;
+  hit.distance = bestT;
+  hit.position = origin + bestT * dir;
```
elalish: This introduces rounding error - it would be better to use the position calculated by Kernel12.

test/manifold_test.cpp
```diff
@@ -1274,3 +1274,188 @@ TEST(Manifold, OpenscadCrash) {
...
+  EXPECT_EQ(hit.faceID, -1);
+}
+
+TEST(Manifold, RayCastNormal) {
+  Manifold cube = Manifold::Cube(vec3(1), true);
+
+  // Ray along -Z should hit bottom face, normal should point -Z
+  RayHit hit = cube.RayCast(vec3(0, 0, -5), vec3(0, 0, 5));
+  EXPECT_GE(hit.faceID, 0);
+  EXPECT_NEAR(hit.normal.z, -1.0, 1e-10);
```
elalish: EXPECT_FLOAT_EQ?

test/manifold_test.cpp
```diff
@@ -1274,3 +1274,188 @@ TEST(Manifold, OpenscadCrash) {
...
+}
+
+TEST(Manifold, RayCastSphere) {
+  Manifold sphere = Manifold::Sphere(1.0, 32);
+
+  // Through center
+  RayHit hit = sphere.RayCast(vec3(0, 0, -5), vec3(0, 0, 5));
+  EXPECT_GE(hit.faceID, 0);
+  EXPECT_NEAR(la::length(hit.position), 1.0, 0.05);  // near the surface
```
elalish: Seems like a pretty wide tolerance - can it get a bit tighter?

test/manifold_test.cpp
```diff
@@ -1274,3 +1274,188 @@ TEST(Manifold, OpenscadCrash) {
...
+}
+
+TEST(Manifold, RayCastWatertightEdge) {
+  // Cast two rays that bracket a shared edge of a cube. Both should hit.
+  // If edge handling were inconsistent, one might miss.
+  Manifold cube = Manifold::Cube(vec3(1), true);
+
+  // Centered unit cube: the diagonal edge of the bottom face passes through
+  // Y=0 (approximately). Rays just above and below should both hit.
+  RayHit hitA = cube.RayCast(vec3(0.25, 1e-15, -5), vec3(0.25, 1e-15, 5));
```
elalish: Why [0.25, 0]? With a centered cube, won't the diagonal go right through [0, 0] instead?

test/manifold_test.cpp
```diff
@@ -1274,3 +1274,188 @@ TEST(Manifold, OpenscadCrash) {
...
+  EXPECT_GE(hitB.faceID, 0);
+
+  // And exactly on the edge
+  RayHit hitEdge = cube.RayCast(vec3(0.25, 0.0, -5), vec3(0.25, 0.0, 5));
+  EXPECT_GE(hitEdge.faceID, 0);
+}
+
+TEST(Manifold, RayCastWatertightVertex) {
+  // Ray exactly through a vertex of the cube. The symbolic perturbation
+  // should assign the hit to exactly one triangle.
```
elalish: If I'm not mistaken, you actually tested these cases in the first few tests, right? Maybe just move these comments / names up there.

test/manifold_test.cpp
```diff
@@ -1274,3 +1274,188 @@ TEST(Manifold, OpenscadCrash) {
...
+
+  auto hits = cube.RayCastAll(vec3(0, 0, -5), vec3(0, 0, 5));
+  ASSERT_EQ(hits.size(), 2);
+  // Sorted by distance
+  EXPECT_LT(hits[0].distance, hits[1].distance);
+  EXPECT_NEAR(hits[0].position.z, -0.5, 1e-10);
+  EXPECT_NEAR(hits[1].position.z, 0.5, 1e-10);
+}
+
+TEST(Manifold, RayCastAllMiss) {
```
elalish: Might be good to add a test of interesting an edge near-exactly at the silhouette, you can test getting two intersections or zero - never one.

elalish (changes requested):
Thanks, looking good!

include/manifold/common.h
zmerlynn: Done, switched to `uint64_t`.

include/manifold/manifold.h
zmerlynn: Good call — switched back to `std::optional<RayHit>`.

src/boolean3.cpp
zmerlynn: Inlined the logic into `RayCastAll` directly, so no separate namespace needed.

src/boolean3.cpp
zmerlynn: Clean under CI-matching ASAN build (`MANIFOLD_DEBUG=ON`, `MANIFOLD_STRICT=ON`, `clang++`).

src/boolean3.cpp
zmerlynn: Good catch — using Kernel12's computed position `v` directly now.

test/manifold_test.cpp
zmerlynn: Updated to `EXPECT_FLOAT_EQ` throughout.

test/manifold_test.cpp
zmerlynn: Bumped to 128-segment sphere with `1e-4` tolerance.

test/manifold_test.cpp
zmerlynn: You're right — moved into `RayCastHitCube` which casts through `(0,0)`.

test/manifold_test.cpp
zmerlynn: Merged the edge case into `RayCastHitCube`. Kept the vertex test separate since it's a distinct degenerate case.

test/manifold_test.cpp
zmerlynn: Added — verifies 0 or 2 hits at the silhouette edge, never 1.

zmerlynn:
`RayCast` now delegates to `RayCastAll().front()` — `Collider::Collisions` visits all AABB-overlapping leaves regardless so there's no early termination to exploit, and this keeps the code in one place.

src/boolean3.cpp
```diff
@@ -528,4 +528,77 @@ Boolean3::Boolean3(const Manifold::Impl& inP, const Manifold::Impl& inQ,
   }
 #endif
 }
+std::optional<RayHit> Manifold::Impl::RayCast(vec3 origin,
+                                              vec3 endpoint) const {
+  // Delegates to RayCastAll since Collider::Collisions visits all
+  // AABB-overlapping leaves regardless — there's no early termination
+  // to exploit, so the only overhead is a small vector and sort.
```
elalish: I think this is a good call, and given that, let's just keep the API simple and remove this one and rename `RayCastAll` to `RayCast` (referring both to these internal methods and the public API).

elalish (changes requested):
Looking great - if you can just fix the conflicts and simplify the API, I think we're about ready to merge.

src/boolean3.cpp
zmerlynn: Done — dropped the optional variant and renamed RayCastAll → RayCast.

zmerlynn:
Simplified the API per your suggestion: dropped the `optional`-returning `RayCast`, renamed `RayCastAll` → `RayCast`. Single function returning `std::vector<RayHit>`, sorted by distance. Since `Collider::Collisions` visits all AABB-overlapping leaves regardless, there's no early-termination advantage to a single-hit variant.

elalish (approved):
Thanks!

PR #1646, opened 2026-04-12
Add Metal GPU compute backend for parallel primitives on macOS

DatanoiseTV:
Adds a Metal compute backend that accelerates parallel primitives (sort, scan, reduce, sequence) on Apple Silicon using the GPU. Leverages unified memory for zero-copy dispatch. Falls back to TBB when Metal is unavailable or data is below threshold.

New files:
- metal_backend.h/mm: Metal device management and kernel dispatch
- metal_kernels.metal: Compute shaders for sort, scan, reduce, etc.

Modified:
- parallel.h: GPU dispatch paths gated on MANIFOLD_METAL
- CMakeLists.txt: MANIFOLD_METAL option, OBJCXX, Metal framework linking

To be used in conjunction with https://github.com/DatanoiseTV/openscad/tree/feature/metal-gpu-compute

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -14,6 +14,9 @@
 
 cmake_minimum_required(VERSION 3.18)
 project(manifold LANGUAGES CXX)
+if(MANIFOLD_METAL AND APPLE)
+  enable_language(OBJCXX)
+endif()
 
 # Use C++17
 set(CMAKE_CXX_EXTENSIONS OFF)
@@ -47,6 +50,7 @@ option(
   ON
 )
 option(MANIFOLD_PAR "Enable Parallel backend" OFF)
+option(MANIFOLD_METAL "Enable Metal GPU compute backend (macOS)" OFF)
 option(
   MANIFOLD_OPTIMIZED
   "Force optimized build, even with debugging enabled"
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -57,6 +57,11 @@ set(
 # Include directories
 set(MANIFOLD_INCLUDE_DIRS ${PROJECT_SOURCE_DIR}/include)
 
+if(MANIFOLD_METAL AND APPLE)
+  list(APPEND MANIFOLD_SRCS metal_backend.mm)
+  list(APPEND MANIFOLD_PRIVATE_HDRS metal_backend.h)
+endif()
+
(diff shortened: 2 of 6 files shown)


zmerlynn:
This is impressive work — the Metal shaders and dispatch infrastructure are clearly well thought out.

I looked into GPU acceleration for manifold for my use case (MeshGL64 workloads on Apple Silicon) and ended up staying with TBB. My concern was context switch overhead — manifold chains many small parallel operations, and I worried that the per-dispatch cost (command buffer creation, encoding, synchronization) would eat into the gains at typical mesh sizes. I didn't get as far as benchmarking it end-to-end though, so I could be wrong.

I notice this PR accelerates int32 scan/reduce/sort operations, which does cover a meaningful chunk of the boolean engine's index manipulation. But the double-precision geometry paths (vertex positions, normals) and uint32 Morton code sorts aren't covered, which limits the overall speedup. For MeshGL64 workloads specifically, Metal is also a tough fit since Apple Silicon GPUs lack native fp64 — those operations would need to stay on CPU regardless. And 1700+ lines of platform-specific Objective-C++ and Metal shading language is a significant maintenance surface — every change to the parallel primitives would need a Metal implementation kept in sync.

I think GPU acceleration for manifold would need a more fundamental approach — keeping data on the GPU across the full operation chain rather than dispatching individual primitives — but that's a much larger redesign. It might be worth starting with an issue to discuss the approach and get benchmark data before investing in a full implementation?

elalish:
Agreed that a bit of discussion first would be nice, especially since I don't believe I've seen you around this repo before. Also, any time you want to suggest a performance improvement, please start by showing your improvements in the form of benchmarking results.

pca006132:
I think an efficient GPU implementation will probably require rewriting a lot of stuff. And I think we should try things like Vulkan or WebGPU that can support more platforms, unless Metal is shown to be performing a lot better.

Anyway, I think we should probably close this for now.

DatanoiseTV:
Closing this in favor of a more fundamental approach. The per-dispatch overhead of individual GPU primitive calls doesn't yield meaningful speedups at typical mesh sizes — the data needs to stay on GPU across the full boolean pipeline. Will revisit with a different architecture. Even though, the performance was 20-30x on some of my models, if not even higher.

PR #1654, opened 2026-04-13
Add ray casting language bindings

### LINKED ISSUE #1640: Add ray casting via existing BVH

zmerlynn:
## Summary
- Add `RayCast` bindings to all in-tree languages (C, Python, WASM/JS) matching the core API from #1645

Builds on #1645. Fixes #1640.

## C API design
Uses the same opaque-vec pattern as `ManifoldManifoldVec` and `ManifoldCrossSectionVec`:
- `ManifoldRayHitVec*` opaque handle wrapping `std::vector<RayHit>`
- `manifold_ray_cast(mem, m, ...)` → returns vec
- `manifold_ray_hit_vec_length(v)` / `manifold_ray_hit_vec_get(v, idx)` → access
- Standard `_size` / `_alloc` / `_destruct` / `_delete` lifecycle

Empty vec = miss (no separate `hit` flag needed).

We considered a sized-buffer approach (`manifold_ray_cast(m, out, max_hits, ...)` returning count) but rejected it — when the returned count equals `max_hits`, there's no way for the caller to know whether that's all the hits or a truncation. The opaque vec avoids this ambiguity.

## Test plan
- [ ] C API: `CBIND.ray_cast` (hit, miss, normals, lifecycle)
- [ ] WASM: `rayCast returns hits through cube` and `rayCast returns empty on miss`
- [ ] Python/WASM: can't be tested locally (no emscripten/nanobind in local build), relies on CI

### DIFF
--- a/bindings/c/conv.cpp
+++ b/bindings/c/conv.cpp
@@ -37,6 +37,10 @@ ManifoldCrossSectionVec* to_c(CrossSectionVec* csv) {
   return reinterpret_cast<ManifoldCrossSectionVec*>(csv);
 }
 
+ManifoldRayHitVec* to_c(RayHitVec* v) {
+  return reinterpret_cast<ManifoldRayHitVec*>(v);
+}
+
 ManifoldSimplePolygon* to_c(manifold::SimplePolygon* m) {
   return reinterpret_cast<ManifoldSimplePolygon*>(m);
 }
@@ -148,6 +152,10 @@ manifold::CrossSection* from_c(ManifoldCrossSection* cs) {
 
 CrossSectionVec* from_c(ManifoldCrossSectionVec* csv) {
   return reinterpret_cast<CrossSectionVec*>(csv);
+}
+
+RayHitVec* from_c(ManifoldRayHitVec* v) {
+  return reinterpret_cast<RayHitVec*>(v);
 }
 
 manifold::SimplePolygon* from_c(ManifoldSimplePolygon* m) {
--- a/bindings/c/conv.h
+++ b/bindings/c/conv.h
@@ -24,6 +24,7 @@
 using namespace manifold;
 using ManifoldVec = std::vector<Manifold>;
 using CrossSectionVec = std::vector<CrossSection>;
+using RayHitVec = std::vector<RayHit>;
 
 ManifoldManifold* to_c(manifold::Manifold* m);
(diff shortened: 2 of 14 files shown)


include/manifold/manifold.h
```diff
@@ -401,6 +401,8 @@ class Manifold {
    */
   ///@{
   std::vector<RayHit> RayCast(vec3 origin, vec3 endpoint) const;
+  std::vector<RayHit> RayCast(vec3 origin, vec3 direction,
```
elalish: It's a lot of duplicated binding code to support this one small bit of ergonomics. I think users can handle doing `endpoint = origin + maxDist * dir` on their own. Let's just skip this API.

elalish (changes requested):
Looks good, but will look better with half the code.

zmerlynn:
Agreed — I actually talked myself out of it and back into it a couple times. Removed in the latest push. 😄

elalish (approved):
Thanks!

PR #1658, opened 2026-04-15
Improve editor load and keep Monaco features stable.

israrkhan921:
Load important Monaco parts first, move heavy parts to background loading, and fix local dev service worker behavior so editor features stay reliable.

### DIFF
--- a/bindings/wasm/examples/editor/editor.js
+++ b/bindings/wasm/examples/editor/editor.js
@@ -78,16 +78,16 @@ if (navigator.serviceWorker) {
   }
 }
 
-let editor = undefined;
-let monaco = undefined;
-let monacoModulesPromise = undefined;
-let monacoSuggestPromise = undefined;
-let monacoNavigationPromise = undefined;
-let monacoContributionsPromise = undefined;
+let editor = null;
+let monaco = null;
+let monacoModulesPromise = null;
+let monacoSuggestPromise = null;
+let monacoNavigationPromise = null;
+let monacoContributionsPromise = null;
 let monacoContributionsReady = false;
-let autoTypings = undefined;
-let autoTypingsPromise = undefined;
-let esbuildWasmPreloadPromise = undefined;
+let autoTypings = null;
+let autoTypingsPromise = null;
+let esbuildWasmPreloadPromise = null;
 let updateTypeIndicator = () => {};
 
 async function loadMonacoModules() {
@@ -126,7 +126,7 @@ function ensureMonacoContributionsLoaded() {
               return module;
             })
             .catch(error => {
-              monacoContributionsPromise = undefined;
+              monacoContributionsPromise = null;
               monacoContributionsReady = false;
               throw error;
             });
@@ -147,7 +147,7 @@ function ensureMonacoSuggestLoaded() {
                   'monaco-editor/esm/vs/editor/contrib/parameterHints/browser/parameterHints.js'),
             ])
             .catch(error => {
-              monacoSuggestPromise = undefined;
+              monacoSuggestPromise = null;
               throw error;
             });
   }
@@ -169,7 +169,7 @@ function ensureMonacoNavigationLoaded() {
                   'monaco-editor/esm/vs/editor/contrib/gotoSymbol/browser/goToCommands.js'),
             ])
             .catch(error => {
-              monacoNavigationPromise = undefined;
+              monacoNavigationPromise = null;
               throw error;
             });
   }
@@ -188,7 +188,7 @@ function ensureEsbuildWasmPreloaded() {
               return response.arrayBuffer();
             })
             .catch(error => {
-              esbuildWasmPreloadPromise = undefined;
+              esbuildWasmPreloadPromise = null;
               throw error;
             });
   }
@@ -586,8 +586,9 @@ function initializeRun() {
 // Editor ------------------------------------------------------------
 
 async function createEditor() {
-  const {monaco: monacoApi, editorWorker, tsWorker} = await loadMonacoModules();
-  monaco = monacoApi;
+  const monacoModules = await loadMonacoModules();
+  monaco = monacoModules.monaco;
+  const {editorWorker, tsWorker} = monacoModules;
   await ensureMonacoSuggestLoaded();
   await ensureMonacoNavigationLoaded();
   self.MonacoEnvironment = {
@@ -679,7 +680,7 @@ async function createEditor() {
     if (!autoTypingsPromise) {
       autoTypingsPromise =
           initializeAutoTypings(showTypeIndicator).catch(error => {
-            autoTypingsPromise = undefined;
+            autoTypingsPromise = null;
             console.error('Failed to initialize auto typings:', error);
           });
     }
--- a/bindings/wasm/examples/editor/editor.js
+++ b/bindings/wasm/examples/editor/editor.js
@@ -96,6 +96,10 @@ async function loadMonacoModules() {
         Promise
             .all([
               import('monaco-editor/esm/vs/editor/editor.api'),
+              // Load TS tokenizer early so first paint has syntax colors
+              // without waiting for hover-driven language-service activation.
+              import(
+                  'monaco-editor/esm/vs/basic-languages/typescript/typescript.contribution'),
               import(
                   'monaco-editor/esm/vs/language/typescript/monaco.contribution'),
               // '?worker' is vite convention to load a module as a web worker.
@@ -103,7 +107,8 @@ async function loadMonacoModules() {
               import(
                   'monaco-editor/esm/vs/language/typescript/ts.worker?worker'),
             ])
-            .then(([monacoModule, _, editorWorkerModule, tsWorkerModule]) => ({
+            .then(([monacoModule, _, __, editorWorkerModule,
+                    tsWorkerModule]) => ({
                     monaco: monacoModule,
                     editorWorker: editorWorkerModule.default,
                     tsWorker: tsWorkerModule.default,
--- a/bindings/wasm/examples/editor/editor.js
+++ b/bindings/wasm/examples/editor/editor.js
@@ -87,6 +87,7 @@ let monacoContributionsPromise = undefined;
 let monacoContributionsReady = false;
 let autoTypings = undefined;
 let autoTypingsPromise = undefined;
+let esbuildWasmPreloadPromise = undefined;
 let updateTypeIndicator = () => {};
 
 async function loadMonacoModules() {
@@ -168,6 +169,25 @@ function ensureMonacoNavigationLoaded() {
             });
   }
   return monacoNavigationPromise;
+}
+
+function ensureEsbuildWasmPreloaded() {
+  if (!esbuildWasmPreloadPromise) {
+    esbuildWasmPreloadPromise =
+        fetch(esbuildWasmUrl, {cache: 'force-cache'})
+            .then(response => {
+              if (!response.ok) {
+                throw new Error(
+                    `Failed to preload esbuild.wasm (${response.status})`);
+              }
+              return response.arrayBuffer();
+            })
+            .catch(error => {
+              esbuildWasmPreloadPromise = undefined;
+              throw error;
+            });
+  }
+  return esbuildWasmPreloadPromise;
 }
 
 // Pane resizing - draggable pane dividers ---------------------
@@ -666,6 +686,9 @@ async function createEditor() {
     if (enhancementsStarted) return;
     enhancementsStarted = true;
     // Start downloads in background; don't block editor interactivity.
+    ensureEsbuildWasmPreloaded().catch(error => {
+      console.warn('Failed to preload esbuild.wasm:', error);
+    });
     ensureMonacoContributionsLoaded().catch(error => {
       console.error('Failed to load Monaco contributions:', error);
     });
--- a/bindings/wasm/examples/editor/editor.js
+++ b/bindings/wasm/examples/editor/editor.js
@@ -15,15 +15,12 @@
 // '?url' is vite convention to reference a static asset.
 // vite will package the asset and provide a proper URL.
 import '@google/model-viewer';
+import 'monaco-editor/esm/vs/editor/browser/widget/codeEditor/editor.css';
+import 'monaco-editor/esm/vs/editor/standalone/browser/standalone-tokens.css';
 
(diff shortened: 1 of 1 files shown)


israrkhan921:
@elalish 
need to explain Sir:

Changed Monaco loading to lazy loading

Before: big Monaco parts were loaded early.
Now: only needed parts are loaded first, so first page load is faster.
Loaded Monaco core modules with dynamic import

Editor API, TypeScript contribution, and workers now load through loadMonacoModules().
This avoids heavy top-level imports.
Suggestions are loaded early (for reliability)

Added ensureMonacoSuggestLoaded().
It loads only:
suggest controller
parameter hints
These are loaded before editor creation, so autocomplete works more consistently.
Navigation features are loaded early

Added ensureMonacoNavigationLoaded().
It loads:
links
hover
go-to-definition commands
This brings blue links / hover / navigation earlier, without loading full editor.all.
Heavy Monaco bundle moved to background

editor.all is still used, but now loaded later with ensureMonacoContributionsLoaded().
It starts in background after editor is interactive.
Auto-typings moved to background

monaco-editor-auto-typings is now imported lazily in initializeAutoTypings().
It is started by ensureAutoTypings() in background, not blocking first paint.
Background enhancement flow added

startEnhancements() runs once and starts:
full Monaco contributions
auto-typings
Triggered by:
setTimeout(..., 0) (as soon as possible, non-blocking)
first editor focus
first typing
Service worker behavior improved for local dev

Added skipServiceWorker logic for:
?no-sw
import.meta.env.DEV
localhost / 127.0.0.1
This avoids local fetch/cache issues that were breaking editor behavior.
Editor styling/color fix

Added Monaco CSS imports for editor and tokens.

Replaced wrong module/moduleResolution setup with:
target: ScriptTarget.ESNext
Keeps TS language behavior correct in Monaco.
Safer initialization checks

Added guard in getModelForScript() if Monaco is not ready yet.
Changed worker ready check to use editor != null before initialize run logic.



In short: this PR makes initial load lighter, keeps suggestions/navigation reliable, fixes local dev service worker issues, and restores proper editor colors.

elalish:
Thanks, before I review this in more detail, would you mind showing some before/after of the chrome dev tools performance metrics and perhaps the page load graphics? Also, can you describe as a user the difference you see, particularly if you throttle your network? How soon do the buttons react? When is the script visible? How long until it runs and shows the model?

israrkhan921:
> Thanks, before I review this in more detail, would you mind showing some before/after of the chrome dev tools performance metrics and perhaps the page load graphics? Also, can you describe as a user the difference you see, particularly if you throttle your network? How soon do the buttons react? When is the script visible? How long until it runs and shows the model?



@elalish 

the videos are on 5.00x speed:

before:   https://drive.google.com/file/d/1PSJj8bFbghHWk_0pfXyWuF0xop7ED-Gc/view?usp=drive_link

after:   https://drive.google.com/file/d/11VwLIDYdlaXV_7mxZzJbYE5oo4eksY-U/view?usp=drive_link



before--



after--

bindings/wasm/examples/editor/editor.js
```diff
@@ -43,7 +44,7 @@ if (navigator.serviceWorker) {
     });
   }
 
-  if (disableServiceWorker) {
+  if (skipServiceWorker) {
```
elalish, line 47: What problem is this solving again? I generally just use the "update on reload" checkbox in the service worker section of chrome dev tools.

bindings/wasm/examples/editor/editor.js
```diff
@@ -467,6 +561,10 @@ function initializeRun() {
 // Editor ------------------------------------------------------------
 
 async function createEditor() {
+  const {monaco: monacoApi, editorWorker, tsWorker} = await loadMonacoModules();
+  monaco = monacoApi;
+  await ensureMonacoSuggestLoaded();
+  await ensureMonacoNavigationLoaded();
```
elalish, line 587: In your after video, it appears that esbuild.wasm is only triggered to download after the Run button is pressed. Can that also be proactively pulled down in the background with everything else?

elalish (commented):
@tonious This is refactoring code you most recently touched - would you be up for giving this a review?

bindings/wasm/examples/editor/editor.js
israrkhan921: > What problem is this solving again? I generally just use the "update on reload" checkbox in the service worker section of chrome dev tools.



This is mainly for local dev stability, not for production behavior.
Sometimes in local dev, an old/active service worker can still intercept requests and cause stale files or fetch errors (I was seeing editor issues unless I used ?no-sw),The skipServiceWorker check makes localhost/dev always clean by default, so we avoid those cache/intercept problems while testing Monaco loading changes.
You are right that “Update on reload” helps in DevTools, but this code-side guard makes behavior consistent even when that checkbox is not enabled, or when someone else runs the branch without DevTools settings.

Sir if you want I can remove this and keep service worker control only through DevTools + ?no-sw.

bindings/wasm/examples/editor/editor.js
israrkhan921: > In your after video, it appears that esbuild.wasm is only triggered to download after the Run button is pressed. Can that also be proactively pulled down in the background with everything else?



yes , i am going to start work on this ,and i will make a video

bindings/wasm/examples/editor/editor.js
elalish: I don't have a strong opinion, this is fine if you find it useful.

israrkhan921:
@elalish 

video is on 2.00x speed:

after update :   https://drive.google.com/file/d/1Xz7yoWi_84SY78WJPn_Rcm8kyNEKEKNJ/view?usp=drive_link

bindings/wasm/examples/editor/editor.js
```diff
@@ -531,70 +652,53 @@ async function createEditor() {
...
-  if (typeIndicator?.textContent) {
-    syncTypeIndicator();
-  }
+  let enhancementsStarted = false;
+  const startEnhancements = () => {
+    if (enhancementsStarted) return;
+    enhancementsStarted = true;
+    // Start downloads in background; don't block editor interactivity.
+    ensureEsbuildWasmPreloaded().catch(error => {
+      console.warn('Failed to preload esbuild.wasm:', error);
```
elalish, line 696: Not for this PR, but esbuild.wasm looks like it's pulling 13MB or so, which is ridiculous. It would be interesting to see if there's some way we could pare that down.

bindings/wasm/examples/editor/editor.js
```diff
@@ -467,6 +581,10 @@ function initializeRun() {
 // Editor ------------------------------------------------------------
 
 async function createEditor() {
+  const {monaco: monacoApi, editorWorker, tsWorker} = await loadMonacoModules();
```
elalish: In your video it looks like we lost syntax highlighting in the editor? Are we missing something that should be loaded?

bindings/wasm/examples/editor/editor.js
```diff
@@ -467,6 +581,10 @@ function initializeRun() {
...
+  monaco = monacoApi;
+  await ensureMonacoSuggestLoaded();
```
elalish, line 592: Not for this PR, but I believe there are multiple ways to load Monaco. I believe we're now loading more of their code than we're actually using, so it would be great to look into minimizing that. Might be good to make a little chart of our download sizes and what they're used for.

elalish (commented):
This is looking pretty ready to me, except for the syntax highlighting issue. Can you do a before/after pass and look for anything else that may have changed in the UX?

bindings/wasm/examples/editor/editor.js
israrkhan921: > In your video it looks like we lost syntax highlighting in the editor? Are we missing something that should be loaded?

its loading later

bindings/wasm/examples/editor/editor.js
israrkhan921: > Not for this PR, but I believe there are multiple ways to load Monaco. I believe we're now loading more of their code than we're actually using, so it would be great to look into minimizing that. Might be good to make a little chart of our download sizes and what they're used for.

yes , we will work on this

israrkhan921:
after :  https://drive.google.com/file/d/1Qj38Z6-4ShbaPjGAYxEbNNexHKwdipY6/view?usp=drive_link

bindings/wasm/examples/editor/editor.js
```diff
@@ -467,6 +586,10 @@ function initializeRun() {
 // Editor ------------------------------------------------------------
 
 async function createEditor() {
+  const {monaco: monacoApi, editorWorker, tsWorker} = await loadMonacoModules();
+  monaco = monacoApi;
```
elalish: I think you're just renaming this and then back again, right?

bindings/wasm/examples/editor/editor.js
```diff
@@ -479,8 +602,12 @@ async function createEditor() {
 
   editor = monaco.editor.create(document.getElementById('editor'), {
     language: 'typescript',
+    theme: 'vs',
     automaticLayout: true,
     minimap: {enabled: false},
+    quickSuggestions: true,
+    suggestOnTriggerCharacters: true,
+    parameterHints: {enabled: true},
```
elalish, line 611: These look interesting - how did you decide which to use?

bindings/wasm/examples/editor/editor.js
```diff
@@ -78,6 +79,121 @@ if (navigator.serviceWorker) {
...
 let editor = undefined;
+let monaco = undefined;
+let monacoModulesPromise = undefined;
+let monacoSuggestPromise = undefined;
+let monacoNavigationPromise = undefined;
+let monacoContributionsPromise = undefined;
+let monacoContributionsReady = false;
+let autoTypings = undefined;
+let autoTypingsPromise = undefined;
+let esbuildWasmPreloadPromise = undefined;
```
elalish: I prefer to use `null` instead of `undefined` pretty much across the board. It tends to be easier to debug and even runs a touch faster, plus it's easier to type.

elalish (approved):
Looks good, just a couple of minor nits.

bindings/wasm/examples/editor/editor.js
israrkhan921: Yes Sir
 I destructured as monacoApi to avoid confusion with the outer monaco variable, then assigned to the module-level monaco,I will simplify naming.

bindings/wasm/examples/editor/editor.js
israrkhan921: These were chosen to restore expected Monaco TS UX while keeping imports minimal:

quickSuggestions:     shows completions while typing (no manual trigger needed).
suggestOnTriggerCharacters:    supports . / import-path trigger behavior.
parameterHints:    function signature help while typing arguments.

I enabled only these because they are low-cost and directly tied to the delayed-intellisense issue.

bindings/wasm/examples/editor/editor.js
israrkhan921: ok , i am going to replace it with null.

bindings/wasm/examples/editor/editor.js
tonious: Yeah... Unfortunately that pretty much means replacing it.

bindings/wasm/examples/editor/editor.js
```diff
@@ -77,7 +78,122 @@ if (navigator.serviceWorker) {
...
+let monacoNavigationPromise = null;
+let monacoContributionsPromise = null;
+let monacoContributionsReady = false;
+let autoTypings = null;
+let autoTypingsPromise = null;
+let esbuildWasmPreloadPromise = null;
+let updateTypeIndicator = () => {};
+
+async function loadMonacoModules() {
+  if (!monacoModulesPromise) {
```
tonious, line 94: These all follow the same kind of pattern, which is basically a memoize...
It's possible to extract that out.  Something like:

```typescript
async function memoizeAsync(fn) {
  let promise = null;
  return async () => {
    if (!promise) promise = fn();
    return promise;
  }
}

async function loadMonacoModules() {
  return Promise.all([
    // .... Stuff goes here.
  ]);
}

const ensureLoadMonacoModules = memoizeAsync(loadMonacoModules);
```

Typed, but not tested.

If this approach makes the PR more readable to you, great!  If not, that's fine too -- this is just another technique that may help someday.

bindings/wasm/examples/editor/editor.js
```diff
@@ -33,6 +30,10 @@ const exampleFunctions = self.examples;
 if (navigator.serviceWorker) {
   const params = new URLSearchParams(window.location.search);
   const disableServiceWorker = params.has('no-sw');
+  const isLocalhost = window.location.hostname === 'localhost' ||
+      window.location.hostname === '127.0.0.1';
+  const skipServiceWorker =
+      disableServiceWorker || import.meta.env.DEV || isLocalhost;
```
tonious, line 36: I understand using the term `skipServiceWorker` as it encompasses more than just `disableServiceWorker`.  But I think I'd prefer to call out the cases individually.

```typescript
const disableServiceWorker = params.has('no-sw');
const isLocalhost = window.location.hostname === 'localhost' ||
      window.location.hostname === '127.0.0.1';
const isDevEnv = import.meta.env.DEV;
```

This makes the check more explicit:

```typescript
// Disable the service worker if asked, in development or
// on localhost (in which case you can work offline anyhow).
if (disableServiceWorker || isDevEnv || isLocalhost) {
```

tonious (approved):
Looks good!  There's a few small comments in here, but they're all optional.

israrkhan921:
@tonious 
Thank you Sir, for the notes
I applied all 3 suggestions:

I changed the service worker check to be more clear (disableServiceWorker, isDevEnv, isLocalhost) and used them directly in the if condition.
I also made one small helper for the repeated async loading pattern (memoizeAsync) and used it for Monaco and esbuild loading.
I kept the same behavior for retry if loading fails.

Thanks again, your suggestion helped make this part cleaner.

israrkhan921:
@elalish @tonious 
Thanks for merging and for your feedback!

israrkhan921:
@elalish 
I am going to start work and make plan, according to the 12-week schedule that i have created for GSoC 2026, starting from where I previously paused, in order to regain a full understanding of the project and create a structured plan for the next steps.

Issue #971, opened 2024-10-01
Progress and cancellation of booleans


pca006132:
### Discussed in https://github.com/elalish/manifold/discussions/957

<div type='discussions-op-text'>

<sup>Originally posted by **wtholliday** September 26, 2024</sup>
I would like to add cancellation of boolean operations to the library. Am I correct that no facility currently exists? Progress reporting would be nice as well, though my primary concern is that I'm throwing big enough meshes at Manifold that iOS tends to kill my app for using too much memory. With a cancellation callback I could (hopefully) react to the OS memory warning in time and cancel the operation.

@elalish @pca006132 can you provide guidance? Is the boolean operation done lazily so an optional callback could be passed to `Manifold::GetMeshGL()`?

thanks!</div>

pca006132:
The plan is to also add a progress report API, as some fraction indicating the number of boolean operations needed and done (in the csg tree). The problem is that we can have multiple threads evaluating different things, and their progress can be different, so I do not plan to use something global for this.

I'm thinking about overloading the `Status` API, which allows users to pass in a progress handle with 2 atomic ints for progress and 1 atomic bool for cancellation. This progress handle can be passed to different threads, e.g,. the GUI thread, for querying, and we will internally pass it around to worker threads, so we can do proper work cancellation.

@t-paul @wtholliday @starseeker @howardt what do you guys think about the API?

starseeker:
So the 2 atomic ints would be total scheduled and processed so far?

pca006132:
> So the 2 atomic ints would be total scheduled and processed so far?

Yes

howardt:
At the moment, Blender's use of Manifold only does one BatchBoolean at a
time, usually with only two arguments. So if I'm understanding the proposal
correctly, there is not much value for us in this added capability.  What
might be of more value would be a progress report for a single boolean
operation.  Maybe in future we will use Manifold more in its intended
multi-operation-tree mode, but that isn't in the near-term future.

wtholliday:
@howardt But wouldn't it still be possible for the Blender user to accidentally do a boolean that takes a long time or uses a lot of memory? In such a case I certainly would want cancellation instead of having to force-quit Blender.

howardt:
[...]

PR #1663, opened 2026-04-18
Add ExecutionContext for boolean progress + cancellation

zmerlynn:
Progress towards #971 — observe progress and request cancellation of long-running boolean evaluations.

## Public API

```cpp
ExecutionContext ctx;
std::thread eval([&]{ result = m.Status(ctx); });  // triggers evaluation
ctx.Progress();   // [0, 1], monotonic within an evaluation
ctx.Cancel();     // result's Status becomes Error::Cancelled
```

`Status(ctx)` is an overload of the existing `Status()`; no other public API is affected. Copies of `ctx` share state, so a UI thread can hold a copy and call `Cancel()` from a button handler while a worker runs the evaluation.

## Design notes

- **Pimpl with `shared_ptr<Impl>`.** Opaque `Impl` (in `src/execution_impl.h`) keeps the ABI stable and makes copies share state naturally.
- **Sticky cancel.** Once cancelled, a context stays cancelled — same shape as Go's `context.WithCancel`. Typical use is a fresh context per op.
- **Per-boolean granularity as starting state.** Cancel is checked between booleans; a single large boolean runs to completion. Finer granularity (into the boolean kernels themselves) is a follow-up.
- **Progress via leaf pre-pass.** A CSG tree with N leaves reduces in N-1 ops, so `totalBooleans = leaves - 1`; each boolean increments `doneBooleans`. Exact denominator, no estimation.
- **Counters are boolean-scoped.** `totalBooleans`/`doneBooleans` count boolean ops specifically. Future progress dimensions (phases, per-op items) will be added as separate fields rather than reinterpreting these.
- **Relaxed atomics.** The three atomics carry no happens-before for other data; progress readers tolerate staleness; missing cancel by one boolean is within the granularity contract.

## Bindings

`Cancelled` added to the error enum in C/Python/WASM. Full progress/cancel API is a follow-up once the C++ API settles.

## Testing

13 new tests covering progress, cancellation, reuse, copy-share-state, concurrent observation, and the N-1 invariant.

### DIFF
--- a/src/csg_tree.cpp
+++ b/src/csg_tree.cpp
@@ -21,6 +21,7 @@
 
 #include "boolean3.h"
 #include "csg_tree.h"
+#include "execution_impl.h"
 #include "impl.h"
 #include "mesh_fixes.h"
 #include "parallel.h"
@@ -101,7 +102,8 @@ std::shared_ptr<const Manifold::Impl> CsgLeafNode::GetImpl() const {
   return pImpl_;
 }
 
-std::shared_ptr<CsgLeafNode> CsgLeafNode::ToLeafNode() const {
+std::shared_ptr<CsgLeafNode> CsgLeafNode::ToLeafNode(
+    ExecutionContext::Impl*) const {
   return std::make_shared<CsgLeafNode>(*this);
 }
 
@@ -144,8 +146,19 @@ std::shared_ptr<CsgLeafNode> ImplToLeaf(Manifold::Impl&& impl) {
       std::make_shared<Manifold::Impl>(std::move(impl)));
 }
 
+// Build a leaf with the given error status — used to short-circuit boolean
+// evaluation on cancellation.
+std::shared_ptr<CsgLeafNode> ErrorLeaf(Manifold::Error err) {
+  Manifold::Impl impl;
+  impl.status_ = err;
+  return ImplToLeaf(std::move(impl));
+}
+
 std::shared_ptr<CsgLeafNode> SimpleBoolean(const Manifold::Impl& a,
-                                           const Manifold::Impl& b, OpType op) {
+                                           const Manifold::Impl& b, OpType op,
+                                           ExecutionContext::Impl* ctx) {
+  if (ctx && ctx->cancel.load(std::memory_order_relaxed))
+    return ErrorLeaf(Manifold::Error::Cancelled);
 #ifdef MANIFOLD_DEBUG
   auto dump = [&]() {
     dump_lock.lock();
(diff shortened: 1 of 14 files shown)


src/csg_tree.cpp
```diff
@@ -683,31 +712,37 @@ std::shared_ptr<CsgLeafNode> CsgOpNode::ToLeafNode() const {
   //     destination->push_back(node->cache_->Transform(transform));
   // }
   while (!stack.empty()) {
+    if (ctx && ctx->cancel.load(std::memory_order_relaxed)) {
```
pca006132, line 715: I think we need to be careful here and not directly discard the stack, since users may try the operation a second time and it will give an incorrect result.

There are two possible ways: Pollute the result so users get an error when they try to evaluate the Boolean operation, or collect the stack so users can resume the evaluation. I don't know what is the best way to implement the second option, so I think maybe we can stick with option 1 first, and switch to option 2 later?

src/csg_tree.cpp
```diff
@@ -778,4 +813,14 @@ CsgNodeType CsgOpNode::GetNodeType() const {
   return CsgNodeType::Leaf;
 }
 
+size_t CsgOpNode::NumLeaves() const {
+  // An already-evaluated CsgOpNode counts as a single leaf for the purposes
+  // of estimating remaining boolean work.
+  if (cache_ != nullptr) return 1;
+  auto impl = impl_.GetGuard();
+  size_t total = 0;
+  for (const auto& child : *impl) total += child->NumLeaves();
```
pca006132: We don't want recursion here, since we have cases where the tree is very deep (when users do `+=` in a loop).

pca006132 (commented):
Looks pretty good to me. The cancellation granularity is not fine enough (IMO), but we can change that later. In the documentation, we may directly include the example code for `ExecutionContextCancelConcurrent` so users know how to do cancellation.

elalish (approved):
LGTM, thanks! Once @pca006132's concerns are addressed, I think we can merge this.

src/csg_tree.cpp
zmerlynn: Going with your option 1. Commit 183ee866 now poisons every op_node on the stack (not just `this`) with a shared `ErrorLeaf(Cancelled)` — any subsequent evaluation of a shared sub-op that was in flight also short-circuits. Safe to share one cancelled leaf across all frames since the empty-cancelled `Impl` is transform-invariant for error-status queries.

src/csg_tree.cpp
zmerlynn: Good catch — rewritten iteratively in 183ee866 using the same explicit-stack pattern as `ToLeafNode`. Added `Manifold.DeepChainDoesNotOverflowNumLeaves` (300k-deep `+=` chain) as a regression test — verified locally that the recursive version SIGSEGVs at that depth on a stock 8MB stack, and the iterative version runs in ~1s.

zmerlynn:
Added a concrete observer-thread example to the `ExecutionContext` docstring in `common.h` in 183ee866. Agreed on the granularity — left per-boolean for now, framed as a starting state in the PR body.

pca006132 (approved):
Thanks

Issue #1662, opened 2026-04-18
Add TSAN to CI (maybe path-filtered)


zmerlynn:
Running TSAN locally against the test suite would catch data races in our parallel code — currently the fuzzer job runs with ASAN (via `MANIFOLD_FUZZ=ON`) but no TSAN coverage. Status quo is ad-hoc: whoever's touching sensitive code runs TSAN locally if they remember. Works, but depends on discipline.

I was poking at thread safety in #971 and realized it'd be worth automating, at least for PRs that touch relevant files.

## Complication: TBB + TSAN is noisy

Stock libtbb produces thousands of TSAN warnings due to `std::atomic_thread_fence` usage (google/sanitizers#1352 — TSAN cannot model standalone fences). This is a sanitizer limitation, not a TBB bug. oneTBB#358 has requested a suppression file since 2021, never shipped.

## Options for handling TBB

1. **Rebuild oneTBB with `TBB_SANITIZE=thread`** in the TSAN job. Most correct; doubles CI build time but gives real coverage.
2. **Run TSAN job with `MANIFOLD_PAR=OFF`**. Cheap, but loses most of what we'd want to check — the parallel code paths just don't run.
3. **Suppression file** (`race:tbb::` / `called_from_lib:libtbb`). Common in practice, but blunt — would also suppress real races in our own lambdas whose stack traverses TBB.

Leaning between (1) and (3). (1) is the clean answer; (3) is the cheap answer with a footgun.

## CI budget

Matrix is already ~34 jobs per PR. To reduce impact, we could **path-filter** the TSAN job so it only runs when PRs touch thread-safety-relevant paths (`src/csg_tree.cpp`, `src/boolean3.cpp`, `src/boolean_result.cpp`, `src/face_op.cpp`, `src/parallel.h`, `src/utils.h`, plus relevant public headers). GitHub Actions supports this natively via `on.pull_request.paths`.

Or nightly instead of per-PR.

## Context

Local TSAN run on the current test suite (with TBB, stock libtbb) shows no races in our own code — all warnings trace into TBB internals. So we're starting from a clean baseline.

elalish:
Hmm, maybe. My approach to automated testing is that it's safer to have no guardrail than a rotten guardrail (the kind you lean on only to fall). So I like a CI that tests everything it says it does with very low noise, or else clearly rely on manual testing. As soon as we start adding caveats, I fear it only encourages us to not bother checking these things manually anymore, and thus increase rather than decrease our blind spots.

[...]

PR #1664, opened 2026-04-19
Fix concurrent-use race on `static` affinity_partitioner in `AddNewEdgeVerts`

zmerlynn:
Follow-up on #1662, poking at the one TSAN warning that remained after rebuilding TBB with `TBB_SANITIZE=thread`. Curious what the `no_sanitize("thread")` on `AddNewEdgeVerts` was masking. Claude came up with a hypothesis and I verified it.

## The bug

`static tbb::affinity_partitioner ap` is shared across every call. `BatchBoolean` dispatches `SimpleBoolean` in parallel via `task_group`, so multiple workers hit the same static concurrently. oneTBB's [`affinity_partitioner_base`](https://github.com/uxlfoundation/oneTBB/blob/master/include/oneapi/tbb/partitioner.h) mutates its internal `my_array`/`my_size` during `parallel_for` without any synchronization — so concurrent use is a real race, and TSAN correctly flagged it when I looked.

## The fix

`static` → `thread_local`. Each worker thread gets its own partitioner; per-thread affinity is preserved across successive calls on the same thread. Also removes the `no_sanitize` attribute that was hiding the signal.

## Results

- TSAN on Boolean + Minkowski: 3 warnings → **0**.
- Regular TBB build: 362/362 pass.
- Test-suite timing A/B on `Boolean.*`: within run-to-run noise.

### DIFF
--- a/src/boolean_result.cpp
+++ b/src/boolean_result.cpp
@@ -199,19 +199,13 @@ struct EdgePos {
   }
 };
 
-// thread sanitizer doesn't really know how to check when there are too many
-// mutex
-#if defined(__has_feature)
-#if __has_feature(thread_sanitizer)
-__attribute__((no_sanitize("thread")))
-#endif
-#endif
 void AddNewEdgeVerts(
     // we need concurrent_map because we will be adding things concurrently
-    concurrent_map<int, std::vector<EdgePos>> &edgesP,
-    concurrent_map<std::pair<int, int>, std::vector<EdgePos>> &edgesNew,
-    const Vec<std::array<int, 2>> &p1q2, const Vec<int> &i12, const Vec<int> &v12R,
-    const VecView<Halfedge> &halfedgeP, bool forward, size_t offset) {
+    concurrent_map<int, std::vector<EdgePos>>& edgesP,
+    concurrent_map<std::pair<int, int>, std::vector<EdgePos>>& edgesNew,
+    const Vec<std::array<int, 2>>& p1q2, const Vec<int>& i12,
+    const Vec<int>& v12R, const VecView<Halfedge>& halfedgeP, bool forward,
+    size_t offset) {
   ZoneScoped;
   // For each edge of P that intersects a face of Q (p1q2), add this vertex to
   // P's corresponding edge vector and to the two new edges, which are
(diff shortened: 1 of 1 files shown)


zmerlynn:
Actually, thinking about this more, it might be better if we settle on #1662 and use the existing warnings as a signal that CI is working, if we decide to go that way.

zmerlynn:
Moving to draft while I play with CI. I'm going to briefly rebase this on #1665 and A/B test to verify that this clears the warnings.

zmerlynn:
Quick update on validation — I temporarily rebased this on top of #1665 (the TSAN CI workflow) to prove the fix end-to-end:

- On #1665 alone (this PR's fix not yet applied), the TSAN check [fails correctly](https://github.com/elalish/manifold/actions/runs/24635568206/job/72030232797) with the affinity_partitioner warnings that the `no_sanitize` on `AddNewEdgeVerts` was masking.
- On #1664 stacked on #1665, the TSAN check [passes](https://github.com/elalish/manifold/actions/runs/24635347319/job/72029647340): 362 tests, 0 warnings.

I'm about to rebase off #1665 so this PR stays minimal — just wanted to have the demo runs in the record.

src/boolean_result.cpp
```diff
@@ -199,19 +199,13 @@ struct EdgePos {
...
-__attribute__((no_sanitize("thread")))
-#endif
-#endif
 void AddNewEdgeVerts(
     // we need concurrent_map because we will be adding things concurrently
-    concurrent_map<int, std::vector<EdgePos>> &edgesP,
-    concurrent_map<std::pair<int, int>, std::vector<EdgePos>> &edgesNew,
-    const Vec<std::array<int, 2>> &p1q2, const Vec<int> &i12, const Vec<int> &v12R,
-    const VecView<Halfedge> &halfedgeP, bool forward, size_t offset) {
+    concurrent_map<int, std::vector<EdgePos>>& edgesP,
```
zmerlynn, line 204: BTW - this chunk is just a clang-format change that appeared after I dropped the chunk above.

pca006132 (approved):

pca006132:
Nice catch! Should have read TBB's documentation better.

PR #1665, opened 2026-04-19
Add TSAN CI workflow

### LINKED ISSUE #1662: Add TSAN to CI (maybe path-filtered)

zmerlynn:
Adds a ThreadSanitizer workflow per #1662. Builds oneTBB with `TBB_SANITIZE=thread` (cached), runs the test suite under `-fsanitize=thread`, fails the job if any warnings fire.

## Validation

- #1664's `affinity_partitioner` race fix is now on master, and this PR's TSAN run (once CI catches up) should be the first demonstration of a clean signal on upstream master.
- Earlier validation on this PR: on master prior to #1664, the TSAN check [failed correctly](https://github.com/elalish/manifold/actions/runs/24635568206/job/72030232797) with the affinity_partitioner warnings.

## Timing

~22–24 min cold, ~15–18 min warm (estimated). A few minutes longer than `code coverage` when warm; not an outlier.

Fork PRs read master's cache once this lands, so external contributors get warm-cache runs without setup.

Fixes #1662.

### DIFF
--- a/.github/workflows/tsan.yml
+++ b/.github/workflows/tsan.yml
@@ -47,10 +47,15 @@ jobs:
         if: steps.cache-tbb.outputs.cache-hit != 'true'
         run: |
           git clone --depth 1 --branch "$TBB_TAG" https://github.com/uxlfoundation/oneTBB.git tbb-src
+          # -Wno-tsan: GCC 14's -Wtsan fires on TBB's own atomic_thread_fence
+          # usage (which TBB_SANITIZE=thread mode replaces internally anyway).
+          # oneTBB master adds this suppression in its own sanitize.cmake but
+          # it didn't land until after v2022.3.0.
           cmake -S tbb-src -B tbb-build \
             -DTBB_TEST=OFF \
             -DTBB_SANITIZE=thread \
             -DCMAKE_BUILD_TYPE=RelWithDebInfo \
+            -DCMAKE_CXX_FLAGS=-Wno-tsan \
             -DCMAKE_INSTALL_PREFIX="$TBB_PREFIX"
           cmake --build tbb-build -j$(nproc)
           cmake --install tbb-build
--- a/.github/workflows/tsan.yml
+++ b/.github/workflows/tsan.yml
@@ -3,9 +3,9 @@ name: TSAN
 # Run ThreadSanitizer against the test suite using a TBB built with
 # TBB_SANITIZE=thread (so TBB's internal fences are swapped for acquire/release
 # ops TSAN can model — otherwise stock libtbb produces thousands of spurious
-# warnings). The run takes ~45 min, so once this workflow is validated we
-# intend to move it off per-PR and onto a nightly/on-demand trigger. For now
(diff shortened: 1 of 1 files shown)


.github/workflows/tsan.yml
```diff
@@ -0,0 +1,102 @@
...
+
+      - name: Run tests under TSAN
+        env:
+          LD_LIBRARY_PATH: ${{ env.TBB_PREFIX }}/lib
+          TSAN_OPTIONS: halt_on_error=0 report_thread_leaks=0
+        run: |
+          set -o pipefail
+          ./build/test/manifold_test 2>&1 | tee tsan.log
+          # Fail the job if TSAN reported any warnings. The test process itself
+          # returns 0 because halt_on_error=0; the log is our signal.
```
pca006132: Maybe just do `halt_on_error=1`?

zmerlynn: Done in 4e593efb. Good simplification — for CI we just need "is it clean?"; developer can rerun locally with `halt_on_error=0` for full diagnostics.

elalish (approved):
Thanks - can you pull master to get the other recent CI update?

PR #1667, opened 2026-04-20
Fix leak in `manifold_alloc_*` (C API)

zmerlynn:
`manifold_alloc_X()` was `new T`, which allocates **and** constructs a default `T`. Every subsequent constructor function placement-new's into that memory without calling the old `T`'s destructor, so the default object's heap allocations leak. For `Manifold` it's a `shared_ptr<CsgLeafNode>`; for `CrossSection` it's a `shared_ptr<PathImpl>` (plus a Clipper2 `PathsD`).

Reproducer:

```cpp
for (int i = 0; i < 100; ++i) {
  ManifoldManifold* m = manifold_alloc_manifold();
  manifold_cube(m, 1.0, 1.0, 1.0, 0);
  manifold_delete_manifold(m);
}
// LSan: 100 leaks, scales linearly
```

Surfaced by a downstream Rust binding's LSan CI — [failed run](https://github.com/zmerlynn/manifold-csg/actions/runs/24645851906) (`403452 byte(s) leaked in 2930 allocation(s)`).

Fix: `manifold_alloc_*` returns raw storage via a small `alloc_raw<T>` helper using [`::operator new`](https://en.cppreference.com/w/cpp/memory/new/operator_new), matching what `malloc(manifold_X_size())` already does in the in-tree C test. Callers placement-new into the storage, then `manifold_delete_*` handles dtor + dealloc — the plain `::operator new` pairs with the plain `::operator delete` that [`delete T*`](https://en.cppreference.com/w/cpp/language/delete) picks for default-aligned types. A `static_assert` guards against over-aligned types slipping in later.

0 leaks after the fix. 356 tests pass; ASan run on the full suite + the repro pattern (2000 cycles) clean.

## History

#1018 switched these alloc functions from `malloc` to `new T` to fix a real `malloc`/`delete` pairing UB, and explicitly flagged the leak it traded for: "this causes memory leak for the default-constructed object… should be keep in mind". I copied the pattern forward in #1654 when adding `manifold_alloc_ray_hit_vec`. This change keeps the `new`/`delete` pairing correct and removes the leak by not constructing in alloc.

## Behavior change

`alloc_X()` + `delete_X()` with no intervening constructor call is now UB (calls `~T()` on uninitialized memory). Not a documented pattern; no in-tree test exercises it.

### DIFF
--- a/bindings/c/manifoldc.cpp
+++ b/bindings/c/manifoldc.cpp
@@ -41,6 +41,18 @@ ManifoldManifold* level_set(
   return to_c(new (mem) Manifold(Manifold::LevelSet(
       fun, *from_c(bounds), edge_length, level, tolerance, !seq)));
 }
+
+// Raw uninitialized storage for a T — callers must placement-new into it
+// before any use or before passing to manifold_destruct_* / manifold_delete_*.
+// Pairs with `delete T*`, which uses the plain `::operator delete` for
+// types with default alignment.
+template <typename T>
+T* alloc_raw() {
+  static_assert(alignof(T) <= __STDCPP_DEFAULT_NEW_ALIGNMENT__,
+                "over-aligned types need the aligned operator new/delete "
+                "pair; extend this helper if a binding type ever needs it");
+  return static_cast<T*>(::operator new(sizeof(T)));
+}
 }  // namespace
 
 #ifdef __cplusplus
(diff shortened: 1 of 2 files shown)


pca006132 (approved):
Good catch, I forgot about this issue.

PR #1668, opened 2026-04-21
Add ExecutionContext language bindings

zmerlynn:
Follow-up to #1663, exposing `ExecutionContext` through the C, Python, and WASM bindings. Progress towards #971.

- **C**: opaque `ManifoldExecutionContext` with the full alloc/destruct/delete plumbing (matching #1667); `manifold_status_with_context`.
- **Python**: `ExecutionContext` class; `status(ctx)` overload via nanobind.
- **WASM**: embind `ExecutionContext` class; `Manifold.statusWithContext(ctx)`. Also plugs a pre-existing gap — the `Cancelled` case wasn't in the `status()` string map.

## Tests

- C: happy path (`malloc`+`destruct`+`free`) and cancel (`alloc`+`delete`) — both ownership patterns downstream uses.
- WASM: happy path + cancel with explicit `.delete()`.
- Python: added to `all_apis.py`.


bindings/c/include/manifold/manifoldc.h
```diff
@@ -249,6 +253,15 @@ ManifoldRayHitVec* manifold_ray_cast(void* mem, ManifoldManifold* m,
 size_t manifold_ray_hit_vec_length(ManifoldRayHitVec* v);
 ManifoldRayHit manifold_ray_hit_vec_get(ManifoldRayHitVec* v, size_t idx);
 
+// ExecutionContext: observe progress and request cancellation of a
+// long-running Manifold evaluation. Pass to manifold_status_with_context.
+// Safe to read/write from any thread. See the ExecutionContext class in
+// common.h for full semantics (sticky cancel, per-boolean granularity).
+ManifoldExecutionContext* manifold_execution_context(void* mem);
```
pca006132, line 260: should we have the word `alloc` in the name? iirc we have that for the other APIs.

pca006132 (approved):
Just a minor naming convention comment.

bindings/wasm/manifold-global-types.d.ts
```diff
@@ -121,3 +121,22 @@ export type ErrorStatus = 'NoError'|'NonFiniteVertex'|'NotManifold'|
...
+ */
+export interface ExecutionContext {
+  /** Request cancellation. Can be called from any context. Idempotent. */
+  cancel(): void;
+  /** Has cancellation been requested? */
+  cancelled(): boolean;
+  /** Normalized progress in [0, 1]. Monotonic within an evaluation. */
+  progress(): number;
+  /** Release the underlying memory. */
+  delete(): void;
```
elalish, line 148: What is this function? Did Claude hallucinate this, or am I missing something?

bindings/c/include/manifold/manifoldc.h
zmerlynn: The placement-new form follows the existing pattern — `manifold_simple_polygon(void* mem, ...)`, `manifold_meshgl(void* mem, ...)`, `manifold_rect(void* mem, ...)`, `manifold_box(void* mem, ...)`. The `alloc` version is the combined alloc+placement-new (`manifold_alloc_execution_context()` at line 511), which does have `alloc` in the name.

bindings/wasm/manifold-global-types.d.ts
zmerlynn: Not hallucinated — it's the embind auto-generated cleanup method, same one declared on `Manifold.delete()` (manifold-encapsulated-types.d.ts:1242) and `CrossSection.delete()` (line 437). I've updated the doctoring in 992bca54 to match the existing "Frees the WASM memory of this X" wording.

(I spent a little while researching this, and this is part of the JS<->WASM contract. So the rule is essentially: if it survives a function call, it needs .delete(); if it's a one-shot return value, it doesn't.)

elalish: Great, thanks for the explanation. 👍

elalish (approved):
Thanks!

PR #1669, opened 2026-04-21
Thread ExecutionContext into Boolean3 for mid-boolean cancel

zmerlynn:
Follow-up to #1663. Progress towards #971.

After #1663, a single boolean between large meshes couldn't be cancelled — `ctx->cancel` was only checked at the CSG tree loop heads (`SimpleBoolean`/`BatchBoolean`/`BatchUnion`), not inside `Boolean3` itself. This threads `ExecutionContext::Impl*` into `Boolean3` and checks at 8 phase boundaries (4 in the ctor — before each `Intersect12` and `Winding03`; 4 in `Result` — assembly entry, `Face2Tri`, `CreateProperties`, `CalculateBBox`), bounding mid-boolean cancel latency to at most one phase.

Cost: 8 relaxed atomic loads per boolean, negligible relative to the work each phase does. Within-phase cancellation (checks inside hot parallel lambdas) is a separate follow-up.

Default ctx is `nullptr` so non-CSG callers (`Manifold::Split`, etc.) don't need to thread it. Those remain uncancellable and are tracked in #971.

## Test

`ExecutionContextCancelMidBoolean` spawns eval on a worker thread and cancels after 1ms. Probabilistic — `Cancelled` may come from an inner `Boolean3` check or from the outer `CsgOpNode::ToLeafNode` check after a fast `SimpleBoolean`, and `NoError` is valid if eval races past the sleep. `MANIFOLD_PAR == 1` guarded like `CancelConcurrent`.

### DIFF
--- a/src/boolean3.cpp
+++ b/src/boolean3.cpp
@@ -479,8 +479,8 @@ Vec<int> Winding03(const Manifold::Impl& inP, const Manifold::Impl& inQ,
 
 namespace manifold {
 Boolean3::Boolean3(const Manifold::Impl& inP, const Manifold::Impl& inQ,
-                   OpType op)
-    : inP_(inP), inQ_(inQ), expandP_(op == OpType::Add) {
+                   OpType op, ExecutionContext::Impl* ctx)
+    : inP_(inP), inQ_(inQ), expandP_(op == OpType::Add), ctx_(ctx) {
   ZoneScoped;
   // Symbolic perturbation:
   // Union -> expand inP, expand inQ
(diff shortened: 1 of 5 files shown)


elalish:
Would you mind providing some example timings of these phases so we can get an idea of how granular this cancellation is? Maybe for our sphere benchmark and for something triangulator-heavy like the Menger sponge?

zmerlynn:
Posted phase timings over on #971 (https://github.com/elalish/manifold/issues/971#issuecomment-4292887065) to keep the discussion searchable. TL;DR: max single phase is 100 ms–5 s on realistic subtracts, dominated by Triangulate on triangulator-heavy inputs — phase-boundary cancel handles most cases, but an ~89 s Triangulate phase in MengerSponge(4) construction shows where within-phase cancel would matter.

elalish (approved):
LGTM, though looks like you need to pull master. @pca006132 I'll let you merge this one.

pca006132 (approved):

PR #1671, opened 2026-04-22
More smoothing fixes

elalish:
Okay, my fillet example is now working better, FYI @tonious. This fixed several numerical precision problems, did some general cleanup, and updated some tests.

Things that still need work: 
- Smoothing only selects quads for planar quads; it would be much better to select quad smoothing more aggressively.
- Converting `Polygons` to `CrossSection` and back loses quite a bit of precision, not sure why.
- Smoothing can cause planar regions to overlap themselves when the triangles are high aspect ratio and the tangents apply a bunch of curvature. Not sure how solvable this is, it's the usual self-intersecting bezier curve problem. Still, might be nice to improve planar triangulation if possible.


elalish:
@zmerlynn Any idea why the TSAN check is failing here? I feel like I didn't change `IsMarkedQuad` in this PR. Do you think it was an existing problem that TSAN missed until something else changed?

Issue #1378, opened 2025-10-09
Python binding manifold3d.Manifold.to_mesh() causes segmentation fault


elektrohenk:
manifold3d.Manifold.to_mesh() sometimes causes (ipython) "The kernel died" errors in Spyder and no further information. I can not reproduce the error in a controllable way. Some observations about the crash behavior: When the problem occurs it is often but not always reproducible in cmd using both python and ipython, the ipython console in Spyder seems to crash quicker. longer runtime and more complex meshes increase the change of a crash. The Manifold instances typically consist out of many union and difference operations after each-other. Reshuffling the order in which to_mesh() is called on different Manifold instances (without modifying their content) may change the instance which causes the crash. The created Manifold instances have boolean operations with certain Manifold instances in common, recreating instead of reusing some of those common Manifold instances sometimes seems to solve the problem.

Environment:
Windows 11
Python 3.13
numpy 2.3.3
manifold3d 2.3.1

GDB back trace:
Thread 1 received signal SIGSEGV, Segmentation fault.
0x00007ff8d35f8e48 in manifold3d.cp313-win_amd64!?what@python_error@nanobind@@UEBAPEBDXZ () from C:\Users\User\AppData\Local\miniforge3\envs\py313\Lib\site-packages\manifold3d.cp313-win_amd64.pyd
(gdb) bt
#0  0x00007ff8d35f8e48 in manifold3d.cp313-win_amd64!?what@python_error@nanobind@@UEBAPEBDXZ () from C:\Users\User\AppData\Local\miniforge3\envs\py313\Lib\site-packages\manifold3d.cp313-win_amd64.pyd
#1  0x00007ff8d35fb59b in manifold3d.cp313-win_amd64!?what@python_error@nanobind@@UEBAPEBDXZ () from C:\Users\User\AppData\Local\miniforge3\envs\py313\Lib\site-packages\manifold3d.cp313-win_amd64.pyd
#2  0x00007ff8d355508c in manifold3d.cp313-win_amd64!?what@python_error@nanobind@@UEBAPEBDXZ () from C:\Users\User\AppData\Local\miniforge3\envs\py313\Lib\site-packages\manifold3d.cp313-win_amd64.pyd
[...]

PR #1673, opened 2026-04-22
Fix CsgOpNode destructor corrupting shared subtrees

### LINKED ISSUE #1672: Crash in CsgOpNode::ToLeafNode()
Hi!

After a series of 7 boolean operations I try to get the result. I call the result.NumTri() and crash happens.
If instead I call NumTri() after each of the 7 operations, then it succeeds.

I dived into the code and found why it happens.
In **csg_tree.cpp** in CsgOpNode::ToLeafNode() there is a code:
```
        switch (frame->op_node->op_) {
          case OpType::Add:
            *impl = {BatchUnion(frame->positive_children)};
            break;
          case OpType::Intersect: {
            *impl = {BatchBoolean(OpType::Intersect, frame->positive_children)};
            break;
          };
```
It is expected that frame->positive_children is not empty. But it appears to be empty in my case. So, later, that BatchUnion() has to return something and it returns null after this: return children.front(); which leads to crash.

The _OpType::Subtract_ case already has check for emptiness. But _Add_ and _Intersect_ cases don't have it. I added these checks and now everything works good, no crashes. So, I suggest to change the above code to the code below:
```
       switch (frame->op_node->op_) {
          case OpType::Add:
            if (frame->positive_children.empty())
              *impl = {std::make_shared<CsgLeafNode>()};
            else
              *impl = {BatchUnion(frame->positive_children)};            
            break;
          case OpType::Intersect: {
            if (frame->positive_children.empty())
              *impl = {std::make_shared<CsgLeafNode>()};
            else
              *impl = {BatchBoolean(OpType::Intersect, frame->positive_children)};
            break;
          };
 

```

zmerlynn:
Fixes #1672 (confirmed by reporter). Related to #1378 (which fixed a different manifestation of the same "shared tree state gets corrupted" class of bug).

The destructor at `src/csg_tree.cpp:560` guarded the destructive `handleChildren(*childImpl)` with `impl_.UseCount() == 1`. That check referred to the outer (being-destroyed) node's impl — which is tautologically true inside this block, because the enclosing `if (impl_.UseCount() == 1)` already established it and nothing in the loop body changes `impl_`'s refcount. So the guard was effectively unconditional, and the intended check — whether the *child* is still held by other live trees — was never performed. Result: when a tree containing a shared, unevaluated `CsgOpNode` was destroyed, it emptied that shared node's impl, breaking any other tree that still referenced it.

Minimal reproducer, segfaults on master:

```cpp
Manifold shared = a + b;
Manifold tree2 = shared - c;
{
  Manifold tree1 = shared + c;
}  // tree1 destroyed -> shared's impl emptied
tree2.NumTri();  // crash: Add finalize with empty positive_children
```

Fix: check both `child.use_count() == 1` (no other tree holds this `CsgOpNode`) and `child->impl_.UseCount() == 1` (no other `CsgOpNode` shares this impl via `Transform()`).

## Test

New regression `Boolean.SharedSubtreeAfterSiblingDestroyed` in `test/boolean_test.cpp`. Builds `tree2 = shared - c` alongside and after a destroyed `tree1 = shared + c`, then compares against an independently-constructed `(a+b)-c` reference to catch partial corruptions as well as crashes. Full suite (384 tests) passes.

### DIFF
--- a/src/csg_tree.cpp
+++ b/src/csg_tree.cpp
@@ -560,7 +560,10 @@ CsgOpNode::~CsgOpNode() {
     while (!toProcess.empty()) {
       auto child = std::move(toProcess.back());
       toProcess.pop_back();
-      if (impl_.UseCount() == 1) {
+      // Only empty the child's impl if we are its last holder. Otherwise
+      // another live tree still references this CsgOpNode and may later
+      // try to evaluate it, which requires its impl intact.
+      if (child.use_count() == 1 && child->impl_.UseCount() == 1) {
         auto childImpl = child->impl_.GetGuard();
         handleChildren(*childImpl);
       }
--- a/test/boolean_test.cpp
+++ b/test/boolean_test.cpp
@@ -825,3 +825,28 @@ TEST(Boolean, BatchBoolean) {
   EXPECT_FLOAT_EQ(subtract.Volume(), 7226.043);
   EXPECT_FLOAT_EQ(subtract.SurfaceArea(), 14904.597);
 }
+
+// Regression for #1672: a sub-tree shared between two trees must survive when
+// one of those trees is destroyed without being evaluated. Previously the
+// CsgOpNode destructor emptied descendant impls regardless of whether they
+// were still referenced by other live trees, which then manifested as empty
+// positive_children in Add/Intersect finalize and a crash in BatchUnion.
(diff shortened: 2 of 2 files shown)


elalish (approved):
Good catch! Look right to you @pca006132?

pca006132 (approved):

PR #1674, opened 2026-04-23
Within-phase cancel via Cancellable<F> wrapper

zmerlynn:
Follow-up on #971, extending #1669.

## Summary

Extends #1669's phase-boundary cancel into the hot loops that run during CSG boolean evaluation. Cancel latency drops from "wait for the next phase boundary" (up to seconds on huge meshes) to ms-scale.

## Approach

Adds ctx-aware `for_each` / `for_each_n` overloads in `src/parallel.h`. Cancel check fires once per `tbb::blocked_range` chunk (or once at the start of the sequential range), not per element. Callsites pass `ctx` directly:

```cpp
for_each(policy, first, last, ctx, body);
if (IsCancelled(ctx)) return ...;
```

`ExecutionContext::Impl*` threads through signatures with a `nullptr` default, so non-CSG callers (Manifold ctors, SDF, hull, smoothing, `MeshGL::Merge`) are unaffected.

Threaded: Assembly phase (`boolean_result.cpp`), Winding03 (`boolean3.cpp`), `Collider::Collisions`, and the Sort phase (`sort.cpp`'s 8 `Impl::` methods).

## Safety: invariant rule

Bug class: a ctx-passing `for_each` skips remaining chunks on cancel → partial output → unconditional downstream consumer (scan, sort, resize, index) → corruption or OOB.

Rule: **every ctx-passing parallel op is followed by `IsCancelled` to keep partial output from feeding unconditional downstream consumers.** Each function carrying a ctx-passing op has its own `Invariant:` comment.

`IsCancelled(ctx)` (in `src/execution_impl.h`) is the canonical reader. `ExecutionContext::Impl::cancel` is private; only `IsCancelled` and `ExecutionContext` are friends. Direct `ctx->cancel.load(...)` from random code is a compile error.

## Perf

A/B (sphere subtract, 3 runs each):
- No-ctx path (default — Python, Rust, OpenSCAD, WASM): zero measurable overhead. Short-circuits on the null check.
- Ctx-active: ~1–3% slowdown. Cost is the relaxed atomic load per chunk.

## Testing

All existing tests pass. Adds \`ForEachCtxSkipsWhenCancelled\` and \`ForEachCtxNullCtxPassesThrough\` for the parallel.h overload.

## Test plan

- [ ] CI passes
- [ ] TSAN clean
- [ ] Manual: large boolean with cancel fired mid-pipeline returns \`Cancelled\`

## Follow-ups (not in this PR)

- \`Face2Tri\`, \`SimplifyTopology\`, \`CalculateBBox\` aren't threaded — cancel latency bounded by "slowest un-threaded phase" on triangulator-heavy workloads.

### DIFF
--- a/src/boolean_result.cpp
+++ b/src/boolean_result.cpp
@@ -107,8 +107,11 @@ std::tuple<Vec<int>, Vec<int>> SizeOutput(
     Manifold::Impl& outR, const Manifold::Impl& inP, const Manifold::Impl& inQ,
     const Vec<int>& i03, const Vec<int>& i30, const Vec<int>& i12,
     const Vec<int>& i21, const Vec<std::array<int, 2>>& p1q2,
-    const Vec<std::array<int, 2>>& p2q1, bool invertQ) {
+    const Vec<std::array<int, 2>>& p2q1, bool invertQ,
+    ExecutionContext::Impl* ctx) {
   ZoneScoped;
+  // Invariant: every ctx-passing parallel op is followed by IsCancelled to
+  // keep partial output from feeding unconditional downstream consumers.
   Vec<int> sidesPerFacePQ(inP.NumTri() + inQ.NumTri(), 0);
   // note: numFaceR <= facePQ2R.size() = sidesPerFacePQ.size() + 1
 
@@ -116,26 +119,27 @@ std::tuple<Vec<int>, Vec<int>> SizeOutput(
   auto sidesPerFaceQ = sidesPerFacePQ.view(inP.NumTri(), inQ.NumTri());
 
   auto policy = autoPolicy(inP.halfedge_.size());
-  for_each(policy, countAt(0_uz), countAt(inP.halfedge_.size() / 3),
+  for_each(policy, countAt(0_uz), countAt(inP.halfedge_.size() / 3), ctx,
            CountVerts({inP.halfedge_, sidesPerFaceP, i03}));
-  for_each(policy, countAt(0_uz), countAt(inQ.halfedge_.size() / 3),
+  for_each(policy, countAt(0_uz), countAt(inQ.halfedge_.size() / 3), ctx,
            CountVerts({inQ.halfedge_, sidesPerFaceQ, i30}));
 
   if (i12.size() >= 1e5) {
-    for_each_n(ExecutionPolicy::Par, countAt(0), i12.size(),
+    for_each_n(ExecutionPolicy::Par, countAt(0), i12.size(), ctx,
                CountNewVerts<false, true>(
                    {sidesPerFaceP, sidesPerFaceQ, i12, p1q2, inP.halfedge_}));
-    for_each_n(ExecutionPolicy::Par, countAt(0), i21.size(),
+    for_each_n(ExecutionPolicy::Par, countAt(0), i21.size(), ctx,
                CountNewVerts<true, true>(
                    {sidesPerFaceQ, sidesPerFaceP, i21, p2q1, inQ.halfedge_}));
   } else {
-    for_each_n(ExecutionPolicy::Seq, countAt(0), i12.size(),
+    for_each_n(ExecutionPolicy::Seq, countAt(0), i12.size(), ctx,
                CountNewVerts<false, false>(
                    {sidesPerFaceP, sidesPerFaceQ, i12, p1q2, inP.halfedge_}));
-    for_each_n(ExecutionPolicy::Seq, countAt(0), i21.size(),
+    for_each_n(ExecutionPolicy::Seq, countAt(0), i21.size(), ctx,
                CountNewVerts<true, false>(
                    {sidesPerFaceQ, sidesPerFaceP, i21, p2q1, inQ.halfedge_}));
   }
+  if (IsCancelled(ctx)) return std::make_tuple(Vec<int>{}, Vec<int>{});
 
   Vec<int> facePQ2R(inP.NumTri() + inQ.NumTri() + 1, 0);
   auto keepFace = TransformIterator(sidesPerFacePQ.begin(),
@@ -206,8 +210,10 @@ void AddNewEdgeVerts(
     concurrent_map<std::pair<int, int>, std::vector<EdgePos>>& edgesNew,
     const Vec<std::array<int, 2>>& p1q2, const Vec<int>& i12,
     const Vec<int>& v12R, const VecView<Halfedge>& halfedgeP, bool forward,
-    size_t offset) {
+    size_t offset, ExecutionContext::Impl* ctx) {
   ZoneScoped;
+  // Invariant: every ctx-passing parallel op is followed by IsCancelled to
+  // keep partial output from feeding unconditional downstream consumers.
   // For each edge of P that intersects a face of Q (p1q2), add this vertex to
   // P's corresponding edge vector and to the two new edges, which are
   // intersections between the face of Q and the two faces of P attached to the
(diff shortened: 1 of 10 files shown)


zmerlynn:
## For reviewers — patterns to watch for

Both rounds of self-review caught issues of the same shape. Worth re-applying these lenses:

**1. Every `cancellable()` site → trace downstream.** A wrapped `for_each` drained mid-iteration produces partial output. If downstream code consumes that unconditionally (scan, sort, resize, index), corruption or OOB follows. Both fixes in this PR match this pattern (SizeOutput counters → `halfedge_` undersize; SortGeometry morton/keep → bogus permutation). If a wrapped loop's output is read without a subsequent `cancelled()` check, flag it.

**2. Functions returning Impl by value → cancel-before-return.** `Boolean3::Result` had a gap: SortGeometry could drain but the function returned `outR` with `NoError`. `BatchUnion`/`BatchBoolean` don't re-check cancel after their last iteration, so corrupt-NoError impls could escape to the user. Invariant now: `Boolean3::Result` never returns corrupt data with NoError status.

**3. Non-cancellable sequential loops between wrapped ones.** `AppendPartialEdges`/`AppendNewEdges` run to completion even on cancel. Safe only if input isn't corrupt — which is what #1 guarantees. If a plain `for` loop reads data produced by a wrapped loop, verify there's a cancel check between.

Happy to split the PR if it'd help review (wrapper + per-phase). Follow-on testing plans: extend the fuzz test with cancel injection, and possibly a DEBUG-only countdown-cancel mechanism for deterministic checkpoint coverage.

pca006132:
Sorry I don't have much time recently, will review in a few days.

src/boolean_result.cpp
```diff
@@ -788,18 +807,26 @@ Manifold::Impl Boolean3::Result(OpType op) const {
...
+  std::tie(faceEdge, facePQ2R) =
+      SizeOutput(outR, inP_, inQ_, i03, i30, i12, i21, xv12_.p1q2, xv21_.p1q2,
+                 invertQ, ctx_);
+
+  // Critical: if cancel fired during SizeOutput, its wrapped for_each loops
+  // drained partway, leaving sidesPerFacePQ undercounted. That sized
+  // outR.halfedge_ too small. The Append* calls below write to halfedge_ via
+  // facePtrR[face]++ indexing -- an undersized buffer means OOB writes. Bail
+  // before touching halfedge_.
+  if (auto c = cancelled()) return *c;
```
elalish, line 821: Feels like it would be safer to have this return check after every cancellable op, rather than having to think about whether or not each one might be safe or not. 

Of course that gets tricky if there are ever several of these in a function. It could be handled pretty nicely with exceptions, but of course we're avoiding those generally.

zmerlynn: Good point — applied uniformly in 263a5f2d. Every `cancellable()` op is now followed by `IsCancelled`, and `IsCancelled(ctx)` is the only path to read `cancel` (it's private; just `IsCancelled`, `Cancellable<F>`, and `ExecutionContext` are friends). Per-function `Invariant:` comments make each site self-documenting.

On exceptions: would you open the door for the cancel path specifically? Throw inside `Cancellable<F>::operator()` propagates out of TBB cleanly; one catch at the top of `Boolean3::Result` would replace every per-callsite `IsCancelled`. The bug class becomes structurally impossible — stack unwinding takes any partial `outR` with it. The trade is a different bug class: exception-safety on the hot paths, where a throw at any iteration must leave RAII-managed state (Vec, shared_ptr) in a destructible condition, and any manual two-phase init must be revisited. Not asking to do it here, just curious if it's a principled "no" or a "maybe, show me." Happy to sketch on a separate branch.

elalish: This isn't entirely my area of expertise, so I'll let @pca006132 weigh in here. We have a lot of users that specifically build with exceptions disabled, so we'd need to make sure everything works properly in that case. 

One thing I do take a principled stand on is friends - I strongly dislike them. Is there a reasonable way to factor this so they aren't necessary?

zmerlynn: I think this PR represents that reasonable way, I was merely bringing it up as an alternative if the noise was too high. That's great context that there are users building without exceptions, though - in that case, I think what's here is actually a right way (short, opinionated way to check cancellation after each cancellable). It reads a lot like straight C code or Go, which is not unexpected.

pca006132:
Haven't yet done a full review, but I'm wondering if the cancellation functor is useful since a lot of those operations are already very fast. I would rather modify parallel.h to provide a variant with `ctx` that can be cancelled, e.g.,

https://github.com/elalish/manifold/blob/396bc5841d3550dbacbc82f9a44be431de798138/src/parallel.h#L382-L403

zmerlynn:
Agreed — the parallel.h overload is cleaner: cancel lives at the primitive layer, callsites lose the wrap, and the per-chunk check is cheaper than per-element. I'll amend this PR with that shortly.

The post-loop `IsCancelled` guards stay regardless — they're a defense against partial output from a cancelled wrapped op feeding unconditional downstream code (resizes, scans, indexed reads), which is orthogonal to where the cancel check fires inside `for_each`.

pca006132:
Yes, we do need those `IsCancelled` early returns.

zmerlynn:
Done in `ed626436` — ctx-aware `for_each` / `for_each_n` overloads in `parallel.h`, all `cancellable(ctx, f)` callsites converted to `for_each(..., ctx, f)`, and `Cancellable<F>` + `cancellable()` deleted. Per-chunk check via `tbb::blocked_range`, sequential branch checks once at start. Post-loop `IsCancelled` guards stay.

zmerlynn:
No rush on reviewing — I know big PRs like this take real time. To make the next pass easier I reorganized into two commits, machinery then mechanical:

- `727acaf4` *Cancellable primitive* — IsCancelled boundary + ctx-aware `for_each` overloads. ~60 LOC, no callers yet.
- `fe3b8913` *Apply* — callsite conversions + post-loop guards + invariant docs + tests. Repetitive, skim-able once you've seen the first few sites.

Tree-state unchanged from the previous push.

src/boolean_result.cpp
pca006132, line 1: Why no early return for `AppendPartialEdges` and `AppendNewEdges`, for example? And for `createProperties`, we don't have early return within the big sequential loop.

It seems that the current change is mainly for parallel code. I am fine with this if the goal is to keep the PR more focused, but we need to mark it down and do a more finegrained change later.

src/boolean_result.cpp
pca006132: Yeah, I wish we had `Result` in c++ with `?` like rust.

src/parallel.h
```diff
@@ -399,25 +410,40 @@ void for_each(ExecutionPolicy policy, Iter first, Iter last, F f) {
...
+                        [&f, ctx](const tbb::blocked_range<Iter>& range) {
+                          if (IsCancelled(ctx)) return;
                           for (Iter i = range.begin(); i != range.end(); i++)
                             f(*i);
                         });
     });
     return;
   }
 #endif
+  if (IsCancelled(ctx)) return;
```
pca006132, line 425: Actually, I think we should also allow cancellation within a sequential for loop.

pca006132 (commented):
sorry for the delay, was focusing on other stuff on the past few days

src/boolean_result.cpp
zmerlynn: Deliberate scope. Happy to extend this PR with the per-iter `IsCancelled` checks for those three, or split into a follow-up. Preference?

src/boolean_result.cpp
zmerlynn: Yeah. `std::expected` (C++23) is the closest we'll get; macro-izing the if-return wasn't worth the obscurity.

src/parallel.h
zmerlynn: Done in `a9b667cc`: periodic check every 1024 iters on the seq path; nullptr ctx still falls through to `std::for_each`.

src/boolean_result.cpp
pca006132: Follow-up is fine.

pca006132 (approved):

PR #1675, opened 2026-04-23
Add --progress interactive demo mode to perfPhases

zmerlynn:
Follow-up on #971, related to #1674.

## Summary

Adds `--progress[=SECONDS]` to `perfPhases`. Each op runs with a live `ExecutionContext`; a background thread prints progress every SECONDS (default 5). A SIGINT handler makes the first ^C call `ctx.Cancel()` and reports the cancel latency (time from `Cancel()` to `Status()` returning). A second ^C force-exits.

Non-interactive mode (the default) is unchanged.

## Demo

```
$ ./build/extras/perfPhases --progress=2
Interactive mode: progress every 2s. Press ^C to cancel the running op (^C again to force exit).
...
=== Sphere(4096): add, nTri(a) = 8388608 ===
[progress] elapsed 2.1s, 0%
----------- 480 ms for   Intersect12 P->Q
[progress] elapsed 4.2s, 0%
[progress] elapsed 10.4s, 0%
^C
^C received, cancelling... (press ^C again to force exit)
total = 14.05 sec, status = 14
cancel latency = 1662.65 ms (Cancel() -> Status returned)
```

## Why

Concrete demo of the ExecutionContext cancellation feature. On current master (phase-boundary cancel only, via #1669), cancel latency is dominated by "time remaining in current phase" — ~seconds on large meshes, as in the demo above. Once #1674 lands (within-phase cancel), the same demo shows ms-scale latency, making this a useful before/after.

## Notes

- Progress % = `doneBooleans / totalBooleans`, so it stays at 0 for the single-boolean ops this tool runs. Elapsed time still ticks, so the user can see work is happening.
- SIGINT handler is only installed when `--progress` is passed. Default invocation preserves normal ^C-exits behavior.

## Test plan

- [ ] CI passes
- [ ] Manual: run with `--progress=2`, send ^C mid-op, verify cancel latency is printed.

### DIFF
--- a/extras/perf_phases.cpp
+++ b/extras/perf_phases.cpp
@@ -15,10 +15,24 @@
 // Prints per-phase Boolean3 wall-clock timings on representative workloads.
 // Requires MANIFOLD_TIMING=ON (or MANIFOLD_DEBUG=ON) so Timer::Print fires.
 // Used for cancel-latency analysis and for identifying dominant phases.
+//
+// With --progress[=SEC], runs each op with a live ExecutionContext,
+// prints progress every SEC seconds (default 5), and installs a SIGINT
+// handler so ^C cancels the running op and reports cancel latency.
+// First ^C requests cancel; second ^C force-exits. Useful as an
+// interactive demo of ExecutionContext cancellation.
 
+#include <unistd.h>
+
+#include <atomic>
 #include <chrono>
+#include <csignal>
+#include <cstdlib>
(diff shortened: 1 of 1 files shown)


zmerlynn:
Closing — opened prematurely. Also, in retrospect ~200 lines for an interactive demo is more code than the demo value justifies. If we want an interactive cancel demo later, a smaller and simpler approach would fit better.

PR #1678, opened 2026-04-25
Drop early return in CMake on emscripten

### LINKED ISSUE #1677: Install manifold library in emscripten mode
Currently, there is 

https://github.com/elalish/manifold/blob/8b173db61ab3ddae1b5ce0c8a259c4780ac404d4/CMakeLists.txt#L253-L256

which skips the cmake package generation and header installation. This prevents using manifolds wasm artifacts to be found in a wasm cmake project down stream.

Is there any particular reason for this? I tried locally patching these lines out, and I don't observe any issues. If there are no objections, I can open a PR.

paulgessinger:
This removes the check and early return in the CMake configuration if emscripten is configured, allowing the installation of the CMake configuration and library headers.

Fixes https://github.com/elalish/manifold/issues/1677

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -250,11 +250,6 @@ set_source_files_properties(
   PROPERTIES GENERATED TRUE
 )
 
-# If it's an EMSCRIPTEN build, we're done
-if(EMSCRIPTEN)
-  return()
-endif()
-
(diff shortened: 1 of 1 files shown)


elalish (approved):
Thanks!

PR #1680, opened 2026-04-27
add non-blocking PR benchmark guard in manifold.yml (base vs head perfTest)

AnshulPatil2005:
## Summary

This PR adds an initial PR benchmark guard to CI for performance visibility during review.
It runs a lightweight benchmark comparison between:
- the PR base commit, and
- the PR head commit,
on the same runner to reduce cross-runner noise.

### working
1. Checkout repository with full history.
2. Create two git worktrees:
   - `wt-base` from `github.event.pull_request.base.sha`
   - `wt-head` from `github.event.pull_request.head.sha`
3. Build and run `extras/perfTest` for each variant (`base`, `head`) with the same settings.
benchmark used -> perfTest
4. Repeat runs (`PERF_REPEATS=3`) and save outputs under:
   - `bench/base/run*.txt`
   - `bench/head/run*.txt`

### reporting
- Parse timing lines (`time = ... sec`) from both sides.
Compute per-run means and compare using mean of run means (primary); median of run means is also reported for context.
- Emit a warning when both thresholds are exceeded:
  - `REGRESSION_WARN_PCT=20`
  - `REGRESSION_WARN_ABS_MS=10`
- Append markdown summary to `GITHUB_STEP_SUMMARY`.
- Upload raw outputs and JSON summary as artifact (`pr-benchmark-guard`).

## expected follow-up
Calibrate thresholds and repeats using data.
Improve output accordingly if needed (case-level clarity, summary polish).

ref - https://github.com/opencax/GSoC/issues/114#issuecomment-4044725388

### DIFF
--- a/.github/workflows/manifold.yml
+++ b/.github/workflows/manifold.yml
@@ -637,7 +637,17 @@ jobs:
           --json-out ./bench/result.json
         cat ./bench/summary.md >> "$GITHUB_STEP_SUMMARY"
     - name: Upload benchmark artifacts
+      if: always()
       uses: actions/upload-artifact@v4
       with:
         name: pr-benchmark-guard
         path: bench/
+    - name: Cleanup benchmark worktrees
+      if: always()
+      shell: bash
+      run: |
+        for wt in wt-base wt-head; do
+          if [ -d "$wt" ]; then
+            git worktree remove "$wt" --force
+          fi
+        done
--- a/.github/workflows/manifold.yml
+++ b/.github/workflows/manifold.yml
@@ -591,3 +591,53 @@ jobs:
     - name: Compare meshes across platforms
       run: |
         bash ./scripts/compare_artifact.sh meshes
+
+  benchmark_guard:
+    name: perfTest base-vs-head guard
+    if: github.event_name == 'pull_request'
+    timeout-minutes: 45
+    runs-on: ubuntu-24.04
+    env:
+      PERF_REPEATS: 3
+      REGRESSION_WARN_PCT: 20
+      REGRESSION_WARN_ABS_MS: 10
+    steps:
+    - uses: actions/checkout@v6
+      with:
+        fetch-depth: 0
+    - name: Install apt packages
+      uses: awalsh128/cache-apt-pkgs-action@v1.6.0
+      with:
+        packages: libgtest-dev libtbb-dev cmake
+        version: 1
+    - *clone-clipper2
+    - name: Prepare base and head worktrees
+      shell: bash
+      run: |
+        set -euo pipefail
+        BASE_SHA="${{ github.event.pull_request.base.sha }}"
+        HEAD_SHA="${{ github.event.pull_request.head.sha }}"
+        git worktree add wt-base "$BASE_SHA"
+        git worktree add wt-head "$HEAD_SHA"
+    - name: Benchmark base and head
+      shell: bash
+      run: |
+        for variant in base head; do
+          bash ./scripts/run_pr_perf_guard.sh "./wt-${variant}" ./clipper2 "./bench/${variant}" "${PERF_REPEATS}"
+        done
+    - name: Compare benchmark results
+      shell: bash
+      run: |
+        python3 ./scripts/compare_pr_perf_guard.py \
+          --base-dir ./bench/base \
+          --head-dir ./bench/head \
+          --warn-pct "${REGRESSION_WARN_PCT}" \
+          --warn-abs-ms "${REGRESSION_WARN_ABS_MS}" \
+          --markdown-out ./bench/summary.md \
+          --json-out ./bench/result.json
+        cat ./bench/summary.md >> "$GITHUB_STEP_SUMMARY"
+    - name: Upload benchmark artifacts
+      uses: actions/upload-artifact@v4
+      with:
+        name: pr-benchmark-guard
+        path: bench/
--- a/.github/workflows/manifold.yml
+++ b/.github/workflows/manifold.yml
@@ -656,3 +656,4 @@ jobs:
         cat <<'EOF' >> "$GITHUB_STEP_SUMMARY"
 ###Inspect Raw Benchmark Logs
 Open the `Compare benchmark results` step in this workflow run.
+        EOF
--- a/.github/workflows/manifold.yml
+++ b/.github/workflows/manifold.yml
@@ -614,44 +614,11 @@ jobs:
     - name: Prepare base and head worktrees
       continue-on-error: true
       shell: bash
-      run: |
-        set -euo pipefail
-        BASE_SHA="${{ github.event.pull_request.base.sha }}"
-        HEAD_SHA="${{ github.event.pull_request.head.sha }}"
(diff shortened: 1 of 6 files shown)


AnshulPatil2005:
artifact upload uses if: always() so benchmark outputs are available even if earlier steps fail,
added an cleanup step to remove wt-base/wt-head worktrees after the job which was missing 

if base/head data is invalid (e.g., different run counts, missing run files, or empty parsed samples), should we fail the benchmark_guard job?

elalish:
>if base/head data is invalid (e.g., different run counts, missing run files, or empty parsed samples), should we fail the benchmark_guard job?

No, if it's only a warning when the perf drops, it should also only be a warning if these other things fail. Out of curiosity, where will these warnings show up?

AnshulPatil2005:
[result.json](https://github.com/user-attachments/files/27365213/result.json)
[summary.md](https://github.com/user-attachments/files/27365214/summary.md)
> > if base/head data is invalid (e.g., different run counts, missing run files, or empty parsed samples), should we fail the benchmark_guard job?
> 
> No, if it's only a warning when the perf drops, it should also only be a warning if these other things fail. Out of curiosity, where will these warnings show up?

Got it,
For visibility: currently warnings show up as GitHub Actions in the benchmark_guard job logs (::warning::) and also in the uploaded bench artifacts (summary.md/result.json).
here is the example of the last output attached

AnshulPatil2005:
I switched the primary comparison metric from **median of run means** to **mean of run means**, since it should better reflect overall timing differences across repeated runs, while still keeping **median of run means** in the summary as a secondary context metric.
Am i going in the right direction here ?should there be any specific change ? @elalish @pca006132

elalish:
Sounds reasonable - @pca006132 you've been doing more on the perf testing side of things, do you have time to review this?

.github/workflows/manifold.yml
```diff
@@ -591,3 +591,67 @@ jobs:
...
+          --markdown-out ./bench/summary.md \
+          --json-out ./bench/result.json
+        cat ./bench/summary.md >> "$GITHUB_STEP_SUMMARY"
+    - name: Upload benchmark artifacts
+      if: always()
+      uses: actions/upload-artifact@v4
+      with:
+        name: pr-benchmark-guard
+        path: bench/
+    - name: Cleanup benchmark worktrees
```
pca006132: is this useful?

.github/workflows/manifold.yml
```diff
@@ -591,3 +591,67 @@ jobs:
...
+      run: |
+        python3 ./scripts/compare_pr_perf_guard.py \
+          --base-dir ./bench/base \
+          --head-dir ./bench/head \
+          --warn-pct "${REGRESSION_WARN_PCT}" \
+          --warn-abs-ms "${REGRESSION_WARN_ABS_MS}" \
+          --markdown-out ./bench/summary.md \
+          --json-out ./bench/result.json
+        cat ./bench/summary.md >> "$GITHUB_STEP_SUMMARY"
+    - name: Upload benchmark artifacts
```
pca006132: what artifact are we uploading? mostly just the `result.json`?

.github/workflows/manifold.yml
```diff
@@ -591,3 +591,67 @@ jobs:
...
+      if: always()
```
pca006132: why do we need `if` here if this is always?

pca006132 (changes requested):
I don't understand what mean you are talking about.

By `run means`, are you running each *single* benchmark multiple times and taking their mean? And `mean of run means` is taking the mean across benchmarks? If yes, I don't think you need the final `mean`, just compare the mean of each benchmark.

AnshulPatil2005:
> I don't understand what mean you are talking about.
> 
> By `run means`, are you running each _single_ benchmark multiple times and taking their mean? And `mean of run means` is taking the mean across benchmarks? If yes, I don't think you need the final `mean`, just compare the mean of each benchmark.

Yes you are understanding it right, honestly I was confused with the approach for the comparison ; I will do the each benchmark mean comparison

.github/workflows/manifold.yml
AnshulPatil2005: It removes `wt-base` and `wt-head` created during the job. If we later add more steps in the same job, cleaning these temporary worktrees can be useful to keep workspace usage predictable.

Otherwise, cleanup is effectively automatic at job end, so we can remove this step to keep the workflow simpler.

.github/workflows/manifold.yml
AnshulPatil2005: The only required output for the guard itself is result.json (and summary.md for human-readable context).
I uploaded the full bench/ mainly to keep raw run*.txt for debugging when results look suspicious.
If you prefer, I can narrow this to just result.json + summary.md ?

.github/workflows/manifold.yml
AnshulPatil2005: yea makes sense, I'll remove it

.github/workflows/manifold.yml
pca006132: Just simplify it, I don't think we need to worry about workspace usage here, as long as things are not cached.

.github/workflows/manifold.yml
pca006132: I feel like it suffices to just print them instead of storing them as artifacts...
And also, what do we have in those raw input files?

pca006132 (commented):
Can we use cached build from master? I remember GitHub action has some complicated cache setup, some may have security concerns, so we may need to be a bit careful about that.

.github/workflows/manifold.yml
AnshulPatil2005: done

.github/workflows/manifold.yml
AnshulPatil2005: each file is just timing output from one full perfTest run (all 8 cases)
it could be printed in the logs

AnshulPatil2005:
> Can we use cached build from master? I remember GitHub action has some complicated cache setup, some may have security concerns, so we may need to be a bit careful about that.

Yes, good point. I am also finding a safe cache strategy for base/master builds to avoid cross-branch/cache-poisoning issues and other issues as you mentioned.

This could help in other sanitizer lane pr as-well 

edit :this job currently takes around 4 minutes while the overall CI critical path is much longer so i dont think we should add a cache to this

.github/workflows/manifold.yml
```diff
@@ -591,3 +591,67 @@ jobs:
...
+      continue-on-error: true
+      shell: bash
+      run: |
+        for variant in base head; do
+          bash ./scripts/run_pr_perf_guard.sh "./wt-${variant}" ./clipper2 "./bench/${variant}" "${PERF_REPEATS}"
+        done
+    - name: Compare benchmark results
+      shell: bash
+      run: |
+        python3 ./scripts/compare_pr_perf_guard.py \
```
pca006132: as mentioned in the meeting, move these to a script.

AnshulPatil2005: done

pca006132 (approved):

PR #1681, opened 2026-04-27
Use std::sqrt in svd.h to avoid an FP-determinism risk

zmerlynn:
## Summary

Replaces three `std::hypot` calls in `src/svd.h` with `std::sqrt(a*a + b*b)`.

`std::hypot` is recommended but not required correctly-rounded (IEEE 754 §9.2), so different libms are free to round differently. These calls sit inside SVD's iterative reductions, so in principle this leaves `SpectralNorm` open to platform-dependent rounding -- though the OBJ-hash determinism CI hasn't caught it in practice.

`std::sqrt(a*a + b*b)` is bit-exact across platforms: `*`, `+`, and `sqrt` are correctly-rounded under existing `-ffp-contract=off`. Overflow protection isn't needed at SVD's input scales (eigenvalues of `A^T A` for typical mesh-transform matrices).

## Test plan

- [x] Full `manifold_test` suite passes locally (Mac, RelWithDebInfo)
- [x] Determinism CI (`#1594` SHA256-hash workflow) should remain green

### DIFF
--- a/src/svd.h
+++ b/src/svd.h
@@ -92,7 +92,7 @@ inline double Dist2(vec3 v) { return manifold::la::dot(v, v); }
 inline Givens ApproximateGivensQuaternion(Symmetric3x3& A) {
   Givens g{2.0 * (A.m_00 - A.m_11), A.m_10};
   bool b = _gamma * g.sh * g.sh < g.ch * g.ch;
-  double w = 1.0 / hypot(g.ch, g.sh);
+  double w = 1.0 / std::sqrt(g.ch * g.ch + g.sh * g.sh);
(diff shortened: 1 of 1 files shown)


zmerlynn:
Local sanity check on macOS (Apple clang 21, ARM64) with manifold's flags: `std::hypot(a, b)` vs `std::sqrt(a*a + b*b)` disagree at 1 ULP on ~5.5% of 10M random inputs in [-2, 2]. Doesn't prove cross-platform divergence on its own, but confirms Apple's libm exercises hypot's §9.2 latitude.

pca006132 (approved):
good catch, thanks.

PR #1685, opened 2026-04-30
Improve import transformation for complex models

tonious:
Some complex glTF models are going to pieces when imported.

For example, here is the LattePanda Mu carrier board model (step file is publicly available) and converted to glTF with opencascade-tools.  Different nodes have different rotations.
Screenshot 2026-04-30 at 9 43 28 am

This PR fixes the problem by iterating over glTF scenes rather than nodes.  The end result is more recognizable:

Screenshot 2026-04-30 at 9 44 02 am

(This one isn't actually on the Villianous Task List, I just happened to notice it on the way by)

### DIFF
--- a/bindings/wasm/lib/import-model.ts
+++ b/bindings/wasm/lib/import-model.ts
@@ -31,7 +31,7 @@
  * including manifold in another project.
  */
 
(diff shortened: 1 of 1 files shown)


elalish (approved):
Good catch, thanks!

PR #1687, opened 2026-05-01
Fix mergeRec stability when pivot comes from right half

zmerlynn:
## Summary

\`mergeRec\` (parallel merge sort in \`src/parallel.h\`) used \`lower_bound\` on the right-pivot branch, which placed left-side equal-to-pivot elements after the pivot in the merged output — breaking stability. Switching that branch to \`upper_bound\` puts them in the first sub-merge, before the pivot.

The mirror branch (left pivot) was already stable via \`lower_bound\` on the right; #1472's stability fix only handled that case.

## Test

\`ParallelStableSortStability\` sorts 50k items across several equal-key bucket sizes and asserts ascending input tags remain ascending within each group. Verified TDD: red on parent under \`MANIFOLD_PAR=ON\`, green after.

## Test plan

- [ ] CI matrix (\`MANIFOLD_PAR ∈ {ON, OFF}\`, TSAN, determinism)

### DIFF
--- a/test/manifold_test.cpp
+++ b/test/manifold_test.cpp
@@ -1785,31 +1785,31 @@ TEST(Manifold, ExecutionContextProgressInvariant) {
 }
 
 // Sweeps equal-key bucket sizes that span the parallel mergeSort's
-// partition boundaries (n > kSeqThreshold under MANIFOLD_PAR=ON).
+// partition boundaries (n > kSeqThreshold under MANIFOLD_PAR=ON). Tags
+// are sequential and unique; keys are scrambled by a multiplicative hash
+// so the input isn't already sorted by key. std::stable_sort is the
+// oracle.
 TEST(Manifold, ParallelStableSortStability) {
   constexpr size_t kN = 50000;  // > kSeqThreshold to hit parallel
   constexpr size_t kBucketSizes[] = {kN, kN / 2, kN / 4, kN / 8, 1000, 100};
   for (size_t bucket : kBucketSizes) {
     struct Item {
-      double key;
+      size_t key;
       int tag;
     };
     std::vector<Item> items;
     items.reserve(kN);
     for (size_t i = 0; i < kN; ++i) {
-      items.push_back({static_cast<double>(i / bucket), static_cast<int>(i)});
+      items.push_back({(i * 2654435761u % kN) / bucket, static_cast<int>(i)});
     }
-    manifold::stable_sort(
-        items.begin(), items.end(),
-        [](const Item& a, const Item& b) { return a.key > b.key; });
-    // Ascending input tags + stable sort: tags must ascend within equal keys.
-    size_t violations = 0;
-    for (size_t i = 1; i < items.size(); ++i) {
-      if (items[i].key == items[i - 1].key && items[i].tag < items[i - 1].tag) {
-        ++violations;
-      }
+    auto expected = items;
+    auto cmp = [](const Item& a, const Item& b) { return a.key > b.key; };
+    std::stable_sort(expected.begin(), expected.end(), cmp);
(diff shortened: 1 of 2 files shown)


test/manifold_test.cpp
```diff
@@ -1783,6 +1784,35 @@ TEST(Manifold, ExecutionContextProgressInvariant) {
...
+  constexpr size_t kBucketSizes[] = {kN, kN / 2, kN / 4, kN / 8, 1000, 100};
+  for (size_t bucket : kBucketSizes) {
+    struct Item {
+      double key;
+      int tag;
+    };
+    std::vector<Item> items;
+    items.reserve(kN);
+    for (size_t i = 0; i < kN; ++i) {
+      items.push_back({static_cast<double>(i / bucket), static_cast<int>(i)});
```
elalish: Is there a reason the key needs to be a double, when it seems it naturally wants to be a size_t?

test/manifold_test.cpp
```diff
@@ -1783,6 +1784,35 @@ TEST(Manifold, ExecutionContextProgressInvariant) {
...
+  constexpr size_t kN = 50000;  // > kSeqThreshold to hit parallel
+  constexpr size_t kBucketSizes[] = {kN, kN / 2, kN / 4, kN / 8, 1000, 100};
+  for (size_t bucket : kBucketSizes) {
+    struct Item {
+      double key;
+      int tag;
+    };
+    std::vector<Item> items;
+    items.reserve(kN);
+    for (size_t i = 0; i < kN; ++i) {
```
elalish, line 1802: Isn't this making an input that's already sorted? Should we use a hash function or something to randomize it?

test/manifold_test.cpp
```diff
@@ -1783,6 +1784,35 @@ TEST(Manifold, ExecutionContextProgressInvariant) {
...
+      items.push_back({static_cast<double>(i / bucket), static_cast<int>(i)});
+    }
+    manifold::stable_sort(
+        items.begin(), items.end(),
+        [](const Item& a, const Item& b) { return a.key > b.key; });
+    // Ascending input tags + stable sort: tags must ascend within equal keys.
+    size_t violations = 0;
+    for (size_t i = 1; i < items.size(); ++i) {
+      if (items[i].key == items[i - 1].key && items[i].tag < items[i - 1].tag) {
```
elalish: I think with the sequential, unique tags you can just check that they're in order, regardless of whether the keys are equal.

elalish (changes requested):
Good catch - how'd you notice this? Just some details on the test.

Also, we're getting enough tests now that it's probably getting to be time to reorganize our tests into more files - this one certainly seems like more of a utility test. If you're interested in doing a cleanup like that, you're more than welcome.

zmerlynn:
> Good catch - how'd you notice this? Just some details on the test.

Building an app on manifold with tests that should catch geometry regressions - some of them are really accidental determinism tests in disguise, flaking under MANIFOLD_PAR=ON.

Bisecting one of the flakes with Claude's help pointed at manifold::stable_sort scrambling equal-keyed elements when the right pivot lands inside a bucket. Apologies on the initial test, I should've factored that better!

Updated the test!

> Also, we're getting enough tests now that it's probably getting to be time to reorganize our tests into more files - this one certainly seems like more of a utility test. If you're interested in doing a cleanup like that, you're more than welcome.

Happy to do the test reorganization as a follow-up!

zmerlynn:
Also to be clear, this was not the determinism issue I ran into. Claude developed it as a theory along the way - not what was happening in my case (issue coming soon) but it was an issue nonetheless.

elalish (approved):
Thanks!

PR #1688, opened 2026-05-01
Snapshot meshIDCounter_ once in CsgLeafNode::Compose

zmerlynn:
## Summary

`CsgLeafNode::Compose` read `Manifold::Impl::meshIDCounter_` twice to compute the per-component offset — once when shifting `triRef.meshID` inside the parallel `for_each_n`, and again in the sequential loop that shifts `meshIDtransform` keys. Under cross-thread CSG, another thread can advance the atomic between the reads, so the two offsets disagree: `triRef` holds meshIDs not present in `meshIDtransform`, `IncrementMeshIDs`'s `HashTable` lookup misses, the meshIDs collapse to 0, and `GetMeshGL64` emits runs with the default `Relation`'s `originalID = -1`. Downstream this can also produce different topology after `SetTolerance` because `SimplifyTopology`'s collapse decisions use `SameFace` on the corrupted refs.

The fix snapshots the counter once and reuses it in both shifts. The other read of `meshIDCounter_` in the codebase (`UpdateReference` in `boolean_result.cpp`) already snapshots once and reuses, so it's unaffected.

## Test

`Boolean.BatchBooleanComposeMeshIDStable`: 8 worker threads each repeatedly run `BatchBoolean(Add)` over three pairwise-disjoint cubes (which forces the `Compose` path). Their internal `ReserveIDs` calls during the CSG ops provide the cross-thread counter contention. Without the fix the test sees runs with `originalID = UINT32_MAX` (the default `Relation`); with the fix none. Verified TDD: red on parent, green after.

## Test plan

- [ ] CI matrix (\`MANIFOLD_PAR ∈ {ON, OFF}\`, TSAN, determinism)

### DIFF
--- a/src/csg_tree.cpp
+++ b/src/csg_tree.cpp
@@ -273,11 +273,15 @@ std::shared_ptr<CsgLeafNode> CsgLeafNode::Compose(
   if (nodes.size() > 1 && policy == ExecutionPolicy::Par)
     policy = ExecutionPolicy::Seq;
 
+  // Snapshot once and reuse for both shifts; two reads can disagree under
+  // cross-thread CSG and leave triRef.meshID values not in meshIDtransform.
+  const uint32_t meshIDCounterSnapshot = Manifold::Impl::meshIDCounter_;
+
   for_each_n(
       nodes.size() > 1 ? ExecutionPolicy::Par : ExecutionPolicy::Seq,
       countAt(0), nodes.size(),
       [&nodes, &vertIndices, &edgeIndices, &triIndices, &propVertIndices,
-       numPropOut, &combined, policy](int i) {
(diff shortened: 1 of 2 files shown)


elalish (approved):
Another good catch, thanks! Minor nit: especially for these PRs that don't have an issue with a repro yet, it's great to push a first commit with just the test, so we can see the CI failure, then push the fix after. Then no need to assert that the test used to fail - it's been recorded publicly.

zmerlynn:
> Another good catch, thanks! Minor nit: especially for these PRs that don't have an issue with a repro yet, it's great to push a first commit with just the test, so we can see the CI failure, then push the fix after. Then no need to assert that the test used to fail - it's been recorded publicly.

Good call - I'll do that in the future!

PR #1690, opened 2026-05-03
Add MANIFOLD_NO_IOSTREAM build-time option

zmerlynn:
## Summary

Adds `MANIFOLD_NO_IOSTREAM` (default OFF) that strips iostream- and
filesystem-using bits from the public API and tests. Useful for
freestanding/embedded consumers — the immediate motivator is
`wasm32-unknown-unknown` via [wasm-cxx-shim], which today carries
near-equivalent patches downstream that this PR consolidates upstream.
Idea floated and well-received in [#1046][discussion].

When ON:

- Defines `MANIFOLD_NO_IOSTREAM` and `MANIFOLD_NO_FILESYSTEM` as PUBLIC
  compile defs on `manifold`, so consumers see the stripped API.
- Triggers `CLIPPER2_NO_IOSTREAM=ON` for the bundled Clipper2 via a
  carry-patch tracking AngusJohnson/Clipper2#1094 — drops once that lands.
- Gates iostream-using `TEST()` blocks at compile time.
- `FATAL_ERROR`s if combined with `MANIFOLD_DEBUG` / `MANIFOLD_TIMING`.

Always-on `manifold_no_iostream_check` / `manifoldc_no_iostream_check`
compile the public headers with the macro defined; build fails if
iostream use sneaks back in.

## Test plan

- [x] Default build: unchanged.
- [x] `MANIFOLD_NO_IOSTREAM=ON + MANIFOLD_TEST=ON`: 270 tests register
  (vs 383 default; tests requiring fixture-file reads are gated out at
  compile time), all pass.
- [x] `nm libmanifold.{a,dylib}`: zero iostream/filesystem symbols.
- [x] `MANIFOLD_DEBUG=ON + MANIFOLD_NO_IOSTREAM=ON` fatals as expected.
- [x] New CI cell `build_no_iostream` exercises the option end-to-end
  on Ubuntu/clang.
- [x] End-to-end with the consumer: wasm-cxx-shim builds against this
  branch with its own iostream patches dropped — full integration
  suite passes.

## Notes

- Windows: carry-patch uses `GIT_CONFIG core.autocrlf=false`, ships
  with `*.patch text eol=lf` in `.gitattributes`, and applies with
  `--whitespace=nowarn`. Without these, default Windows checkouts
  EOL-convert and the patch fails. CI is Ubuntu-only; happy to add a
  Windows lane on request.

[wasm-cxx-shim]: https://github.com/zmerlynn/wasm-cxx-shim
[discussion]: https://github.com/elalish/manifold/discussions/1046#discussioncomment-16793640

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,4 @@
+# Carry-patches must keep LF line endings even when checked out on
+# Windows with default core.autocrlf=true; otherwise `git apply`
+# refuses them on EOL mismatch with the (non-CRLF) target tree.
+*.patch text eol=lf
--- a/.github/workflows/manifold.yml
+++ b/.github/workflows/manifold.yml
@@ -175,6 +175,40 @@ jobs:
         sudo cmake --install build
         ./scripts/test-cmake.sh
 
+  build_no_iostream:
+    name: NO_IOSTREAM (clang)
+    timeout-minutes: 20
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v6
+    - name: Install apt packages
+      uses: awalsh128/cache-apt-pkgs-action@v1.6.0
+      with:
+        packages: cmake clang libgtest-dev
+    - name: Configure + build with MANIFOLD_NO_IOSTREAM=ON, MANIFOLD_TEST=ON
+      run: |
+        cmake \
+          -DCMAKE_BUILD_TYPE=Release \
+          -DCMAKE_C_COMPILER=clang \
+          -DCMAKE_CXX_COMPILER=clang++ \
+          -DMANIFOLD_NO_IOSTREAM=ON \
+          -DMANIFOLD_USE_BUILTIN_CLIPPER2=ON \
+          -DMANIFOLD_STRICT=ON \
+          -DMANIFOLD_TEST=ON \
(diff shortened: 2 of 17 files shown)


src/no_iostream_check.cpp
pca006132, line 1: Do we need this? Or can we do it by compiling examples?

pca006132 (approved):

src/no_iostream_check.cpp
zmerlynn: Fair point — quick defense was "catches creep on every build," but the new `build_no_iostream` CI cell already builds manifold + manifoldc with `MANIFOLD_NO_IOSTREAM=ON`, so iostream creep in public headers fails there. Dropped the targets + .cpp files in 40586ef1.

pca006132 (approved):
Thanks. Btw, it will be better if you just commit the changes without force-push. We will squash the updates anyway.

zmerlynn:
> Thanks. Btw, it will be better if you just commit the changes without force-push. We will squash the updates anyway.

Noted - still getting used to local preferences. If I need to rebase, would you prefer I keep the commit stack so it's easy to re-review?

pca006132:
yes, it is easier if I can see the diff, especially if the change is large.

PR #1692, opened 2026-05-03
Topological sort for 3MF export

tonious:
Some 3MF parsers (like PrusaSlicer and descendants) expect child nodes to be defined before their parents.  This has come up before.  The previous approach -- reversing the node list -- is not robust.

This PR topologically sorts components ensuring that children are defined before their parents, independent of their order in the source glTF document.

### DIFF
--- a/bindings/wasm/lib/export-3mf.ts
+++ b/bindings/wasm/lib/export-3mf.ts
@@ -94,7 +94,58 @@ export interface Export3MFOptions extends ExportOptions {
 }
 
 /**
+ * Sort 3MF components topologically.
+ *
+ * Some 3MF parsers (like PrusaSlicer and descendants) expect child nodes to be
+ * defined before their parents.  This function sorts the component list
+ * accordingly.
+ */
+const toposort = (components: Component3MF[]) => {
+  const graph: Record<string, string> = {};
+  for (const parent of components) {
+    for (const {objectID: child} of parent.children) {
+      graph[child] = parent.id;
+    }
+  }
+
+  const children = (id: string) =>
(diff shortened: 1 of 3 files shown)


bindings/wasm/lib/export-3mf.ts
```diff
@@ -93,9 +93,59 @@ export interface Export3MFOptions extends ExportOptions {
...
+      graph.filter(([p, c]) => !(p.id === parent.id && c.id === child.id));
+
+  const queue = unsorted.filter(c => orphaned(c));
+  const sorted = [];
+  while (queue.length) {
+    const parent = queue.shift()!;
+    sorted.unshift(parent);
+    for (const child of children(parent)) {
+      disown(parent, child);
+      if (orphaned(child)) queue.push(child);
```
elalish: I assume you know what you're doing, but for some reason I just cannot grok this algorithm right now. Mind adding a test?

elalish (approved):

bindings/wasm/lib/export-3mf.ts
tonious: I'm happy to write tests for this. 

It's a version of Kahn's algorithm -- a stripped down breadth first search.  It collects root nodes, adds them to the result and removes them from the graph.  It then moves on to their children -- which are now root nodes themselves.  Rinse, lather, repeat, and the final result is a list of nodes ordered by generation.

Unshifting nodes on to the sorted list puts earlier generations (with more descendants) at the end of the list.  Order within a generation is not guaranteed, and is not required in this particular case.

elalish (approved):
Thanks, looks great!

PR #1694, opened 2026-05-05
Sub-Boolean progress granularity via per-phase increments

zmerlynn:
Follow-up on #971, extending #1674 / #1693. With cancellation now sub-Boolean, `Progress()` is the one part still stuck at boolean-boundary granularity — a single long Boolean plateaus at `(N-1)/N` until completion.

## What changes

`Progress()` is now `donePhases / totalPhases`. `totalPhases` is a new atomic, set at `GetCsgLeafNode` to `totalBooleans * kPhasesPerBoolean` for Boolean trees. The denominator is op-agnostic — future non-Boolean ops (Hull, LevelSet, Refine, ...) can add to `totalPhases` and increment `donePhases` independently.

`doneBooleans` / `totalBooleans` semantics are unchanged and remain useful for test introspection; they no longer drive `Progress()`.

## How sub-Boolean progress is reported

`Boolean3::Result`'s existing `cancelled()` lambda is renamed to `phase()`, which both advances `donePhases` and checks cancel. Each `if (auto c = phase()) return *c;` replaces the old `cancelled()` call — **zero new lines per phase site**. `kPhasesPerBoolean = 11` must equal the count of `phase()` sites; a `DEBUG_ASSERT` on the happy-path return catches drift.

A function-local `PhaseBalance` scope guard tops up `donePhases` to `kPhasesPerBoolean` on any non-cancelled return, so trivial Booleans (status errors, empty/degenerate inputs) that early-exit before the first `phase()` site still credit a full Boolean. On cancel, partial publication is intentional and `Progress < 1.0` correctly reflects incomplete work.

`Compose` mirrors its existing `doneBooleans += set.size() - 1` with `donePhases += (set.size() - 1) * kPhasesPerBoolean`.

## Out of scope (Phase 2)

The `totalPhases` decoupling enables non-Boolean op coverage via ctx-aware overloads `Op(args..., ExecutionContext& ctx)` mirroring the existing `Status(ExecutionContext&)` pattern. Priority order by user value: `LevelSet` (resolution³), `Hull`, `MinkowskiSum/Difference`, `RefineToLength/Tolerance`.

## Test plan

- [x] All 392 PAR / 387 no-PAR tests pass locally (RelWithDebInfo + TBB).
- [x] `EXPECT_DOUBLE_EQ(Progress(), 1.0)` cross-check added to `ConcurrentProgress` and to `ProgressInvariant`'s three sub-cases. The Compose case is load-bearing — without the `* kPhasesPerBoolean` multiplier in the Compose path, that case fails.
- [x] New `ProgressReachesOneOnTrivialBooleans`: empty + cube hits the IsEmpty fast-path, asserts PhaseBalance credits `kPhasesPerBoolean`.
- [x] New `SubBooleanProgress` (PAR): single overlapping-spheres Boolean; observer thread polls best-effort, final-state invariants pin `totalPhases == donePhases == kPhasesPerBoolean`.
- [x] `CancelConcurrent` strengthened: `Progress() < 1.0` strictly when `Cancelled`.
- [ ] CI green.

### DIFF
--- a/src/boolean_result.cpp
+++ b/src/boolean_result.cpp
@@ -694,6 +694,33 @@ Manifold::Impl Boolean3::Result(OpType op) const {
   const int c2 = op == OpType::Add ? 1 : 0;
   const int c3 = op == OpType::Intersect ? 1 : -1;
 
+  // Sub-Boolean progress accounting. `phase()` advances `donePhases` by 1
+  // and checks cancel; `balance` (declared just below) tops up `donePhases`
+  // to `kPhasesPerBoolean` on any non-cancelled return so `Progress()`
+  // reaches 1.0 even after early-exit paths (status errors, empty/degenerate
+  // inputs) that skip most or all of the phase sites.
+  int phasesPublished = 0;
+  bool fullPath = false;
+  struct PhaseBalance {
+    ExecutionContext::Impl* ctx;
+    int& published;
+    bool& fullPath;
+    ~PhaseBalance() {
+      if (!ctx) return;
+      if (IsCancelled(ctx)) return;  // partial publication is intentional
+      if (fullPath) {
+        DEBUG_ASSERT(published == kPhasesPerBoolean, logicErr,
+                     "Boolean3::Result phase count drift; "
+                     "update kPhasesPerBoolean to match phase() site count");
+        return;
+      }
+      if (published < kPhasesPerBoolean) {
+        ctx->donePhases.fetch_add(kPhasesPerBoolean - published,
+                                  std::memory_order_relaxed);
+      }
+    }
+  } balance{ctx_, phasesPublished, fullPath};
+
   if (inP_.status_ != Manifold::Error::NoError) {
     auto impl = Manifold::Impl();
     impl.status_ = inP_.status_;
@@ -717,7 +744,9 @@ Manifold::Impl Boolean3::Result(OpType op) const {
     return inP_;
   }
 
-  auto cancelled = [&]() -> std::optional<Manifold::Impl> {
+  auto phase = [&]() -> std::optional<Manifold::Impl> {
+    if (ctx_) ctx_->donePhases.fetch_add(1, std::memory_order_relaxed);
+    ++phasesPublished;
     if (IsCancelled(ctx_)) {
       auto impl = Manifold::Impl();
       impl.status_ = Manifold::Error::Cancelled;
@@ -728,7 +757,7 @@ Manifold::Impl Boolean3::Result(OpType op) const {
 
   // Invariant: every ctx-passing parallel op is followed by IsCancelled to
   // keep partial output from feeding unconditional downstream consumers.
-  if (auto c = cancelled()) return *c;
+  if (auto c = phase()) return *c;
 
   if (!valid) {
     auto impl = Manifold::Impl();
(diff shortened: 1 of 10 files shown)


src/boolean_result.cpp
```diff
@@ -694,6 +694,33 @@ Manifold::Impl Boolean3::Result(OpType op) const {
...
+                     "Boolean3::Result phase count drift; "
+                     "update kPhasesPerBoolean to match phase() site count");
+        return;
+      }
+      if (published < kPhasesPerBoolean) {
+        ctx->donePhases.fetch_add(kPhasesPerBoolean - published,
+                                  std::memory_order_relaxed);
+      }
+    }
+  } balance{ctx_, phasesPublished, fullPath};
```
elalish: I've never seen a variable declared this way - cute, but I think for readability, let's put the variable on its own line.

src/boolean_result.cpp
```diff
@@ -694,6 +694,33 @@ Manifold::Impl Boolean3::Result(OpType op) const {
...
+  // and checks cancel; `balance` (declared just below) tops up `donePhases`
+  // to `kPhasesPerBoolean` on any non-cancelled return so `Progress()`
+  // reaches 1.0 even after early-exit paths (status errors, empty/degenerate
+  // inputs) that skip most or all of the phase sites.
+  int phasesPublished = 0;
+  bool fullPath = false;
+  struct PhaseBalance {
+    ExecutionContext::Impl* ctx;
+    int& published;
+    bool& fullPath;
```
elalish: Let's just make these normal members and access them as e.g. `balance.published`.

src/execution_impl.cpp
```diff
@@ -32,11 +32,9 @@ void ExecutionContext::Cancel() {
 bool ExecutionContext::Cancelled() const { return IsCancelled(impl_.get()); }
 
 double ExecutionContext::Progress() const {
-  const int total = impl_->totalBooleans.load(std::memory_order_relaxed);
-  return total > 0
-             ? double(impl_->doneBooleans.load(std::memory_order_relaxed)) /
-                   total
-             : 0.0;
+  const int total = impl_->totalPhases.load(std::memory_order_relaxed);
+  if (total == 0) return 0.0;
```
elalish: In what case would this get hit? In the case where there's zero total work to do, might that imply progress is done, i.e. 100%?

elalish (commented):
Thanks, just some style nits.

src/boolean_result.cpp
zmerlynn: Done in 896b16be, bundled with the `balance.published` change since they touched the same struct.

src/boolean_result.cpp
zmerlynn: Done in 896b16be. `PhaseBalance` now owns `published` and `fullPath` directly; the `phase()` lambda and the happy-path return access them via `balance.published` / `balance.fullPath`. No more references-to-locals.

src/execution_impl.cpp
zmerlynn: Good point — flipped to return 1.0 in 896b16be. Updated the docstring in `common.h`, the WASM TS docstring, and the three tests that pinned 0.0 (`ExecutionContextNoWorkNeeded`, `CBIND.execution_context_happy_path`, `bindings.test.ts`).

One thing worth flagging: this also changes pre-eval `Progress()` (called before `Status(ctx)`) from 0.0 to 1.0, since the counter can't distinguish "trivially complete" from "haven't started yet." The new docstring covers both cases under "no pending work." Users polling `Progress()` from another thread should already gate on a separate "started" signal — the previous 0.0 was equally racy under that pattern — so I don't think this is a regression in practice, but happy to add a sentinel if you'd rather distinguish the two states.

elalish (approved):
Thanks!

PR #1699, opened 2026-05-07
Thread ExecutionContext into Face2Tri for in-flight cancel

zmerlynn:
## Summary

Thread `ExecutionContext` into `Manifold::Impl::Face2Tri` so cancel during the per-face triangulation phase (level 6 of `Boolean3::Result`) is honored mid-sweep. Without this, cancel requests arriving while triangulation is running wait for the full sweep to finish before the next phase boundary observes them.

Addresses item (2) of [#971](https://github.com/elalish/manifold/issues/971#issuecomment-4396198791), narrowed to `Face2Tri` only per the suggestion in that comment.

## Changes

- `Face2Tri` gains a defaulted `ExecutionContext::Impl* ctx` parameter, mirroring the convention in `sort.cpp` ops.
- `IsCancelled(ctx)` checks at function entry, inside the `tbb::task_group` lambda, after `group.wait()`, before `CreateHalfedges`, and inside the sequential `MANIFOLD_PAR=OFF` loop. Both `for_each` calls use the existing ctx-aware `parallel.h` overload.
- `boolean_result.cpp` forwards `ctx_` to `Face2Tri`.
- New test `ExecutionContextCancelDuringFace2Tri`.

If cancel fires inside `Face2Tri`, the surrounding `phase()` check observes `Cancelled` and discards `outR`; partial state never escapes.

## Phase accounting

No new phase counter. `Face2Tri` stays accounted as part of the existing level-6 phase. `kPhasesPerBoolean` unchanged.

## Verified

- Full test suite passes on `MANIFOLD_PAR=ON` and `MANIFOLD_PAR=OFF`.
- New test passes 10/10 under TSAN-instrumented TBB; no new races.

### DIFF
--- a/test/context_test.cpp
+++ b/test/context_test.cpp
@@ -145,36 +145,6 @@ TEST(Manifold, ExecutionContextCancelMidBoolean) {
 }
 #endif  // MANIFOLD_PAR == 1
 
-// Cancel during a workload heavy in Face2Tri. With ctx threaded into
-// Face2Tri the request is honored mid-sweep instead of waiting for the
-// full triangulation to finish. This test exercises the new code path
-// so sanitizer lanes catch UB in the partial-state path; it does not
-// strictly distinguish pre/post-fix outcomes (existing phase-boundary
-// checks already yield Cancelled).
-#if MANIFOLD_PAR == 1
-TEST(Manifold, ExecutionContextCancelDuringFace2Tri) {
-  // High-segment-count spheres bias the boolean's wall time toward
-  // Face2Tri (many sub-edges per face from the intersection curve).
-  Manifold a = Manifold::Sphere(1.0, 384);
-  Manifold b = Manifold::Sphere(1.0, 384).Translate(vec3(0.5, 0, 0));
-  Manifold u = a + b;
-
-  ExecutionContext ctx;
-  std::atomic<Manifold::Error> result{Manifold::Error::NoError};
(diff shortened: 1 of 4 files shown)


test/context_test.cpp
```diff
@@ -145,6 +145,36 @@ TEST(Manifold, ExecutionContextCancelMidBoolean) {
...
+// Cancel during a workload heavy in Face2Tri. With ctx threaded into
+// Face2Tri the request is honored mid-sweep instead of waiting for the
+// full triangulation to finish. This test exercises the new code path
+// so sanitizer lanes catch UB in the partial-state path; it does not
+// strictly distinguish pre/post-fix outcomes (existing phase-boundary
+// checks already yield Cancelled).
+#if MANIFOLD_PAR == 1
+TEST(Manifold, ExecutionContextCancelDuringFace2Tri) {
+  // High-segment-count spheres bias the boolean's wall time toward
+  // Face2Tri (many sub-edges per face from the intersection curve).
```
elalish: Is this true? I would expect large flat faces to have more triangulation time. Anyway, this test is nearly identical to the one above - let's just pick one.

elalish (approved):
Thanks!

test/context_test.cpp
zmerlynn: Fair. Dropped in 53a05ef9 -- `ExecutionContextCancelMidBoolean` covers the same path.

elalish (approved):

PR #1702, opened 2026-05-08
Add ctx-aware Refine overloads for progress + cancellation

zmerlynn:
## Summary

Adds `Refine`/`RefineToLength`/`RefineToTolerance` overloads taking `ExecutionContext&`, with matching C/Python/WASM bindings. Each runs the eager refinement with mid-flight cancel checks and reports a single progress phase reserved up front so `Progress()` rises monotonically.

Addresses item (1) of [#971](https://github.com/elalish/manifold/issues/971#issuecomment-4396198791). Hull, Minkowski, LevelSet to follow as separate PRs using the same pattern.

## Changes

- `Impl::Refine` gains a defaulted `ctx` with cancel checks at entry, after `Subdivide`, and after the `InterpTri` for_each_n. The for_each_n threads ctx; `SortGeometry` was already ctx-aware.
- `GetCsgLeafNode` gains an `extraPhases` parameter so non-Boolean callers reserve their phase count up front. Refine wrappers pass `extraPhases=1` so Progress rises monotonically through the boolean tree and Refine without dipping.
- Private `RefineCommon` helper plus three static callback factories (`EdgeDivByN`/`ByLength`/`ByTolerance`) consolidate the boilerplate. `ctx == nullptr` keeps existing behavior; existing API surface preserved.
- Bindings: C (`manifold_refine_with_context`, mirroring `manifold_status_with_context`), Python (overload via `static_cast`), WASM (`refineWithContext` with matching `.d.ts`). External bindings (OCaml/Rust/Java) not updated -- out of tree.

## Tests

- Happy path: leaf Refine, `totalPhases = donePhases = 1`, Cube + Refine(2): 12 -> 48 tris.
- Pre-cancelled ctx returns Cancelled for all three variants.
- Pre-cancel a tree refinement: `totalPhases = kPhasesPerBoolean + 1` immediately (validates no-dip end-to-end).
- C-API smoke test for `manifold_refine_with_context`.

## Verified

396 PAR / 390 no-PAR tests pass. Python module builds.

### DIFF
--- a/bindings/python/manifold3d.cpp
+++ b/bindings/python/manifold3d.cpp
@@ -331,11 +331,31 @@ NB_MODULE(manifold3d, m) {
       .def("smooth_out", &Manifold::SmoothOut, nb::arg("min_sharp_angle") = 60,
            nb::arg("min_smoothness") = 0,
            manifold__smooth_out__min_sharp_angle__min_smoothness)
-      .def("refine", &Manifold::Refine, nb::arg("n"), manifold__refine__n)
-      .def("refine_to_length", &Manifold::RefineToLength, nb::arg("length"),
-           manifold__refine_to_length__length)
-      .def("refine_to_tolerance", &Manifold::RefineToTolerance,
+      .def("refine",
+           static_cast<Manifold (Manifold::*)(int) const>(&Manifold::Refine),
+           nb::arg("n"), manifold__refine__n)
+      .def("refine",
+           static_cast<Manifold (Manifold::*)(int, ExecutionContext&) const>(
+               &Manifold::Refine),
+           nb::arg("n"), nb::arg("ctx"), manifold__refine__n__ctx)
+      .def("refine_to_length",
+           static_cast<Manifold (Manifold::*)(double) const>(
+               &Manifold::RefineToLength),
+           nb::arg("length"), manifold__refine_to_length__length)
+      .def("refine_to_length",
+           static_cast<Manifold (Manifold::*)(double, ExecutionContext&) const>(
+               &Manifold::RefineToLength),
+           nb::arg("length"), nb::arg("ctx"),
+           manifold__refine_to_length__length__ctx)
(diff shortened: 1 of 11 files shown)


bindings/python/manifold3d.cpp
```diff
@@ -331,11 +331,31 @@ NB_MODULE(manifold3d, m) {
...
            manifold__smooth_out__min_sharp_angle__min_smoothness)
-      .def("refine", &Manifold::Refine, nb::arg("n"), manifold__refine__n)
-      .def("refine_to_length", &Manifold::RefineToLength, nb::arg("length"),
-           manifold__refine_to_length__length)
-      .def("refine_to_tolerance", &Manifold::RefineToTolerance,
+      .def("refine",
+           static_cast<Manifold (Manifold::*)(int) const>(&Manifold::Refine),
+           nb::arg("n"), manifold__refine__n)
+      .def("refine",
+           static_cast<Manifold (Manifold::*)(int, ExecutionContext&) const>(
```
elalish, line 338: Ugh, this is so much code, especially in the bindings. Looking at multiplying this across all the other functions, and having to add all of these anytime we add a new function feels a bit daunting from a maintenance perspective. I wonder if there's a different API we could make to simplify this?

One thought: could we attach an execution context to a Manifold instead, and then use that for any operation that's done to it? We haven't made a release with any cancellation stuff yet, so we don't have to worry about breaking compatibility.

zmerlynn: Agreed it's a lot, and reminds me of Go with how much the ctx has to be carried through. What a builder pattern, something like
`m.With(ctx).Foo()`? I'd have to play with what makes sense for the bindings but happy to try to sketch another option.

elalish: Yes, I think that's what I had in mind. If it cleans up the bindings too then I think we'll have a winner. I look forward to seeing what you come up with.

PR #1704, opened 2026-05-09
Attach ExecutionContext to Manifold; remove per-call Status(ctx)

zmerlynn:
Replacement PR for the ExecutionContext lifetime discussion in #971. Following the Options 0 / 1 / 5 thread, this implements Option 1 (attach the context to the Manifold) and reverts the existing per-call `Status(ExecutionContext&)` API to demonstrate that the data-attachment model fully subsumes it.

Two commits:

**1. Attach ExecutionContext to Manifold; remove per-call Status(ctx) variant**

Adds `WithContext` / `WithoutContext` / `GetContext` / `HasContext` to `Manifold`. The attachment travels with the data: copying propagates it; ops like `Boolean` / `Translate` / `Hull` inherit it from the primary operand; static factories (`BatchBoolean`, `Hull(vector)`) inherit from `manifolds[0]`. The no-arg `Status()` routes through the attached context.

Removes the `Status(ExecutionContext&)` overload and its mirror in C / Python / WASM bindings (`manifold_status_with_context`, `statusWithContext`, the second `.def("status", ...)`); existing tests migrate to `m.WithContext(ctx).Status()`.

**2. Refine: route attached ExecutionContext through Subdivide post-pass (#1702)**

The motivating example for per-call ctx params in #1702 was mid-Refine cancellation. With ctx attached, `Refine` / `RefineToLength` / `RefineToTolerance` snapshot `ctx_` and thread it through `GetCsgLeafNode` and `Impl::Refine`, which now checks `IsCancelled` between `Subdivide` and the post-pass and routes through the existing ctx-aware `SortGeometry`.

### Cross-binding parity

All four bindings expose the full set: C++ `WithContext` / `WithoutContext` / `GetContext` / `HasContext`, with matching semantics in C (`manifold_with_context` / `_without_context` / `_has_context` / `_get_context`, the last with `_has_context() == 1` as precondition), Python (`with_context` / `without_context` / `has_context` / `get_context`), and WASM (`withContext` / `withoutContext` / `hasContext` / `getContext`, same precondition).

### Threading

`Manifold::ctx_` is a `shared_ptr<ExecutionContext::Impl>` accessed via `std::atomic_load` / `std::atomic_store`. Lock-free, and the snapshot pins the `Impl` across long-running evaluations even if a concurrent `op=` reseats `ctx_` mid-eval. Rationale (and the contrast with `pNode_`'s mutex pattern) is documented at the field. The 10 new ctx-attach tests pass under TSAN with no warnings.

## Test plan

- [x] Full test suite: 403/403 pass (was 398 before)
- [x] 10 new ctx-attach tests pass under TSAN with no warnings
- [x] clang-format-20 clean
- [x] C-binding has/with/get/without round-trip covered (`CBIND.execution_context_attach_detach_roundtrip`)
- [x] WASM round-trip covered (`hasContext / getContext round-trip`)


include/manifold/manifold.h
```diff
@@ -381,7 +382,32 @@ class Manifold {
    */
   ///@{
   Error Status() const;
-  Error Status(ExecutionContext& ctx) const;
+
+  /// Returns a copy of this Manifold with the given ExecutionContext attached.
+  /// Subsequent operations on the result (and its derived Manifolds) observe
+  /// progress and cancellation through this context. The attachment travels
+  /// with the data: copying the Manifold copies the attachment, and operations
+  /// like Boolean / Translate / Hull propagate the attachment to their
```
elalish: Definitely an improvement. Thinking of your other options, could we do something a little simpler? I prefer the idea of context not propagating at all - it is used to observe / cancel exactly one operation. However, because of our internal CSG tree and lazy evaluation, for Booleans, one op is not one Boolean. In fact the evaluation is triggered by calling `Status`, so can we simply build up some pile of boolean ops ending with `result` and then:
```
result.With(ctx).Status();
```
We can document that any context attached to a manifold that goes into a Boolean gets ignored, e.g. `a.With(ctx) + b.With(ctx)` observes nothing and internally clears their ctx pointers (no special type, everything is still just a `Manifold`).

zmerlynn: Like it — drops the `Derived()` inheritance and the static-factory propagation, straight subtraction.

One wrinkle: deferred vs. eager ops land differently. Booleans defer, so your recipe works:

- `(a + b).With(ctx).Status()` — ctx observed during eval.

But Refine is eager — densification runs synchronously inside the call:

- `m.Refine(n).With(ctx).Status()` — too late, work's done.
- `m.With(ctx).Refine(n)` — only works if Refine reads `this->ctx_`, the propagation you're suggesting deleting.

How would you like to handle the eager case?

- Input-ctx exception: `m.With(ctx).Refine(n)` works; eager ops read input `ctx_`, deferred ops still ignore it.
- Per-call ctx on eager ops: `m.Refine(n, ctx)` — clean rule, two-knob API, but takes us back to #1702.
- Something else?

elalish: >Input-ctx exception: m.With(ctx).Refine(n) works; eager ops read input ctx_, deferred ops still ignore it.

This is exactly what I had in mind. So basically, defered ops ignore and clear any context, and only eager ops (those that trigger immediate evaluation) use the context. So in that sense, Refine and Status are in the same category.

pca006132: My hope is to later make most ops lazy and deferred. Not really a priority, but I think it can make things more consistent?

zmerlynn: Pushed `0412748c` implementing this. Public surface trimmed to just `With` / `manifold_with_context` / `with_context` / `withContext`; `WithoutContext` / `HasContext` / `GetContext` and their mirrors are dropped, since under the new model the user always knows what ctx they attached. Internal `Derived()` helper and `InheritContextFrom` go with the inheritance step.

Refine continues working as the eager-case demonstration — its existing `ctx_` snapshot threads input ctx through unchanged; only the result no longer carries it.

src/manifold.cpp
```diff
@@ -529,7 +536,7 @@ Manifold Manifold::Mirror(vec3 normal) const {
   if (leafImpl->status_ != Error::NoError)
     return PropagateStatus(leafImpl->status_);
   if (la::length(normal) == 0.) {
-    return Manifold();
+    return Manifold(Manifold());
```
elalish: What is this doing?

bindings/wasm/manifold-encapsulated-types.d.ts
```diff
@@ -1168,13 +1168,16 @@ export class Manifold {
...
+   * The attachment is consumed by the next eager op invoked on the result —
+   * status() for a deferred CSG tree, or refine() / refineToLength() /
+   * refineToTolerance() for those eager ops. Deferred ops (Boolean
+   * operators, transforms, batch ops) ignore any attached ctx and produce
+   * a result with no attached ctx. See ExecutionContext for the full model.
    *
    * @group Information
    */
-  statusWithContext(ctx: ExecutionContext): ErrorStatus;
+  withContext(ctx: ExecutionContext): Manifold;
```
elalish, line 1180: Is there a reason we call the function `WithContext` in the bindings, while we use just `With` in the core API?

elalish (approved):
LGTM - @pca006132 what do you think?

src/manifold.cpp
zmerlynn: Oops - search-and-replace leftover from `Derived(...)` -> `Manifold(...)`. Fixed in `8b28b6c3`.

bindings/wasm/manifold-encapsulated-types.d.ts
zmerlynn: Original rationale: C++ overload resolution future-proofs `With(...)` against later shapes; the bindings can't overload, and Python `with` is a hard keyword that would need a `with_` workaround anyway. The consistency cost outweighed the future-proofing - collapsed to `WithContext` everywhere in `ee04311a`.

src/manifold.cpp
```diff
@@ -964,7 +979,7 @@ Manifold Manifold::MinkowskiSum(const Manifold& other) const {
   if (aImpl->status_ != Error::NoError) return PropagateStatus(aImpl->status_);
   auto bImpl = other.GetCsgLeafNode().GetImpl();
   if (bImpl->status_ != Error::NoError) return PropagateStatus(bImpl->status_);
-  return aImpl->Minkowski(*bImpl, false);
+  return Manifold(aImpl->Minkowski(*bImpl, false));
```
pca006132: Is the `Manifold` constructor doing anything here?

src/manifold.cpp
```diff
@@ -981,7 +996,7 @@ Manifold Manifold::MinkowskiDifference(const Manifold& other) const {
   if (aImpl->status_ != Error::NoError) return PropagateStatus(aImpl->status_);
   auto bImpl = other.GetCsgLeafNode().GetImpl();
   if (bImpl->status_ != Error::NoError) return PropagateStatus(bImpl->status_);
-  return aImpl->Minkowski(*bImpl, true);
+  return Manifold(aImpl->Minkowski(*bImpl, true));
```
pca006132: Same here.

pca006132 (commented):
Just the two redundant `Manifold` constructor calls. Otherwise this LGTM, I think the behavior is very similar to the old approach, just with ctx moved from status to `WithContext`, which is nice.

src/manifold.cpp
zmerlynn: Sed leftover from collapsing `Derived(...)` to `Manifold(...)` -- `Impl::Minkowski` already returns a `Manifold`, so the wrap is just a move ctor. Same class as the earlier Mirror fix. Fixed in `db954b5d`.

src/manifold.cpp
zmerlynn: Same fix as the previous one, both in `db954b5d`.

pca006132 (approved):

PR #1709, opened 2026-05-13
Halfedge Refactoring

pca006132:
This is a vibe-coded halfedge refactoring that changes the AoS representation to SoA (without the end vertex storage, so memory usage for halfedge should be lower by 25%). This also modifies the triangulator to directly output halfedges, so we can directly construct halfedges without going through the sorting part.

This gets some performance improvement for large meshes. I will show the detailed benchmark data later. This is a draft pull request because I still haven't had enough time to review the changes. I only know that the changes should be rather mechanical, and most of the additional lines are in the new `Halfedges` struct, which manages the AoS representation and provides helper functions.

In particular, the convex hull code still uses the old representation, because it needs the end vertex result in the halfedge.


src/boolean3.cpp
```diff
@@ -368,10 +402,9 @@ Intersections Intersect12_(const Manifold::Impl& inP, const Manifold::Impl& inQ,
   Kernel12<expandP, forward> k12{a, b, k02, k11};
   Kernel12Recorder<expandP, forward> recorder{k12, {}};
   auto f = [&a](int i) {
-    return a.halfedge_[i].IsForward()
-               ? Box(a.vertPos_[a.halfedge_[i].startVert],
-                     a.vertPos_[a.halfedge_[i].endVert])
-               : Box();
+    const int start = a.halfedge_.Start(i);
+    const int end = a.halfedge_.Start(NextHalfedge(i));
```
pca006132: just use `End(i)`.

src/boolean3.cpp
```diff
@@ -116,6 +135,12 @@ struct Kernel11 {
   const Manifold::Impl& inQ;
 
   std::pair<int, vec4> operator()(int p1, int q1) {
+    return (*this)(p1, inP.halfedge_.Start(p1), inP.halfedge_.End(p1), q1,
```
pca006132: Check if the overload is really useful, or we can just move the initialization of `p1s, p1e, q1s, q1e` inside the code.

src/boolean3.cpp
```diff
@@ -182,6 +209,12 @@ struct Kernel02 {
   const Manifold::Impl& inB;
 
   std::pair<int, double> operator()(int a0, int b2) {
+    FaceEdge edgeB[3];
```
pca006132, line 183: Again, check if the overload is useful.

src/boolean3.cpp
```diff
@@ -419,16 +452,16 @@ Vec<int> Winding03_(const Manifold::Impl& inP, const Manifold::Impl& inQ,
   DisjointSets uA(a.vertPos_.size());
   for_each(autoPolicy(a.halfedge_.size()), countAt(0),
            countAt(a.halfedge_.size()), ctx, [&](int edge) {
-             const Halfedge& he = a.halfedge_[edge];
-             if (!he.IsForward()) return;
+             const int start = a.halfedge_.Start(edge);
+             const int end = a.halfedge_.Start(NextHalfedge(edge));
```
pca006132: again, directly use `End(edge)`.

src/boolean_result.cpp
```diff
@@ -476,32 +474,32 @@ struct DuplicateHalfedges {
...
       int forwardEdge = AtomicAdd(facePtr[newFace], 1);
       int backwardEdge = AtomicAdd(facePtr[faceRight], 1);
-      halfedge.pairedHalfedge = backwardEdge;
 
-      halfedgesR[forwardEdge] = halfedge;
-      halfedgesR[backwardEdge] = {halfedge.endVert, halfedge.startVert,
-                                  forwardEdge};
+      halfedgesR[forwardEdge] = {startVert, endVert, backwardEdge, propVert};
+      halfedgesR[backwardEdge] = {endVert, startVert, forwardEdge,
+                                  pairPropVert};
```
pca006132, line 480: IDK if this is an old bug, it seems that we never set the propVert for the backward edge before.

elalish:
This looks cool - nice first big use of the AI tools. Unrelated to this PR, but on the topic of large, vibe-coded external contributions, which will probably only become more common: I wonder if adding something like a [bundle-size analyzer](https://github.com/relative-ci/bundle-stats) to our CI would be a good check?

src/boolean_result.cpp
```diff
@@ -388,13 +383,12 @@ void AppendPartialEdges(Manifold::Impl& outR, Vec<char>& wholeHalfedgeP,
 }
 
 void AppendNewEdges(
-    Manifold::Impl& outR, Vec<int>& facePtrR,
+    Manifold::Impl& outR, Vec<Halfedge>& halfedgeR, Vec<int>& facePtrR,
```
pca006132, line 386: Should probably document why this is using the old AoS halfedges. This is because for AppendPartialEdges and AppendNewEdges, we can form polygons instead of triangles, and the new SoA halfedges cannot directly handle that (it assumes we can get the end vertex by indexing the next halfedge in the triangle).

src/polygon_internal.h
```diff
@@ -0,0 +1,118 @@
...
+        // Store the exterior contour halfedge, opposite the filled contour.
+        AddHalfedge(end, start);
+      }
+    }
+    contourEnd = halfedges.size();
+  }
+
+  void ReserveTriangles(size_t numTri) {
+    halfedges.reserve(contourEnd + 3 * numTri);
+    edge2halfedge.reserve(edge2halfedge.size() + numTri);
```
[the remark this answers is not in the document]
pca006132: Not really, the map entries are removed when a matching halfedge is added, so the map is hopefully not as large as the number of triangles * 3.

src/shared.h
```diff
@@ -130,20 +132,175 @@ inline vec3 GetBarycentric(const vec3& v, const mat3& triPos,
...
   int propVert;
   bool IsForward() const { return startVert < endVert; }
-  bool operator<(const Halfedge& other) const {
+  bool operator<(const HalfedgeData& other) const {
     return startVert == other.startVert ? endVert < other.endVert
                                         : startVert < other.startVert;
   }
 };
 
+using Halfedge = HalfedgeData;
```
pca006132: Not sure if we should keep this one.

src/shared.h
```diff
@@ -130,20 +132,175 @@ inline vec3 GetBarycentric(const vec3& v, const mat3& triPos,
...
+  }
+
+  void Set(int idx, HalfedgeData edge) {
+    SetStart(idx, edge.startVert);
+    SetEnd(idx, edge.endVert);
+    SetPair(idx, edge.pairedHalfedge);
+    SetProp(idx, edge.propVert);
+  }
+
+  Ref operator[](int idx) {
```
pca006132: These are mostly backward-compatible helpers. Maybe our refactor should remove these and move to the new API.

src/shared.h
```diff
@@ -130,20 +132,175 @@ inline vec3 GetBarycentric(const vec3& v, const mat3& triPos,
...
+    return (*this)[static_cast<int>(idx)];
+  }
+
+  void push_back(HalfedgeData edge) {
+    start_.push_back(edge.startVert);
+    paired_.push_back(edge.pairedHalfedge);
+    propVert_.push_back(edge.propVert);
+  }
+
+  void resize(size_t newSize, HalfedgeData edge = {}) {
```
pca006132: We probably never want to clone the halfedge over to new entries.

pca006132 (commented):
temporary review

src/shared.h
pca006132: Another issue here is that these allows modifying endVerts, but with the current implementation, this is just modifying the startVerts, and may not work if the user is expecting something similar to the old representation that can in fact represent arbitrary polygonal halfedges. We should remove these helpers even though it adds more changes.

pca006132:
> This looks cool - nice first big use of the AI tools. Unrelated to this PR, but on the topic of large, vibe-coded external contributions, which will probably only become more common: I wonder if adding something like a [bundle-size analyzer](https://github.com/relative-ci/bundle-stats) to our CI would be a good check?

Forgot to say, yes, this seems good.

src/boolean_result.cpp
elalish: It does have that look about it. I think duplicating halfedges is rare enough and often degenerate that probably we just didn't notice.

elalish (approved):
So, I'm thinking now that the progress/cancel API is updated and merged, now might be a good time for a release, before we get into updating the CrossSection API and the decimator, which may or may not deserve a major. I'm thinking merge this and ideally deal with #1712 before release. Does that sound good to you, or would you rather push these after release?

pca006132:
The release plan sounds good to me. @zmerlynn may also have stuff that he wants to get into the upcoming release?

We can keep the discussion here if it is short, or move to an issue if more discussion is needed.

zmerlynn:
> The release plan sounds good to me. @zmerlynn may also have stuff that he wants to get into the upcoming release?

Thanks for checking!

#1713 is a good stopping point for cancellation, but otherwise I'm just working on CrossSection. I'd actually be happy for a release ~soon (no rush), the Rust binding is pinned to an intermediate hash right now.

PR #1697, opened 2026-05-06
Reorganize tests: extract ExecutionContext, parallel-utility, and RayCast clusters

zmerlynn:
Per the [suggestion on #1687](https://github.com/elalish/manifold/pull/1687#pullrequestreview-4210873692). As the test suite grows: more files.

Three extractions from `manifold_test.cpp`:

- **`test/context_test.cpp`** — 18 `ExecutionContext*` tests. Accumulated across #1669, #1674, #1693 and ongoing progress/cancel work.
- **`test/parallel_test.cpp`** — 3 `ForEachCtx*` tests + `ParallelStableSortStability`. The "utility test" cluster the comment specifically pointed at.
- **`test/ray_cast_test.cpp`** — 12 `RayCast*` tests. Self-contained feature area.

`ForEachCtx*` lives in `parallel_test.cpp` rather than `context_test.cpp` because it tests `parallel.h`'s overloads — the ctx is the input, not the subject. End-to-end ctx integration tests stay in `context_test.cpp`.

Pure file moves; tests preserved verbatim, no behavior change, counts unchanged.

## Deliberately not extracted

- `ErrorPropagation*` (16 tests) — cross-cutting; splitting fragments the "all error-handling in one place" view.
- Construction (Sphere/Cylinder/Extrude/Revolve/...) — entangled with downstream Boolean and MeshGL tests in the file; higher reorg cost, unclear benefit.
- `OppositeFace`, `DeepChainDoesNotOverflowNumLeaves` — singletons.

## Test plan

- [x] 390 PAR / 386 no-PAR tests pass locally; counts identical to master.
- [x] clang-format clean.
- [ ] CI green.

PR #1713, opened 2026-05-15
Thread ExecutionContext into Hull and Minkowski

zmerlynn:
Picks up the remaining non-static eager ops from the #971 eager-ops backlog (`Refine` landed in #1704, `Face2Tri` in #1699). After this, every eager op the data-attached model can naturally observe is wired up. Static factories (`Hull(vector<vec3>)`, `Hull(vector<Manifold>)`, `LevelSet`, `MeshGL` constructors) are intentionally out of scope: they have no input `Manifold` to source ctx from, and picking the static-factory convention is a separate design conversation.

#### Hull

`Impl::Hull` gains an optional `ctx` parameter. Cancel checks fire at entry, after `QuickHull::buildMesh`, after `SortGeometry`, plus inside `SortGeometry` itself (already ctx-aware via #1697 / #1699). `QuickHull`'s own loop stays cancel-blind -- granularity is "after `buildMesh` returns," matching how `Refine` handles `Subdivide`. Public `Manifold::Hull()` snapshots `this->ctx_` and threads it through `GetCsgLeafNode` and `Impl::Hull`.

#### Minkowski

`Impl::Minkowski` gains an optional `ctx` parameter. Cancel checks fire at function entry, per batch in the convex/non-convex path, per A-face in the non-convex/non-convex path, and before the final `BatchBoolean`. Internal `BatchBoolean` calls are evaluated with ctx via a local `evalBatch` helper that forces lazy eval through `tree.GetCsgLeafNode(ctx)`, routing into Boolean3's existing per-phase cancel machinery -- so mid-Boolean cancel fires inside Minkowski, not just between batches. Per-hull `Manifold::Hull(simpleHull)` calls (the static factory) remain ctx-blind; granularity inside a batch is "one hull."

#### Tests

Pre-cancel tests for the up-front checks (`ManifoldContextCancelMidHull`, `ManifoldContextCancelMidMinkowski`) and concurrent-cancel tests under `MANIFOLD_PAR == 1` that fire cancel from an observer thread mid-evaluation (`ManifoldContextCancelConcurrentHull`, `ManifoldContextCancelConcurrentMinkowski`).

## Test plan

- [x] 397/397 tests pass under `MANIFOLD_PAR=OFF`.
- [ ] Confirm 397/397 under `MANIFOLD_PAR=ON` (concurrent tests will compile in and run).
- [x] clang-format-20 clean.

### DIFF
--- a/bindings/wasm/manifold-encapsulated-types.d.ts
+++ b/bindings/wasm/manifold-encapsulated-types.d.ts
@@ -1169,11 +1169,12 @@ export class Manifold {
 
   /**
    * Returns a copy of this Manifold with the given ExecutionContext attached.
-   * The attachment is consumed by the next eager op invoked on the result —
-   * status() for a deferred CSG tree, or refine() / refineToLength() /
-   * refineToTolerance() for those eager ops. Deferred ops (Boolean
-   * operators, transforms, batch ops) ignore any attached ctx and produce
-   * a result with no attached ctx. See ExecutionContext for the full model.
+   * The attachment is consumed by the next eager op invoked on the result:
+   * status() for a deferred CSG tree, refine() / refineToLength() /
+   * refineToTolerance(), hull(), or minkowskiSum() / minkowskiDifference().
+   * Deferred ops (Boolean operators, transforms, batch ops) ignore any
+   * attached ctx and produce a result with no attached ctx. See
+   * ExecutionContext for the full model.
    *
    * @group Information
    */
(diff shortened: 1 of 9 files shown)


elalish:
This looks promising, but I'd like to have the conversation about how this will look in the public API first. What's your proposal?

zmerlynn:
Good question - I think no new public surface is needed here, this just slots into the `WithContext` mechanic from #1704. Hull and Minkowski become additional eager ops:

```cpp
m.WithContext(ctx).Hull();
m.WithContext(ctx).MinkowskiSum(other);
```

Same shape as `m.WithContext(ctx).Refine(n)` - the public methods `atomic_load(&ctx_)` and pass the snapshot through (manifold.cpp:1043, :978, :996).

I just pushed a follow-up that documents this in the doxygen on `ExecutionContext` / `WithContext` and mirrors it into the WASM `.d.ts` files, plus a Minkowski cancellation `@code` example and a rewrite of the old "per-boolean granularity" caveat to name the actual check points per op. Happy to adjust any of the wording if it doesn't read right - let me know what you think!

elalish (approved):
LGTM. Perhaps we should discuss how to cancel static functions. I wonder if they should all be members of a context? We could consider something like that for the next major, which probably isn't far off.

Issue #1712, opened 2026-05-14
Incorrect normals after difference


twanvl:
After performing a manifold difference operation, the vertices coming from the subtracted manifold have incorrect normals. It looks like the normal vectors are inverted. Calling `computeNormals` afterwards does not recompute them. Only converting to a MeshGL and back seems to correct the normals.

[Here is an example that shows the problem](https://manifoldcad.org/#Incorrect%20normals%3Ccode%3Eimport%20%7B%20Manifold%2C%20setMaterial%20%7D%20from%20'manifold-3d%2FmanifoldCAD'%3B%0Aconst%20%7B%20union%2C%20intersection%2C%20difference%2C%20sphere%20%7D%20%3D%20Manifold%3B%0A%0A%2F%2F%20Difference%20of%20two%20spheres%0Aconst%20model%20%3D%20difference(sphere(10)%2C%20sphere(10).translate(%5B10%2C10%2C10%5D))%3B%0A%0A%2F%2F%20Convert%20to%20mesh%20and%20back%0Aconst%20model2%20%3D%20Manifold.ofMesh(model.getMesh())%3B%0A%0Aconst%20nodes%20%3D%20%5B%0A%20%20%2F%2F%20model%20has%20incorrect%20normals%0A%20%20setMaterial(model.calculateNormals(0)%2C%20%7B%20baseColorFactor%3A%20%5B1%2C%201%2C%201%5D%2C%20attributes%3A%20%5B%22COLOR_0%22%5D%20%7D).translate(%5B-15%2C0%2C0%5D)%2C%0A%20%20%2F%2F%20converting%20to%20mesh%20and%20back%20gives%20correct%20normals%0A%20%20setMaterial(model2.calculateNormals(0)%2C%20%7B%20baseColorFactor%3A%20%5B1%2C%201%2C%201%5D%2C%20attributes%3A%20%5B%22COLOR_0%22%5D%20%7D).translate(%5B15%2C0%2C0%5D)%2C%0A%5D%3B%0A%0Aexport%20default%20nodes%3B):
```
import { Manifold, setMaterial } from 'manifold-3d/manifoldCAD';
const { union, intersection, difference, sphere } = Manifold;

// Difference of two spheres
const model = difference(sphere(10), sphere(10).translate([10,10,10]));

// Convert to mesh and back
const model2 = Manifold.ofMesh(model.getMesh());

const nodes = [
  // model has incorrect normals
[...]

PR #1716, opened 2026-05-16
Some partial fixes for normals

elalish:
A start on #1712. We also need a better normals update API.

I've also got some refactors in here, like splitting MeshGL to its own header, which as been long overdue.


src/smoothing.cpp
```diff
@@ -532,122 +535,124 @@ void Manifold::Impl::SetNormals(int normalIdx, double minSharpAngle) {
...
+        [&](int, const FaceEdge& here, FaceEdge& next) {
+          const double dihedral = degrees(
+              AngleBetween(faceNormal_[here.face], faceNormal_[next.face]));
+          if (dihedral > minSharpAngle ||
+              triIsFlatFace[here.face] != triIsFlatFace[next.face] ||
+              (triIsFlatFace[here.face] && triIsFlatFace[next.face] &&
+               !meshRelation_.triRef[here.face].SameFace(
+                   meshRelation_.triRef[next.face]))) {
+            normals.push_back(vec3(0.0));
+            meshIds.push_back(meshRelation_.triRef[next.face].meshID);
```
elalish, line 606: In addition to the normals on the difference that were backward, there were also some bad normals right near the intersection. This was my attempt to fix that, but it only fixed it partially. I need to come back to this until the test passes all the way (currently has 17 bad normals).

src/smoothing.cpp
```diff
@@ -532,122 +535,124 @@ void Manifold::Impl::SetNormals(int normalIdx, double minSharpAngle) {
...
 
-      int lastGroup = 0;
-      int lastProp = -1;
-      int newProp = -1;
-      int idx = 0;
-      ForVert(endEdge, [&](int current1) {
-        const int prop = oldHalfedgeProp[current1];
-        auto start = oldProperties.begin() + prop * oldNumProp;
+    for (int i = 0; i < normals.size(); ++i) {
+      normals[i] = getTransform(meshIds[i]) * SafeNormalize(normals[i]);
```
elalish, line 620: The above logic is feeding this change. The rest of this function's diff is mostly just a change in indentation.

PR #1717, opened 2026-05-16
Retry transient fetch failures (fixes #1711)

### LINKED ISSUE #1711: CI failure
It seems that we are getting random CI failures due to

```
Failed to fetch version info for @thi.ng/vectors
```

@tonious any idea?

zmerlynn:
Three fetch sites in `bindings/wasm/lib/` (bundler, `fetchModel`, `loadTexture`) were single-shot, so transient blips fatalled CI and production alike. Trying retry-with-backoff (3 attempts, 200/600ms, retry on 5xx/429, fail fast on other 4xx) as the smallest fix.

Also threading status through to error text - the #1711 log only shows the jsDelivr body, no status, so it's hard to know whether retries would have helped this specific failure. Added a small `FetchError` carrying status/url/body for the model/texture paths.

Verified the full vitest suite (325 passed including the live jsDelivr test). Happy to narrow scope or tune the policy if any of this feels off.


bindings/wasm/lib/fetch-with-retry.ts
```diff
@@ -0,0 +1,48 @@
...
+
+/**
+ * Wraps `fetch` with retry-on-transient-failure. Retries on network errors,
+ * HTTP 5xx, and HTTP 429; returns immediately on 2xx and on 4xx (other than
+ * 429) so typo'd URLs fail fast. Aborts via `init.signal` are not retried.
+ *
+ * Exponential backoff: 200ms, 600ms, then give up (~800ms max added latency
+ * before failure when the upstream is fully down).
+ */
+export async function fetchWithRetry(
```
tonious: Would it make sense to put this in `util.ts`?

bindings/wasm/lib/fetch-with-retry.ts
```diff
@@ -0,0 +1,48 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+const ATTEMPTS = 3;
+const BASE_DELAY_MS = 200;
+const FACTOR = 3;
```
tonious: This is a stylistic comment, rather than a refactoring request.  Feel free to completely ignore it :grin:

It might be more idiomatic to build these out as an interface, and then give `fetchWithRetry()` a parameter:

```typescript
interface FetchRetryOptions = {
  attempts: number;
  base_delay_ms: number;
  factor: number;
};

const FetchRetryDefaults: FetchRetryOptions = {
    attempts: 3,
    base_delay_ms: 200,
    factor: 3,
}

export async function fetchWithRetry(
    input: RequestInfo|URL,
    options: Partial<FetchRetryOptions> 
): Promise<Response> {

  const {attempts, base_delay_ms, factor} = {
    ...FetchRetryDefaults, ...options
  };

}
```

In this case, omitting `RequestInit` as that is somewhat redundant... `RequestInfo` evaluates to `Request|string` and all of `RequestInit` can be provided in the `Request` object.

That all said, this is a call used four times.  It's fair to say I'm overthinking it and that global constants are acceptable in this case.

bindings/wasm/lib/fetch-with-retry.ts
```diff
@@ -0,0 +1,48 @@
...
+      if (!isRetryable(response.status) || attempt === ATTEMPTS - 1) {
+        return response;
+      }
+    } catch (err) {
+      if (init?.signal?.aborted) throw err;
+      if (attempt === ATTEMPTS - 1) throw err;
+    }
+    await new Promise(
+        resolve => setTimeout(resolve, BASE_DELAY_MS * FACTOR ** attempt));
+  }
```
tonious: Looking at the `catch` logic... It will potentially swallow exceptions.  The MDN docs say that `fetch()` can only throw `AbortError`, `TypeError` or `NotAllowedError`.  And we don't really want to catch any of those.  So let's get rid of that...


```typescript
for (let attempt = 0; attempt < ATTEMPTS; attempt++) {
 const response = await fetch(input);

  if (!isRetryable(response.status) || attempt === ATTEMPTS - 1) {
    return response;
  }

  await new Promise(resolve => setTimeout(resolve, BASE_DELAY_MS * FACTOR ** attempt));
}
```

If we raise `response` to a higher scope, we can eliminate that extra exit condition.

```typescript
let response:Response|null = null;
for (let attempt = 0; attempt < ATTEMPTS; attempt++) {
  response = await fetch(input);
  if (!isRetryable(response.status)) break;
  await new Promise(resolve => setTimeout(resolve, BASE_DELAY_MS * FACTOR ** attempt));
}
return response as Response;
```

The loop will exit if the response cannot be retried or the number of attempts is met.  The whole thing will throw if `fetch()` throws.  We know that `response` will be a valid `Response` object at the end of it all.

This also eliminates the unreachable code at the end of the function.

bindings/wasm/lib/gltf-io.ts
```diff
@@ -471,7 +473,11 @@ export function disposeMesh(mesh: GLTFTransform.Mesh) {
  * @param uri The location of the image to download
  */
 export async function loadTexture(texture: GLTFTransform.Texture, uri: string) {
-  const response = await fetch(uri);
+  const response = await fetchWithRetry(uri);
+  if (!response.ok) {
+    throw new FetchError(
+        response.status, response.statusText, uri, await response.text());
+  }
```
tonious: Does it make sense to promote this into `fetchWithRetry()`, rather than duplicating it?

That would imply refactoring the call in `bundler.ts` to catch the exception.  Given that we have an in-framework `FetchError()`, just give that an accessor method to return an `esbuild` compatible `errors` array?

tonious (changes requested):
I was thinking of this as a vitest/CI thing, but it's perfectly reasonable to do retries within ManifoldCAD.

I did a little experimental refactoring here to try and make it a little more idiomatically typescript.

bindings/wasm/lib/fetch-with-retry.ts
zmerlynn: Done in the follow-up commit. `fetchWithRetry` lives in `util.ts` now alongside the other helpers; `FetchError` stays in `error.ts`.

bindings/wasm/lib/fetch-with-retry.ts
zmerlynn: Keeping the module constants for now since the four call sites all use the same policy - the interface idea makes total sense once we have a fifth caller that wants different timeouts/attempts, so I'll keep the pattern in mind.

bindings/wasm/lib/fetch-with-retry.ts
zmerlynn: Good catch. Mostly accepted - dropped the broad catch, but kept retry on `TypeError` specifically (the "Failed to fetch" / DNS / connection-reset family) since that's a meaningful resilience win for ManifoldCAD.org users behind flaky networks. `AbortError` and other exceptions propagate untouched now. The unreachable `throw` at the end is gone, and the loop is restructured as `for (;;)` with explicit return/throw per iteration to avoid the `response!` non-null assertions in the previous draft.

bindings/wasm/lib/gltf-io.ts
zmerlynn: Accepted. `fetchWithRetry` now throws `FetchError` on non-2xx directly, so `fetchModel` and `loadTexture` are clean drop-ins (no manual `response.ok` check). Bundler catches `FetchError` and adapts to the esbuild errors-array shape inline at the call site - kept that adapter at the call site rather than adding a `FetchError.toEsbuildErrors()` method since it's two lines.

zmerlynn:
Yeah - figured a fix at the bundler level would be more general (every CDN consumer benefits, not just CI) and surface better diagnostics through the new `FetchError` class. The vitest `test.tags` retry idea you floated in #1711 is still a good belt-and-suspenders option if this turns out not to be enough on its own. Thanks for the refactoring suggestions; addressed each below.

bindings/wasm/lib/util.ts
```diff
@@ -169,3 +171,55 @@ export function formatVolume(mm3: number): string {
...
+ * non-2xx. Retries on HTTP 5xx, HTTP 429, and `TypeError` (network errors
+ * like "Failed to fetch"); returns immediately on other 4xx so typo'd URLs
+ * fail fast. AbortError and other exceptions propagate untouched.
+ *
+ * Exponential backoff: 200ms, 600ms; ~800ms worst-case added latency.
+ */
+export async function fetchWithRetry(
+    input: RequestInfo|URL, init?: RequestInit): Promise<Response> {
+  for (let attempt = 0;; attempt++) {
+    const isLast = attempt === FETCH_ATTEMPTS - 1;
```
tonious, line 203: Another way to do this is to count down...

```typescript
for (let remaining = FETCH_ATTEMPTS; remaining >= 0; --remaining) {
}
```

Then, checking if you're on the last iteration becomes `!remaining`.

bindings/wasm/lib/util.ts
```diff
@@ -169,3 +171,55 @@ export function formatVolume(mm3: number): string {
...
+    if (isRetryableStatus(response.status) && !isLast) {
+      await delay(FETCH_BASE_DELAY_MS * FETCH_FACTOR ** attempt);
+      continue;
+    }
+    if (!response.ok) {
+      throw new FetchError(
+          response.status, response.statusText, inputToUrl(input),
+          await response.text());
+    }
+    return response;
```
tonious: If you move all of this outside of the for loop, you can avoid the duplicated delay.  Like so:

```typescript
  let response = null;
  for (let attempt = 0;; attempt++) {
    const isLast = attempt === FETCH_ATTEMPTS - 1;
    try {
      response = await fetch(input, init);
      if (!isRetryableStatus(response.status)) break; 
    } catch (err) {
      // TypeError covers "Failed to fetch" / DNS / connection-reset
      // errors. Retry those unless out of attempts. AbortError and
      // NotAllowedError propagate untouched.
      if (!(err instanceof TypeError) || isLast) throw err;
    }
    await delay(FETCH_BASE_DELAY_MS * FETCH_FACTOR ** attempt);
  }

  if (response.ok) {
    return response;
  } else {
    throw new FetchError(
        response.status, response.statusText, inputToUrl(input),
        await response.text());
  }
```

This also drops the `continue` statements, making the loop a little easier to follow.

bindings/wasm/lib/util.ts
```diff
@@ -169,3 +171,55 @@ export function formatVolume(mm3: number): string {
...
+  for (let attempt = 0;; attempt++) {
+    const isLast = attempt === FETCH_ATTEMPTS - 1;
+    let response: Response;
+    try {
+      response = await fetch(input, init);
+    } catch (err) {
+      // TypeError covers "Failed to fetch" / DNS / connection-reset
+      // errors. Retry those unless out of attempts. AbortError and
+      // NotAllowedError propagate untouched.
+      if (!(err instanceof TypeError) || isLast) throw err;
```
tonious, line 211: Does it make sense to convert the `TypeError` to a `FetchError`?

bindings/wasm/lib/util.ts
zmerlynn: Took counting up since the `FACTOR ** attempt` backoff reads more naturally that direction (the countdown form would need `FETCH_ATTEMPTS - 1 - remaining` in the exponent). Stylistic, agreed though.

bindings/wasm/lib/util.ts
zmerlynn: Done in 3d288eda. Went with `Response | undefined` plus a single `const r = response!` narrowing after the loop, so the `FetchError` construction stays clean. The `break` paths always assign `response` first, so the assertion is safe.

bindings/wasm/lib/util.ts
zmerlynn: Leaning toward keeping them distinct: `FetchError` carries `status`/`statusText`/`body` which do not really fit a network-layer failure (we never got a response). Letting `TypeError` propagate also lets callers tell "no network" apart from "got 503". Happy to revisit if you have a callsite in mind that would benefit from the unification.

tonious (approved):
Looks good.  Thanks, @zmerlynn!

PR #1718, opened 2026-05-16
Record normals on Manifold; auto-substitute on GetMeshGL; round-trip via MeshGL

### LINKED ISSUE #1712: Incorrect normals after difference

### LINKED ISSUE #1719: Bad normals at seam verts where two surfaces meet just under `minSharpAngle`
Distinct from #1712 (cavity-side inversion, fixed in #1718). Same `Properties.CalculateNormals` test added in #1716, different failure mode caught by enabling its commented-out FIXME assertions.

Repro: `(sphere - sphere.Translate(vec3(10))).CalculateNormals().GetMeshGL()` produces 17 vertices on the seam between outer-A and cut-B surfaces (both spheres radius 10) where `dot(pos, expected_outward) <= 0`.

Example vert `(1.96, 4.27, 8.76)` lies exactly on both spheres. Dihedral between adjacent faces is ~59° (just under default `minSharpAngle = 60`), so `SetNormals` averages instead of splitting into sharp normals. The observed averaged value doesn't match the expected average of the two contributing face normals either - something beyond simple averaging is going wrong (possibly per-meshID transform applied incorrectly to one contributor, propVert merging across the seam, or a sign bug in the averaging algorithm for opposite-orientation faces).

Three plausible causes worth tracing; haven't narrowed down which yet.

zmerlynn:
Fixes #1712, fixes #1719.

## Summary

Make "where are my normals?" a property of the `Manifold` rather than something the caller threads through every call. After `CalculateNormals()`, `GetMeshGL()` returns world-frame normals without an explicit `normalIdx`, and a `MeshGL` <-> `Manifold` round-trip preserves the recording per-run even across mixed-input Booleans (e.g., a cube without normals composed with a sphere that has them).

The #1712 repro from `(sphere - sphere.Translate(10)).CalculateNormals().GetMeshGL()` now returns the expected outward-from-cavity normals. The underlying cross-mesh seam-vert sign issue (#1719) is fixed in `SetNormals`.

## Mechanism

- `runFlags` becomes a bitmask: bit 0 = backside (existing), bit 1 = hasNormals (new). New `MeshGLP::HasNormals(run)` accessor mirrors `Backside(run)`. Fixed `Backside` from `runFlags[run] == 1` to `& 1` so it survives the addition of a second bit.
- `CalculateNormals(0)` marks the per-meshID `Relation.hasNormals` flag, which `Impl::HasNormals()` ANDs across to give the impl-wide answer. `GetMeshGL(-1)` auto-substitutes slot 0 on output when `HasNormals()` is true.
- Normals at the standard slot are stored world-frame on the Impl. `Impl::Transform` and `csg_tree.cpp::Compose` eager-transform `properties_[0..2]` per-meshID so the slot stays in sync with `vertPos_` and `faceNormal_` across any sequence of transforms, including mixed-input Boolean/Compose outputs where some meshIDs carry normals and others don't.
- `SetNormals` multi-normal walk sign-flips `cross(next, here)` when it disagrees with `faceNormal_[next.face]`. This handles Boolean subtractee triangles (winding unchanged, faceNormal_ negated by the Boolean post-pass) without losing the existing "flair" curving for smooth surfaces. This is what fixes #1719.
- `CreateProperties` negates Q's slot 0..2 per source-meshID for subtractee cavity inversion.
- `GetNormal` returns `properties_[0..2]` directly when the meshID's `Relation.hasNormals` is true (world-frame already); applies the per-meshID inverse-normal-transform when false, preserving the pre-PR contract for hand-built MeshGL inputs that don't set the bit.
- Non-zero `normalIdx` to `CalculateNormals` uses the legacy deferred-transform path: stores per-mesh-frame, recovered to world-frame by `GetMeshGL`'s per-run transform on export. Docstring notes this path will be removed in a future release.

## API additions

- `MeshGLP::HasNormals(run)` C++ accessor plus C and Python bindings.
- `MeshGLP::Backside(run)` C and Python bindings (previously C++-only).

## Tests

- `Properties.CalculateNormals`: previously-FIXME'd `dot(pos, norm) > 0` asserts enabled (#1719 regression).
- `Manifold.CalculateNormalsRoundTripsThroughGetMeshGL`: ofMesh+getMesh preserving the per-run flag, Rotate before and after CalculateNormals, no-arg invocation, non-zero idx.
- `Manifold.GetNormalLegacyContract`: per-mesh-frame fallback in `GetNormal` for MeshGLs without bit 1.
- `Manifold.TransformMixedNormalsPerMeshID`, `Manifold.ComposeMixedNormalsPerMeshID`, `Manifold.BooleanSubtractMixedQPerMeshIDNegation`: per-meshID iteration in `Impl::Transform` / `Compose` / `CreateProperties` for mixed-input Boolean outputs.
- `Manifold.CalculateNormalsNonZeroIdxSurvivesTransform`: legacy non-zero `normalIdx` path across post-calc transforms.
- `CBIND.run_flag_accessors`: exercises the new C accessors.

## Known limitations

- Hand-built MeshGL inputs that share propVerts across runs with differing hasNormals settings produce one-interpretation-wins behavior at the shared propVerts (the stored value can only have one semantic). Standard `CalculateNormals` / Boolean / Compose outputs never produce this shape. Documented on `MeshGLP::HasNormals(run)`.
- `Warp` / `WarpBatch` / `SetProperties` keep the recording set; if the user replaces slot 0..2 or warps non-rigidly, the stored normals are stale until they call `CalculateNormals()` again. Documented on each method.

### DIFF
--- a/bindings/wasm/bindings.js
+++ b/bindings/wasm/bindings.js
@@ -252,8 +252,12 @@ Module.setup = function() {
   };
 
   Module.Manifold.prototype.calculateNormals = function(
-      normalIdx, minSharpAngle = 60) {
+      normalIdx = 0, minSharpAngle = 60) {
     return this._CalculateNormals(normalIdx, minSharpAngle);
+  };
+
+  Module.Manifold.prototype.smoothByNormals = function(normalIdx = 0) {
+    return this._SmoothByNormals(normalIdx);
   };
 
   Module.Manifold.prototype.setProperties = function(numProp, func) {
--- a/bindings/wasm/bindings.js
+++ b/bindings/wasm/bindings.js
@@ -451,6 +451,16 @@ Module.setup = function() {
       mat4[15] = 1;
       return mat4;
     }
+
+    backside(run) {
+      return this.runFlags != null && run < this.runFlags.length &&
+          (this.runFlags[run] & 1) !== 0;
+    }
+
+    hasNormals(run) {
+      return this.runFlags != null && run < this.runFlags.length &&
+          (this.runFlags[run] & 2) !== 0;
+    }
   }
 
   Module.Mesh = Mesh;
--- a/include/manifold/mesh.h
+++ b/include/manifold/mesh.h
@@ -154,6 +154,11 @@ struct MeshGLP {
   /// the bounding box. Any edge shorter than tolerance may be collapsed.
   /// Tolerance may be enlarged when floating point error accumulates.
   Precision tolerance = 0;
+  /// True if the first three extra-property channels (slots 3, 4, 5) hold
+  /// vertex normals. Set on output by GetMeshGL/GetMeshGL64 when normals were
+  /// recorded on the Manifold (via CalculateNormals); read on input by the
+  /// MeshGL Manifold constructor so the recording survives a round-trip.
+  bool hasNormals = false;
 
   MeshGLP() = default;
 
@@ -164,9 +169,12 @@ struct MeshGLP {
    *
    * @param normalIdx Specifies the first of the three consecutive property
    * channels forming the (x, y, z) normals to update. NumProp must be at least
-   * normalIdx + 3 and normalIdx must be >= 3.
+   * normalIdx + 3. Default is -1, meaning "use the standard slot (3, 4, 5) if
+   * `hasNormals` is set; otherwise no-op." Pass a value >= 3 explicitly to
+   * update normals at a non-standard channel; that path is retained for
+   * compatibility and will not be supported in a future release.
    */
-  void UpdateNormals(int normalIdx);
+  void UpdateNormals(int normalIdx = -1);
 
   /**
    * Updates the mergeFromVert and mergeToVert vectors in order to create a
--- a/include/manifold/mesh.h
+++ b/include/manifold/mesh.h
@@ -154,27 +154,8 @@ struct MeshGLP {
   /// the bounding box. Any edge shorter than tolerance may be collapsed.
   /// Tolerance may be enlarged when floating point error accumulates.
   Precision tolerance = 0;
-  /// True if the first three extra-property channels (slots 3, 4, 5) hold
-  /// vertex normals. Set on output by GetMeshGL/GetMeshGL64 when normals were
-  /// recorded on the Manifold (via CalculateNormals); read on input by the
-  /// MeshGL Manifold constructor so the recording survives a round-trip.
-  bool hasNormals = false;
 
   MeshGLP() = default;
-
-  /**
-   * Updates the normals of the mesh based on the runTransform matrices and
-   * backside flags, which are then cleared to avoid double-applying them when
-   * round-tripping.
-   *
-   * @param normalIdx Specifies the first of the three consecutive property
-   * channels forming the (x, y, z) normals to update. NumProp must be at least
-   * normalIdx + 3. Default is -1, meaning "use the standard slot (3, 4, 5) if
-   * `hasNormals` is set; otherwise no-op." Pass a value >= 3 explicitly to
-   * update normals at a non-standard channel; that path is retained for
-   * compatibility and will not be supported in a future release.
-   */
-  void UpdateNormals(int normalIdx = -1);
 
   /**
    * Updates the mergeFromVert and mergeToVert vectors in order to create a
@@ -243,13 +224,34 @@ struct MeshGLP {
 
   /**
    * Returns true if this triangle run is on the backside compared to the
-   * original mesh, e.g. from a subtraction. In this case vertex normals will
-   * need to be flipped. UpdateNormals() will take care of this.
+   * original mesh, e.g. from a subtraction. In this case vertex normals stored
+   * for the run point opposite to the result surface and consumers reading
+   * them via `runTransform` should multiply by -1.
    *
    * @param run The index of the triangle run (0 <= run < runFlags.size()).
    */
   bool Backside(size_t run) const {
-    return run < runFlags.size() && runFlags[run] == 1;
+    return run < runFlags.size() && (runFlags[run] & 1) != 0;
+  }
+
+  /**
+   * Returns true if the first three extra-property channels (slots 3, 4, 5)
+   * of this run carry world-frame vertex normals (set by
+   * `Manifold::CalculateNormals(0)` and round-tripped via `runFlags` bit 1).
+   * Consumers should treat the slot as normals and skip re-applying
+   * `runTransform` to it.
+   *
+   * For hand-built MeshGL inputs: if multiple runs share propVerts (same
+   * vertex index used by triangles from different runs), all sharing runs
+   * must agree on what slot 3..5 represents. Mixing hasNormals=true and
+   * hasNormals=false across shared propVerts is undefined - the stored
+   * value can only have one interpretation. Outputs from CalculateNormals
+   * / Boolean / Compose never share propVerts across differing-flag runs.
+   *
+   * @param run The index of the triangle run (0 <= run < runFlags.size()).
+   */
+  bool HasNormals(size_t run) const {
+    return run < runFlags.size() && (runFlags[run] & 2) != 0;
   }
 };
 
--- a/include/manifold/mesh.h
+++ b/include/manifold/mesh.h
@@ -241,6 +241,13 @@ struct MeshGLP {
    * Consumers should treat the slot as normals and skip re-applying
    * `runTransform` to it.
    *
+   * hasNormals is per-run, so different runs may set it differently.
+   * Behavior is undefined when a single propVert is shared by triangles
+   * from runs that disagree - the slot has one interpretation, and a
+   * Transform rotates it for hasNormals=true and clobbers any
+   * hasNormals=false sharer. Standard `CalculateNormals` / Boolean /
+   * Compose outputs never produce that shape.
+   *
    * @param run The index of the triangle run (0 <= run < runFlags.size()).
    */
   bool HasNormals(size_t run) const {
--- a/include/manifold/mesh.h
+++ b/include/manifold/mesh.h
@@ -241,6 +241,12 @@ struct MeshGLP {
    * Consumers should treat the slot as normals and skip re-applying
    * `runTransform` to it.
    *
+   * Mixed hasNormals values across runs are supported. If a propVert is
+   * shared between a hasNormals=true run and a hasNormals=false run (e.g. a
+   * hand-built MeshGL stitches a normals-bearing run to a color-only run at
+   * common vertices), the `Manifold(MeshGL)` ctor duplicates the propVert
+   * so each camp's slot 0..2 transforms independently.
+   *
    * @param run The index of the triangle run (0 <= run < runFlags.size()).
    */
   bool HasNormals(size_t run) const {
--- a/include/manifold/mesh.h
+++ b/include/manifold/mesh.h
@@ -241,12 +241,6 @@ struct MeshGLP {
    * Consumers should treat the slot as normals and skip re-applying
    * `runTransform` to it.
    *
-   * Mixed hasNormals values across runs are supported. If a propVert is
-   * shared between a hasNormals=true run and a hasNormals=false run (e.g. a
-   * hand-built MeshGL stitches a normals-bearing run to a color-only run at
-   * common vertices), the `Manifold(MeshGL)` ctor duplicates the propVert
-   * so each camp's slot 0..2 transforms independently.
-   *
    * @param run The index of the triangle run (0 <= run < runFlags.size()).
    */
   bool HasNormals(size_t run) const {
--- a/include/manifold/mesh.h
+++ b/include/manifold/mesh.h
@@ -224,9 +224,9 @@ struct MeshGLP {
 
   /**
    * Returns true if this triangle run is on the backside compared to the
-   * original mesh, e.g. from a subtraction. In this case vertex normals stored
-   * for the run point opposite to the result surface and consumers reading
-   * them via `runTransform` should multiply by -1.
+   * original mesh, e.g. from a subtraction. Informational only - the framework
+   * already orients stored normals so the standard `getMesh()` flow returns
+   * world-frame values regardless of this bit.
    *
    * @param run The index of the triangle run (0 <= run < runFlags.size()).
    */
@@ -240,13 +240,6 @@ struct MeshGLP {
    * `Manifold::CalculateNormals(0)` and round-tripped via `runFlags` bit 1).
    * Consumers should treat the slot as normals and skip re-applying
    * `runTransform` to it.
-   *
-   * For hand-built MeshGL inputs: if multiple runs share propVerts (same
-   * vertex index used by triangles from different runs), all sharing runs
-   * must agree on what slot 3..5 represents. Mixing hasNormals=true and
-   * hasNormals=false across shared propVerts is undefined - the stored
-   * value can only have one interpretation. Outputs from CalculateNormals
-   * / Boolean / Compose never share propVerts across differing-flag runs.
    *
    * @param run The index of the triangle run (0 <= run < runFlags.size()).
    */
--- a/src/boolean_result.cpp
+++ b/src/boolean_result.cpp
@@ -569,7 +569,8 @@ struct Barycentric {
 };
 
 void CreateProperties(Manifold::Impl& outR, const Manifold::Impl& inP,
-                      const Manifold::Impl& inQ, ExecutionContext::Impl* ctx) {
+                      const Manifold::Impl& inQ, bool invertQ,
+                      ExecutionContext::Impl* ctx) {
   ZoneScoped;
   // Invariant: every ctx-passing parallel op is followed by IsCancelled to
   // keep partial output from feeding unconditional downstream consumers.
@@ -607,6 +608,15 @@ void CreateProperties(Manifold::Impl& outR, const Manifold::Impl& inP,
     const int oldNumProp = PQ ? numPropP : numPropQ;
     const auto& properties = PQ ? inP.properties_ : inQ.properties_;
     const auto& halfedge = PQ ? inP.halfedge_ : inQ.halfedge_;
+
+    // For Subtract, Q's triangles are flipped in the result, so Q's
+    // world-frame vertex normals (slot 0..2 when hasNormals) need a sign
+    // flip to point outward from the result's solid (into the cavity).
+    // Check is per-source-triangle, not whole input - inQ may be a mixed
+    // Boolean result.
+    const bool negateNormals =
+        !PQ && invertQ && oldNumProp >= 3 &&
+        Manifold::Impl::TriHasNormals(inQ.meshRelation_, ref.faceID);
 
     for (const int i : {0, 1, 2}) {
       const int vert = outR.halfedge_.Start(3 * tri + i);
@@ -665,7 +675,9 @@ void CreateProperties(Manifold::Impl& outR, const Manifold::Impl& inP,
           for (const int j : {0, 1, 2})
             oldProps[j] =
                 properties[oldNumProp * halfedge.Prop(3 * ref.faceID + j) + p];
-          outR.properties_.push_back(la::dot(uvw, oldProps));
+          double val = la::dot(uvw, oldProps);
+          if (negateNormals && p < 3) val = -val;
+          outR.properties_.push_back(val);
         } else {
           outR.properties_.push_back(0);
         }
@@ -936,7 +948,7 @@ Manifold::Impl Boolean3::Result(OpType op) const {
     DEBUG_ASSERT(outR.IsManifold(), logicErr,
                  "triangulated mesh is not manifold!");
 
-  CreateProperties(outR, inP_, inQ_, ctx_);
+  CreateProperties(outR, inP_, inQ_, invertQ, ctx_);
   if (auto c = phase()) return *c;
 
   UpdateReference(outR, inP_, inQ_, invertQ, ctx_);
--- a/src/impl.cpp
+++ b/src/impl.cpp
@@ -167,6 +167,36 @@ void Manifold::Impl::RemoveUnreferencedVerts() {
   });
 }
 
+void Manifold::Impl::EagerTransformPropNormals(
+    const Halfedges& halfedge, const MeshRelationD& meshRelation,
+    const mat3& normalTransform, Vec<double>& properties, int numPropVert,
+    int stride, int offset) {
+  // Short-circuit when no meshID carries normals. Note this is OR semantics
+  // (any has it), not AND like Impl::HasNormals() - mixed inputs still need
+  // the per-meshID iteration below to rotate the with-normals subset.
+  bool anyHasNormals = false;
+  for (const auto& m : meshRelation.meshIDtransform) {
+    if (m.second.hasNormals) {
+      anyHasNormals = true;
+      break;
+    }
+  }
+  if (!anyHasNormals) return;
+  Vec<bool> propVisited(numPropVert, false);
+  for (size_t e = 0; e < halfedge.size(); ++e) {
+    if (!TriHasNormals(meshRelation, e / 3)) continue;
+    const int prop = halfedge.Prop(e);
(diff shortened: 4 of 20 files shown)


zmerlynn:
Follow-up on the `Properties.CalculateNormals` FIXME I mentioned in the description: enabling those assertions catches 17 seam verts where two surfaces meet at ~59° dihedral (just under the `minSharpAngle = 60` default), and the averaged normal `SetNormals` produces doesn't match what you'd expect from the two contributing face normals. Distinct from this PR's cavity-inversion fix - failures persist with or without the changes here. Filed as #1719 for the separate trace.

include/manifold/mesh.h
```diff
@@ -164,9 +169,12 @@ struct MeshGLP {
...
    * @param normalIdx Specifies the first of the three consecutive property
    * channels forming the (x, y, z) normals to update. NumProp must be at least
-   * normalIdx + 3 and normalIdx must be >= 3.
+   * normalIdx + 3. Default is -1, meaning "use the standard slot (3, 4, 5) if
+   * `hasNormals` is set; otherwise no-op." Pass a value >= 3 explicitly to
+   * update normals at a non-standard channel; that path is retained for
+   * compatibility and will not be supported in a future release.
    */
-  void UpdateNormals(int normalIdx);
+  void UpdateNormals(int normalIdx = -1);
```
elalish: Since this function hasn't gone out in a release yet, I think we should just remove it. It'll have no need now that the normal transforms are automated.

include/manifold/mesh.h
```diff
@@ -154,6 +154,11 @@ struct MeshGLP {
   /// the bounding box. Any edge shorter than tolerance may be collapsed.
   /// Tolerance may be enlarged when floating point error accumulates.
   Precision tolerance = 0;
+  /// True if the first three extra-property channels (slots 3, 4, 5) hold
+  /// vertex normals. Set on output by GetMeshGL/GetMeshGL64 when normals were
+  /// recorded on the Manifold (via CalculateNormals); read on input by the
+  /// MeshGL Manifold constructor so the recording survives a round-trip.
+  bool hasNormals = false;
```
elalish: Will this work? What if you have a mesh with normals and another without, but with, say colors? When you boolean them together, what value would this get? Perhaps we should put this as another runFlag next to backfacing.

src/manifold.cpp
```diff
@@ -600,6 +605,10 @@ Manifold Manifold::SetProperties(
   auto pImpl = std::make_shared<Impl>(*leafImpl);
   const int oldNumProp = NumProp();
   const Vec<double> oldProperties = pImpl->properties_;
+  // TODO: hasNormals_ is preserved across SetProperties, which is wrong if
+  // the user overwrote channels 0-2. Detecting this by comparing old vs new
+  // slot values is straightforward but deferred to a follow-up. For now,
+  // callers who rewrite normals should re-call CalculateNormals.
```
elalish: To be fair, vertex normals are not well-defined and CalculateNormals is just one way to generate them. Users might be adjusting them on purpose, rather than turning them into something other than normals.

src/manifold.cpp
```diff
@@ -763,6 +783,10 @@ Manifold Manifold::Refine(int n) const {
   auto pImpl = std::make_shared<Impl>(*leafImpl);
   if (n > 1) {
     pImpl->Refine([n](vec3, vec4, vec4) { return n - 1; }, false, ctx.get());
+    // Subdivided verts get linearly interpolated property values, so any
+    // recorded normals at the standard slot no longer point at the surface.
+    // Caller should re-call CalculateNormals().
```
elalish: Linear interpolation of normals may not be the most accurate, but it's not crazy. I think these are still normals, but just a bit less precise.

test/manifold_test.cpp
```diff
@@ -227,6 +227,144 @@ TEST(Manifold, ErrorPropagationCalculateNormals) {
             Manifold::Error::NonFiniteVertex);
 }
 
+// CalculateNormals(idx) followed by GetMeshGL() (no idx) used to drop
+// the transform-on-export step, returning input-frame data.
+TEST(Manifold, CalculateNormalsRoundTripsThroughGetMeshGL) {
```
elalish: I see a bunch of separate tests in here that don't seem to share anything - might be better as separate `TEST`s? Also, let's try to keep the test names a little more pithy, since we have to type these on the command line when debugging.

test/manifold_test.cpp
```diff
@@ -227,6 +227,144 @@ TEST(Manifold, ErrorPropagationCalculateNormals) {
...
+  }
+  EXPECT_EQ(afterBad, 0);
+
+  // No-arg invocation: defaults to slot 0 and sets hasNormals.
+  MeshGL noIdxMesh = Manifold::Sphere(10.0, 32).CalculateNormals().GetMeshGL();
+  EXPECT_TRUE(noIdxMesh.hasNormals);
+  ASSERT_GE(noIdxMesh.numProp, 6);
+
+  // Roundtrip: getMesh -> ofMesh -> getMesh should preserve hasNormals,
+  // so the second getMesh still applies the transform on export.
```
elalish: This is close, but I think there may still be a broken case in here somewhere. When I was thinking about this before, I convinced myself that the system we currently have where the normals are only updated during GetMesh will inevitably have problems during some combination of applied transforms, booleans, and round-trips. I was thinking it would be safer to transform the now known normal channels during our existing `Transform` updates. This would also allow us to remove the confusing "un-transform normals" step I put into `CalculateNormals`. What do you think?

elalish (commented):
Thanks! Have you been doing any manual testing with ManifoldCAD? I've found that makes it quicker to find surprising normals situations, a la the example in the bug report.

zmerlynn:
Substantially reworked (AI-driven implementation, human reviewed as best I could). Updated description has the full scope. Short version: per-run hasNormals via `runFlags` bit 1, eager-transform per-meshID in `Impl::Transform` and `Compose`, and the cross-mesh seam-vert sign issue from #1719 fixed in `SetNormals` via a sign-flip against `faceNormal_`. Refine / SetProperties / `MeshGL::hasNormals` field / `UpdateNormals` all dropped per your other comments. `MeshGLP::HasNormals(run)` / `Backside(run)` are the new accessors (C and Python bindings included).

Two latent things came up along the way:
- `Backside(run)` used `runFlags[run] == 1` which silently broke once a second bit landed; fixed to `& 1`.
- `AsOriginal` was dropping the normals recording through `InitializeOriginal`; fixed by AND'ing the old Relations onto the new one.

412 tests pass, including the previously-FIXME'd asserts in `Properties.CalculateNormals`. Open to anything you want to push on.

include/manifold/mesh.h
```diff
@@ -235,13 +224,34 @@ struct MeshGLP {
 
   /**
    * Returns true if this triangle run is on the backside compared to the
-   * original mesh, e.g. from a subtraction. In this case vertex normals will
-   * need to be flipped. UpdateNormals() will take care of this.
+   * original mesh, e.g. from a subtraction. In this case vertex normals stored
+   * for the run point opposite to the result surface and consumers reading
+   * them via `runTransform` should multiply by -1.
```
elalish: I think this comment is wrong? It's letting you know that this came from the backside of the original object, but the normals have already been fixed automatically, right?

include/manifold/mesh.h
```diff
@@ -235,13 +224,34 @@ struct MeshGLP {
...
+   * Returns true if the first three extra-property channels (slots 3, 4, 5)
+   * of this run carry world-frame vertex normals (set by
+   * `Manifold::CalculateNormals(0)` and round-tripped via `runFlags` bit 1).
+   * Consumers should treat the slot as normals and skip re-applying
+   * `runTransform` to it.
+   *
+   * For hand-built MeshGL inputs: if multiple runs share propVerts (same
+   * vertex index used by triangles from different runs), all sharing runs
+   * must agree on what slot 3..5 represents. Mixing hasNormals=true and
+   * hasNormals=false across shared propVerts is undefined - the stored
```
elalish: I'm hoping this isn't true either? Part of the purpose of specifying this per run was so these would not have to agree.

src/boolean_result.cpp
```diff
@@ -608,6 +609,15 @@ void CreateProperties(Manifold::Impl& outR, const Manifold::Impl& inP,
     const auto& properties = PQ ? inP.properties_ : inQ.properties_;
     const auto& halfedge = PQ ? inP.halfedge_ : inQ.halfedge_;
 
+    // For Subtract, Q's triangles are flipped in the result, so Q's
+    // world-frame vertex normals (slot 0..2 when hasNormals) need a sign
+    // flip to point outward from the result's solid (into the cavity).
+    // Check is per-source-triangle, not whole input - inQ may be a mixed
+    // Boolean result.
```
elalish, line 616: 👍

src/impl.cpp
```diff
@@ -176,8 +206,13 @@ void Manifold::Impl::InitializeOriginal() {
              [meshID, &triRef](const int tri) {
                triRef[tri] = {meshID, meshID, -1, triRef[tri].coplanarID};
              });
+  // Preserve the AND-across-old-Relations state so AsOriginal keeps the
+  // recording when it builds a fresh Relation. Primitives start with an
+  // empty map, which HasNormals() returns false for.
+  const bool wasHasNormals = HasNormals();
```
elalish: nit: I might call this `hadNormals`...

src/impl.cpp
```diff
@@ -606,6 +641,14 @@ Manifold::Impl Manifold::Impl::Transform(const mat3x4& transform_) const {
   transform(vertNormal_.begin(), vertNormal_.end(), result.vertNormal_.begin(),
             TransformNormals({normalTransform}));
 
+  // Eager-transform property normals per-meshID. Mixed Boolean/Compose
+  // outputs leave HasNormals() false overall but still have hasNormals=true
+  // meshIDs whose stored world-frame values would otherwise go stale.
```
elalish: I think this code would be clearer without this comment. Agree?

src/impl.h
```diff
@@ -53,6 +57,26 @@ struct Manifold::Impl {
   double tolerance_ = -1;
   int numProp_ = 0;
   Error status_ = Error::NoError;
+
+  // True iff every meshID in meshRelation_.meshIDtransform has hasNormals
+  // set (i.e., slot 0..2 of properties_ holds world-frame normals across the
+  // whole impl). Set by CalculateNormals at idx 0; AND'd across inputs by
+  // Boolean and Compose; preserved by Transform/Refine.
+  bool HasNormals() const {
```
elalish: This function is surprising to me - I would have expected it to return true if any had normals. Just a nit, but worth checking if this is actually used as !HasNormals, in which case we might refactor slightly.

src/impl.h
```diff
@@ -573,13 +616,25 @@ inline MeshGLP<Precision, I> GetMeshGLImpl(const manifold::Manifold::Impl& impl,
           out.vertProperties.push_back(impl.properties_[prop * numProp + p]);
         }
 
+        // Normalize the requested normal slot. For runs that already carry
+        // world-frame normals (hasNormals bit), just normalize; for legacy
+        // callers asking to interpret a slot as normals on a run without
+        // hasNormals, apply the per-run inverse-frame transform first.
```
elalish, line 623: Maybe put a TODO to remove this logic once we remove the old API.

src/impl.h
```diff
@@ -573,13 +616,25 @@ inline MeshGLP<Precision, I> GetMeshGLImpl(const manifold::Manifold::Impl& impl,
...
-          normal = SafeNormalize(runNormalTransform[run] * normal);
+          const bool runHasN = !isOriginal && (out.runFlags[run] & 2) != 0;
+          if (!isOriginal && !runHasN) {
+            const Precision* m = out.runTransform.data() + 12 * run;
+            const mat3x4 t({m[0], m[1], m[2]}, {m[3], m[4], m[5]},
+                           {m[6], m[7], m[8]}, {m[9], m[10], m[11]});
+            normal = NormalTransform(t) *
+                     ((out.runFlags[run] & 1) ? -1.0 : 1.0) * normal;
+          }
+          normal = SafeNormalize(normal);
```
elalish, line 641: Is there any need for this normalization? Or was this not applied during the eager transformations? I think it should be applied during the eager transformations so that interpolation of normals works as expected, e.g. at the edges of booleans.

src/manifold.cpp
```diff
@@ -242,17 +242,20 @@ Manifold::Manifold(const MeshGL64& meshGL64)
...
  * automatically update these normals according to the applied transforms and
  * front/back side. normalIdx + 3 must be <= numProp, and all original meshes
- * must use the same channels for their normals. Default is -1, meaning no
- * normals. If normals are selected, the runTransform matrices will be removed
- * from the output, to avoid them being double-applied when round-tripped.
+ * must use the same channels for their normals. Default is -1: if
+ * `CalculateNormals()` recorded normals at the standard slot, that slot is
+ * used automatically; otherwise no normals are applied. If normals are
+ * selected, the runTransform matrices will be removed from the output, to
+ * avoid them being double-applied when round-tripped.
```
elalish, line 249: Let's put in some deprecation notices about how this input is going away.

src/smoothing.cpp
```diff
@@ -501,27 +505,41 @@ void Manifold::Impl::SetNormals(int normalIdx, double minSharpAngle) {
                halfedge_.SetProp(i, -1);
              });
 
+  // Cached per-meshID inverse-normal-transform for the legacy non-zero
+  // normalIdx path. Lazily populated on first lookup; reused across all
+  // verts in the loop below.
```
elalish, line 510: Perhaps another TODO to remove.

src/smoothing.cpp
```diff
@@ -607,17 +625,25 @@ void Manifold::Impl::SetNormals(int normalIdx, double minSharpAngle) {
           }
           groups.push_back(normals.size() - 1);
           if (std::isfinite(next.normalizedEdge.x)) {
+            // Boolean subtractee triangles keep their original winding but
+            // have faceNormal_ negated, so the edge-cross points opposite to
+            // faceNormal_ there - flip to keep the accumulation outward-from-
+            // result.
```
elalish, line 633: Ah, this is what I had missed - thank you!

test/manifold_test.cpp
elalish: I think this comment still applies.

bindings/wasm/bindings.js
elalish, line 1: Can we get the backside and hasNormals functions bound here too?

elalish (approved):
Okay, this looks totally excellent! A bunch of comments, but they're all minor and quick.

test/manifold_test.cpp
zmerlynn: Done in `ce0e18d8` - split into `NormalsCavity`, `NormalsRotateBeforeCalc`, `NormalsRotateAfterCalc`, `NormalsAutoSubstitute`, `NormalsRoundTrip`, `NormalsRefinePreserved`, `NormalsSmoothByNormalsNoArg`, `NormalsNonStandardSlotNotRecorded`. Shared `CountSphereNormalAlignment` helper takes the boilerplate.

include/manifold/mesh.h
zmerlynn: Right - rewritten in `ce0e18d8` to "Informational only - the framework already orients stored normals so the standard `getMesh()` flow returns world-frame values regardless of this bit."

include/manifold/mesh.h
zmerlynn: You're right about intent and I owe a clearer answer than "documented limitation." I tried to demonstrate the conflict with a test - `Manifold.NormalsSharedPropVertMixedFlags` in `ce0e18d8`, which is intentionally failing in CI right now (you'll see one red test). It hand-builds a tetrahedron MeshGL with Run 0 (`hasNormals=true`, +Z is the vertex normal) and Run 1 (`hasNormals=false`, +Z is the user color), sharing all 4 propVerts, then `Rotate(90,X)`. Only one stored value per propVert can survive: my eager-transform picks Run 0's interpretation and Run 1's color gets corrupted to (0, -1, 0). Master's deferred-transform contract avoided this because `properties_` was never mutated - per-run interpretation happened on the consumer side after `runTransform` was applied.

Working on a follow-up commit that splits propVerts in the `Manifold(MeshGL)` ctor when triangles from runs with differing `hasNormals` flags share the same propVert index. Each meshID then gets its own copy and the eager-transform contract works per-meshID. The failing test should flip to green when that lands.

src/boolean_result.cpp
zmerlynn: Thanks!

src/impl.cpp
zmerlynn: Done.

src/impl.cpp
zmerlynn: Done - dropped the eager-transform comment block; the helper call is self-documenting.

src/impl.h
zmerlynn: Renamed `Impl::HasNormals()` -> `Impl::AllHaveNormals()` in `ce0e18d8`. Per-run `MeshGLP::HasNormals(run)` stays as-is - different class, different semantics. The "All" prefix makes the AND explicit at the call site.

src/impl.h
zmerlynn: Done in `ce0e18d8`: TODO notes "collapse the !runHasN branch into a no-op once the explicit-normalIdx parameter on GetMeshGL is removed."

src/impl.h
zmerlynn: Right - moved `SafeNormalize` inside `EagerTransformPropNormals` so non-orthogonal transforms and the CreateProperties barycentric path leave unit values in `properties_`. Export-time SafeNormalize is now defensive.

src/manifold.cpp
zmerlynn: Added inside the `@param normalIdx` doc on both `GetMeshGL` and `GetMeshGL64`. Kept it out of a function-level `@deprecated` tag since the no-arg form isn't deprecated.

src/smoothing.cpp
zmerlynn: Done.

src/smoothing.cpp
zmerlynn: Took me a while to track that one down too.

bindings/wasm/bindings.js
zmerlynn: Added in `ce0e18d8`: `Mesh.backside(run)` and `Mesh.hasNormals(run)` in `bindings.js`, plus the `runFlags` field doc and method docs in `manifold-encapsulated-types.d.ts`. Mirrors the C and Python accessor shapes.

include/manifold/mesh.h
zmerlynn: Fixed. The MeshGL ctor scans triangles and, when a propVert is referenced by two runs with differing hasNormals flags, appends a duplicate - first-seen camp keeps the original index, opposite camp points at the new copy. DedupePropVerts already skips cross-meshID pairs and EagerTransformPropNormals is gated per-triangle on TriHasNormals, so each camp's slot 0..2 transforms independently. `NormalsSharedPropVertMixedFlags` passes; doc paragraph re-added to `MeshGLP::HasNormals`.

One wrinkle worth flagging, specific to the case the split actually fires (mixed-flag runs sharing a propVert): the duplicate carries the same values as the original right after the ctor. `Manifold(gl).GetMeshGL()` with no operations between produces an extra vertProperty row paired to its twin via mergeFromVert/mergeToVert, even though the twin doesn't visibly need to exist until a Transform happens. A caller diffing input vs output verbatim would see the extra row before any operation justifies it. Uniform-flag inputs and inputs without shared propVerts round-trip unchanged.

Considered deferring the scan to a Transform/Compose preamble to keep identity round-trip byte-stable. Lazy works cleanly for `Impl::Transform` (mutates the freshly-built result impl), but Compose materializes deferred leaf transforms by calling `EagerTransformPropNormals` on a shared `node->pImpl_` - lazy-in-Compose then needs either an in-place mutation of the shared source (hidden side-effect, race-prone) or a copy-then-split per source (memory overhead). Neither felt cleaner than the wrinkle. Open to revisiting.

elalish: Oh, I'm sorry, I think I misread your original comment - I thought you meant different runs couldn't have different hasNormals values. I don't think a given propVert index should ever be shared between different runs. I guess you're now automatically fixing that for folks whose input is weird. My one concern is that we're adding yet more processing to MeshGL import that isn't usually needed, while users are already complaining that our import is slower than our booleans. It's also a fair amount of code to maintain. 

Let's revert this change. The rest looks great and ready to merge.

zmerlynn: Reverted in fcfaf0fd, with 2f41a281 re-adding a caveat to `MeshGLP::HasNormals` (and the WASM equivalent): hasNormals is per-run, but sharing propVerts across disagreeing runs is undefined - the hasNormals=true camp's interpretation wins on Transform. Test (`NormalsSharedPropVertMixedFlagsUndefined`) repurposed to pin that behaviour. Thanks for the call.

elalish (approved):

zmerlynn:
Is there an umbrella issue tracking "remove on next release" cleanup items? This PR leaves a couple of TODOs (the legacy `normalIdx` path in `SetNormals` / `GetMeshGLImpl`) and a deprecation on the explicit `normalIdx` parameter to `GetMeshGL`/`GetMeshGL64`. Happy to file one if not - a checkbox-list tracking issue tends to work well for this kind of thing.

elalish:
Not yet, go for it. I'd say it should actually be a v4.0 list, listing major new features, but especially breaking changes and these corresponding cleanups.

PR #1720, opened 2026-05-16
Fix seam-vert normals in SetNormals (fixes #1719)

### LINKED ISSUE #1719: Bad normals at seam verts where two surfaces meet just under `minSharpAngle`

zmerlynn:
`Properties.CalculateNormals`'s commented-out FIXME assertions caught 17 verts on master where `(sphere - sphere.Translate(10)).CalculateNormals(0).GetMeshGL64(0)` produced inward-pointing normals. Two coupled fixes in `SetNormals`:

1. **Split on meshID mismatch as a sharp-edge trigger.** Boolean seam verts at the intersection of two operands need the multi-normal path so each meshID gets its own group with its own inverse-transform applied. The existing dihedral / flat-face triggers don't catch this for curved-curved seams under the ~60° default.

2. **Use `faceNormal_` accumulation for cross-mesh seam verts** (tracked via a new `vertIsCrossMesh` flag). The default cross-product pseudo-normal accumulation produces an inward-pointing result for CCW fan walks; single-meshID multi-normal verts (cube corners, etc.) tolerate this end-to-end through the smoothing pipeline, but cross-mesh seam groups need an outward direction so the per-meshID inverse-transform produces values that round-trip correctly on export.

Enables the previously-FIXMEd assertions in `Properties.CalculateNormals`. Adds five new regression tests covering Union, Intersection, Mirror'd operand, 3-way `BatchBoolean`, and SmoothByNormals-after-Boolean. All 412 tests pass.

### DIFF
--- a/src/smoothing.cpp
+++ b/src/smoothing.cpp
@@ -465,6 +465,12 @@ void Manifold::Impl::SetNormals(int normalIdx, double minSharpAngle) {
   Vec<bool> triIsFlatFace = FlatFaces();
   Vec<int> vertFlatFace = VertFlatFace(triIsFlatFace);
   Vec<int> vertNumSharp(NumVert(), 0);
+  // Verts whose incident faces span at least two meshIDs need different
+  // per-group treatment than ordinary sharp-edge verts (e.g. cube corners),
+  // because the cross-product pseudo-normal accumulation produces an inward
+  // result that the rest of the smoothing pipeline tolerates for
+  // single-meshID groups but not for multi-meshID seam groups.
+  Vec<bool> vertIsCrossMesh(NumVert(), false);
   for (size_t e = 0; e < halfedge_.size(); ++e) {
     if (!halfedge_.IsForward(e)) continue;
     const int pair = halfedge_.Pair(e);
@@ -472,7 +478,13 @@ void Manifold::Impl::SetNormals(int normalIdx, double minSharpAngle) {
     const int tri2 = pair / 3;
     const double dihedral =
         degrees(AngleBetween(faceNormal_[tri1], faceNormal_[tri2]));
-    if (dihedral > minSharpAngle) {
+    const bool meshIDMismatch =
+        meshRelation_.triRef[tri1].meshID != meshRelation_.triRef[tri2].meshID;
+    if (meshIDMismatch) {
+      vertIsCrossMesh[halfedge_.Start(e)] = true;
+      vertIsCrossMesh[halfedge_.End(e)] = true;
+    }
+    if (dihedral > minSharpAngle || meshIDMismatch) {
       ++vertNumSharp[halfedge_.Start(e)];
       ++vertNumSharp[halfedge_.End(e)];
     } else {
@@ -555,6 +567,8 @@ void Manifold::Impl::SetNormals(int normalIdx, double minSharpAngle) {
       const double dihedral =
           degrees(AngleBetween(faceNormal_[face], faceNormal_[prevFace]));
       if (dihedral > minSharpAngle ||
+          meshRelation_.triRef[face].meshID !=
+              meshRelation_.triRef[prevFace].meshID ||
           triIsFlatFace[face] != triIsFlatFace[prevFace] ||
           (triIsFlatFace[face] && triIsFlatFace[prevFace] &&
            !meshRelation_.triRef[face].SameFace(
@@ -598,6 +612,8 @@ void Manifold::Impl::SetNormals(int normalIdx, double minSharpAngle) {
           const double dihedral = degrees(
               AngleBetween(faceNormal_[here.face], faceNormal_[next.face]));
           if (dihedral > minSharpAngle ||
+              meshRelation_.triRef[here.face].meshID !=
+                  meshRelation_.triRef[next.face].meshID ||
               triIsFlatFace[here.face] != triIsFlatFace[next.face] ||
               (triIsFlatFace[here.face] && triIsFlatFace[next.face] &&
                !meshRelation_.triRef[here.face].SameFace(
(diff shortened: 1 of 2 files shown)


src/smoothing.cpp
```diff
@@ -607,10 +623,20 @@ void Manifold::Impl::SetNormals(int normalIdx, double minSharpAngle) {
...
           if (std::isfinite(next.normalizedEdge.x)) {
-            normals.back() +=
-                SafeNormalize(
-                    la::cross(next.normalizedEdge, here.normalizedEdge)) *
-                AngleBetween(here.normalizedEdge, next.normalizedEdge);
+            if (vertIsCrossMesh[vert]) {
+              // Use face_normal directly for seam verts where the multi-
+              // normal split groups faces from different meshIDs. The
+              // cross-product pseudo-normal produces inward-pointing
+              // results for partial-fan groups across a Boolean seam.
```
elalish, line 630: I think we might be able to simplify this logic based on a change I suggested in your other PR. Sound reasonable?

test/properties_test.cpp
```diff
@@ -144,9 +144,8 @@ TEST(Properties, CalculateNormals) {
       norm2[j] = out2.vertProperties[np * v + 3 + j];
       ASSERT_FLOAT_EQ(pos[j], pos2[j]);
     }
-    // FIXME
-    // ASSERT_GT(dot(pos2, norm2), 0);
-    // ASSERT_GT(dot(pos, norm), 0);
+    if (dot(pos, norm) <= 0) ++numBad;
+    if (dot(pos2, norm2) <= 0) ++numBad2;
```
elalish, line 148: Your PR comment refers to some newly added tests that I don't see here. Not that I'm sure more tests are really needed?

elalish (commented):
Good find.

src/smoothing.cpp
zmerlynn: Agreed - if `Transform` updates `faceNormal_` eagerly so it stays in world frame across meshIDs, the per-group `getTransform(meshIds[i])` at line 620 goes away, and with it the whole `vertIsCrossMesh` branch I added here. The meshID-mismatch sharp-edge trigger also stops being load-bearing (dihedral is what should drive splits; meshID was only a proxy because frames disagreed).

Best ordering: land #1718 first, then this either drops to just enabling the FIXME asserts, or becomes moot. Happy to close this and re-open the simplified version on top of #1718 if that is cleaner.

test/properties_test.cpp
zmerlynn: Stale PR body, sorry - I drafted 5 new tests (Union, Intersection, Mirror, 3-way BatchBoolean, Smooth-after-Boolean) but dropped them after verifying that none of them actually caught the bug pre-fix; only the existing FIXME asserts here did. Agreed the additions are not needed. Updating the body.

src/smoothing.cpp
elalish: Sounds good, I'm not picky. We can come back to this once the other is merged.

elalish: Is this needed anymore, or did this make it into #1718 already?

zmerlynn:
Nope, obviated by #1718

elalish:
hmm, it does look mostly fixed, but there's still a small problem: look around the outside of the boolean seam on this [example](https://manifoldcad.org/#Incorrect%20normals%3Ccode%3Eimport%20%7B%20Manifold%2C%20setMaterial%20%7D%20from%20'manifold-3d%2FmanifoldCAD'%3B%0Aconst%20%7B%20union%2C%20intersection%2C%20difference%2C%20sphere%20%7D%20%3D%20Manifold%3B%0A%0A%2F%2F%20Difference%20of%20two%20spheres%0Aconst%20model%20%3D%20difference(sphere(10)%2C%20sphere(10).translate(%5B10%2C10%2C10%5D))%3B%0A%0A%2F%2F%20Convert%20to%20mesh%20and%20back%0Aconst%20model2%20%3D%20Manifold.ofMesh(model.getMesh())%3B%0A%0Aconst%20nodes%20%3D%20%5B%0A%20%20%2F%2F%20model%20has%20incorrect%20normals%0A%20%20setMaterial(model.calculateNormals()%2C%20%7B%20baseColorFactor%3A%20%5B1%2C%201%2C%201%5D%2C%20attributes%3A%20%5B%22COLOR_0%22%5D%20%7D).translate(%5B-15%2C0%2C0%5D)%2C%0A%20%20%2F%2F%20converting%20to%20mesh%20and%20back%20gives%20correct%20normals%0A%20%20setMaterial(model2.calculateNormals()%2C%20%7B%20baseColorFactor%3A%20%5B1%2C%201%2C%201%5D%2C%20attributes%3A%20%5B%22COLOR_0%22%5D%20%7D).translate(%5B15%2C0%2C0%5D)%2C%0A%5D%3B%0A%0Aexport%20default%20nodes%3B). 
image

elalish:
Ah, I think I see - that dir flip you added to SetNormals shouldn't be needed; it's a symptom of a bug further up in SetNormals, but the Boolean is actually generating the right faceNormals, as verified by `MatchesTriNormals`. I should have a fix in soon, but not tonight.

zmerlynn:
I dug into this more, and I think your read was right: the Boolean `faceNormal_`s look correct, and the remaining issue seems to be in `SetNormals()`.

My current theory is that Boolean seam edges need to be treated as normal-split boundaries even when their geometric dihedral is below `minSharpAngle`. In the sphere-minus-sphere case, some seam-adjacent faces come from different `meshID`s but still get smoothed together, so the exported vertex normal becomes an average of the outer sphere and subtractee cavity directions.

The fix I’d try is to make `SetNormals()` split normal groups on `meshID` changes, not just on dihedral / flat-face boundaries, and use the existing `faceNormal_` direction when accumulating each group.

Does that direction sound reasonable? I can open a PR if you’d like.

elalish:
Thanks, but I'm on it. This is less a bug and more a choice between different smoothing behaviors. I think I tried to get a little too fancy.

PR #1723, opened 2026-05-18
Thread ExecutionContext into MeshGL ingest via ExecutionContext::FromMeshGL

zmerlynn:
Adds `ExecutionContext::FromMeshGL(MeshGL)` / `FromMeshGL(MeshGL64)` -
eager, ctx-aware analog of `Manifold(MeshGL)`. First installment of
the static-factories-on-ctx direction from
#971 (https://github.com/elalish/manifold/issues/971#issuecomment-4472693075);
MeshGL ingest is stephomi's headline perf concern in that thread.

`Impl(MeshGLP, ctx)` threads cancel + `donePhases` through the 7
heavy ingest steps. Entry-time Cancel beats empty/malformed input;
past the gate, validation errors beat racing Cancel. New
`Manifold::FromImpl(shared_ptr<Impl>)` (vein of `Invalid()` /
`PropagateStatus()`) lets factories wrap Impls without friending.

Follow-ups: bindings (Python/WASM/C), other static factories (Hull,
LevelSet, Smooth) - separate PRs per #971.

## Test plan

- [ ] CI green
- [ ] `./test/manifold_test` (430/430 locally; 10 new tests)

### DIFF
--- a/src/impl.h
+++ b/src/impl.h
@@ -461,10 +461,8 @@ Manifold::Impl::Impl(const MeshGLP<Precision, I>& meshGL,
     }
   }
 
-  // Phase boundaries: ADVANCE_PHASE_OR_RETURN (from execution_impl.h)
-  // does cancel-check + MakeEmpty + return on cancel, then credits one
-  // phase on the continue path. Count of calls below must equal
-  // kPhasesPerFromMesh.
+  // Phase boundaries; count of ADVANCE_PHASE_OR_RETURN calls below must
+  // equal kPhasesPerFromMesh.
   CreateHalfedges(triProp, triVert);
   if (!IsManifold()) {
     MakeEmpty(Error::NotManifold);
--- a/src/impl.h
+++ b/src/impl.h
@@ -461,19 +461,16 @@ Manifold::Impl::Impl(const MeshGLP<Precision, I>& meshGL,
     }
   }
 
-  // Phase boundaries: cancel-check then credit the phase only on
-  // continue. Count of fetch_add sites below must equal
+  // Phase boundaries: ADVANCE_PHASE_OR_RETURN (from execution_impl.h)
+  // does cancel-check + MakeEmpty + return on cancel, then credits one
+  // phase on the continue path. Count of calls below must equal
   // kPhasesPerFromMesh.
   CreateHalfedges(triProp, triVert);
   if (!IsManifold()) {
     MakeEmpty(Error::NotManifold);
     return;
   }
-  if (IsCancelled(ctx)) {
-    MakeEmpty(Error::Cancelled);
-    return;
-  }
-  if (ctx) ctx->donePhases.fetch_add(1, std::memory_order_relaxed);
+  ADVANCE_PHASE_OR_RETURN(ctx);
 
   CalculateBBox();
   SetEpsilon(-1, std::is_same<Precision, float>::value);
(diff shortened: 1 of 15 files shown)


src/impl.h
```diff
@@ -451,25 +461,66 @@ Manifold::Impl::Impl(const MeshGLP<Precision, I>& meshGL) {
...
 
+  // Phase boundaries: cancel-check then credit the phase only on
+  // continue. Count of fetch_add sites below must equal
+  // kPhasesPerFromMesh.
   CreateHalfedges(triProp, triVert);
   if (!IsManifold()) {
     MakeEmpty(Error::NotManifold);
     return;
   }
+  if (IsCancelled(ctx)) {
```
pca006132: thinking about it, while macro is bad, it seems legit to use them for this kind of repetitive patterns

pca006132 (approved):
LGTM, I'm thinking maybe we should just use macro to deal with the repeated early returns. @elalish what do you think?

zmerlynn:
Demo of what this could look like, happy to revert if you'd rather keep #1723 scope tight. I come from a vanilla C background, so these read just fine to me, but I know they smell to other people.

- `RETURN_IF_CANCELLED(ctx)` / `RETURN_IF_CANCELLED_VAL(ctx, val)` in `src/execution_impl.h`, applied to the 44 existing one-line cancel sites (`boolean3.cpp`, `boolean_result.cpp`, `csg_tree.cpp`, `face_op.cpp`, `minkowski.cpp`, `parallel.h`, `sort.cpp`).
- `ADVANCE_PHASE_OR_RETURN(ctx)` for `Manifold::Impl` methods with a phase counter (cancel-check + `MakeEmpty(Cancelled)` + `donePhases.fetch_add(1)`). Applied to the seven phase boundaries in the MeshGL ctor; doc note says call count must match `kPhasesPer*`.
- Split rather than variadic to stay -Wpedantic clean for downstream consumers.

Six multi-line cancel sites left inline (quickhull x3, smoothing.cpp:1100, csg_tree.cpp:735, boolean_result.cpp:758) - they do more than `return`.

pca006132:
I also worked with old C libs with macros flying everywhere, so this looks fine to me :)

src/impl.h
elalish: Yeah, that seems reasonable.

src/impl.h
```diff
@@ -451,25 +461,37 @@ Manifold::Impl::Impl(const MeshGLP<Precision, I>& meshGL) {
...
   }
 
+  // Phase boundaries; count of ADVANCE_PHASE_OR_RETURN calls below must
+  // equal kPhasesPerFromMesh.
   CreateHalfedges(triProp, triVert);
   if (!IsManifold()) {
     MakeEmpty(Error::NotManifold);
     return;
   }
+  ADVANCE_PHASE_OR_RETURN(ctx);
```
elalish, line 471: I like this macro since it reduced repetition, but I think the others can be removed, as they didn't really reduce the callsite code or make it more readable.

elalish (changes requested):
Other than the one revert, this LGTM.

zmerlynn:
Agreed - reverted the bulk migration and dropped both `RETURN_IF_CANCELLED` forms in 3d1417ac / b230eb42 / 9ee890ed. Only `ADVANCE_PHASE_OR_RETURN` remains, applied at the 7 phase boundaries in the MeshGL ctor. It'll plug in cleanly when the other ctors (Hull, Smoothing, LevelSet) grow per-phase counters too - I'll do that in a follow-on.

elalish (approved):
Thanks!

PR #1724, opened 2026-05-19
Fix CalculateNormals

elalish:
I noticed some ugliness near the edges of Booleans when using CalculateNormals (see https://github.com/elalish/manifold/pull/1720#issuecomment-4472810306), which ultimately stemmed from my trying to do too much fancy stuff. CalculateTangents(sharpenedEdges) identifies flat faces as sets of three or more neighboring triangles with the same normal and uses this normal instead of the pseudonormal. I tried to do the same thing with CalculateNormals and SmoothByNormals, but it's a poor heuristic because Boolean ops often cut triangles into polygons, which then get stuck as flat faces, breaking with the curving surface around them.

I fixed this by simplifying the logic considerably, which should make it more robust and less surprising. It does loose some of the cool features I was going for, but I think much of that can now come by using the missing normals approach. I also changed the default min sharp angle from 60 degrees to 52.5 (halfway between 60 and 45), since it's annoying to have it right on the edge for relatively common angles.

I still have more work to do in this space, especially around better handling of quads, but first I need better decimation.

### DIFF
--- a/include/manifold/manifold.h
+++ b/include/manifold/manifold.h
@@ -245,7 +245,8 @@ class Manifold {
       int numProp,
       std::function<void(double*, vec3, const double*)> propFunc) const;
   Manifold CalculateCurvature(int gaussianIdx, int meanIdx) const;
-  Manifold CalculateNormals(int normalIdx = 0, double minSharpAngle = 60) const;
+  Manifold CalculateNormals(int normalIdx = 0,
+                            double minSharpAngle = 52.5) const;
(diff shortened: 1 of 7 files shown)


mmiscool:
Would this have any impact on doing fillets using a chamfer?

elalish:
Yes, this and several other of my recent PRs are in pursuit of this goal, but I haven't quite attained it yet. Stay tuned.

PR #1726, opened 2026-05-19
Fix MSVC strict conversion warnings

### LINKED ISSUE #1189: Type mismatch: MSVC: warning C4267: 'argument': conversion from 'size_t' to 'int', possible loss of data
Dear manifold team,
we are trying to use manifold within our codebase and usually compile third-party code on MSVC 2019/2022 with `/W0 /WX`, meaning disable all warnings and make every warning an error. 
Interestingly, Manifold produces warnings if _with_ all warnings disabled (with TBB disabled):

```
2>cl /c /I"E:\playground\manifold-3.0.1\include" /I"E:\playground\manifold-3.0.1\build\include" /Zi /W1 /WX- /diagnostics:column /O2 /Ob1 /D _WINDLL /D _MBCS /D WIN32 /D _WINDOWS /D NDEBUG /D NOMINMAX /D "MANIFOLD_PAR=-1" /D "CMAKE_INTDIR=\"RelWithDebInfo\"" /D manifold_EXPORTS /Gm- /EHsc /MD /GS /fp:precise /Zc:wchar_t /Zc:forScope /Zc:inline /GR /std:c++17 /Fo"manifold.dir\RelWithDebInfo\\" /Fd"manifold.dir\RelWithDebInfo\vc143.pdb" /external:W1 /Gd /TP /errorReport:prompt  /bigobj "E:\playground\manifold-3.0.1\src\boolean3.cpp" "E:\playground\manifold-3.0.1\src\boolean_result.cpp" "E:\playground\manifold-3.0.1\src\constructors.cpp" "E:\playground\manifold-3.0.1\src\csg_tree.cpp" "E:\playground\manifold-3.0.1\src\edge_op.cpp" "E:\playground\manifold-3.0.1\src\face_op.cpp" "E:\playground\manifold-3.0.1\src\impl.cpp" "E:\playground\manifold-3.0.1\src\manifold.cpp" "E:\playground\manifold-3.0.1\src\polygon.cpp" "E:\playground\manifold-3.0.1\src\properties.cpp" "E:\playground\manifold-3.0.1\src\quickhull.cpp" "E:\playground\manifold-3.0.1\src\sdf.cpp" "E:\playground\manifold-3.0.1\src\smoothing.cpp" "E:\playground\manifold-3.0.1\src\sort.cpp" "E:\playground\manifold-3.0.1\src\subdivision.cpp"
2>boolean3.cpp
2>C:\Program Files\Microsoft Visual Studio\2022\Professional\VC\Tools\MSVC\14.43.34808\include\algorithm(372,15): warning C4267: 'argument': conversion from 'size_t' to 'int', possible loss of data
2>(compiling source file '../../src/boolean3.cpp')
2>    C:\Program Files\Microsoft Visual Studio\2022\Professional\VC\Tools\MSVC\14.43.34808\include\algorithm(372,15):
2>    the template instantiation context (the oldest one first) is
2>        E:\playground\manifold-3.0.1\src\collider.h(285,5):
2>        see reference to function template instantiation 'void manifold::for_each_n<manifold::CountingIterator<size_t>,manifold::collider_internal::CreateRadixTree>(manifold::ExecutionPolicy,Iter,size_t,F)' being compiled
2>        with
2>        [
2>            Iter=manifold::CountingIterator<size_t>,
2>            F=manifold::collider_internal::CreateRadixTree
2>        ]
2>        E:\playground\manifold-3.0.1\src\parallel.h(399,3):
2>        see reference to function template instantiation 'void manifold::for_each<Iter,F>(manifold::ExecutionPolicy,Iter,Iter,F)' being compiled
2>        with
2>        [
2>            Iter=manifold::CountingIterator<size_t>,
2>            F=manifold::collider_internal::CreateRadixTree
2>        ]
2>        E:\playground\manifold-3.0.1\src\parallel.h(389,8):
2>        see reference to function template instantiation '_Fn std::for_each<Iter,F>(_InIt,_InIt,_Fn)' being compiled
2>        with
2>        [
2>            _Fn=manifold::collider_internal::CreateRadixTree,
2>            Iter=manifold::CountingIterator<size_t>,
2>            F=manifold::collider_internal::CreateRadixTree,
2>            _InIt=manifold::CountingIterator<size_t>
2>        ]
```

The problem seems to be `Vec<int>` which are then dereferenced in the for_each and used with size_t's.

You can find the build log without `/WX` here:
[Manifold_build.txt](https://github.com/user-attachments/files/19226226/Manifold_build.txt)

zmerlynn:
Fixes #1189

## Summary

This enables MSVC warnings-as-errors for strict builds and fixes the C4267 size/sign conversion warnings reported when building with MSVC.

The main change is to make `countAt()` preserve the integral type it is given. This keeps existing int-indexed loops as `int` ranges when they use `countAt(0)`, while allowing genuinely size-indexed loops to opt into `size_t` with `countAt(size_t{0})` / `countAt(0_uz)`.

## Details

- Add `/WX` for MSVC when `MANIFOLD_STRICT` is enabled.
- Make `countAt(T)` return `CountingIterator<T>` instead of always producing a `size_t` iterator.
- Fix call sites to choose the intended index type explicitly.
- Avoid adding a mixed `CountingIterator<T>, CountingIterator<U>` `for_each` overload, so mixed begin/end types do not silently create extra template instantiations.
- Add explicit casts where the surrounding data structures are already int-indexed.

## Testing

- `cmake --build build --target manifold_test -j2`
- Fork CI Windows matrix passed: https://github.com/zmerlynn/manifold/actions/runs/26119491509

### DIFF
--- a/extras/minimize_testcase.cpp
+++ b/extras/minimize_testcase.cpp
@@ -89,7 +89,7 @@ bool safeToRemove(const Polygons& polys, size_t i, size_t j, double precision) {
 std::pair<int, int> findIndex(const Polygons& polys, size_t i) {
   size_t outer = 0;
   while (i >= polys[outer].size()) i -= polys[outer++].size();
-  return std::make_pair(outer, i);
+  return std::make_pair(static_cast<int>(outer), static_cast<int>(i));
 }
 
 void Dump(const Polygons& polys, const double epsilon) {
--- a/extras/minimize_testcase.cpp
+++ b/extras/minimize_testcase.cpp
@@ -86,10 +86,10 @@ bool safeToRemove(const Polygons& polys, size_t i, size_t j, double precision) {
   return true;
 }
 
-std::pair<int, int> findIndex(const Polygons& polys, size_t i) {
+std::pair<size_t, size_t> findIndex(const Polygons& polys, size_t i) {
   size_t outer = 0;
   while (i >= polys[outer].size()) i -= polys[outer++].size();
-  return std::make_pair(static_cast<int>(outer), static_cast<int>(i));
+  return std::make_pair(outer, i);
 }
 
 void Dump(const Polygons& polys, const double epsilon) {
@@ -117,9 +117,9 @@ void DumpTriangulation(const Polygons& polys, double precision) {
   std::vector<ivec3> result = Triangulate(polys);
   ManifoldParams() = oldParams;
   for (const ivec3& tri : result) {
-    std::pair<int, int> xid = findIndex(polys, tri.x);
-    std::pair<int, int> yid = findIndex(polys, tri.y);
-    std::pair<int, int> zid = findIndex(polys, tri.z);
+    std::pair<size_t, size_t> xid = findIndex(polys, tri.x);
+    std::pair<size_t, size_t> yid = findIndex(polys, tri.y);
+    std::pair<size_t, size_t> zid = findIndex(polys, tri.z);
     vec2 x = polys[xid.first][xid.second];
     vec2 y = polys[yid.first][yid.second];
     vec2 z = polys[zid.first][zid.second];
--- a/src/boolean3.cpp
+++ b/src/boolean3.cpp
@@ -387,7 +387,7 @@ Intersections Intersect12_(const Manifold::Impl& inP, const Manifold::Impl& inQ,
   sequence(i12.begin(), i12.end());
 
   int index = forward ? 0 : 1;
-  stable_sort(i12.begin(), i12.end(), [&](size_t a, size_t b) {
+  stable_sort(i12.begin(), i12.end(), [&](auto a, auto b) {
     return p1q2[a][index] < p1q2[b][index] ||
            (p1q2[a][index] == p1q2[b][index] &&
             p1q2[a][1 - index] < p1q2[b][1 - index]);
@@ -420,7 +420,7 @@ Vec<int> Winding03_(const Manifold::Impl& inP, const Manifold::Impl& inQ,
 
   // Invariant: every ctx-passing parallel op is followed by IsCancelled to
   // keep partial output from feeding unconditional downstream consumers.
-  DisjointSets uA(static_cast<uint32_t>(a.vertPos_.size()));
+  DisjointSets uA(a.vertPos_.size());
   for_each(autoPolicy(a.halfedge_.size()), countAt(0),
            countAt(static_cast<int>(a.halfedge_.size())), ctx, [&](int edge) {
              const int start = a.halfedge_.Start(edge);
@@ -437,26 +437,26 @@ Vec<int> Winding03_(const Manifold::Impl& inP, const Manifold::Impl& inQ,
   if (IsCancelled(ctx)) return Vec<int>{};
 
   // find components, the hope is the number of components should be small
-  std::unordered_set<int> components;
+  std::unordered_set<size_t> components;
 #if (MANIFOLD_PAR == 1)
   if (a.vertPos_.size() > 1e5) {
-    tbb::combinable<std::unordered_set<int>> componentsShared;
-    for_each(autoPolicy(a.vertPos_.size()), countAt(0),
-             countAt(static_cast<int>(a.vertPos_.size())), ctx,
-             [&](int v) { componentsShared.local().insert(uA.find(v)); });
-    componentsShared.combine_each([&](const std::unordered_set<int>& data) {
+    tbb::combinable<std::unordered_set<size_t>> componentsShared;
+    for_each(autoPolicy(a.vertPos_.size()), countAt(0_uz),
+             countAt(a.vertPos_.size()), ctx,
+             [&](size_t v) { componentsShared.local().insert(uA.find(v)); });
+    componentsShared.combine_each([&](const std::unordered_set<size_t>& data) {
       components.insert(data.begin(), data.end());
     });
   } else
 #endif
   {
     for (size_t v = 0; v < a.vertPos_.size(); v++)
-      components.insert(uA.find(static_cast<uint32_t>(v)));
+      components.insert(uA.find(v));
   }
   if (IsCancelled(ctx)) return Vec<int>{};
   Vec<int> verts;
   verts.reserve(components.size());
-  for (int c : components) verts.push_back(c);
+  for (size_t c : components) verts.push_back(static_cast<int>(c));
 
   Vec<int> w03(a.NumVert(), 0);
   Kernel02<expandP, forward> k02{a, b};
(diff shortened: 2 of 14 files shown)


src/subdivision.cpp
```diff
@@ -116,7 +116,7 @@ class Partition {
     const int offset = interiorOffset - newVerts.size();
     size_t old = newVerts.size();
     newVerts.resize_nofill(vertBary.size());
-    std::iota(newVerts.begin() + old, newVerts.end(), old + offset);
+    std::iota(newVerts.begin() + old, newVerts.end(), interiorOffset);
```
elalish, line 119: 👍

src/parallel.h
```diff
@@ -1205,7 +1207,7 @@ void gather(InputIterator mapFirst, InputIterator mapLast,
 // Write `[0, last - first)` to the range `[first, last)`.
 template <typename Iterator>
 void sequence(ExecutionPolicy policy, Iterator first, Iterator last) {
-  for_each(policy, countAt(0),
+  for_each(policy, countAt(size_t{0}),
```
elalish: Do we have anything like 0_uz for size_t, for consistency?

src/iters.h
```diff
@@ -208,6 +208,8 @@ struct CountingIterator {
   }
 };
 
+constexpr CountingIterator<int> countAt(int i) { return CountingIterator(i); }
```
elalish: Should this just be a template function?

src/impl.cpp
```diff
@@ -824,9 +824,9 @@ static std::ostream& WriteOBJWithEpsilon(std::ostream& stream,
   std::vector<ivec3> triangles;
   triangles.reserve(mesh.NumTri());
   for (size_t i = 0; i < mesh.NumTri(); i++)
-    triangles.emplace_back(mesh.triVerts[3 * i] + 1,
-                           mesh.triVerts[3 * i + 1] + 1,
-                           mesh.triVerts[3 * i + 2] + 1);
+    triangles.emplace_back(static_cast<int>(mesh.triVerts[3 * i] + 1),
```
elalish: Should we change the type of triangles instead?

src/disjoint_sets.h
```diff
@@ -31,9 +31,9 @@
 class DisjointSets {
  public:
   DisjointSets(uint32_t size) : mData(size) {
-    using namespace manifold;
-    for_each(autoPolicy(size), countAt(0), countAt(size),
-             [this](uint32_t i) { mData[i] = i; });
+    for (uint32_t i = 0; i < size; ++i) {
```
elalish: Is there a reason to unparallelize this? Then again, it is oddly trivial...

src/constructors.cpp
```diff
@@ -529,12 +529,12 @@ std::vector<Manifold> Manifold::Decompose() const {
...
     const auto& halfedge = pImpl_->halfedge_;
-    const int nFace =
-        copy_if(countAt(0_uz), countAt(NumTri()), faceNew2Old.begin(),
-                [i, &vertLabel, &halfedge](int face) {
-                  return vertLabel[halfedge.Start(3 * face)] == i;
-                }) -
-        faceNew2Old.begin();
+    const int nFace = copy_if(countAt(0), countAt(static_cast<int>(NumTri())),
+                              faceNew2Old.begin(),
+                              [i, &vertLabel, &halfedge](int face) {
```
elalish: Should we change the type of `face` instead?

src/boolean3.cpp
```diff
@@ -471,10 +471,10 @@ Vec<int> Winding03_(const Manifold::Impl& inP, const Manifold::Impl& inQ,
   b.collider_.Collisions<false>(recorder, f, verts.size(), true, ctx);
   if (IsCancelled(ctx)) return Vec<int>{};
   // flood fill
-  for_each(autoPolicy(w03.size()), countAt(0), countAt(w03.size()), ctx,
+  for_each(autoPolicy(w03.size()), countAt(0_uz), countAt(w03.size()), ctx,
            [&](size_t i) {
-             size_t root = uA.find(i);
-             if (root == i) return;
+             uint32_t root = uA.find(static_cast<uint32_t>(i));
```
elalish: Seems like root and i should have the same type?

elalish (changes requested):
Thanks!

extras/minimize_testcase.cpp
```diff
@@ -89,7 +89,7 @@ bool safeToRemove(const Polygons& polys, size_t i, size_t j, double precision) {
 std::pair<int, int> findIndex(const Polygons& polys, size_t i) {
   size_t outer = 0;
   while (i >= polys[outer].size()) i -= polys[outer++].size();
-  return std::make_pair(outer, i);
+  return std::make_pair(static_cast<int>(outer), static_cast<int>(i));
```
pca006132: fix function return tyoe instead?

src/boolean3.cpp
```diff
@@ -387,7 +387,7 @@ Intersections Intersect12_(const Manifold::Impl& inP, const Manifold::Impl& inQ,
   sequence(i12.begin(), i12.end());
 
   int index = forward ? 0 : 1;
-  stable_sort(i12.begin(), i12.end(), [&](int a, int b) {
+  stable_sort(i12.begin(), i12.end(), [&](size_t a, size_t b) {
```
pca006132: actually can we just auto these? computer not with me now, will check later.

src/disjoint_sets.h
pca006132: no, probably we just forgot about it

pca006132: actually, why not just make disjoint sets accept usize?

we still need to cast internally, but at least it moves the casts from outside to inside and reducing the number of casts needed

src/parallel.h
zmerlynn: Good call. I moved the `0_uz` literal definition down into `iters.h`, since `parallel.h` already includes that, and switched these call sites to `countAt(0_uz)`.

src/disjoint_sets.h
zmerlynn: Restored the parallel initialization. I also moved the `size_t` boundary into `DisjointSets`, so callers do not need to scatter casts around.

src/impl.cpp
zmerlynn: Agreed. I changed the temporary OBJ triangle container to use a size-capable index type instead of casting each index down to `int`.

src/iters.h
zmerlynn: Done. I changed `countAt` to a constrained integral template rather than keeping explicit overloads.

src/constructors.cpp
zmerlynn: Agreed. I changed this to use `size_t face` and made the final `size_t` to `int` conversion explicit when pushing into the `Vec<int>` expected by `GatherFaces()`.

src/boolean3.cpp
zmerlynn: Fixed via the `DisjointSets` public `size_t` API. `root` and `i` are now both `size_t` in the flood-fill path.

src/boolean3.cpp
zmerlynn: Changed to `auto` as suggested.

extras/minimize_testcase.cpp
zmerlynn: Done. `findIndex()` now returns `std::pair<size_t, size_t>` instead of casting the result down to `int`.

src/disjoint_sets.h
zmerlynn: Done. `DisjointSets` now accepts `size_t` at the public boundary and keeps the existing `uint32_t` internals, with an explicit `ToIndex()` helper/assert at the narrowing point.

pca006132 (approved):

PR #1666, opened 2026-04-20
dedicated Linux Clang ASan+UBSan workflow lane

AnshulPatil2005:
Draft PR for now, will add full description when marking ready for review.

PR #1728, opened 2026-05-20
ASan+UBSan workflow lane Pt2 

AnshulPatil2005:
## Summary
some improvements on #1666 

## Improvements 

1. Move gtest filter to env var  
Use `SANITIZER_GTEST_FILTER` in workflow so test-set edits don’t require command-line rewrites.

2. Extract filter to script  
Store sanitizer test sets in `scripts/sanitizer_cases.sh` and source it from workflow for easier tuning.

3. Add explicit sanitizer options  
Use stable defaults such as:
- `ASAN_OPTIONS=detect_container_overflow=0:strict_init_order=1`
- `UBSAN_OPTIONS=print_stacktrace=1`

4. Add clear failure diagnostics  
On failure, print a short local reproduction block directly in job logs.

5. Add runtime guardrails  
Track runtime and keep the lane within budget; make test-set trimming/expansion data-driven.

6. Split core and extended subsets  
Run a small core subset on PRs; keep broader subset for scheduled/manual runs.

7. Pin compiler/setup consistency  
Keep clang setup explicit to reduce drift and cross-runner variability.

8. (Optional) Separate build/test stages with artifact reuse  
Improve retry speed and log clarity by decoupling build from execution.

## Why this is needed

The sanitizer lane is now part of routine CI, so the next step is to make it:
- simpler to update,
- less noisy/flaky,
- easier to debug when failures happen,
- predictable in runtime.

## Scope

- CI/workflow + helper-script refactor only
- no product/runtime behavior changes in manifold code


pca006132:
do you want to cache the build when `src` and `bindings/c` is not changed? otherwise I don't see how this can be useful

AnshulPatil2005:
Local sanitizer timing looks good: core subset runs in ~56 ms (6 tests) and extended subset in ~119 ms (7 tests), so the core/extended split keeps PR checks fast

AnshulPatil2005:
Sanitizer build (ASan+UBSan) builds the sanitizer binaries (reusing cache/artifacts when available), and Sanitizer (ASan+UBSan) runs the sanitizer test subset against that build.

.github/workflows/manifold.yml
```diff
@@ -434,21 +441,185 @@ jobs:
...
+          echo "- budget_sec: \`${SANITIZER_RUNTIME_BUDGET_SEC}\`"
+        } >> "$GITHUB_STEP_SUMMARY"
+
+        if [ "${ELAPSED_SEC}" -gt "${SANITIZER_RUNTIME_BUDGET_SEC}" ]; then
+          echo "::warning::Sanitizer runtime exceeded budget: ${ELAPSED_SEC}s > ${SANITIZER_RUNTIME_BUDGET_SEC}s."
+        fi
+        if [ "${TEST_RC}" -eq 124 ]; then
+          echo "::warning::Sanitizer test timed out after ${SANITIZER_TEST_TIMEOUT_SEC}s."
+        fi
+        exit "${TEST_RC}"
```
elalish: Please move anything complex out of the yml and into its own shell script.

elalish (changes requested):
I'll defer to @pca006132 on the details of this one.

pca006132:
1. How much time does it take to build the sanitized version of the library? If it is in a couple minutes, I don't think it is worth making it cached. Caching stuff feels quite fragile to me.
2. I doubt if those additional data you print to the output can aid error discovery. I think we should just keep it simple, maybe add some information to documentation instead of making this shell script.

AnshulPatil2005:
> 1. How much time does it take to build the sanitized version of the library? If it is in a couple minutes, I don't think it is worth making it cached. Caching stuff feels quite fragile to me.

Adding cache changed the time from around 5 minutes to some 19-20 seconds.
I can add a cache toggle so caching can be turned off for debugging when needed ? it will add some workflow complexity though

> 2. I doubt if those additional data you print to the output can aid error discovery. I think we should just keep it simple, maybe add some information to documentation instead of making this shell script.

Yes, The extra output is not actually providing enough value as i expected earlier before execution. I’ll remove that diagnostic output, and add any useful guidance into documentation.

AnshulPatil2005:
@pca006132 @elalish 
Should I add anything else in this PR, or is the current sanitizer CI scope good for now?

README.md
```diff
@@ -166,6 +166,34 @@ Offline building (with missing dependencies/dependency version override):
...
+  -DMANIFOLD_CROSS_SECTION=ON \
+  -DMANIFOLD_PAR=OFF \
+  -DCMAKE_C_COMPILER=clang-18 \
+  -DCMAKE_CXX_COMPILER=clang++-18 \
+  -DCMAKE_C_FLAGS="-fsanitize=address,undefined" \
+  -DCMAKE_CXX_FLAGS="-fsanitize=address,undefined" \
+  -DCMAKE_EXE_LINKER_FLAGS="-fsanitize=address,undefined" \
+  -DCMAKE_SHARED_LINKER_FLAGS="-fsanitize=address,undefined" \
+  . -B build
+cmake --build build
```
elalish: can this be replaced by running the sanitizer_build script?

elalish (approved):
I think the scope is fine for now as long as we're planning to expand it in the next PR.

README.md
AnshulPatil2005: yes

scripts/ci/sanitizer_test.sh
```diff
@@ -0,0 +1,100 @@
...
+  echo "::error::Could not find manifold_test after downloading sanitizer artifacts."
+  echo "::group::Artifact layout"
+  find . -maxdepth 4 -type d | sort
+  echo "::endgroup::"
+  exit 127
+fi
+chmod +x "${SANITIZER_TEST_BIN}" || true
+
+BIN_DIR="$(dirname "${SANITIZER_TEST_BIN}")"
+LIB_CANDIDATES=(
```
pca006132: do we really need these lib directory finder?

AnshulPatil2005: yes we can simplify this. One advantage of the lib finder is if we later add or move library output directories we would not necessarily need to update the script with new paths.

but for now since the sanitizer artifact layout is predictable after download, I can replace the dynamic lib-directory finder with a fixed `LD_LIBRARY_PATH` based on `./build/src`, `./build/lib`, and `./build/bindings/c`, and keep a fallback if the test is missing ?

pca006132: The question here is do we need to update `LD_LIBRARY_PATH`? Did you get failures not updating that? I think the rpath should be relative and can handle this.

AnshulPatil2005: I tested removing LD_LIBRARY_PATH, and the sanitizer test failed to start with libsamples.so.3: cannot open shared object file.

AnshulPatil2005: The original script used both fixed candidates and a dynamic `find` to locate library directories.

now I kept the dynamic `find` and removed the fixed paths

pca006132:
not really required, but I think it will be nice if you can try the build on docker, look at the rpath issue. maybe it can be fixed by cmake args.

AnshulPatil2005:
> not really required, but I think it will be nice if you can try the build on docker, look at the rpath issue. maybe it can be fixed by cmake args.

I'll do it, and report you the findings or fixes before implementing

AnshulPatil2005:
> not really required, but I think it will be nice if you can try the build on docker, look at the rpath issue. maybe it can be fixed by cmake args.

I checked this in Docker and the issue was the runtime path after the artifact gets downloaded to a different location. The sanitizer build was embedding absolute build-time library paths, so `manifold_test` could not find `libsamples.so.3` without `LD_LIBRARY_PATH`.

So I changed the build to use `-DCMAKE_BUILD_RPATH_USE_ORIGIN=ON`, which makes the binaries use `$ORIGIN`-relative paths instead. Since that passed in CI, I then removed the `LD_LIBRARY_PATH` workaround from the test script and now it just relies on the embedded runtime path.

pca006132 (approved):

pca006132:
So I think this is good to go? Do you have anything else?

AnshulPatil2005:
> So I think this is good to go? Do you have anything else?

Its done for this pr, we can make a follow up later if needed.

PR #1734, opened 2026-05-26
Thread ExecutionContext into Smooth via ExecutionContext::Smooth

zmerlynn:
Follow-up to #1723: adds `ExecutionContext::Smooth(MeshGL)` and `Smooth(MeshGL64)` as the next ctx-aware static factories. Same pattern, same cancel-vs-validation precedence. Tracks the static-factory direction from #971.

## Changes

- New `MakeSmoothImpl` template in `impl.h` replaces the anonymous `SmoothImpl` in `constructors.cpp`; shared by `Manifold::Smooth` (ctx=nullptr) and `ExecutionContext::Smooth`.
- `Manifold::Impl::CreateTangents(std::vector<Smoothness>)` gains an optional `ExecutionContext::Impl* ctx = nullptr` parameter and 7 `ADVANCE_PHASE_OR_RETURN(ctx)` sites at the heavy boundaries. The two parallel `for_each_n` loops pass `ctx` to the existing ctx-aware overload in `parallel.h` for per-chunk cancel.
- `LinearizeFlatTangents` and `DistributeTangents` stay non-ctx - the boundary belongs at their call site, since both are also called from the unrelated `CreateTangents(int normalIdx)` path.
- New `kPhasesPerSmooth = kPhasesPerFromMesh + 7` (= 14) in `execution_impl.h`.

## Test plan

- [x] `ExecutionContextSmooth.*` (11 new tests) mirror the FromMeshGL cluster: HappyPath, MatchesCtor, CancelBeforeIngest, CancelBeforeEmptyInputWinsOverNoError, CancelBeforeMalformedWinsOverValidation, ValidationErrorWithoutCancel, CancelConcurrent, ReuseForBoolean, ReuseForSecondSmooth, StickyCancelAcrossCalls, plus a Smooth-specific CancelWithSharpenedEdges asserting `NumTri() == 0` on cancel (the faceID-restoration loop is a natural no-op after MakeEmpty).
- [x] Existing 18 `Smooth.*` tests still pass.
- [x] clang-format-20 clean.
- [ ] CI green

### DIFF
--- a/test/context_test.cpp
+++ b/test/context_test.cpp
@@ -21,7 +21,6 @@
 #include <atomic>
 #include <chrono>
 #include <thread>
-#include <type_traits>
 
 #include "../src/execution_impl.h"
 #include "../src/parallel.h"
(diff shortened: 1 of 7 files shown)


test/context_test.cpp
```diff
@@ -911,3 +911,155 @@ TEST(ExecutionContextFromMeshGL, StickyCancelAcrossCalls) {
...
+  ASSERT_EQ(ctx.impl_->donePhases.load(), kPhasesPerSmooth);
+  Manifold second = ctx.Smooth(TetGL());
+  EXPECT_EQ(second.Status(), Manifold::Error::NoError);
+  EXPECT_EQ(ctx.impl_->totalPhases.load(), kPhasesPerSmooth);
+  EXPECT_EQ(ctx.impl_->donePhases.load(), kPhasesPerSmooth);
+  EXPECT_DOUBLE_EQ(ctx.Progress(), 1.0);
+}
+
+// Sticky cancel survives across Smooth calls.
+TEST(ExecutionContextSmooth, StickyCancelAcrossCalls) {
```
elalish: Food for thought: could we find some generic way to generate these tests that are similar for each function? Of course the goal is simplicity and maintainability, so don't bother if there isn't a simple, clean way.

elalish (approved):
LGTM

test/context_test.cpp
zmerlynn: Took a swing in 6bde95ed: `TYPED_TEST_SUITE` over `FromMeshGLTraits` / `SmoothTraits`, with the Smooth-only `CancelWithSharpenedEdges` left outside since no other factory takes that arg. Net -66 lines. A static_assert hardens the trait contract; a TODO notes LevelSet / Hull (non-MeshGL shapes) won't fit until we widen to fixture-based traits or spin up a parallel typed suite.

Genuinely not sure I prefer it to the duplication. The indirection shows up at the call site, and the savings only really compound when a 3rd MeshGL-shaped factory lands.

A third option worth raising: just cut the per-factory test surface. The shared cancel/counter machinery is exercised by the FromMeshGL cluster, so each new factory could get away with maybe `HappyPath` + `MatchesPlain` + `CancelConcurrent` - around 30 lines without any typed-test machinery.

What's your read? Happy to keep, revert, or trim.

elalish: Yeah, I prefer trim 👍 . Code coverage is good, but let's focus heavy testing on the areas most likely to produce bugs, which is generally the algorithms.

zmerlynn: Done in ba3a8c6e - reverted the typed-test machinery and trimmed Smooth to its unique paths: happy path, factory parity, concurrent cancel through the extra tangent phases, and sharpenedEdges. The shared counter/reuse/validation coverage stays in the full FromMeshGL cluster.

Kept one pre-cancel test (`CancelBeforeIngest`) since `MakeSmoothImpl` has its own entry gate ahead of the shared ctor gate; the other precedence/reuse cases are identical plumbing, so FromMeshGL covers them.

elalish (approved):
Thanks!

PR #1739, opened 2026-05-29
Thread ExecutionContext into LevelSet via ExecutionContext::LevelSet

zmerlynn:
Adds `ExecutionContext::LevelSet`, an eager ctx-aware factory for
`Manifold::LevelSet` so a build can report `Progress()` and honor `Cancel()`.
Continues the #971 eager-ops work, same shape as `FromMeshGL` (#1723) and
`Smooth` (#1734).

- Factors the body into `Impl::CreateLevelSet`; `Manifold::LevelSet` becomes a
  thin `ctx=nullptr` wrapper (behavior unchanged).
- 5 progress phases (`kPhasesPerLevelSet`); cancel checked at an entry gate, the
  resize loop, the four grid loops (per chunk), and `SortGeometry` in finalize.
- Tests `ExecutionContextLevelSet`: HappyPath, MatchesFactory,
  CancelBeforeSampling, CancelConcurrent (trimmed to LevelSet-unique paths).

Python + wasm bindings for the ctx variant follow in a separate PR.

### DIFF
--- a/include/manifold/common.h
+++ b/include/manifold/common.h
@@ -14,6 +14,7 @@
 
 #pragma once
 #include <cmath>
+#include <functional>
 #include <limits>
 #include <memory>
 #include <vector>
@@ -35,6 +36,7 @@ template <typename Precision, typename I = uint32_t>
 struct MeshGLP;
 using MeshGL = MeshGLP<float>;
 using MeshGL64 = MeshGLP<double, uint64_t>;
+struct Box;  // defined below; needed by ExecutionContext::LevelSet
 
 /** @addtogroup Math
  * @ingroup Core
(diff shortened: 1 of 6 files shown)


elalish (commented):
Something I probably should have asked for on the last few PRs like this as well: can you give some examples of the time taken for each phase to give an idea of the granularity/evenness of the phases? Within SDF, I'd expect the vast majority of the time to be spent in NearSurface, but would be good to check. If that's true, should it have an inner-loop cancel like the booleans?

zmerlynn:
Good call, gathered numbers with a perfPhases SDF section (will land as a small follow-up so this PR stays a clean API change). Per-phase wall-clock, single-threaded:

| regime (grid) | sampling | NearSurface | ComputeVerts | BuildTris | finalize |
|---|---|---|---|---|---|
| cheap sphere (128^3) | 9% | **73%** | 3% | 3% | 12% |
| cheap gyroid, high area (128^3) | 12% | **37%** | 11% | 9% | 31% |
| expensive sphere (64^3) | **77%** | 16% | 1% | 1% | 5% |

So NearSurface dominates as you expected for cheap SDFs. The one wrinkle: the SDF is evaluated in two places - once per voxel in the sampling pass, plus inside NearSurface/ComputeVerts via FindSurface's root-finding when tolerance > 0 - so for an expensive SDF the sampling pass dominates instead, and a positive tolerance shifts cost into ComputeVerts (11 -> 113 ms here).

On the inner-loop cancel: I don't think it needs one. The dominant phase is always one of the four grid `for_each_n` loops, and those already use the ctx-aware overload, so cancel lands per parallel chunk (Par) or every 1024 elements (Seq) - finer than the booleans' per-phase-boundary cancel. The five phases are just the Progress() denominator, not the cancel granularity. The only uninterruptible unit is a single in-flight SDF callback, same as every other eager op.

elalish (approved):
Oh right, you get granular cancel by threading ctx into the for_each_n calls - missed that when I was going too fast. Thanks for the perf breakdown. One of these days we should try to optimize finalize a bit, since every op goes through it.

PR #1741, opened 2026-05-31
Fix originalID races on shared CSG nodes (#1740)

### LINKED ISSUE #1740: OriginalID racing in parallel execution
I'm encountering what appears to be a thread-safety issue when calling Manifold operations concurrently from multiple threads. Specifically, the assignment of originalID seems to suffer from race conditions, leading to missing IDs and unexpected geometry behavior.

I'm [still](https://github.com/elalish/manifold/issues/1248) working on a project that aims to replicate the [destructible procedural cave terrain of Deep Rock Galactic](https://store.steampowered.com/news/app/548430/view/4593196713081471258) (which I suspect uses Manifold under the hood, given the presence of tbb12.dll next to the executable).

To allow real-time CSG deformations to the terrain, the world is divided into cubic chunks.
Each chunk can hold n manifolds, each representing a distinct material (rock, dirt, etc.).
Modifying the terrain applies a CSG operation only to the affected chunks, which then update their visual and collision meshes.
Each chunk holds a boundary cube manifold. Its originalID is used to exclude boundary faces during the meshing process.

When executing these chunk operations in parallel, it causes racy originalID errors. Sometimes, faces that should have a known originalID end up with -1. This causes holes to appear in the terrain because the boundary faces are not correctly identified and excluded.

Running everything sequentially works exactly as expected with no missing IDs. Running chunk operations concurrently across multiple threads causes the issue described above.

Parallelizing the chunk operations mainly speeds up the creation of visual and collision meshes, since I enabled Manifold TBB parallelization anyways.

I couldn't find any explicit documentation regarding the safety of performing multiple independent Manifold operations in parallel across different threads. Is Manifold intended to be thread-safe for concurrent operations on separate threads?

Any insight into what might be going on would be greatly appreciated!

zmerlynn:
## Problem

#1740 reports faces coming back with `originalID = -1` when CSG ops run across threads on master. It's a data race on lazily-evaluated CSG nodes, separate from #1688.

`Manifold` is meant to be safe for concurrent const access (#1636), but lazy evaluation mutates shared node state through `const` methods with no per-node lock:
- `CsgLeafNode::GetImpl` realizes a pending transform by rewriting `mutable pImpl_`/`transform_`.
- `CsgOpNode::ToLeafNode` publishes `mutable cache_`.

A `Manifold` and its copies share the underlying node (`pNode_`) but each has its own `pNodeMutex_`, and `GetCsgLeafNode` drops that lock before `GetImpl`/`cache_` runs. So two threads touching one shared node at once - e.g. reading `OriginalID()` to exclude boundary faces, the reporter's case - tear the meshID map; `IncrementMeshIDs` collapses the orphaned IDs and `GetMeshGL` emits `originalID = -1`. Single-threaded, or with copies that don't share a node, it's fine, which matches the report.

## Fix

- A per-node `std::mutex` guards `CsgLeafNode`'s one-shot realize and the const accessors that read `pImpl_`/`transform_` (`GetImpl`, `Transform`, `GetBoundingBox`, `NumVert`, plus a copy ctor). The lock lives on the node, not the `Manifold`, because nodes are shared across different `Manifold`s via `CsgOpNode` children.
- `CsgOpNode` reads and publishes `cache_` under `impl_`'s existing guard (the writes were already guarded; this makes the read consistent), including the cancellation path.
- `Compose`'s direct field reads stay unlocked but are documented as requiring thread-local leaves.

The boolean path is untouched: it copies leaf operands (`node->Transform`) before `GetImpl`, so it never mutates a shared node.

## Testing

Two regression tests, committed first (without the fix) so the failure is visible, then fixed in the second commit:
- `Boolean.ConcurrentSharedNodeOriginalIDStable` - the leaf `GetImpl` path.
- `Boolean.ConcurrentSharedOpNodeOriginalIDStable` - the op-node `cache_` path.

Both halt the TSAN lane without the fix and pass with it (8 threads barrier-synced onto one shared node). Full suite passes; clang-format-20 clean.

Fixes #1740. (@elalish edited this to fixed so it closes the issue; seems likely, and we can always reopen if the reporter continues to have trouble).

### DIFF
--- a/src/csg_tree.cpp
+++ b/src/csg_tree.cpp
@@ -94,7 +94,14 @@ CsgLeafNode::CsgLeafNode(std::shared_ptr<const Manifold::Impl> pImpl_,
                          mat3x4 transform_)
     : pImpl_(pImpl_), transform_(transform_) {}
 
+CsgLeafNode::CsgLeafNode(const CsgLeafNode& other) {
+  std::lock_guard<std::mutex> lock(other.mutex_);
+  pImpl_ = other.pImpl_;
+  transform_ = other.transform_;
+}
+
 std::shared_ptr<const Manifold::Impl> CsgLeafNode::GetImpl() const {
+  std::lock_guard<std::mutex> lock(mutex_);
   if (transform_ == mat3x4(la::identity)) return pImpl_;
   pImpl_ =
       std::make_shared<const Manifold::Impl>(pImpl_->Transform(transform_));
@@ -108,6 +115,7 @@ std::shared_ptr<CsgLeafNode> CsgLeafNode::ToLeafNode(
 }
 
 std::shared_ptr<CsgNode> CsgLeafNode::Transform(const mat3x4& m) const {
+  std::lock_guard<std::mutex> lock(mutex_);
   return std::make_shared<CsgLeafNode>(pImpl_, m * Mat4(transform_));
 }
 
@@ -117,6 +125,7 @@ Box CsgLeafNode::GetBoundingBox() const {
   // Compute transformed bounding box without triggering full mesh transform.
   // This is an approximation - the actual bounding box of the transformed mesh
   // may be tighter, but this is sufficient for overlap checks.
+  std::lock_guard<std::mutex> lock(mutex_);
   const Box& box = pImpl_->bBox_;
   if (transform_ == mat3x4(la::identity)) {
     return box;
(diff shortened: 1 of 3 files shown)


zmerlynn:
The first commit adds the regression tests *without* the fix, so the bug is visible failing in CI before the fix lands:

- Failed run (test commit only): https://github.com/elalish/manifold/actions/runs/26720951764
- The test that trips it: [`Boolean.ConcurrentSharedNodeOriginalIDStable`](https://github.com/zmerlynn/manifold/blob/5102ee52/test/boolean_test.cpp#L881)

The non-sanitized lanes crash outright on the torn `shared_ptr`; the TSAN lane catches the underlying data race in `CsgLeafNode::GetImpl`. The second commit (the fix) brings it back green.

elalish (approved):
Certainly looks likely to be the source of the reported bug, and fixes something in any case. FYI @pca006132, but I believe you're busy at the moment, so I'm going to go ahead and merge this anyway.

PR #1742, opened 2026-06-01
Per-phase timing in ADVANCE_PHASE_OR_RETURN; extend perfPhases

zmerlynn:
Follow-up to #1739. Rather than hand-placing `Timer`s per op, `ADVANCE_PHASE_OR_RETURN` now records the time since the previous boundary (baseline at `ResetForStaticFactory`, `lastPhase` on `ExecutionContext::Impl`) and prints `phase N (file:line) = X ms` under `verbose >= 2`. All behind `MANIFOLD_DEBUG || MANIFOLD_TIMING`, so release is unchanged. One change times every phased op, in sync with the cancel/Progress boundaries: LevelSet (5), FromMesh (7), Smooth (14). It records only with a ctx attached; `extras/perfPhases` drives the ops through an `ExecutionContext` and adds FromMesh/Smooth sections.

**Per-phase timings** (single-threaded, ms; % of phase sum)

LevelSet:

| workload | sampling | NearSurface | ComputeVerts | BuildTris | finalize |
|---|---|---|---|---|---|
| cheap sphere n=128 | 196 (9%) | **1597 (73%)** | 77 | 49 | 274 (12%) |
| cheap gyroid n=128 | 413 (12%) | **1440 (41%)** | 376 | 352 | 902 (26%) |
| expensive sphere n=64 | **1004 (76%)** | 213 (16%) | 11 | 9 | 78 |

FromMesh ingest:

| mesh (tris) | CreateHalfedges | Cleanup | Dedupe | SetNormals | RemoveDegen | RemoveUnref | Sort |
|---|---|---|---|---|---|---|---|
| Sphere(256), 33k | 10 | 2 | 0 | 7 | **15** | 0 | 10 |
| MengerSponge(4), 395k | 80 | 54 | 0 | 130 | **166** | 9 | 82 |

(Smooth's 14 phases are sub-ms on a small control cage; the work is the unphased `Refine` after.)

Matches #1739: NearSurface dominates cheap SDFs, sampling dominates the expensive one. Release compiles with timing guarded out (verified).

### DIFF
--- a/extras/perf_phases.cpp
+++ b/extras/perf_phases.cpp
@@ -12,13 +12,19 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 //
-// Prints per-phase Boolean3 wall-clock timings on representative workloads.
-// Requires MANIFOLD_TIMING=ON (or MANIFOLD_DEBUG=ON) so Timer::Print fires.
-// Used for cancel-latency analysis and for identifying dominant phases.
+// Prints per-phase wall-clock timings on representative workloads for the
+// eager ops that carry phase instrumentation: Boolean3 (its own Timers) and
+// the ADVANCE_PHASE_OR_RETURN ops driven through an ExecutionContext -
+// LevelSet, FromMesh ingest, and Smooth. Requires MANIFOLD_TIMING=ON (or
+// MANIFOLD_DEBUG=ON). Used for cancel-latency analysis and identifying dominant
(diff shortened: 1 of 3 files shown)


elalish (approved):
Good idea! Can we use this to remove the existing timing prints we have in e.g. Boolean3? Would be nice to have consistency.

PR #1743, opened 2026-06-01
Expose ExecutionContext factories in bindings (C/Python/WASM)

### LINKED ISSUE #971: Progress and cancellation of booleans

zmerlynn:
The `ExecutionContext` object and `Manifold::WithContext` are already bound, but the ctx-aware static factories - `FromMeshGL`, `LevelSet`, `Smooth` - aren't, so ops with no source Manifold to attach a context to (notably MeshGL ingest) can't be observed or cancelled from the bindings. This adds them as methods on the context in all three layers, each documented and a thin pass-through to the already-reviewed `ExecutionContext` methods.

- **C**: `manifold_execution_context_{level_set,level_set_seq,of_meshgl,of_meshgl64,smooth,smooth64}`. Names mirror the existing `manifold_execution_context_cancel/cancelled/progress`; the SDF user-data arg is renamed `sdf_context` to keep it distinct from the `ExecutionContext`. (Happy to switch to `manifold_<op>_with_context` if you'd rather.)
- **Python**: `ExecutionContext.{from_mesh, level_set, smooth}`.
- **WASM**: `ExecutionContext.{fromMesh, smooth, levelSet}` (embind + JS wrapper + `.d.ts`).

Each layer's tests assert a pre-cancelled context propagates through a factory (`status() == Cancelled`) - the meaningful check, since `NoError` / `progress() == 1` also hold for an untouched context.

External C-API bindings (Java, OCaml, Rust) will need out-of-tree updates for the six new functions.

Closes #971.

### DIFF
--- a/bindings/c/include/manifold/manifoldc.h
+++ b/bindings/c/include/manifold/manifoldc.h
@@ -268,6 +268,30 @@ void manifold_execution_context_cancel(ManifoldExecutionContext* ctx);
 int manifold_execution_context_cancelled(ManifoldExecutionContext* ctx);
 double manifold_execution_context_progress(ManifoldExecutionContext* ctx);
 
+// ctx-aware static factories. These ops have no source manifold to attach via
+// manifold_with_context, so they run on the ExecutionContext directly to report
+// progress / observe cancellation. Mirror manifold_level_set /
+// manifold_of_meshgl / manifold_smooth; `sdf_context` is the SDF callback's
+// user-data.
+ManifoldManifold* manifold_execution_context_level_set(
+    void* mem, ManifoldExecutionContext* ec, ManifoldSdf sdf,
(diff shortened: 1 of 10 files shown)


elalish (approved):
Thanks!

PR #1745, opened 2026-06-02
Route Boolean3::Result timing through the shared phase mechanism

zmerlynn:
Follow-up to #1742, addressing the [review ask](https://github.com/elalish/manifold/pull/1742#pullrequestreview-4397896723): reuse the new per-phase timing to drop Boolean's hand-placed Timers, for consistency.

`Result`'s four Timers (Assembly/Triangulation/Simplification/Sorting) bracketed spans that were already unions of `phase()` boundaries, so the `phase()` lambda now records through the same path as the `ADVANCE_PHASE_OR_RETURN` ops, via a new explicit-line `MANIFOLD_RECORD_PHASE_AT` (a lambda called from 11 sites would otherwise report a single line). `BeginPhaseTiming` seeds the baseline at `Result` entry, since Boolean never hits the static-factory reset that seeds it for LevelSet/FromMesh/Smooth. `perfPhases` now drives its booleans through an `ExecutionContext` so the lines print.

Every phased op now emits the one `phase N (file:line) = X ms` format. The intersection-stage Timers are kept; the 11 phase lines replace the old four named Timers:

```
=== Sphere(32): subtract ===
----------- 0 ms for Intersect12 P->Q
----------- 0 ms for Intersections (total)
  phase 1 (boolean_result.cpp:772) = 0.0002 ms
  ...
  phase 11 (boolean_result.cpp:959) = 3.38 ms
366 verts and 728 tris
```

Notes:
- Kept the intersection-stage Timers in the `Boolean3` ctor (Intersect12 P->Q / Q->P, Winding03 P / Q): a P/Q sub-op breakdown, not sequential phases.
- The file:line keys drop the old "Assembly"/etc. labels.

Progress/cancel semantics and the release build are unchanged - timing only records with a ctx attached, under `MANIFOLD_DEBUG`/`_TIMING`.

### DIFF
--- a/extras/perf_phases.cpp
+++ b/extras/perf_phases.cpp
@@ -13,11 +13,13 @@
 // limitations under the License.
 //
 // Prints per-phase wall-clock timings on representative workloads for the
-// eager ops that carry phase instrumentation: Boolean3 (its own Timers) and
-// the ADVANCE_PHASE_OR_RETURN ops driven through an ExecutionContext -
-// LevelSet, FromMesh ingest, and Smooth. Requires MANIFOLD_TIMING=ON (or
-// MANIFOLD_DEBUG=ON). Used for cancel-latency analysis and identifying dominant
-// phases.
+// eager ops that carry phase instrumentation, all driven through an
+// ExecutionContext (phases only record with a ctx attached): Boolean, LevelSet,
+// FromMesh ingest, and Smooth. Boolean's Result stage reports through the
+// shared phase timing like the others; its intersection stage keeps its own
+// Timers (a P/Q sub-op breakdown, not sequential phases). Requires
+// MANIFOLD_TIMING=ON (or MANIFOLD_DEBUG=ON). Used for cancel-latency analysis
+// and identifying dominant phases.
 
 #include <algorithm>
 #include <chrono>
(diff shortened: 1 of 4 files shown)


extras/perf_phases.cpp
```diff
@@ -49,8 +54,9 @@ void BenchAll(const std::string& workload, const Manifold& a,
     std::cout << "=== " << workload << ": " << op.name
               << ", nTri(a) = " << a.NumTri() << " ===" << std::endl;
     auto t0 = std::chrono::high_resolution_clock::now();
-    Manifold result = op.apply(a, b);
-    result.NumTri();
+    ExecutionContext ctx;
+    Manifold result = op.apply(a, b).WithContext(ctx);
+    result.Status();  // force evaluation under ctx so phases time and print
```
elalish, line 59: What will happen if `a` and `b` are themselves the result of a tree of boolean operations and everything gets evaluated in parallel at `Status`? What will the timing prints look like?

zmerlynn: Good catch - it was a real bug. The timing baseline lived on the shared `ExecutionContext`, so parallel `Result` calls clobbered each other's deltas and the lines interleaved unattributably.

Fixed: the baseline is now a per-`Result` stack local, and each line carries a per-boolean uid. Still interleaved (eval is parallel), but each boolean's deltas are correct and splittable by tag:

```
  [b0] phase 7 = 0.026 ms
  [b1] phase 1 = 0.001 ms
  [b0] phase 8 = 0.228 ms   <- since b0's phase 7, not b1's
  [b1] phase 2 = 0.017 ms
```

`grep '\[b0\]'` recovers b0's clean 1..11. Static factories (single-threaded) are unchanged. Verified race-free under TSan.

elalish (approved):
Thanks!

PR #1746, opened 2026-06-02
Add Manifold::WindingNumber

### LINKED ISSUE #1440: Determine whether or not a given point is inside of a manifold
Hey, it's been a while 👋

I'm still working [a terrain system for the Godot game engine using manifold](https://www.reddit.com/r/godot/comments/1k8nlav/im_building_a_pretty_cool_new_kind_of_terrain/). Just to clarify, it doesn't use the CSG features Godot already provides using manifold.

For implementing navigation/pathfinding, especially for flying entities, I need to determine which areas are occupied by the manifold terrain in order to build an octree representation.
I was quite surprised to find there's no `bool Manifold::Contains(vec3 point)` method present in the library.

Is there an efficient recommended way to obtain that information?
I suppose retrieving a cross section at the correct height and using the winding number algorithm on the resulting polygons would work, but that seems overly complicated.

I'd like to humbly request a `Contains(vec3 point)` method or similar if there isn't a better way already.

Robert-K:
Closes #1440.

## What

Adds point-in-mesh containment testing to the public API:

```cpp
// Scalar
int  Manifold::WindingNumber(vec3 point) const;
bool Manifold::Contains(vec3 point) const;

// Batch (C++ and Python)
std::vector<int>  Manifold::WindingNumber(const std::vector<vec3>& points) const;
std::vector<bool> Manifold::Contains(const std::vector<vec3>& points) const;
```

## How

The core is `Manifold::Impl::PointWinding` in `boolean3.cpp`, which reuses the `Kernel02` kernel already at the heart of CSG booleans — the same kernel that computes `w03` during a Boolean operation.

Key design decisions:

- **XY-projected BVH query**: passing a `vec3` to the collider triggers `DoesOverlap(vec3)`, which checks only X and Y. This means all faces with XY overlap are returned regardless of Z, so the +Z ray winding accumulation correctly finds faces above the query point.
- **3D bounding-box pre-filter**: points outside `bBox_` are returned as 0 immediately, skipping the BVH entirely.
- **Zero vertex normals on query points**: `expandP=false` with zero normals means no symbolic perturbation from the query side — consistency at shared edges/vertices depends on the mesh's own normals, matching the convention used by `RayCast`.
- **Batch via single BVH traversal**: all query points are submitted together; `for_each_n` parallelises over query indices, each of which accumulates into a distinct winding slot (no atomics needed).
- **True winding number, not ray parity**: correctly returns 0 for the cavity of a hollow shell (`Sphere(2) - Sphere(1)`), where ray parity would return 1.

Bindings: C (scalar), Python (scalar + batch overloads), WASM (scalar, TypeScript types updated).

## Tests

Seven new tests in `test/measurement_test.cpp`:

| Test | What it proves |
|---|---|
| `ContainsBasic` | Basic inside/outside on a unit cube |
| `ContainsEmpty` | Empty manifold returns false/0 |
| `ContainsTallBox` | XY-projected collider query - point at z=-4 inside a z∈[-5,5] box correctly reports winding=1 (fails if a 3D Box query is used instead) |
| `WindingWatertight` | No double-counting at shared cube edges/vertices |
| `WindingHollowShell` | Real winding number: cavity=0, wall=1, outside=0 |
| `ContainsBatch` | Batch results match scalar for every point |
| `WindingMatchesRayCastParity` | Cross-checked against the existing `RayCast` implementation |

Full test suite passes (100%, 0 failures).

### DIFF
--- a/bindings/wasm/manifold-encapsulated-types.d.ts
+++ b/bindings/wasm/manifold-encapsulated-types.d.ts
@@ -1163,25 +1163,17 @@ export class Manifold {
   rayCast(origin: Vec3, endpoint: Vec3): RayHit[];
 
   /**
-   * Returns the winding number of this manifold around the given point.
-   * Returns 0 for points outside and a non-zero value (typically ±1) for
-   * points inside. May exceed ±1 for manifolds with overlapping regions.
+   * Returns the winding number of this manifold around each of the given
+   * points. Returns 0 for points outside and a non-zero value (typically ±1)
+   * for points inside a simple closed surface. May exceed ±1 in magnitude for
+   * manifolds with overlapping regions.
    *
-   * @param point The 3D query point.
+   * @param points The 3D query points.
    *
    * @group Spatial Queries
    */
-  windingNumber(point: Vec3): number;
+  windingNumber(points: Vec3[]): number[];
 
-  /**
-   * Returns true if the given point is inside this manifold, i.e. its winding
-   * number is non-zero.
-   *
-   * @param point The 3D query point.
-   *
-   * @group Spatial Queries
-   */
-  contains(point: Vec3): boolean;
 
 
   /**
--- a/bindings/wasm/manifold-encapsulated-types.d.ts
+++ b/bindings/wasm/manifold-encapsulated-types.d.ts
@@ -1163,6 +1163,28 @@ export class Manifold {
   rayCast(origin: Vec3, endpoint: Vec3): RayHit[];
 
   /**
+   * Returns the winding number of this manifold around the given point.
+   * Returns 0 for points outside and a non-zero value (typically ±1) for
+   * points inside. May exceed ±1 for manifolds with overlapping regions.
+   *
+   * @param point The 3D query point.
+   *
+   * @group Spatial Queries
+   */
+  windingNumber(point: Vec3): number;
+
+  /**
+   * Returns true if the given point is inside this manifold, i.e. its winding
+   * number is non-zero.
+   *
+   * @param point The 3D query point.
+   *
+   * @group Spatial Queries
+   */
+  contains(point: Vec3): boolean;
+
+
+  /**
    * Returns the reason for an input Mesh producing an empty Manifold. This
    * Status will carry on through operations like NaN propogation, ensuring an
    * errored mesh doesn't get mysteriously lost. Empty meshes may still show
--- a/src/boolean3.cpp
+++ b/src/boolean3.cpp
@@ -594,9 +594,10 @@ Vec<int> Manifold::Impl::PointWinding(VecView<const vec3> points) const {
     pointImpl.vertPos_[i] = points[active[i]];
 
   // expandP=false: no symbolic perturbation from the query side (zero normals).
-  // forward=true: project along +Z to count signed face crossings above each point.
-  // f returns vec3 → collider uses DoesOverlap(vec3), which is XY-projected,
-  // so all faces with XY overlap are returned regardless of Z (correct for +Z winding).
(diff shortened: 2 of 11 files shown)


src/manifold.cpp
```diff
@@ -1136,4 +1136,55 @@ double Manifold::MinGap(const Manifold& other, double searchLength) const {
...
+ */
+int Manifold::WindingNumber(vec3 point) const {
+  const auto impl = GetCsgLeafNode().GetImpl();
+  Vec<int> w = impl->PointWinding(VecView<const vec3>(&point, 1));
+  return w.empty() ? 0 : w[0];
+}
+
+/**
+ * Returns whether each of the given points is inside this manifold, i.e. its
+ * winding number is non-zero.
```
elalish: So, this assumes a fill rule, and the one I would tend to default to would be strictly positive, meaning negative winding numbers count as not contained. Still, I think really this is just a reason not to bother including these helpers - it's really not much work for someone to use `WindingNumber` to get this directly, and it forces them to consider what inside should mean for them. I'd prefer to cut down the amount of binding surface we need to maintain anyway.

src/manifold.cpp
```diff
@@ -1136,4 +1136,55 @@ double Manifold::MinGap(const Manifold& other, double searchLength) const {
...
+ * Returns the winding number of this manifold around the given point. Returns
+ * 0 for points outside, and a non-zero value (typically ±1) for points inside.
+ * For manifolds with overlapping regions the magnitude may exceed 1.
+ *
+ * @param point The 3D query point.
+ */
+int Manifold::WindingNumber(vec3 point) const {
+  const auto impl = GetCsgLeafNode().GetImpl();
+  Vec<int> w = impl->PointWinding(VecView<const vec3>(&point, 1));
+  return w.empty() ? 0 : w[0];
```
elalish: Again, I think users can handle this themselves - putting it as it's own API might make someone think it's somehow more efficient in some case when it's not. Let's keep this PR to adding only a single function.

src/boolean3.cpp
```diff
@@ -573,6 +573,45 @@ Boolean3::Boolean3(const Manifold::Impl& inP, const Manifold::Impl& inQ,
   }
 #endif
 }
+Vec<int> Manifold::Impl::PointWinding(VecView<const vec3> points) const {
+  ZoneScoped;
+  Vec<int> winding(points.size(), 0);
+  if (points.empty() || IsEmpty()) return winding;
+
+  // Points outside the bounding box have winding 0 for any closed manifold.
+  Vec<int> active;
```
elalish: I might call this `query2input` since it's a mapping.

bindings/wasm/manifold-encapsulated-types.d.ts
```diff
@@ -1162,6 +1162,28 @@ export class Manifold {
...
+  /**
+   * Returns the winding number of this manifold around the given point.
+   * Returns 0 for points outside and a non-zero value (typically ±1) for
+   * points inside. May exceed ±1 for manifolds with overlapping regions.
+   *
+   * @param point The 3D query point.
+   *
+   * @group Spatial Queries
+   */
+  windingNumber(point: Vec3): number;
```
elalish: I think there will be a pretty serious performance penalty to doing these one at a time through WASM - can we plumb through the vector version instead?

elalish (changes requested):
Thanks! Let's simplify it a bit first.

Robert-K:
I don't have clang-format setup yet, sorry 😅

elalish (approved):
Looks great, thank you!

PR #1722, opened 2026-05-17
Add boolean2 core implementation

zmerlynn:
## Summary

Adds the core Boolean2 2D polygon arrangement implementation behind the experimental `MANIFOLD_CROSS_SECTION_BACKEND=boolean2` path.

This PR is the Boolean2 core implementation step in the staged #1707 stack, which is building a Manifold-native replacement path for the current Clipper2-backed `CrossSection` implementation. It lands the internal arrangement pipeline and design document, but intentionally does not switch production `CrossSection` behavior over to Boolean2 yet.

On this branch, the selected Boolean2 backend is still placeholder-only; behavior-level `CrossSection` validation remains Clipper2-backed until the follow-up backend-wiring PR.

Context:

- Original overlap-removal sketch: #289
- Clipper2 replacement / Boolean2 backend tracker: #1707

## What’s Included

- Boolean2 core data flow:
  - vertex merge / edge collapse
  - edge-edge broad phase and intersection insertion
  - near-vertex edge splitting
  - fixed-point cleanup iteration
  - canonical sub-edge generation
  - winding-rule face filtering
  - polygon conversion helpers
  - offset and containment utilities for later `CrossSection` integration

- Design documentation:
  - documents the Boolean2 approach and branch boundary
  - relates the implementation to the overlap-removal sketch in #289
  - clarifies that public `CrossSection` wiring lands in a later PR

- Build integration:
  - adds Boolean2 sources and private headers to CMake
  - keeps `MANIFOLD_CROSS_SECTION_BACKEND=boolean2` selectable
  - preserves the backend as placeholder-only on this branch

## Not Included

- No production `CrossSection` behavior is switched to Boolean2 in this PR.
- No public API change is intended.
- CrossSection behavior tests on this branch should still be interpreted as Clipper2-backed validation.

## Follow-Up

The next stack branch wires this core into the public `CrossSection` backend:

- Backend wiring draft: https://github.com/zmerlynn/manifold/tree/pr/boolean2-backend-wiring

A later stack branch adds behavior coverage against the backend:

- Tests draft: https://github.com/zmerlynn/manifold/tree/pr/boolean2-tests

## Performance

Boolean2 is not expected to outperform Clipper2 in this staged PR. The last recorded prototype measurements still show Boolean2 slower than Clipper2, especially for offset-heavy workloads.

[...]

Issue #1707, opened 2026-05-13
Add experimental boolean2 CrossSection backend


zmerlynn:
`CrossSection` currently depends on Clipper2 for 2D clipping and offsetting. This issue proposes staging an in-tree `boolean2` backend that moves 2D polygon cleanup/boolean work closer to Manifold's own Smith-style robustness model, with the goal of eventually removing the Clipper2 dependency.

### Motivation

The main motivation is architectural: Manifold's 3D boolean design is based on Smith-style symbolic perturbation and topology-preserving assembly, but `CrossSection` currently delegates its core 2D cleanup/boolean behavior to Clipper2. The `boolean2` work brings the 2D path into the same family of floating-point algorithms and APIs as the rest of Manifold.

This connects directly to #289. That issue sketches overlap removal as a future capability for cleaning up self-intersections after operations like `Warp()` and `Refine()`, and possibly as part of a manifold repair pipeline. The proposed `boolean2` backend is the 2D portion of that direction: remove overlaps with Manifold-native robust predicates and explicit topology handling rather than treating polygon cleanup as an external black box.

This also gives us a path to avoid #1683, where `CrossSection` currently loses double precision through the Clipper2 path. Keeping the 2D path in Manifold's floating-point robustness framework should avoid that kind of precision mismatch at the 2D/3D boundary.

Moving this in-tree should also make it possible to add user-definable `Tolerance` support for 2D, matching the direction of the 3D APIs more closely.

Secondary motivations:

- Reduce binary size, especially for WASM.
- Reduce optional dependency and packaging surface.
- Make CrossSection behavior easier to test, fuzz, and evolve in-tree.
- Provide a backend path for future 2D behavior that aligns with Manifold's own robustness model.

### Current prototype status

I have an AI-driven `boolean2` CrossSection prototype that I have reviewed, tested, refactored, and split toward reviewable PRs. The intent is not to land it as a large replacement PR.

Future-removal size measurements from a WASM build using the new backend, with Clipper2 removed from the linked output:

- Native stripped `libmanifold.so`: about 3,253 KB -> 2,797 KB, roughly -456 KB.
- WASM MinSizeRel: about 515 KB -> 478 KB, roughly -37 KB.
- WASM Release: about 660 KB -> 617 KB, roughly -43 KB.

Correctness and determinism status:

- Existing `CrossSection.*` tests pass locally under both the Clipper2 and boolean2 backends.
[...]

PR #1751, opened 2026-06-08
Cut over CrossSection from Clipper2 to boolean2

zmerlynn:
Makes the in-tree boolean2 backend (core #1722, offset/decompose #1749) the sole `CrossSection` implementation and removes the Clipper2 dependency, the `MANIFOLD_CROSS_SECTION_BACKEND` selector, and the related CI and packaging plumbing.

The public `CrossSection` API is unchanged. `FillRule` stays in the signature, but boolean2 is Positive-only, so a non-Positive argument is ignored (and rejected under `MANIFOLD_ASSERT`); the constructor docs note this.

`Samples.Sponge4` extrudes a simplified projection of the cut sponge and bounds its near-degenerate triangle count; the bound is raised 32 -> 80. boolean2 and clipper2 simplify to the same vertices, area, and topology, but their coordinates differ at the sub-tolerance (~1e-13) level, so a few more triangles fall in the degenerate band. `Genus` is unchanged, so it's a numerical artifact rather than a geometry change.

Additional `CrossSection` tests and a fuzz harness follow in a later PR. Full `manifold_test` passes and `manifold3d` builds; no Clipper2 dependency remains.


src/cross_section/cross_section_boolean2.cpp
```diff
@@ -31,84 +39,305 @@ struct PathImpl {
 }  // namespace manifold
 
 namespace {
-std::shared_ptr<const PathImpl> empty_paths() {
-  return std::make_shared<const PathImpl>(Polygons{});
+
+bool AllFinite(const mat2x3& m) {
+  for (int c = 0; c < 3; ++c) {
```
elalish: nit: I prefer `const int c : {0,1,2}` for these small loops.

src/cross_section/cross_section_boolean2.cpp
```diff
@@ -31,84 +39,305 @@ struct PathImpl {
...
+    case CrossSection::JoinType::Bevel:
+      return b2::OffsetJoinType::Bevel;
+  }
+  return b2::OffsetJoinType::Square;
+}
+
+// ||M||_2 (largest singular value) of the linear part of an affine 2D
+// transform. Closed 2x2 form rather than svd.h SpectralNorm, which is mat3-only
+// (embedding via Mat3 would fold in the homogeneous block and floor the norm).
+double Mat2SpectralNorm(const mat2& m) {
```
elalish: This could live in svd.h.

src/cross_section/cross_section_boolean2.cpp
```diff
@@ -31,84 +39,305 @@ struct PathImpl {
...
+void HullBacktrack(const vec2& point, std::vector<vec2>& stack) {
+  size_t size = stack.size();
+  while (size >= 2 &&
+         CCW(stack[size - 2], stack[size - 1], point, 0.0) <= 0.0) {
+    stack.pop_back();
+    size = stack.size();
+  }
+}
+
+SimplePolygon HullImpl(SimplePolygon& points) {
```
elalish: Are these functions all just copies of the ones we already have, but modified to our data structures? I guess that's fine, but I'm not sure this is the easiest way to review. Is it time to simply cut over from Clipper to this? Then we can actually see diffs of these files instead of duplicating and removing in separate PRs. Plus it'd be nice to see our tests still passing. I'd also like to get us using the new path early so we can find any problems well before the next release. Still, feel free to push back if this would slow your development.

src/cross_section/cross_section_boolean2.cpp
```diff
@@ -31,84 +39,305 @@ struct PathImpl {
...
+  path.reserve(lower.size() + upper.size());
+  for (const vec2& point : lower) path.push_back(point);
+  for (const vec2& point : upper) path.push_back(point);
+  return path;
+}
+
+// Drop the vertex closest to the line through its current neighbors, repeating
+// while under eps (Clipper2 SimplifyPaths semantics). Re-checking current, not
+// original, neighbors each step is what keeps a curve from collapsing.
+SimplePolygon SimplifyRing(SimplePolygon ring, double eps) {
```
elalish: Since this is input, I might change `eps` to `tol`, since I think it can be user specified. I think of eps as meaning machine precision.

src/cross_section/cross_section_boolean2.cpp
```diff
@@ -31,84 +39,305 @@ struct PathImpl {
...
+  const double eps2 = eps * eps;
+  while (ring.size() > 3) {
+    const int n = static_cast<int>(ring.size());
+    int victim = -1;
+    double bestD2 = eps2;
+    for (int i = 0; i < n; ++i) {
```
elalish: Might want to put in a note about this being O(n^2) and might be worth optimizing at some point.

elalish (approved):
Thanks! Let me know how you'd like to proceed.

src/cross_section/cross_section_boolean2.cpp
zmerlynn: Agreed - I wasn't sure you wanted to start landing breaking changes for 4.0 yet, but I'm game if you are. This turns the PR into a full cutover, breaking changes and all: replace the clipper2 backend with boolean2 so the diff lands on the real files and the suite runs against it, drop `MANIFOLD_CROSS_SECTION_BACKEND`, and pull Clipper2 out (nothing else depends on it). I've got at least one more PR after this with additional tests. I'll fold your nits (range-for, `eps`->`tol`, the O(n^2) note, SpectralNorm into svd.h) in as I go. Sound good?

elalish: Yeah, that sounds great, thanks!

zmerlynn: Quick scope update: I went with the cutover, but backed off the full API breaks (dropping `FillRule`, `Compose`, etc.) - they churned the C/Python/WASM bindings, the examples, and CI more than was worth bundling into this PR. The public `CrossSection` API is unchanged here; it's just the backend swap and the Clipper2 removal. I'll do the API cleanups (the #1707 list) as a focused follow-up soon.

src/cross_section/cross_section_boolean2.cpp
zmerlynn: Done.

src/cross_section/cross_section_boolean2.cpp
zmerlynn: Done - added it as a `mat2` `SpectralNorm` overload in svd.h, next to the mat3 one.

src/cross_section/cross_section_boolean2.cpp
zmerlynn: Done - renamed the param to `tol`.

src/cross_section/cross_section_boolean2.cpp
zmerlynn: Done - added a note on the per-removal rescan.

cmake/info.cmake
```diff
@@ -63,7 +63,6 @@ message(STATUS " ")
 message(STATUS "MANIFOLD_VERSION:              ${MANIFOLD_VERSION}")
 message(STATUS "MANIFOLD_PAR:                  ${MANIFOLD_PAR}")
 message(STATUS "MANIFOLD_CROSS_SECTION:        ${MANIFOLD_CROSS_SECTION}")
-message(STATUS "MANIFOLD_CROSS_SECTION_BACKEND: ${MANIFOLD_CROSS_SECTION_BACKEND}")
```
elalish, line 66: We can also get rid of MANIFOLD_CROSS_SECTION, since it was only optional in order to remove the dependency, but now we can include it always.

src/cross_section/cross_section.cpp
```diff
@@ -228,55 +202,59 @@ CrossSection& CrossSection::operator=(CrossSection&&) noexcept = default;
...
  */
 CrossSection::CrossSection(const SimplePolygon& contour, FillRule fillrule) {
-  auto ps = C2::PathsD{(pathd_of_contour(contour))};
-  paths_ = shared_paths(C2::Union(ps, fr(fillrule), precision_));
+  DEBUG_ASSERT(fillrule == FillRule::Positive, logicErr,
+               "Boolean2 CrossSection supports only Positive fill");
+  (void)fillrule;
+  Polygons input{contour};
+  tolerance_ = b2::InferEps(input, {});
+  paths_ = shared_paths(b2::ApplyFillRule(input, tolerance_));
```
elalish: This could just call the other constructor.

elalish (commented):
This is looking good, but I should pour over it a little more.

src/cross_section/cross_section.cpp
zmerlynn: Good call - done, the single-contour ctor now delegates to the `Polygons` one.

cmake/info.cmake
zmerlynn: Done - with Clipper2 gone it no longer needs to be optional, so CrossSection always builds now; dropped the option and the `cross_section: OFF` CI lane.

src/cross_section/cross_section.cpp
```diff
@@ -14,211 +14,185 @@
...
 
-void decompose_hole(const C2::PolyTreeD* outline,
-                    std::vector<C2::PathsD>& polys, C2::PathsD& poly,
-                    size_t n_holes, size_t j) {
-  if (j < n_holes) {
-    auto child = outline->Child(j);
-    decompose_outline(child, polys, 0);
-    poly[j + 1] = child->Polygon();
-    decompose_hole(outline, polys, poly, n_holes, j + 1);
+b2::OffsetJoinType JoinTypeOf(CrossSection::JoinType jointype) {
```
elalish: Can we remove this and just have the one JoinType?

src/cross_section/cross_section.cpp
```diff
@@ -14,211 +14,185 @@
...
+      const double pnLen2 = la::dot(pn, pn);
+      const double cross = la::cross(ring[i] - P, pn);
+      const double d2 = pnLen2 > 0.0 ? cross * cross / pnLen2 : 0.0;
+      if (d2 < bestD2) {
+        bestD2 = d2;
+        victim = i;
+      }
+    }
+    if (victim < 0) break;  // every vertex deviates by >= tol
+    ring.erase(ring.begin() + victim);
```
elalish: I know this isn't optimized, but this hurts my soul. At least a linked list or something? Oh well, we can always wait until we're checking performance on our corpus. Also, if my new 3D decimator works out well, we may want to change this one to work more like that, so perhaps best not to worry about it too much yet.

src/cross_section/cross_section.cpp
```diff
@@ -672,22 +617,17 @@ CrossSection CrossSection::Simplify(double epsilon) const {
...
-    // (radius) in order to get back the same number of segments in Clipper2:
-    // steps_per_360 = PI / acos(1 - arc_tol / abs_delta)
-    const double abs_delta = std::fabs(delta);
-    arc_tol = (math::cos(Clipper2Lib::PI / n) - 1) * -abs_delta;
-  }
-  auto ps =
-      C2::InflatePaths(GetPaths()->paths_, delta, jt(jointype),
-                       C2::EndType::Polygon, miter_limit, precision_, arc_tol);
-  return CrossSection(shared_paths(ps));
+  Polygons offset = b2::Offset(GetPaths()->paths_, delta, JoinTypeOf(jointype),
```
elalish: Do we need the `b2::` namespace? Actually, general question: do we need any subnamespaces below `manifold` at all? I've never really understood the point within a single project. The only thing I can think of is to avoid name collisions, but I would much rather keep our project free of actual name collisions for the sake of making it searchable, so I'd rather it fail to build if someone collides a name. Am I missing something? I'm a fan of anonymous namespaces, but that's about it.

src/cross_section/cross_section.cpp
```diff
@@ -739,31 +689,27 @@ CrossSection CrossSection::Hull(SimplePolygon pts) {
...
+  return Hull(points);
 }
 
 /**
  * Return the total area covered by complex polygons making up the
  * CrossSection.
  */
-double CrossSection::Area() const { return C2::Area(GetPaths()->paths_); }
+double CrossSection::Area() const {
+  return b2::TotalSignedArea(GetPaths()->paths_);
```
elalish: Can we remove some more of these wrapper functions?

elalish (approved):
This looks great! Love the cleanup.

src/cross_section/cross_section.cpp
zmerlynn: Done: JoinType now lives in common.h next to OpType, and CrossSection::JoinType is an alias for it, so nothing else had to change. Converter deleted.

src/cross_section/cross_section.cpp
zmerlynn: Couldn't resist - it's a min-heap with lazy invalidation now (O(n log n), same current-neighbor semantics). Differential-tested against the old loop on ~22k cases, identical output. Refitting it to your decimator later still makes sense if that lands.

src/cross_section/cross_section.cpp
zmerlynn: Agreed - it only existed because this grew up as a separate backend. Flattened; no collisions, nothing needed renaming. Related: the files are also a level deeper than needed (src/cross_section/boolean2/). Want me to move the engine up (src/boolean2/, or flat like boolean3) in a follow-up PR?

src/cross_section/cross_section.cpp
zmerlynn: Took a pass with call counts: Area() is the only pure forwarder left, and it's public API. TotalSignedArea stays because the white-box tests in the next PR call it directly. The rest of boolean2.h is used directly by those tests and the fuzz harness. Point me at any that still bug you.

zmerlynn:
WASM size at this head vs master (324bec7bd), MinSizeRel, JSBIND=ON, PAR=OFF - same method as the #1707 numbers:

- manifold.wasm: 543,051 -> 531,736 raw (-11.3 KB, -2.1%); gzipped 205,248 -> 197,529 (-7.7 KB, -3.8%)
- manifold.js: +0.7 KB raw

Smaller win than the -37 KB I posted in #1707, so I dug into where it went. That estimate predates the offset and decompose implementations and the review hardening. Breaking down today's bundle per-function: the code section is near a wash (our engine ~87 KB vs Clipper2 plus glue ~83 KB, net -2 KB) and most of the win is actually Clipper2's static data leaving (-9.4 KB). Two notable engine costs: the regularization pipeline inlines into one 26.8 KB function, and deterministic stable sorts instantiate 37 separate sort chains totaling ~34 KB. Funneling those through fewer comparator types is the clearest lever if we want more back later.

src/cross_section/cross_section.cpp
elalish: Yeah, probably so - we used to have more directory structure, but someone flattened it for us and had some reasoning about why it made things easier downstream, though I have now forgotten the details. As part of that we should also reduce it to fewer files.

elalish (approved):
Thanks for looking into the binary size. I would like to try to keep that down, though not at the cost of too much performance. If you find a reasonable tradeoff to shrink it, that would be great, and even better if we can apply those lessons learned to the rest of the repo.

PR #1754, opened 2026-06-10
Flatten the boolean2 engine into src/ and consolidate to fewer files

zmerlynn:
Follow-up from the layout discussion on #1751 ([thread](https://github.com/elalish/manifold/pull/1751#discussion_r3391145219)) and the file-layout item in #1707: the engine moves up from `src/cross_section/boolean2/` to flat `src/`, consolidated from 24 files to 8. The `src/cross_section/` directory only existed to hold the optional-feature gate that #961 left behind, and #1751 removed the option - so the tree goes fully flat again.

| target | absorbs |
| --- | --- |
| `src/cross_section.cpp` | `cross_section/cross_section.cpp` (pure move) |
| `src/boolean2.h` | all nine engine headers |
| `src/boolean2.cpp` | driver, BVH, vertex merge, edge-vertex lists, intersections, canonicalize |
| `src/boolean2_winding.cpp` | winding_filter.cpp |
| `src/boolean2_offset.cpp` | offset.cpp + containment.cpp |
| `src/boolean2_predicates.cpp` | predicates.cpp (kept separable for the eventual merge with the 3D epsilon logic, #1707) |
| `src/boolean2_diagnostics.{h,cpp}` | diagnostics.{h,cpp} |

Pure movement - no code, namespace, or signature changes. Each commit builds and passes tests on its own, and reviewing commit-by-commit is probably easiest: renames first, then the header merge, then the source merges, then path touchpoints (gen_docs.py, the Python docstring deps, Doxyfile, docs/Boolean2.md, and a now-dead `src/*/*.cpp` glob in scripts/format.sh).

Two things to know while reviewing:
- The merged `boolean2.cpp` keeps one section per absorbed module, with the old module contract comments preserved as section headers. The sections also keep their own anonymous-namespace blocks rather than coalescing into one, to keep the diff mechanical - happy to collapse them here or in a follow-up if you'd rather.
- `git log --follow` survives for the 1:1 renames; the modules merged into `boolean2.cpp` and `boolean2_offset.cpp` trace back to the merge commit.

Full `manifold_test` passes, and the follow-up white-box test suite and ASan+UBSan fuzz smoke pass against this layout on the stacked branches.


src/boolean2.cpp
```diff
@@ -255,4 +261,959 @@ Polygons Boolean2D(const Polygons& a, const Polygons& b, OpType op, double eps,
...
+    }
+    out.Add(prev, edges[e].v1, edges[e].mult);
+  }
+  out.Finalize();
+  return out;
+}
+
+// ===== vertex_merge (was cross_section/boolean2/vertex_merge.cpp) =====
+
+using manifold::la::dot;
```
elalish, line 361: Can we avoid `using` directives?

elalish (approved):
Thanks, this looks good. I've heard you mention a large test suite somewhere - that might be a good next PR. It would give me more confidence to see this running on our corpus and some interesting fuzzed examples.

src/boolean2.cpp
zmerlynn: I'll sweep those soon!

zmerlynn:
> Thanks, this looks good. I've heard you mention a large test suite somewhere - that might be a good next PR. It would give me more confidence to see this running on our corpus and some interesting fuzzed examples.

Yup, this was just sequencing, I wanted to flatten the structure first so the churn was less. Test PR up at #1755

src/boolean2.cpp
zmerlynn: Done in #1756.

PR #1755, opened 2026-06-11
Add boolean2 and CrossSection regression and corpus tests

zmerlynn:
Test suite for the boolean2 cutover - #1707 staging item 5, following #1722/#1749/#1751/#1754. Adds 79 tests:

- `test/boolean2_test.cpp` (new): 37 white-box tests of the engine in suite `Boolean2` - graph order, vertex merge, edge-vertex lists, intersection insertion, the retained-graph validator, and winding filtering. The validator's geometric predicates are deliberately independent of the engine's (using `IntersectSegments` to check its own output would be circular); a comment in the file records how their epsilon band differs from `CCW`'s.
- `test/cross_section_test.cpp`: extended from its existing 15 tests with 41 new public-API tests - Boolean/BatchBoolean invariants (commutativity, distributivity, area conservation), Offset join types and edge cases, Decompose containment, Simplify, and degenerate/non-finite inputs.
- `test/cross_section_offset_corpus_test.cpp` (new): offsets the polygon triangulation corpus across every join type and gates on clean triangulation; runs on WASM via the preloaded corpus.

Most regression cases are seeds from a fuzz campaign against the backend, kept as literal geometry with a short note on what each shape triggers; the fuzz harness itself follows in a separate PR. Two of these - tiny-feature point contacts near a host vertex - are marked `DISABLED_`; they expose an arrangement/winding-robustness issue tracked separately. White-box `../src` includes follow the existing pattern (boolean_test, manifold_test, and three others). Suite runs green on Linux/Mac/Windows/WASM and under ASan+UBSan.

### DIFF
--- a/test/boolean2_test.cpp
+++ b/test/boolean2_test.cpp
@@ -1168,3 +1168,10 @@ TEST(Boolean2, IntersectSegmentsSeeds) {
     }
   }
 }
+
+TEST(Boolean2, NonFiniteInputReturnsEmpty) {
+  const double inf = std::numeric_limits<double>::infinity();
+  Polygons bad = {{{0.0, 0.0}, {1.0, 0.0}, {inf, 1.0}, {0.0, 1.0}}};
+  Polygons finite = {{{0.0, 0.0}, {1.0, 0.0}, {1.0, 1.0}, {0.0, 1.0}}};
+  EXPECT_TRUE(Boolean2D(bad, finite, OpType::Add).empty());
+}
--- a/test/boolean2_test.cpp
+++ b/test/boolean2_test.cpp
@@ -270,7 +270,6 @@ std::vector<EdgeM> EdgesFromOverlapResult(const OverlapResult& result) {
 
 OverlapResult CleanupPassLikeIterate(const OverlapResult& result, double eps) {
   return RemoveOverlaps2D(result.verts, EdgesFromOverlapResult(result), eps,
-                          /*tolerance=*/0.0,
                           /*debug=*/false, WindRule::Add);
 }
 
@@ -518,8 +517,8 @@ TEST(Boolean2, CleanupPassMatchesValidAddSinglePass) {
   Polygons polys{RandomTopologicalRing(8, 15)};
   const double eps = InferEps(polys, {});
   const auto [verts, edges] = PolygonsToInput(polys);
-  const auto pass1 = RemoveOverlaps2D(verts, edges, eps, /*tolerance=*/0.0,
-                                      /*debug=*/false, WindRule::Add);
+  const auto pass1 =
+      RemoveOverlaps2D(verts, edges, eps, /*debug=*/false, WindRule::Add);
   EXPECT_TRUE(CheckRetainedGraphValidity(pass1, edges, pass1.inputVert2Merged,
                                          pass1.numMergedVerts, eps));
 
@@ -778,8 +777,8 @@ TEST(Boolean2, BooleanRobustnessMergeTopologyBalance) {
   const auto [verts, edges] = CombinedInput(a, b, /*bMult=*/1);
   ASSERT_FALSE(verts.empty());
   const double eps = InferEps(a, b);
-  const auto overlap = RemoveOverlaps2D(verts, edges, eps, /*tolerance=*/0.0,
-                                        /*debug=*/false, WindRule::Add);
+  const auto overlap =
+      RemoveOverlaps2D(verts, edges, eps, /*debug=*/false, WindRule::Add);
   EXPECT_TRUE(CheckRetainedGraphValidity(
       overlap, edges, overlap.inputVert2Merged, overlap.numMergedVerts, eps));
 }
@@ -811,8 +810,8 @@ TEST(Boolean2, BooleanRobustnessDirectCastKeepsExpectedArea) {
   const auto [verts, edges] = CombinedInput(a, b, /*bMult=*/1);
   ASSERT_FALSE(verts.empty());
   const double eps = InferEps(a, b);
-  const auto overlap = RemoveOverlaps2D(verts, edges, eps, /*tolerance=*/0.0,
-                                        /*debug=*/false, WindRule::Add);
+  const auto overlap =
+      RemoveOverlaps2D(verts, edges, eps, /*debug=*/false, WindRule::Add);
   EXPECT_TRUE(CheckRetainedGraphValidity(
       overlap, edges, overlap.inputVert2Merged, overlap.numMergedVerts, eps));
 
@@ -875,48 +874,14 @@ TEST(Boolean2, KeepsNearDistinctPresentationVertex) {
   constexpr double kDelta = 1e-12;
   const Polygons input = {
       {{0, 0}, {1, 0}, {1, -1}, {2, -1}, {1 + kDelta, 0}, {1, 1}}};
-  const auto polys = Boolean2D(input, {}, OpType::Add, /*eps=*/1e-15,
-                               /*tolerance=*/1e-9);
+  const auto polys = Boolean2D(input, {}, OpType::Add, /*eps=*/1e-15);
   ASSERT_EQ(polys.size(), 1u);
   EXPECT_EQ(polys[0].size(), 6u);
   EXPECT_NEAR(TotalSignedArea(polys), 1.0 + kDelta, 1e-12);
 }
 
-// The arrangement is eps-only: a crossing a tolerance-scale distance from an
-// old endpoint (5e-4 from corner (0,0), eps 1e-6) is a real tiny feature, kept
-// whatever the op tolerance. Merging it is Simplify's job, not the boolean's.
-TEST(Boolean2, GeneratedCrossingNearEndpointStaysDistinct) {
-  constexpr double eps = 1e-6;
-  const Polygons a = {{{0, 0}, {10, 0}, {10, 1}, {0, 1}}};
-  const Polygons b = {{{5e-4, -0.5}, {0.5, -0.5}, {0.5, 0.5}, {5e-4, 0.5}}};
-  const auto [verts, edges] = CombinedInput(a, b, /*bMult=*/1);
-
-  const auto atEps = RemoveOverlaps2D(verts, edges, eps, /*tolerance=*/eps,
-                                      /*debug=*/false, WindRule::Add);
-  const auto atTol = RemoveOverlaps2D(verts, edges, eps, /*tolerance=*/1e-3,
-                                      /*debug=*/false, WindRule::Add);
-
-  auto countNear = [](const std::vector<vec2>& points, vec2 target, double r) {
-    int count = 0;
-    for (const vec2& p : points)
-      if (dot(p - target, p - target) <= r * r) ++count;
-    return count;
-  };
-  const vec2 generatedNearOld{5e-4, 0.0};
-
-  // A 1000x-larger tolerance changes nothing.
-  EXPECT_GT(countNear(atEps.verts, generatedNearOld, 10 * eps), 0);
-  EXPECT_GT(countNear(atTol.verts, generatedNearOld, 10 * eps), 0);
-  EXPECT_EQ(atEps.verts.size(), atTol.verts.size());
-  EXPECT_TRUE(CheckRetainedGraphValidity(atEps, edges, atEps.inputVert2Merged,
-                                         atEps.numMergedVerts, eps));
-  EXPECT_TRUE(CheckRetainedGraphValidity(atTol, edges, atTol.inputVert2Merged,
-                                         atTol.numMergedVerts, eps));
-}
-
 // The new-to-old snap is a 2*eps budget: a generated crossing within 2*eps of
-// an old corner fuses into it; beyond 2*eps it stays a distinct vertex. Op
-// tolerance does not enter - this is purely the eps-scale merge.
+// an old corner fuses into it; beyond 2*eps it stays a distinct vertex.
 TEST(Boolean2, NewToOldMergeBudgetIsTwoEps) {
   constexpr double eps = 1e-6;
   const Polygons a = {{{0, 0}, {10, 0}, {10, 1}, {0, 1}}};
@@ -925,8 +890,8 @@ TEST(Boolean2, NewToOldMergeBudgetIsTwoEps) {
   auto runWithLeftEdgeAt = [&](double xLeft) {
     const Polygons b = {{{xLeft, -0.5}, {0.5, -0.5}, {0.5, 0.5}, {xLeft, 0.5}}};
     auto [verts, edges] = CombinedInput(a, b, /*bMult=*/1);
-    auto result = RemoveOverlaps2D(verts, edges, eps, /*tolerance=*/eps,
-                                   /*debug=*/false, WindRule::Add);
+    auto result =
+        RemoveOverlaps2D(verts, edges, eps, /*debug=*/false, WindRule::Add);
     return std::make_pair(std::move(result), std::move(edges));
   };
   auto countNear = [](const std::vector<vec2>& points, vec2 target, double r) {
@@ -950,50 +915,13 @@ TEST(Boolean2, NewToOldMergeBudgetIsTwoEps) {
                                          distinct.numMergedVerts, eps));
 }
 
-// Regression: an elevated tolerance used to snap a shared crossing onto an old
-// corner and fold this two-star union to empty. Eps-scale snapping cannot move
-// old geometry that far, so it stays a valid ~2.04 union.
-TEST(Boolean2, ElevatedToleranceDoesNotFoldOverlap) {
-  const Polygons a = {{{-0.057041, 0.823614},
-                       {-0.285626, 0.308351},
-                       {-0.835764, 0.185488},
-                       {-0.416356, -0.191134},
-                       {-0.469509, -0.752313},
-                       {0.018284, -0.469815},
-                       {0.535572, -0.693780},
-                       {0.417637, -0.142565},
-                       {0.790491, 0.280197},
-                       {0.229810, 0.338368}}};
-  const Polygons b = {{{-0.210685, -0.690574},
-                       {-0.062302, -0.062027},
-                       {0.582263, -0.021725},
-                       {0.112117, 0.421051},
-                       {0.399497, 0.999411},
-                       {-0.219032, 0.813641},
-                       {-0.576217, 1.351700},
-                       {-0.724599, 0.723153},
-                       {-1.369164, 0.682852},
-                       {-0.899018, 0.240075},
-                       {-1.186398, -0.338285},
-                       {-0.567870, -0.152515}}};
-  const auto [verts, edges] = CombinedInput(a, b, /*bMult=*/1);
-  constexpr double eps = 1e-9;
-  const auto result = RemoveOverlaps2D(verts, edges, eps, /*tolerance=*/0.08,
-                                       /*debug=*/false, WindRule::Add);
-  EXPECT_FALSE(result.edges.empty());
-  EXPECT_TRUE(CheckRetainedGraphValidity(result, edges, result.inputVert2Merged,
-                                         result.numMergedVerts, eps));
-  EXPECT_NEAR(TotalSignedArea(OutEdgesToPolygons(result.verts, result.edges)),
-              2.04, 0.05);
-}
-
 TEST(Boolean2, RemoveOverlapsMergesExactDuplicateCoordinates) {
   const Polygons polys = {{{0, 0}, {1, 0}, {0, 1}}, {{0, 0}, {-1, 0}, {0, -1}}};
   const auto [verts, edges] = PolygonsToInput(polys);
   ASSERT_EQ(verts.size(), 6u);
   const double eps = InferEps(polys, {});
-  const auto overlap = RemoveOverlaps2D(verts, edges, eps, /*tolerance=*/0.0,
-                                        /*debug=*/false, WindRule::Add);
+  const auto overlap =
+      RemoveOverlaps2D(verts, edges, eps, /*debug=*/false, WindRule::Add);
   ASSERT_EQ(overlap.inputVert2Merged.size(), verts.size());
   ASSERT_GE(overlap.inputVert2Merged[0], 0);
   ASSERT_LT(overlap.inputVert2Merged[0], overlap.numMergedVerts);
--- a/test/boolean2_test.cpp
+++ b/test/boolean2_test.cpp
@@ -129,6 +129,11 @@ bool PointInSegmentInteriorBand(vec2 p, vec2 a, vec2 b, double eps) {
   return std::fabs(la::cross(d, p - a)) <= eps * std::sqrt(len2);
 }
 
+// Independent oracle for the arrangement RemoveOverlaps2D retains - its
+// predicates are not reused from the engine, so it can't rubber-stamp a bug.
+// Checks: retained verts finite and >eps apart; edges valid and non-eps-zero;
+// edge balance conserved (per-vertex signed multiplicity matches the input's,
+// zero for introduced verts); no two non-adjacent edges strictly cross.
 ::testing::AssertionResult CheckRetainedGraphValidity(
     const OverlapResult& result, const std::vector<EdgeM>& inputEdges,
     const std::vector<int>& inputVert2Merged, int numMergedVerts, double eps) {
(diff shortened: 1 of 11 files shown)


zmerlynn:
Cherry-picked the fix from #1757 here so these regression tests pass on the branch. Merge #1757 first; rebasing afterward drops the cherry-pick (identical to base) and leaves just the tests.

zmerlynn:
Also, sorry for the noise here: fuzz testing was stable for days, then of course the minute I put up the PR, it finds something. (ETA: but what it found was a good research project and I extended #1757 as well after that first comment.)

test/cross_section_offset_corpus_test.cpp
```diff
@@ -0,0 +1,101 @@
...
+// only throws on overlap under MANIFOLD_DEBUG, so the non-overlap teeth need a
+// debug build. Raw first pass: it runs the whole corpus, so some pathological
+// cases may still need filtering before this becomes a hard gate.
+
+namespace {
+using namespace manifold;
+
+// Read the polygon corpus (name, expectedNumTri, epsilon, numPolys, then the
+// polygons), keeping just the name and the polygon set.
+std::vector<std::pair<std::string, Polygons>> ReadCorpus(
```
elalish: We must already have this function and the next somewhere, right? Can we reuse?

test/cross_section_offset_corpus_test.cpp
```diff
@@ -0,0 +1,101 @@
...
+  const auto corpus = ReadCorpus(CorpusPath());
+  ASSERT_FALSE(corpus.empty());
+  const CrossSection::JoinType joins[] = {
+      CrossSection::JoinType::Round, CrossSection::JoinType::Miter,
+      CrossSection::JoinType::Square, CrossSection::JoinType::Bevel};
+  int nonEmpty = 0;
+  for (const auto& entry : corpus) {
+    const CrossSection cs(entry.second);
+    const vec2 size = cs.Bounds().Size();
+    const double diag = std::sqrt(size.x * size.x + size.y * size.y);
```
elalish: `la::length(size)`?

test/cross_section_offset_corpus_test.cpp
```diff
@@ -0,0 +1,101 @@
...
+    if (!std::isfinite(diag) || diag <= 0) continue;
+    // Offset out and in, by fractions of the bbox diagonal, on every join type.
+    for (const double frac : {0.05, -0.02}) {
+      for (const CrossSection::JoinType jt : joins) {
+        SCOPED_TRACE(entry.first + " frac=" + std::to_string(frac));
+        const CrossSection off = cs.Offset(frac * diag, jt);
+        const Polygons result = off.ToPolygons();
+        if (!result.empty()) ++nonEmpty;
```
elalish, line 68: Could we tighten this further to say an outset has more area than the input and an inset less? In fact, can we also say the join types are always ordered by area? Seems like Miter should be the largest area, followed by Round, then Square and Bevel, though not quite sure what order those are in - can never remember the distinction. Makes me wonder if we should have both in our API...

test/cross_section_test.cpp
```diff
@@ -16,14 +16,152 @@
...
+// Expand run-length radius specs for seeds with long repeated-radius runs:
+// Runs({{3, 1000.}, {21, 332.47}}) is 3x 1000. then 21x 332.47.
+std::vector<double> Runs(std::initializer_list<std::pair<int, double>> runs) {
+  std::vector<double> out;
+  for (const auto& run : runs) out.insert(out.end(), run.first, run.second);
+  return out;
+}
+
+double AreaTol(const CrossSection& a, const CrossSection& b) {
+  return 1e-6 * (1.0 + std::fabs(a.Area()) + std::fabs(b.Area()));
```
elalish, line 61: If these functions are only called once, let's just inline them.

test/cross_section_test.cpp
```diff
@@ -16,14 +16,152 @@
...
+    const CrossSection b(shiftedFeature);
+    if (a.IsEmpty() || b.IsEmpty()) continue;
+
+    const auto aUb = a + b;
+    const auto aIb = a.Boolean(b, OpType::Intersect);
+    const CrossSection soup(Polygons{shiftedHost, shiftedFeature});
+    const double scale = 1.0 + std::fabs(a.Area()) + std::fabs(b.Area());
+    EXPECT_NEAR(aUb.Area(), a.Area() + b.Area() - aIb.Area(), 1e-3 * scale)
+        << "inclusion-exclusion violated at offset " << offset;
+    EXPECT_NEAR(soup.Area(), aUb.Area(), 1e-5 * scale)
```
elalish, line 140: Where do these various near tolerance values come from? If you're really just testing for the presence/absence of a feature, would a bounding box check be more precise?

test/cross_section_test.cpp
```diff
@@ -93,6 +231,1695 @@ TEST(CrossSection, BevelOffset) {
...
+  auto square = CrossSection::Square({20., 20.}, true);
+  auto offset = square.Offset(5., CrossSection::JoinType::Miter);
+  auto result = Manifold::Extrude(offset.ToPolygons(), 1.);
+
+  EXPECT_EQ(result.Genus(), 0);
+  EXPECT_NEAR(result.Volume(), 30. * 30., 0.01);
+  // Offset keeps collinear verts on the miter edges; stripping them is
+  // Simplify's job, so check the simplified corner count (4)
+  // backend-agnostically.
+  EXPECT_EQ(offset.Simplify().NumVert(), 4);
```
elalish, line 252: Based on the comment, perhaps we should also check that NumVert is a different value before Simplify?

test/cross_section_test.cpp
```diff
@@ -93,6 +231,1695 @@ TEST(CrossSection, BevelOffset) {
...
+}
+
+TEST(CrossSection, OffsetWithHole) {
+  SimplePolygon outer = {{-10, -10}, {10, -10}, {10, 10}, {-10, 10}};
+  SimplePolygon hole = {{-2, -2}, {-2, 2}, {2, 2}, {2, -2}};
+  // CCW outer + CW hole are already oriented for Positive fill, which yields
+  // the same annulus as EvenOdd here and is the only rule CrossSection
+  // supports.
+  CrossSection cs({outer, hole}, CrossSection::FillRule::Positive);
```
elalish, line 261: We're removing these fill rules from the public API right? I suppose that can happen in a separate PR if you want, but just checking.

test/cross_section_test.cpp
```diff
@@ -93,6 +231,1695 @@ TEST(CrossSection, BevelOffset) {
...
+  for (double angle : {0., 45., 90., 135.}) polys.push_back(rhomb(angle));
+  CrossSection cs(polys, CrossSection::FillRule::Positive);
+
+  EXPECT_EQ(cs.NumContour(), 1);
+  EXPECT_NEAR(cs.Area(), 0.644423, 1e-4);
+  for (const auto& loop : cs.ToPolygons()) {
+    for (const vec2& v : loop) {
+      EXPECT_TRUE(std::isfinite(v.x));
+      EXPECT_TRUE(std::isfinite(v.y));
+      EXPECT_LE(la::length(v), 1.05);
```
elalish: Seems a little loose? Unless I misunderstand this test. Also, I don't think you need to check for finite, since NaN should trip this expect anyway.

test/cross_section_test.cpp
```diff
@@ -93,6 +231,1695 @@ TEST(CrossSection, BevelOffset) {
...
+    v.y += 4096.0;
+  }
+  for (auto& v : feature) {
+    v.x += 4096.0;
+    v.y += 4096.0;
+  }
+  const CrossSection ca(host), cb(feature);
+  const auto sum =
+      ca.Area() + cb.Area() - ca.Boolean(cb, OpType::Intersect).Area();
+  EXPECT_NEAR((ca + cb).Area(), sum, 1e-3 * (1.0 + ca.Area() + cb.Area()));
```
elalish: Can you show an image of the problem here? I'm curious what the "correct" answer should be here. I'm also curious if the Area calculations are introducing too much error of their own to be reliable tests here. And from your comment it sounds like you expect the intersection to be empty, so we might EXPECT that as well? And where does 1e-3 come from? Does it relate to the eps values of the inputs somehow?

test/cross_section_test.cpp
```diff
@@ -93,6 +231,1695 @@ TEST(CrossSection, BevelOffset) {
...
+}
+
+// Host-drop only at large offset (passes at the origin): the StarRing host plus
+// an 8-vertex feature anchored 1e-9 from host[1].
+TEST(CrossSection, DISABLED_TinyFeatureNearCornerHostDropAtOffset1024) {
```
elalish, line 477: Ditto here.

test/boolean2_test.cpp
```diff
@@ -0,0 +1,1161 @@
...
+// distance-from-line band: |cross(d, p - a)| <= eps * |d|.
+bool ValidVertId(int v, const std::vector<vec2>& verts) {
+  return 0 <= v && v < static_cast<int>(verts.size());
+}
+
+bool PointsCollinearWithinEps(vec2 a, vec2 b, vec2 p, double eps) {
+  const vec2 d = b - a;
+  const double len2 = dot(d, d);
+  if (len2 == 0.0) return false;
+  return std::fabs(la::cross(d, p - a)) <= eps * std::sqrt(len2);
```
elalish: What about when a and b are less than eps apart and p is distant from them? That's why CCW is built how it is - I believe this function will sometimes say they are not collinear. Unless you restrict the kind of input you give this function?

test/boolean2_test.cpp
```diff
@@ -0,0 +1,1161 @@
...
+bool PointInSegmentInteriorBand(vec2 p, vec2 a, vec2 b, double eps) {
+  const vec2 d = b - a;
+  const double len2 = dot(d, d);
+  if (len2 == 0.0) return false;
+  const double along = dot(p - a, d);
+  if (along <= 0.0 || along >= len2) return false;
+  return std::fabs(la::cross(d, p - a)) <= eps * std::sqrt(len2);
+}
+
+::testing::AssertionResult CheckRetainedGraphValidity(
```
elalish, line 141: This looks pretty complicated - can we have a description of what it's doing? Also, I subscribe to the philosophy that simple testing is preferable, since otherwise there is uncertainty when a test fails as to whether the bug is in the algorithm or the test itself. However, you've been using these for awhile, so you probably know better than I what this has been useful in catching.

test/boolean2_test.cpp
```diff
@@ -0,0 +1,1161 @@
...
+  }
+  return edges;
+}
+
+OverlapResult CleanupPassLikeIterate(const OverlapResult& result, double eps) {
+  return RemoveOverlaps2D(result.verts, EdgesFromOverlapResult(result), eps,
+                          /*debug=*/false, WindRule::Add);
+}
+
+void ExpectSameFingerprint(const OverlapResult& a, const OverlapResult& b,
```
elalish, line 287: Some description here would be nice too.

test/boolean2_test.cpp
```diff
@@ -0,0 +1,1161 @@
...
+  IntersectionInsertion inserted = FindAndInsertIntersections(
+      edges, std::move(verts), std::move(narrow.lists), eps, edgeBoxes, BVH{},
+      narrow.intersections);
+
+  ASSERT_EQ(inserted.verts.size(), 12);
+  EXPECT_EQ(inserted.lists[4].size(), 2);
+}
+
+// RemoveOverlaps2D produces an edge-balance imbalance for some
+//   vertices on these mixed-scale inputs - 1e-6 alongside 1024.
```
elalish: What is edge balance?

elalish (changes requested):
Thanks, really helpful to see all this. I've scanned over these pretty quickly, so please highlight any particularly interesting ones that I haven't commented on.

test/cross_section_offset_corpus_test.cpp
zmerlynn: Extracted a shared `ReadPolygonCorpus` into `test/polygon_corpus.h`; offset_corpus and polygon_test now use it (with a guard so a missing corpus still fails loudly rather than silently registering no tests). polygon_fuzz keeps its own copy (separate fuzz harness, doesn't pull the gtest test infra) - can adopt it in a follow-up.

test/cross_section_offset_corpus_test.cpp
zmerlynn: Done - and while here I did a broader `la::` pass across the boolean2 tests and source (length/cross/isfinite/maxelem, vector add/sub); an earlier review pass had missed a bunch of these.

test/cross_section_offset_corpus_test.cpp
zmerlynn: Tightened - asserted outset >= input >= inset per join type, plus the outset join-type ordering, which the corpus confirms is Miter >= Square >= Round >= Bevel (not what I'd have guessed either). The Square-vs-Bevel API question is fair but out of scope here.

test/cross_section_test.cpp
zmerlynn: Right - `ExpectBooleanInvariants` and `ExpectDistributesOverUnion` were each single-use; inlined both. `AreaTol` stays (used across ~10 tests).

test/cross_section_test.cpp
zmerlynn: The tolerances are area-relative (`scale = 1+|a|+|b|`): `1e-3` for the inclusion-exclusion identity (generous - it's a sub-eps near-coincidence) and `1e-5` for parity between the binary `a + b` and building one CrossSection from both polygon sets at once (same arrangement, so tighter). A bbox check wouldn't discriminate here - the tiny polygon sits right at the big one's corner, so dropping it barely moves the union's bbox; the lost area is what catches it.

test/cross_section_test.cpp
zmerlynn: Added `EXPECT_GT(offset.NumVert(), 4)` before the Simplify check, so it's explicit that Simplify is what reduces to 4.

test/cross_section_test.cpp
zmerlynn: Good to confirm - yes, my understanding is the same (dropping to Positive-only). I've kept the tests using it here so that API change can land separately.

test/cross_section_test.cpp
zmerlynn: Dropped the finite checks (NaN trips the bound anyway), and tightened to `1.0 + 1e-6` - the rhombi are on the unit circle, so the union is too; it's a no-blowup guard.

test/cross_section_test.cpp
zmerlynn: Good questions - in order:

The two pieces are genuinely disjoint. The tiny piece's nearest vertex sits ~1e-9 from one of the big piece's vertices but doesn't cross it, so the correct union is just the two added together and the intersection is empty. The bug drops one of the two whole pieces (here the ~30107-area big one), so the failure is a huge area swing, not Area-calculation noise - which is also why a loose bound is fine here.

Added the explicit `EXPECT_NEAR(inter.Area(), 0, tol)` you suggested.

1e-3 isn't eps-derived - it's just a loose area-relative bound (`1e-3 * (1 + areas)`), generous on purpose since the bug blows right past it. I'm still researching the underlying winding issue, so these stay DISABLED_ for now; I'll try to get a trace-viewer link up soon so the drop is actually visible.

test/cross_section_test.cpp
zmerlynn: Same answers as above. This one's the offset-sweep variant: it passes at the origin and only drops the piece once everything's shifted out to a large coordinate, which is why it sweeps {0, 1024, 4096} - same winding instability, surfaced by absolute-coordinate magnitude rather than the merge details. Same loose bound and explicit intersect-empty EXPECT; stays DISABLED_ while I'm still digging into the winding issue.

test/boolean2_test.cpp
zmerlynn: Good catch - you're right it'd call them non-collinear in that case. It's saved by the input: these oracle predicates only run on retained output, where the validator has already dropped any eps-zero edge, so |b - a| > eps. I added a note making that precondition explicit - it's not meant as a general collinearity test.

test/boolean2_test.cpp
zmerlynn: Added a contract. The aim is an independent oracle - predicates not reused from the engine - that's caught missed splits, dropped/stacked verts and sign errors area checks miss. Happy to trim anything murky.

test/boolean2_test.cpp
zmerlynn: Added one - a determinism/idempotence check (arrangements equal up to eps-quantization).

test/boolean2_test.cpp
zmerlynn: Signed sum of incident edge multiplicities at a vertex (out minus in); a valid arrangement conserves it. Defined in the comment, and fixed the stale wording - the test asserts the balance holds.

zmerlynn:
Thanks for the review. A few tests that didn't get their own comments but are worth a look:

- `SubtractInvariantsEmptyIntersectionDrop` - standalone regression for the conservation law `area(A - B) + area(A intersect B) == area(A)`, on a near-degenerate StarRing pair where the empty-intersection path used to drop area.
- `BooleanAssociativitySeeds` and `BooleanDistributivitySeeds` - table-driven associativity of union and distributivity of intersection over union, over fuzzer-derived seeds. These are the strongest checks in the file: the algebraic identity is its own oracle, so there's no precomputed expected area to get wrong.
- `SubtractInvariantsSeeds` - the subtraction conservation law across a seed table (one seed, `DominantSpikeStars`, checks the inclusion-exclusion identity instead).

They're all area-based algebraic invariants rather than fixed-value assertions, so they keep holding as the arrangement internals shift.

test/cross_section_test.cpp
zmerlynn: Revising my earlier reply here - I've made the area check independent of the engine since.

The bbox answer still holds: the small polygon sits at the big one's corner, so dropping it barely moves the union's bbox; the lost area is what catches it. But I was measuring that area with inclusion-exclusion, every term from the same engine - self-referential, so a uniform collapse passes `0 == 0 + 0 - 0` vacuously. I added an independent floor: the union must retain at least the raw shoelace area of the larger input, computed from the input coordinates with no engine call (`ExpectUnionRetainsArea`). That catches both a dropped polygon and a full collapse.

test/cross_section_test.cpp
```diff
@@ -250,6 +2268,66 @@ TEST(CrossSection, BatchBoolean) {
   EXPECT_FLOAT_EQ(subtract.NumVert(), 42);
 }
 
+// A is a 37-vertex star with many zero-radius vertices producing a dense
+// near-degenerate shape; B is a 5-vertex star translated by ~5e-5 in x. The
+// intersection A ∩ B returns area=0 even though both inputs are non-empty and
+// overlap geometrically. (A - B) + (A ∩ B) ends up at 39605 versus
+// a.Area=90895, missing ~51290 area units. The tiny x translation puts B's
+// vertices just inside the eps band of A's; the intersect path appears to
+// mis-classify the overlap region as empty.
```
elalish: I'm confused - the comment says there's a problem, but the test appears to be passing. Did this get fixed? If so, perhaps update the comment?

test/cross_section_test.cpp
```diff
@@ -93,6 +214,1749 @@ TEST(CrossSection, BevelOffset) {
...
+    {"TinyStars",
+     {{0., 0., 0., 2.9078938473355343e-13}},
+     {{1.4795645772678251e-12, 2.4342935254638573e-13, 1.4800031437954507e-12,
+       1.1368244372782033e-305},
+      {0., 0.}},
+     {{0., 0., 0., 0.}}},
+};
+}  // namespace
+
+TEST(CrossSection, BooleanAssociativitySeeds) {
```
elalish, line 1006: What is the distinction between `cross_section_test` and `boolean2_test`, particularly if this test is here?

test/cross_section_test.cpp
```diff
@@ -93,6 +214,1749 @@ TEST(CrossSection, BevelOffset) {
...
+      {0., 4.667562921730494e-33}}},
+    // Two spiky stars at origin - A
+    //   is 5-pointed with mixed magnitudes including two ~1000-radius
+    //   spikes and a near-zero vertex; B is 4-pointed with three
+    //   ~1000-scale spikes plus a near-zero vertex. One of the boolean
+    //   algebraic invariants fails. Likely a different code path than
+    //   the TinyVsLargeStars needle case - this is two large spiky
+    //   shapes, not a tiny shape vs a needle. May share the off-axis
+    //   T-junction root cause diagnosed earlier in this table, but the spike-
+    //   collision geometry could be its own failure mode.
```
elalish: Can we update all these comments to their resolution?

test/cross_section_test.cpp
```diff
@@ -93,6 +214,1749 @@ TEST(CrossSection, BevelOffset) {
...
+}  // namespace
+
+TEST(CrossSection, SubtractInvariantsSeeds) {
+  for (const auto& c : kSubtractInvariantsSeeds) {
+    SCOPED_TRACE(c.name);
+    // Flush the seed name so a hard crash is still attributable.
+    std::cerr << "[seed] " << c.name << std::endl;
+    const CrossSection a = MakeShape(c.a), b = MakeShape(c.b);
+    if (c.kind == SubtractKind::InclusionExclusion) {
+      const auto aIb = a.Boolean(b, OpType::Intersect);
```
elalish: Can we use `a ^ b` instead?

test/cross_section_test.cpp
```diff
@@ -93,6 +214,1749 @@ TEST(CrossSection, BevelOffset) {
...
+      const auto aUb = a + b;
+      const double tol = AreaTol(a, b);
+      EXPECT_NEAR(aUb.Area(), a.Area() + b.Area() - aIb.Area(), tol)
+          << "inclusion-exclusion violated";
```
elalish: Why does this branch test fewer things? Do they fail?

elalish (approved):
This is looking pretty good, perhaps we should get this merged so we can take a more serious look at https://github.com/elalish/manifold/issues/1707#issuecomment-4701435633?

zmerlynn:
> This is looking pretty good, perhaps we should get this merged so we can take a more serious look at https://github.com/elalish/manifold/issues/1707#issuecomment-4701435633?

Yup, I will work on cleaning this up. Ignore the force push just now, just reconciling branches- no new content yet.

zmerlynn:
Sorry for the churn - I hadn't run the full style pass on these test files before putting the PR up, which is why a couple of your comments turned into broader cleanup. Done now: one anonymous namespace instead of six, the fuzz-diagnosis prose trimmed out of the seed comments, `^` in the subtract tests, and the `SubtractInvariants` kind split collapsed (every seed runs all three identities). Replies inline.

test/cross_section_test.cpp
zmerlynn: Fixed - passes on the boolean2 backend this is built on. I didn't bisect which commit closed it, so I left the attribution out of the comment and just rewrote it to past tense.

test/cross_section_test.cpp
zmerlynn: Good call - trimmed each to the distinguishing geometry plus the past-tense outcome, dropped the speculative root-cause notes. All green now.

test/cross_section_test.cpp
zmerlynn: Done in the subtract-invariant tests (here and `SubtractInvariantsEmptyIntersectionDrop`). There are other `Boolean(.., Intersect)` sites in the file if you'd like the whole thing converted - happy to, just kept this focused.

test/cross_section_test.cpp
zmerlynn: They hold - tested it (subtract residuals ~1e-14 against a ~3e-6 tol). Inclusion-exclusion was just the identity that originally broke for this seed, so its standalone test only checked that one. Unified the loop to run all three for every seed and dropped the `SubtractKind` split.

test/cross_section_test.cpp
zmerlynn: Good point - moved the one direct-engine test (the `Boolean2D` non-finite check) to `boolean2_test` in a separate commit, so this file is now just the `CrossSection` public API while `boolean2_test` keeps the engine internals. This seed's a public-API algebraic identity, so it stays here.

test/cross_section_test.cpp
elalish: Yes, I generally prefer to update all related code at once, simply so it's not forgotten. Not a big deal though.

elalish (approved):
LGTM - I'll probably run across other things in here eventually, but better to have it in and iterate. Thanks!

PR #1756, opened 2026-06-11
Drop using directives from boolean2.cpp

zmerlynn:
Follow-up to the review note on #1754 (https://github.com/elalish/manifold/pull/1754#discussion_r3395019386): drop the `using` directives/declarations from `boolean2.cpp`.

- `using manifold::la::dot;` removed - the `dot()` calls resolve by ADL on their `la::vec` arguments.
- `using collider_internal::SpreadBits3;` and `using namespace collider_internal;` removed - the handful of `collider_internal::` symbols are now spelled out at their call sites.

Pure qualification, no behavior change. Full suite green.

### DIFF
--- a/src/boolean2.cpp
+++ b/src/boolean2.cpp
@@ -282,19 +282,19 @@ Box2 BoxOf2DEdge(vec2 p0, vec2 p1, double eps) {
 }
 
 uint32_t MortonCode2(vec2 position, Box2 bBox) {
-  using collider_internal::SpreadBits3;
   const vec2 size = bBox.max - bBox.min;
   const double xNorm = size.x > 0 ? (position.x - bBox.min.x) / size.x : 0.5;
   const double yNorm = size.y > 0 ? (position.y - bBox.min.y) / size.y : 0.5;
   const double xClamped = std::min(1023.0, std::max(0.0, 1024.0 * xNorm));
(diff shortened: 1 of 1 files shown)


src/boolean2.cpp
```diff
@@ -282,19 +282,19 @@ Box2 BoxOf2DEdge(vec2 p0, vec2 p1, double eps) {
...
-  using collider_internal::SpreadBits3;
   const vec2 size = bBox.max - bBox.min;
   const double xNorm = size.x > 0 ? (position.x - bBox.min.x) / size.x : 0.5;
   const double yNorm = size.y > 0 ? (position.y - bBox.min.y) / size.y : 0.5;
   const double xClamped = std::min(1023.0, std::max(0.0, 1024.0 * xNorm));
   const double yClamped = std::min(1023.0, std::max(0.0, 1024.0 * yNorm));
-  const uint32_t x = SpreadBits3(static_cast<uint32_t>(xClamped));
-  const uint32_t y = SpreadBits3(static_cast<uint32_t>(yClamped));
+  const uint32_t x =
+      collider_internal::SpreadBits3(static_cast<uint32_t>(xClamped));
```
elalish, line 291: This isn't really a 2D Morton code if it's using SpreadBits3. Not a big deal, but we should probably either factor these colliders into the same file, or just use the 3D one for 2D directly.

elalish (approved):
Thanks

PR #1757, opened 2026-06-11
Make the intersection-vertex merge eps-only

zmerlynn:
MergeNearbyIntersectionVerts merged intersection verts at two scales, and each
could annihilate a valid arrangement:

- new-to-new: fused crossings sharing an edge within 10*eps
  (kIntersectionMergeEpsFactor). A tiny feature near a host vertex crosses one
  host edge at two distinct points a few eps apart; fusing them pinches the
  arrangement and the winding filter collapses it to empty
  (CrossSectionFuzz.TinyFeatureNearCorner). Dropped the reach to eps.

- new-to-old: snapped a fresh intersection onto an old endpoint within
  tolerance+eps and emitted the cluster centroid, dragging the old endpoint by up
  to ~tolerance. An elevated (propagated) tolerance then folds the arrangement -
  an old corner crosses an incident edge, the winding inverts, and the filter
  returns empty. Two clean ~1-area stars Add to 0.0 at tolerance >= 0.08. Snap at
  2*eps instead and drop tolerance from the merge.

The merge is now eps-only. A near-coincidence more than ~eps apart is a real tiny
feature the arrangement keeps; decimating it is Simplify's job, not the boolean's
(tolerance no longer affects the arrangement at all).

Validated: full Boolean2 + CrossSection suite green, including
SubtractInvariantsEmptyIntersectionDrop (the near-coincidence the old tolerance
snap was added to resolve). Tests on #1755.

### DIFF
--- a/src/boolean2.cpp
+++ b/src/boolean2.cpp
@@ -138,7 +138,7 @@ void AppendInput(const Polygons& polys, int mult, std::vector<vec2>& verts,
 
 // `bSign` is +1 for Add/Intersect-style accumulation and -1 for Subtract.
 Polygons ApplyFillRule(const Polygons& a, const Polygons& b, int bSign,
-                       WindRule rule, double eps, double tolerance) {
+                       WindRule rule, double eps) {
   DEBUG_ASSERT(bSign == 1 || bSign == -1, logicErr,
                "Boolean2 input multiplicity must be +/-1");
   if (!AllFinite(a) || !AllFinite(b)) return {};
@@ -146,7 +146,6 @@ Polygons ApplyFillRule(const Polygons& a, const Polygons& b, int bSign,
   Polygons localA = TranslatePolygons(a, -origin);
   Polygons localB = TranslatePolygons(b, -origin);
   if (eps <= 0.0) eps = InferEps(localA, localB);
-  if (tolerance < eps) tolerance = eps;
 
   std::vector<vec2> verts;
   std::vector<EdgeM> edges;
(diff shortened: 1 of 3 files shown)


zmerlynn:
Expanded this beyond the new-to-new pinch to also cover the new-to-old fold. The merge was snapping a fresh intersection onto an old endpoint at tolerance+eps and emitting the centroid, which drags the old vertex by ~tolerance and folds the arrangement under an elevated tolerance. It now snaps at 2*eps and drops tolerance from the merge, so the whole pass is eps-only. Retitled and updated the description to cover both modes; tests on #1755.

zmerlynn:
One more annihilation in this merge, courtesy of the fuzzer. A tiny feature anchored ~1e-9 from a host vertex made a union drop the whole host - `a + b` returned the feature's area alone. The new-to-old cluster emits its centroid, drifting the old host vertex by ~eps/2; at that near-coincident tip the sub-eps drift folds the winding face-walk, so the host's interior face never forms. Fix: pin a cluster containing an old vertex to that vertex (nearest the centroid) instead of averaging.

This also brings the new-to-old merge in line with the rest of the design as I read it: the #289 sketch snaps crossings onto neighboring vertices and `MergeVerts` already pins clusters to a source vertex - the centroid here was the one spot that averaged old geometry. Pre-existing (the old tolerance+eps merge centroided too), not introduced here.

Validated: suite green incl. `SubtractInvariantsEmptyIntersectionDrop`, and the boolean fuzz properties replay clean over the producer corpus. Test on #1755.

zmerlynn:
Reverted the pin (last commit) - it didn't really fix the host-drop, just relocated it (a feature-drop on the same input, plus a new drop at large offsets), so the branch is eps-only/centroid again. The point-contact winding issue behind it is captured as `DISABLED_` seeds in #1755 and tracked separately. I'll stop touching this PR now - sorry for the churn.

elalish (approved):
Thanks, this simplification makes me feel better about our robustness. I'm having a little trouble following your prose descriptions of the problems you're facing/fixing. The interactive polygon website you made feels quite promising for this kind of thing, but maybe needs a touch more description, or maybe just screenshots with trouble areas highlighted? Would be good if we can work out a quick way to compare notes on problem polygons.

PR #1758, opened 2026-06-16
Weekly benchmarks

AnshulPatil2005:
Adds weekly benchmark workflow and reporting structure.

ref-#1738


AnshulPatil2005:
I am currently working on running the Ember benchmarks in the workflow first. Once that is done, I will move on to adding the other tests.

The structure is similar to the previous benchmark guard CI, with a few changes for the Ember cases.

pca006132:
did you try it on your fork? say, once a day, to see if it works?

AnshulPatil2005:
> did you try it on your fork? say, once a day, to see if it works?

I tested it manually on my fork and the workflow passed. I also added a temporary schedule on the fork default branch to test the scheduled trigger I set it at a very short time(10 mins), but it did not emit any runs so i increased the time so see if it would work that way.

i checked regarding this and i found on github's docs that scheduled workflows are best-effort and can be delayed or dropped under load, I will leave it enabled for longer and confirm once the scheduled run appears.

but this seems to be very unreliable, what if it dropped the weekly run ?

AnshulPatil2005:
Update: the scheduled workflow is working on my fork now. I changed the temporary schedule to a 12-hour interval, and it has produced 3 scheduled Ember benchmark runs so far.

AnshulPatil2005:
I have run the workflow several times on my fork now(using schedule).

The weekly benchmark  currently lands on different runners, so the absolute timings vary between runs. I will compile the results from these runs to estimate how much difference this is causing.

scripts/run_weekly_benchmarks.py
```diff
@@ -0,0 +1,299 @@
...
+import json
+import os
+import shutil
+import subprocess
+import sys
+import urllib.request
+from dataclasses import dataclass
+from pathlib import Path
+
+THINGI10K_URL = (
```
pca006132: I think if we want to run them regularly, we should probably just bring in the meshes.

scripts/run_weekly_benchmarks.py
```diff
@@ -0,0 +1,299 @@
...
+        "--threads",
+        str(threads),
+        "--verbose",
+        str(verbose),
+    ]
+    result = subprocess.run(args, check=True, capture_output=True, text=True, cwd=cwd)
+    return result.stdout
+
+
+def run_ember_suite(ctx: BuildContext, binary: Path) -> None:
```
pca006132, line 194: also, if I understand this correctly, this is kind of just what the python script in `extras/ember_tests` is doing? maybe we can remove that script if this is sufficient?

scripts/run_weekly_benchmarks.py
```diff
@@ -0,0 +1,299 @@
...
+    for suite in SUITES:
+        print(f"::group::Run weekly benchmark suite: {suite.name}")
+        binary = find_binary(ctx, suite.binary)
+        suite.run(ctx, binary)
+        print("::endgroup::")
+    return 0
+
+
+if __name__ == "__main__":
+    raise SystemExit(main())
```
pca006132: This is interesting. I thought we typically just do `sys.exit(code)`. I have never seen this `SystemExit` exception... Googling tells me they are the same, but this raise looks extremely suspicious. Maybe just use `sys.exit`.

scripts/parse_weekly_benchmarks.py
```diff
@@ -0,0 +1,467 @@
...
+        "mean_ms": mean(values),
+        "median_ms": statistics.median(values),
+        "stdev_ms": stdev(values),
+        "min_ms": min(values),
+        "max_ms": max(values),
+        "n_runs": len(values),
+    }
+
+
+def summarize_ratio(values: list[float]) -> dict:
```
pca006132: so this is effectively the same thing except no `_ms` in the field names?

scripts/parse_weekly_benchmarks.py
```diff
@@ -0,0 +1,467 @@
...
+        raise RuntimeError(f"No Ember benchmark cases found in {run_path}")
+
+    for case in cases:
+        for phase in INDEPENDENT_PHASES:
+            case["phases_ms"].setdefault(phase, 0.0)
+
+    return {"path": str(run_path), "run_index": run_index, "cases": cases}
+
+
+def parse_ember_suite(suite_dir: Path) -> dict:
```
pca006132: why not just use `ember-benchmark-cases.json`?

scripts/parse_weekly_benchmarks.py
```diff
@@ -0,0 +1,467 @@
...
+        "type": "existing_gtests",
+        "description": "Selected existing regression tests used as weekly benchmark signals",
+        "test_order": test_order,
+        "tests": tests,
+        "runs": runs,
+    }
+
+
+def detect_compiler() -> str | None:
+    for binary in ("c++", "g++", "clang++"):
```
pca006132: There is no guarantee that cmake uses this compiler. Why not just parse cmake output? we have the info:

```
-- ====================================
-- Manifold Build Configuration Summary
-- ====================================
--  
-- Environment: Linux
--  
-- CMAKE_VERSION:                 4.1.2
-- CMAKE_TOOLCHAIN_FILE:          
-- CMAKE_GENERATOR:               Ninja
-- CPACK_CMAKE_GENERATOR:         
-- CMAKE_BUILD_TYPE:              RelWithDebInfo
-- CMAKE_PREFIX_PATH:             
-- CMAKE_CXX_COMPILER_ID:         GNU
-- CMAKE_CXX_COMPILER_VERSION:    15.2.0
-- BUILD_SHARED_LIBS:             ON
--  
-- MANIFOLD_VERSION:              3.5.1
-- MANIFOLD_PAR:                  ON
-- MANIFOLD_TEST:                 ON
-- MANIFOLD_DEBUG:                OFF
-- MANIFOLD_ASSERT:               OFF
-- MANIFOLD_CBIND:                ON
-- MANIFOLD_PYBIND:               OFF
-- MANIFOLD_JSBIND:               OFF
-- MANIFOLD_FLAGS:                -Wall;-Wno-unknown-warning-option;-Wno-unused;-Wno-shorten-64-to-32;-Wno-sign-compare;-Wno-alloc-size-larger-than;-O3;-fno-omit-frame-pointer;-ffp-contract=off;-fexcess-precision=standard
-- MANIFOLD_COMPILE_OPTIONS:      -Wall;-Wno-unknown-warning-option;-Wno-unused;-Wno-shorten-64-to-32;-Wno-sign-compare;-Wno-alloc-size-larger-than;-O3;-fno-omit-frame-pointer;-ffp-contract=off;-fexcess-precision=standard
-- MANIFOLD_LINK_OPTIONS:         
-- MANIFOLD_DOWNLOADS:            ON
-- MANIFOLD_USE_BUILTIN_TBB:      OFF
-- MANIFOLD_USE_BUILTIN_NANOBIND: OFF
-- MANIFOLD_FUZZ:                 OFF
-- TRACY_ENABLE:                  ON
-- TRACY_MEMORY_USAGE:            OFF
-- ASSIMP_ENABLE:                 OFF
```

scripts/report_weekly_benchmarks.sh
```diff
@@ -0,0 +1,38 @@
+#!/usr/bin/env bash
```
pca006132: is there a reason why we don't want to do everything in python?

scripts/run_weekly_benchmarks.py
AnshulPatil2005: Yes i planned adding it.
store them in the repo and pass the directory as source_dir

scripts/run_weekly_benchmarks.py
AnshulPatil2005: yes, we can remove the do_ember_tests.py in ember_tests : run_ember_suite(new script) is an adapted benchmark version that supersedes it.

from what i saw in do_ember_tests.py it's a multi-thread correctness sweep. It runs all cases across 9 thread counts [1, 2, 4, 6, 8, 10, 12, 14, 16] just to check that man_bench doesn't crash or return a bad exit code.

run_ember_suite only runs at a single thread count by default but it could do the sweep by changing the WEEKLY_BENCHMARK_THREADS.

scripts/run_weekly_benchmarks.py
AnshulPatil2005: yes they are same, ill use sys.exit

scripts/parse_weekly_benchmarks.py
AnshulPatil2005: Yes,it is the same, but the values are ratios/shares rather than timings, so I kept the field names without `_ms` to avoid making the JSON look like these are millisecond values.

scripts/parse_weekly_benchmarks.py
AnshulPatil2005: Yes, that makes sense. I can change the parser to load the JSON file and use the selected case indices for metadata, while keeping the run logs only for timing output.

scripts/parse_weekly_benchmarks.py
AnshulPatil2005: Yes will do that

scripts/report_weekly_benchmarks.sh
AnshulPatil2005: No strong reason. I kept the small shell wrapper as it was convenient for printing the summary/result/raw logs and appending to `GITHUB_STEP_SUMMARY`, but this can be moved into the Python parser/runner too.

doing everything in Python would probably be cleaner

scripts/parse_weekly_benchmarks.py
pca006132: You have a simpler way to do it: just map the field names, excluding the final count.

scripts/run_weekly_benchmarks.py
pca006132: `do_ember_test.py` is not really that much about correctness, the sweep was to check multithreaded performance.

scripts/run_weekly_benchmarks.py
pca006132: Note that we only want the required files. Bringing in everything will be too much.

scripts/run_weekly_benchmarks.py
AnshulPatil2005: oh, in that case we can remove it.

scripts/run_weekly_benchmarks.py
AnshulPatil2005: yes I will only include the STL files needed by the selected Ember benchmark cases

AnshulPatil2005:
[xlx_processed.xlsx](https://github.com/user-attachments/files/29640998/xlx_processed.xlsx)
image

image

image

MacOS(named "arm" here) seems to be more noisier due to the OS scheduler moving the benchmark thread to an E-core(efficiency core rather than performance core) for even one repeat, that repeat is 3–4× slower. 

@pca006132 @elalish

elalish:
Nice plots, though the labels could use a little work. If "arm" is MacOS, maybe just call it that? And what are EPYC####? And what are C###? Are those particular test models?

AnshulPatil2005:
> Nice plots, though the labels could use a little work. If "arm" is MacOS, maybe just call it that? And what are EPYC####? And what are C###? Are those particular test models?

Yes, I should have mentioned the label meanings. This was discussed in the previous meeting, so I missed adding the context here.

`EPYC####` refers to the GitHub-hosted Linux runner CPU models that were assigned during CI runs. `arm` is the macOS 15 ARM runner; I should rename that label to `macOS 15` to make it clearer.

I am also avoiding `macos-latest` for the weekly benchmark because the macOS version can change over time and add extra noise.

And yes, `C667` means Ember benchmark case `667`.

Let me know any improvements or confusion is there, I will change it accordingly

pca006132:
@AnshulPatil2005 what about the macos timing anomaly for the largest sphere in perfTest? Any idea about that?

AnshulPatil2005:
> @AnshulPatil2005 what about the macos timing anomaly for the largest sphere in perfTest? Any idea about that?

Root cause is the macOS-15 runners run on Apple Silicon (M1/M2/M3). The OS moves threads from P-cores (~3.5GHz) to E-cores (~2.0GHz) at any point during execution. A 33-second computation gives the scheduler many opportunities to do this so a mid-run migration to E-cores can cost 10–15 seconds.

I think we could try pinning the process to P-cores only if thats possible in macOS

pca006132:
Interesting. How did you check that?

AnshulPatil2005:
> Interesting. How did you check that?

ok so, i found this
P-cores run at ~4,512 MHz; E-cores run at ~1,050 MHz for background tasks which is a 4.3× frequency ratio

and the Observed worst-case ratio for nTri=512: 4.60ms ÷ 1.10ms = 4.16× which is very similar to that ratio 

 For nTri=8388608: Observed ratio: 44.16s ÷ 20.17s = 2.19× this doesnt match the 4.3x because The likely reason: a 30+ second computation is long enough that the OS migrates the thread back to a P-core mid-run thus we are noticing less effect for larger size.
 
But the effect is clear for the smallest sphere

pca006132:
oh btw, maybe you can convert those STLs to binary STLs, so it will not mess up the line count.

AnshulPatil2005:
@pca006132 @elalish Would it make sense to also run the benchmark workflow on release publish events?(version-wise data)

The idea is to reuse the same benchmark runner, but store release results separately in the same `benchmark-data` branch, for example under `releases/<tag>/<run_id>/`, while keeping weekly runs under `weekly/`.

then plot the data on the dashboard
rust has done similar thing in their perf dashboard

pca006132:
I think we can mark the release in the UI, but the release triggered run should still add to the same dataset. E.g., weekly data but tag when we have releases.

AnshulPatil2005:
> I think we can mark the release in the UI, but the release triggered run should still add to the same dataset. E.g., weekly data but tag when we have releases.

So when a release is published, we trigger the same benchmark workflow and store the result in the same `weekly/` dataset, just like a scheduled run, but add release metadata such as `release_tag: "v3.5.1"`.

Then the dashboard can mark that run as a release point in the same timeline.

Am I understanding this correctly?

pca006132:
Yes, this is what I am thinking.

AnshulPatil2005:
> Yes, this is what I am thinking.

Makes sense, I'll do that

.github/workflows/manifold.yml
```diff
@@ -461,6 +461,38 @@ jobs:
...
+      WEEKLY_BENCHMARK_THREADS: 1
+      CMAKE_OSX_ARCHITECTURES: arm64
+    steps:
+    - uses: actions/checkout@v6
+    - uses: actions/setup-python@v5
+      with:
+        python-version: '3.x'
+    - name: Verify macOS ARM runner
+      run: |
+        test "$(uname -s)" = "Darwin"
```
pca006132: I think this is needed for debug only?

.github/workflows/manifold.yml
```diff
@@ -461,6 +461,38 @@ jobs:
...
+      shell: bash
+      run: python3 ./scripts/run_weekly_benchmarks.py . ./bench/weekly "${WEEKLY_BENCHMARK_REPEATS}"
+    - name: Report weekly benchmarks
+      run: python3 ./scripts/report_weekly_benchmarks.py . ./bench/weekly "${WEEKLY_BENCHMARK_REPEATS}"
+    - name: Upload weekly benchmark artifacts
+      uses: actions/upload-artifact@v4
+      with:
+        name: weekly-benchmark
+        path: ./bench/weekly/
+        retention-days: 30
```
pca006132: do we need 30 days of retention?

.github/workflows/manifold.yml
```diff
@@ -461,6 +461,38 @@ jobs:
       shell: bash
       run: bash ./scripts/ci/sanitizer_test.sh
 
+  weekly_benchmark:
```
pca006132: I think github action supports reusing an action. We can just move the weekly benchmark action out.

.github/workflows/weekly_benchmark.yml
```diff
@@ -0,0 +1,151 @@
...
+  cancel-in-progress: false
+
+jobs:
+  build_sanitizer:
+    name: Sanitizer build (ASan+UBSan)
+    timeout-minutes: 30
+    runs-on: macos-15
+    env:
+      CMAKE_OSX_ARCHITECTURES: arm64
+      SANITIZER_C_COMPILER: clang
```
pca006132: Do we need these? It is fine if we use gcc on linux.

.github/workflows/weekly_benchmark.yml
```diff
@@ -0,0 +1,151 @@
...
+    env:
+      WEEKLY_BENCHMARK_REPEATS: 5
+      WEEKLY_BENCHMARK_THREADS: 1
+      CMAKE_OSX_ARCHITECTURES: arm64
+      DATA_BRANCH: benchmark-data
+      RELEASE_TAG: ${{ github.event_name == 'release' && github.event.release.tag_name || '' }}
+      SANITIZER_SUMMARY_SUBSET: extended
+      SANITIZER_SUMMARY_RUNNER: macos-15
+      SANITIZER_SUMMARY_OS: macOS
+      SANITIZER_SUMMARY_COMPILER: AppleClang
```
pca006132: again, do we need to specify the compiler?

.github/workflows/weekly_benchmark.yml
```diff
@@ -0,0 +1,151 @@
...
+      uses: actions/upload-artifact@v4
+      with:
+        name: weekly-benchmark
+        path: ./bench/weekly/
+        retention-days: 30
+    - name: Prepare benchmark data branch
+      if: github.event_name == 'release' || github.ref == format('refs/heads/{0}', github.event.repository.default_branch)
+      shell: bash
+      run: |
+        set -euo pipefail
```
pca006132: do we need this? I think by default it will fail if any command failed?

.github/workflows/weekly_benchmark.yml
```diff
@@ -0,0 +1,151 @@
...
+        git config user.name "github-actions[bot]"
+        git config user.email "41898282+github-actions[bot]@users.noreply.github.com"
```
pca006132, line 62: what is this magic email?

.github/workflows/weekly_benchmark.yml
```diff
@@ -0,0 +1,151 @@
...
+        if git ls-remote --exit-code --heads origin "${DATA_BRANCH}" >/dev/null 2>&1; then
```
pca006132: mind explaining this code? e.g., why worktree, and why this if condition.

.github/workflows/weekly_benchmark.yml
```diff
@@ -0,0 +1,151 @@
...
+          git worktree add --detach benchmark-data
+          git -C benchmark-data checkout --orphan "${DATA_BRANCH}"
+          git -C benchmark-data rm -rf . >/dev/null 2>&1 || true
+          find benchmark-data -mindepth 1 -maxdepth 1 ! -name .git -exec rm -rf {} +
+        fi
+    - name: Stage weekly benchmark data
+      if: github.event_name == 'release' || github.ref == format('refs/heads/{0}', github.event.repository.default_branch)
+      run: >
+        python3 ./scripts/publish_weekly_benchmark_results.py
+        ./bench/weekly ./benchmark-data
```
pca006132: is this tested? this looks wrong to me, you need an ending backslash to get newlines for a single command.

.github/workflows/weekly_benchmark.yml
```diff
@@ -0,0 +1,151 @@
...
+      run: python3 ./scripts/run_weekly_benchmarks.py . ./bench/weekly "${WEEKLY_BENCHMARK_REPEATS}"
+    - name: Report weekly benchmarks
+      run: python3 ./scripts/report_weekly_benchmarks.py . ./bench/weekly "${WEEKLY_BENCHMARK_REPEATS}"
+    - name: Upload weekly benchmark artifacts
+      uses: actions/upload-artifact@v4
+      with:
+        name: weekly-benchmark
+        path: ./bench/weekly/
+        retention-days: 30
+    - name: Prepare benchmark data branch
```
pca006132: Not sure if it makes too much sense to split the steps like this. They cannot fail independently.

dashboard/dashboard.js
```diff
@@ -0,0 +1,571 @@
...
+  if (!response.ok) {
+    throw new Error(`${response.status} ${response.statusText}: ${url}`);
+  }
+  return response.json();
+}
+
+async function loadIndex(dataRoot) {
+  const root = normalizeRoot(dataRoot);
+  const rootIsWeeklyDir = root.endsWith('/weekly');
+  const candidate = rootIsWeeklyDir ? {
```
pca006132: shouldn't we have a fixed file path?

dashboard/dashboard.js
```diff
@@ -0,0 +1,571 @@
...
+  try {
+    const index = await fetchJson(candidate.indexUrl);
+    return {...candidate, index};
+  } catch (error) {
+    throw new Error(`Unable to load benchmark index.\n${error.message}`);
+  }
+}
+
+function resultUrl(entry) {
+  const path = entry.result_path || '';
```
pca006132: While I understand it is good to do defensive programming, I think this file is a bit too defensive for my liking. Nearly everything has a fallback, and I don't think they are tested.

dashboard/index.html
```diff
@@ -0,0 +1,13 @@
+<!doctype html>
```
pca006132, line 1: Do we really need 6 html pages for a dashboard?

scripts/parse_weekly_benchmarks.py
```diff
@@ -0,0 +1,409 @@
+#!/usr/bin/env python3
```
pca006132, line 1: Can we just use json and avoid a custom parser?

scripts/parse_weekly_benchmarks.py
```diff
@@ -0,0 +1,409 @@
...
+        "Peak RSS mean (MB) | Peak RSS min (MB) | Peak RSS max (MB) | Runs |"
+    )
+    lines.append("|---:|---:|---:|---:|---:|---:|---:|---:|---:|")
+    for workload in suite["workloads"]:
+        timing = workload["time_ms"]
+        peak_rss = workload["peak_rss_mb"]
+        lines.append(
+            "| {n_tri} | {mean:.2f} | {median:.2f} | {min_:.2f} | {max_:.2f} | "
+            "{rss_mean:.2f} | {rss_min:.2f} | {rss_max:.2f} | {runs} |".format(
+                n_tri=workload["n_tri"],
```
pca006132, line 305: I think you can just spread the dictionary as arguments.

.gitignore
```diff
@@ -133,7 +133,23 @@ CMakeSettings.json
 [._]sw[a-p]
 
 # Benchmark stl files
-extras/ember_tests/testfiles/raw_meshes
+extras/ember_tests/testfiles/raw_meshes/*
+!extras/ember_tests/testfiles/raw_meshes/58011.stl
```
pca006132: I think gitignore does not affect already added files, so these entries are probably not necessary. And in the future if we want to add more, we can just git add force.

.github/workflows/manifold.yml
AnshulPatil2005: yep will remove that, it was just temporary addition.

.github/workflows/manifold.yml
AnshulPatil2005: the actual weekly_benchmark.yml has 1 day, this is the test one so didn't bother changing as the whole job will be removed anyways

.github/workflows/manifold.yml
AnshulPatil2005: Oh, anyways I will remove this workflow as i have confirmed this works on my fork.

.github/workflows/weekly_benchmark.yml
AnshulPatil2005: yea this was an unnecessary macOS duplicate of the sanitizer lane that already runs on every PR

dropped the macos setup.

.github/workflows/weekly_benchmark.yml
AnshulPatil2005: removed it

.github/workflows/weekly_benchmark.yml
AnshulPatil2005: yeah, 41898282 is the git action bot's actual permanent user id,

as for this step git just won't let you commit without an author identity set, and the runner doesn't have one by default. git needs to know whose name to put on the commit itself.

and for the email format - github gives every account a noreply address like <user-id>+<username>@users.noreply.github.com

.github/workflows/weekly_benchmark.yml
AnshulPatil2005: worktree so benchmark-data can be checked out in its own folder without switching away from the branch we're actually running benchmarks on. the if/else handles first run vs later runs - if the branch already exists we just fetch and attach to it, if it doesn't we create it as a fresh branch

.github/workflows/weekly_benchmark.yml
AnshulPatil2005: ran a real workflow_dispatch against this "Prepare/Stage/Push benchmark data branch" all succeeded, publish_weekly_benchmark_results.py parsed every sanitizer arg correctly. confirmed working for real

https://github.com/AnshulPatil2005/manifold/actions/runs/29854404591

although test_sanitizer failed in this same run manifold_test crashes on startup with dyld: Library not loaded: @rpath/libsamples.3.dylib, I am working on it 

update : its fixed

dashboard/dashboard.js
AnshulPatil2005: Yeah, I missed this file when reviewing the other dead fallbacks thus lot of dead code in these js files. I've made some changes now, it should be good.

dashboard/index.html
AnshulPatil2005: yeah, that's actually what it was before I split it (i thought it was better to tweak pages with different files as sidderent pages) went back to one page

dashboard/dashboard.js
AnshulPatil2005: yea,i missed the js files when I went through to check for dead fallbacks in all the files written.

It should be good now

scripts/parse_weekly_benchmarks.py
AnshulPatil2005: That would be better
gtest already has a --gtest_output=json flag built in, so switched that one to consume real json instead of regexing 

I'll do the changes for the perftest and manbench output

update :  Ember and perfTest turned out bigger than expected, Ember means touching core boolean algorithm internals (not just test code), and perfTest's parser is shared with the now-merged benchmark guard, so I didn't want to touch either without a heads-up maybe.

.gitignore
AnshulPatil2005: Oh right confirmed, all are already tracked.

.github/workflows/weekly_benchmark.yml
pca006132: Thanks. Maybe add a comment pointing to https://github.com/actions/checkout#push-a-commit-using-the-built-in-token, this looks very confusing without the context.

.github/workflows/weekly_benchmark.yml
pca006132: Oh ok, it is the yaml `>` vs `|`. Wasn't aware of that.

.github/workflows/weekly_benchmark.yml
```diff
@@ -92,6 +86,24 @@ jobs:
...
+    - name: Resolve sanitizer compiler
+      run: |
+        python3 - <<'PY' >> "$GITHUB_ENV"
+        import sys
+        sys.path.insert(0, "scripts")
+        from pathlib import Path
+        from system_metadata import cmake_compiler, parse_cmake_configure_log
+
+        cmake = parse_cmake_configure_log(Path("sanitizer-cmake-log/cmake_configure.log"))
+        print(f"SANITIZER_SUMMARY_COMPILER={cmake_compiler(cmake) or 'unknown'}")
```
pca006132: I think the question here is, why do we need to know the compiler?

scripts/parse_weekly_benchmarks.py
pca006132: I see. Not using JSON is fine, I looked at it and the parsing code is not too bad, I just want to avoid custom parsers when we can just use JSON.

scripts/ci/sanitizer_build.sh
```diff
@@ -1,6 +1,10 @@
 #!/usr/bin/env bash
 set -euo pipefail
 
+SANITIZER_C_COMPILER="${SANITIZER_C_COMPILER:-clang-18}"
```
pca006132: why should we set the compiler? I think the default compiler just supports asan.

.github/workflows/weekly_benchmark.yml
```diff
@@ -0,0 +1,157 @@
...
+        PY
+    - name: Verify macOS ARM runner
+      run: |
```
pca006132: again, why do we need this? it will be fine if we are not using ARM runners later, our timing plots should distinguish between OS and CPUs, we have the metadata for that..

.github/workflows/weekly_benchmark.yml
```diff
@@ -0,0 +1,157 @@
...
+      uses: actions/upload-artifact@v4
+      with:
+        name: weekly-benchmark
+        path: ./bench/weekly/
+        retention-days: 30
+    - name: Publish weekly benchmark data
+      if: github.event_name == 'release' || github.ref == format('refs/heads/{0}', github.event.repository.default_branch)
+      shell: bash
+      run: |
+        set -u
```
pca006132: why?

scripts/ci/sanitizer_build.sh
```diff
@@ -1,6 +1,10 @@
 #!/usr/bin/env bash
 set -euo pipefail
```
pca006132: btw normally you can just set it in the shebang.

dashboard/index.html
pca006132: It makes sense when you have a large file. But for an HTML file with 100LoC, I don't think it makes too much sense to split it into 6 files, when all of them should just be on the same page.

scripts/parse_weekly_benchmarks.py
AnshulPatil2005: sounds good then

scripts/ci/sanitizer_build.sh
AnshulPatil2005: it's what the existing PR sanitizer lane in manifold.yml already uses so maybe changing it could introduce some errors, I'll do the change and report the findings.

.github/workflows/weekly_benchmark.yml
AnshulPatil2005: makes sense

.github/workflows/weekly_benchmark.yml
AnshulPatil2005: set -u makes the script error out immediately if it references a variable that was never set. It was added more for future typo avoidance rather than catching a current error.

scripts/ci/sanitizer_build.sh
AnshulPatil2005: wouldnt shebang line just becomes a plain comment to bash and never gets executed as flags ?

pca006132: if you invoke it as `./script.sh` then it will invoke the shebang.

.github/workflows/weekly_benchmark.yml
pca006132: I am personally not particularly worried about this kind of typo, since it will show up in the diff. But well, it is fine to keep it as is. I just don't like the idea of putting `set -u` everywhere.

AnshulPatil2005:
@pca006132 
tried switching the sanitizer build to gcc, but it fails - gcc 13.3.0 hits a false-positive -Wmaybe-uninitialized inside libstdc++'s own <regex> code (used in impl.cpp for obj parsing), and it only shows up because of -fsanitize=address,undefined + -O3 + -Werror together. the regular gcc build job compiles the same file fine, so it's not a real bug in our code, just the gcc tripping over itself in sanitizer mode.

easiest fix is adding -Wno-maybe-uninitialized(ignore the error) just for the sanitizer build. ok to do that ?

pca006132:
Sounds fine to me, I somehow recall I saw this error a while ago but forgot what did I do to avoid that.

AnshulPatil2005:
@pca006132 I think this is ready to merge

AnshulPatil2005:
oh also where does the dashboard gets hosted? it's 3 static files, data is fetched from raw.githubusercontent.com so it can go anywhere.

easiest would be adding `mv ./dashboard ./public` to deploy.yml so it ships with the existing gh-pages deploy at manifoldcad.org/dashboard - but that's domain so figured i'd ask.

.github/workflows/weekly_benchmark.yml
```diff
@@ -0,0 +1,154 @@
...
+permissions:
+  contents: write
+
+concurrency:
+  group: weekly-benchmark
+  cancel-in-progress: false
+
+jobs:
+  build_sanitizer:
+    name: Sanitizer build (ASan+UBSan)
```
pca006132: is this the same as in our regular CI? If yes, maybe reuse the action?

.github/workflows/weekly_benchmark.yml
```diff
@@ -0,0 +1,154 @@
...
+        name: sanitizer-build
+        path: build
+    - name: Test extended sanitizer subset
+      shell: bash
+      run: bash ./scripts/ci/sanitizer_test.sh
+
+  weekly_benchmark:
+    name: Weekly benchmarks
+    needs: [build_sanitizer, test_sanitizer]
+    if: ${{ always() && !cancelled() }}
```
pca006132: what are these conditions?

.github/workflows/weekly_benchmark.yml
```diff
@@ -0,0 +1,154 @@
...
+      SANITIZER_TEST_RESULT: ${{ needs.test_sanitizer.result }}
+    steps:
+    - uses: actions/checkout@v6
+      with:
+        fetch-depth: 0
+    - uses: actions/setup-python@v5
+      with:
+        python-version: '3.x'
+    - name: Download sanitizer cmake configure log
+      if: always()
```
pca006132: same here

.github/workflows/weekly_benchmark.yml
pca006132: why is this resolved?

.github/workflows/weekly_benchmark.yml
```diff
@@ -0,0 +1,154 @@
...
+      continue-on-error: true
```
pca006132: what is the use of the log, and why should we continue if we cannot download it?

.github/workflows/weekly_benchmark.yml
```diff
@@ -0,0 +1,154 @@
...
+    - name: Report weekly benchmarks
+      run: python3 ./scripts/report_weekly_benchmarks.py . ./bench/weekly "${WEEKLY_BENCHMARK_REPEATS}"
+    - name: Upload weekly benchmark artifacts
+      uses: actions/upload-artifact@v4
+      with:
+        name: weekly-benchmark
+        path: ./bench/weekly/
+        retention-days: 30
+    - name: Publish weekly benchmark data
+      if: github.event_name == 'release' || github.ref == format('refs/heads/{0}', github.event.repository.default_branch)
```
pca006132, line 56: why are the conditions here? shouldn't this CI only be triggered in these situations?

.github/workflows/weekly_benchmark.yml
```diff
@@ -0,0 +1,154 @@
...
+      shell: bash
+      run: |
+        # github-actions[bot]'s identity for committing with the built-in
+        # token, see https://github.com/actions/checkout#push-a-commit-using-the-built-in-token
+        git config user.name "github-actions[bot]"
+        git config user.email "41898282+github-actions[bot]@users.noreply.github.com"
+        if git ls-remote --exit-code --heads origin "${DATA_BRANCH}" >/dev/null 2>&1; then
```
pca006132: I thought our consensus was to just manually add a branch, to removev the condition in the CI?

.github/workflows/weekly_benchmark.yml
```diff
@@ -0,0 +1,154 @@
...
+          --release-tag "${RELEASE_TAG}" \
+          --sanitizer-subset "${SANITIZER_SUMMARY_SUBSET}" \
+          --sanitizer-build-result "${SANITIZER_BUILD_RESULT}" \
+          --sanitizer-test-result "${SANITIZER_TEST_RESULT}" \
+          --sanitizer-runner "${SANITIZER_SUMMARY_RUNNER}" \
+          --sanitizer-os "${SANITIZER_SUMMARY_OS}" \
+          --sanitizer-compiler "${SANITIZER_SUMMARY_COMPILER}"
+
+        git -C benchmark-data add weekly
+        if git -C benchmark-data diff --cached --quiet; then
```
pca006132: why would this happen?

scripts/parse_weekly_benchmarks.py
```diff
@@ -0,0 +1,429 @@
...
+    lines.append("")
+    lines.append("| Test | Mean (ms) | Median (ms) | Min (ms) | Max (ms) | Runs |")
+    lines.append("|---|---:|---:|---:|---:|---:|")
+    if not suite["tests"]:
+        lines.append("| No tests selected | 0 | 0 | 0 | 0 | 0 |")
+    for test in suite["tests"]:
+        timing = test["time_ms"]
+        lines.append(
+            "| {name} | {mean:.2f} | {median:.2f} | {min_:.2f} | {max_:.2f} | {runs} |".format(
+                name=test["name"],
```
pca006132, line 329: spread this?

scripts/publish_weekly_benchmark_results.py
```diff
@@ -0,0 +1,149 @@
...
+        / run_id
+    )
+    dated_dir = args.data_dir / dated_rel
+    dated_dir.mkdir(parents=True, exist_ok=True)
+
+    entry = {
+        "run_id": run_id,
+        "timestamp": timestamp_iso,
+        "date": timestamp.strftime("%Y-%m-%d"),
+        "commit_sha": metadata.get("commit_sha"),
```
pca006132: can you spread metadata or just do dictionary union?

.github/workflows/weekly_benchmark.yml
AnshulPatil2005: Yes, Reuse is possible only difference is for build_sanitizer only difference is retention days which is not that important anyways  

we can do the same for test_sanitizer it differs a bit more - asan options, subset (weekly is always extended), and the artifact download path.

I'll do it for both tho

.github/workflows/weekly_benchmark.yml
AnshulPatil2005: If the sanitizer breaks, that's no reason to throw away data. On top of that, the report records whether the sanitizer passed or failed so the condition 

but i realised !cancelled() can be used alone here always() is already true on cancel, so having both is redundant.

.github/workflows/weekly_benchmark.yml
AnshulPatil2005: > why is this resolved?

my bad missed it, we weren't actually using it the compiler got written to index.json but nothing ever displayed it

i was thinking maybe we could use it to detect when GitHub bumps the ubuntu-24.04 image but its near constant anyways

.github/workflows/weekly_benchmark.yml
AnshulPatil2005: fair on both counts. the log only existed to derive the sanitizer compiler (same thing as the thread above)

removed the artifact, the download step and the parsing.

.github/workflows/weekly_benchmark.yml
AnshulPatil2005: schedule and release do imply it, but workflow_dispatch doesn't - it can run from any branch, so without this a manual test run from a feature branch would push real data to benchmark-data.

both halves are needed: on a release github.ref is the tag, not refs/heads/master, so the ref check alone would skip publishing on releases.

.github/workflows/weekly_benchmark.yml
AnshulPatil2005: right, my bad  removed it

.github/workflows/weekly_benchmark.yml
AnshulPatil2005: it can't - index.json gets a generated_at timestamp on every publish, and each run writes a new dated directory, so the staged diff is never empty. dead branch, removed it.

elalish:
@AnshulPatil2005 You have some replies above saying you've fixed things, but I don't see any commits since then - have you pushed?

AnshulPatil2005:
> @AnshulPatil2005 You have some replies above saying you've fixed things, but I don't see any commits since then - have you pushed?

Yep they are sitting local, I have some more changes so will commit that together only.

AnshulPatil2005:
The sanitizer build sets MANIFOLD_STRICT=ON, which turns on -Werror and that's what caused the GCC failure. -fsanitize changes how GCC inlines, and it threw a false-positive uninitialized warning inside libstdc++'s own regex headers. Not our code.

So my question is whether this lane should enforce warnings at all. Its job is runtime memory and UB bugs, and warning enforcement is already covered by the regular build jobs GCC 13 and 14, TBB on and off.

Setting MANIFOLD_STRICT=OFF should be better than -Wno-maybe-uninitialized ?

elalish:
>So my question is whether this lane should enforce warnings at all. Its job is runtime memory and UB bugs, and warning enforcement is already covered by the regular build jobs GCC 13 and 14, TBB on and off. Setting MANIFOLD_STRICT=OFF should be better than -Wno-maybe-uninitialized ?

I think that sounds reasonable.

elalish:
Okay, is this ready to merge? @pca006132, you have more context than I on this one - is it ready for approval?

AnshulPatil2005:
> Okay, is this ready to merge? @pca006132, you have more context than I on this one - is it ready for approval?

>easiest would be adding mv ./dashboard ./public to deploy.yml so it ships with the existing gh-pages deploy at >manifoldcad.org/dashboard - but that's domain so figured i'd ask.

yes should be good with this and some more minor changes

pca006132 (approved):
looks fine to me. just let me know what I need to do to setup the other branch, before merging.

AnshulPatil2005:
> looks fine to me. just let me know what I need to do to setup the other branch, before merging.

just need an empty orphan branch called benchmark-data on the repo:

git checkout --orphan benchmark-data
git rm -rf .
git commit --allow-empty -m "Initialize benchmark data branch"
git push origin benchmark-data
git checkout master

nothing else - the workflow fetches it, adds a worktree and writes weekly/ into it. it needs to not be branch-protected so github-actions[bot] can push to it.

the dashboard also ships to manifoldcad.org/dashboard via deploy.yml in this PR. it'll show an error until the first weekly run publishes data, since it reads from that branch.

pca006132:
https://github.com/elalish/manifold/tree/benchmark-data

Done. I guess we can merge this and see!

PR #1759, opened 2026-06-16
Simplify the boolean2 winding and crossing-insertion pipeline

zmerlynn:
## Summary

Simplifies the boolean2 `CrossSection` arrangement engine, following the 2D
overlap-removal sketch in #289 and working toward #1707. The change deletes more
than it adds: the winding filter, crossing insertion, and the crossing decision
all get simpler. `docs/Boolean2.md` is updated to match.

## Changes

- **Winding: per-edge ray-cast.** Replaces the halfedge face-walk with a
  per-sub-edge ray-cast filter (the #289 model), BVH-accelerated over the
  canonical sub-edges (~O(E log E)). On a general-position arrangement the two
  approaches agree, so this is a simplification, not a behavior change.
- **Crossing insertion: one pass.** Each crossing reuses an existing within-eps
  vertex id, and a crossing is skipped when an endpoint of one edge already lies
  in the other's vertex list (#289 step 4). This removes the separate post-hoc
  reconcile pass and its magnitude-sensitive gates.
- **Crossing decision: by sign.** A transversal crossing is decided by the
  strict sign of the straddle rather than an epsilon near-endpoint band. The band
  rejected a genuine crossing that landed within eps of an endpoint as a sliver -
  the failure behind the near-corner `SquareAnnihilation` case, where the dropped
  crossing left an edge unsplit and collapsed the whole larger shape. A crossing
  within eps of an endpoint now snaps to that endpoint at insertion, the same
  metric the band used, so nothing changes away from the degenerate case. The
  vertex-on-edge merge stays epsilon-based - that one is a bounded-distance
  decision, not a sign one.
- **Stress tests** for near-degenerate inputs, a comment and test-name tidy, and
  the doc refresh.

## Testing

`manifold_test` passes locally, including an assertion build
(`MANIFOLD_ASSERT=ON`). The sign-based crossing decision fixes
`SquareAnnihilation`, now enabled, along with
`TinyFeatureNearCornerHostDropAtOffset1024`. The `4096` variant stays
`DISABLED_`: at that offset the arrangement comes out non-planar - the
large-offset magnitude class this change does not address - which assertion
builds reject. `InclusionExclusionSliver` and `ShortEdgeFusionAddIntersect`
remain `DISABLED_` as eps-resolution floors.


src/boolean2_winding.cpp
```diff
@@ -52,503 +36,101 @@ bool IsInside(WindRule rule, int w) {
...
+    const int e = bvh.leafToOrig[leafIdx];
+    const auto& edge = canon.edges[e];
+    vec2 a = verts[edge.vMin];
+    vec2 b = verts[edge.vMax];
+    const bool upward = a.y < b.y;
+    if (!upward) std::swap(a, b);
+    if (a.y == b.y) return;  // horizontal edges contribute nothing
+    if (!(a.y <= p.y && p.y < b.y)) return;
+    const double t = (p.y - a.y) / (b.y - a.y);
+    const double xi = a.x + t * (b.x - a.x);
```
elalish: Can we reuse Smith's `Interpolate` function here? He made some effort to avoid all rounding errors at both endpoints, so I think it's worth reusing for safety. The more we can harmonize with Boolean3, the better.

src/boolean2_winding.cpp
```diff
@@ -52,503 +36,101 @@ bool IsInside(WindRule rule, int w) {
...
+  for (int ei = 0; ei < nE; ++ei) {
+    const auto& edge = canon.edges[ei];
+    const vec2 p = verts[edge.vMin];
+    const vec2 q = verts[edge.vMax];
+    const vec2 mid = (p + q) * 0.5;
+    const vec2 d = q - p;
+    const double L = la::length(d);
+    if (L == 0.0) continue;
+    // Perpendicular sample offset, bounded by a quarter of the clearance to the
+    // nearest other sub-edge so each side-point lands in the adjacent face.
```
elalish: I'd prefer to avoid this tolerancy offset. Can we simply use the start vert, and for any edge the collider finds that shares that vert, just fall back to doing the smith shadow function for the other end-points?

src/boolean2_winding.cpp
```diff
@@ -52,503 +36,101 @@ bool IsInside(WindRule rule, int w) {
...
+      const Box2 query(mid - vec2(L, L), mid + vec2(L, L));
+      auto nearest = [&](int /*queryIdx*/, int leafIdx) {
+        const int f = bvh.leafToOrig[leafIdx];
+        if (f == ei) return;
+        clr = std::min(clr, DistPointSeg(mid, verts[canon.edges[f].vMin],
+                                         verts[canon.edges[f].vMax]));
+      };
+      auto recorder = MakeSimpleRecorder(nearest);
+      auto qf = [&](int) { return query; };
+      BVHCollisions(bvh, recorder, qf, /*n=*/1, /*parallel=*/false);
```
elalish: I think we should be able to remove this entirely, based on the above.

src/boolean2_winding.cpp
```diff
@@ -52,503 +36,101 @@ bool IsInside(WindRule rule, int w) {
...
+    }
+    const vec2 nrm = la::normalize(vec2(-d.y, d.x));  // left unit normal
+    const double delta = 0.25 * clr;
+    const vec2 leftP = mid + nrm * delta;
+    const vec2 rightP = mid - nrm * delta;
+    const int leftW = WindingNumberAt(leftP, bvh, canon, verts, globalMaxX);
+    const int rightW = WindingNumberAt(rightP, bvh, canon, verts, globalMaxX);
```
elalish: `rightW == leftW - 1` always, or our algorithm is unstable. I don't think there's any reason to calculate a second time except maybe in the midst of a debug.

test/boolean2_test.cpp
```diff
@@ -889,39 +889,57 @@ TEST(Boolean2, KeepsNearDistinctPresentationVertex) {
...
-// an old corner fuses into it; beyond 2*eps it stays a distinct vertex.
-TEST(Boolean2, NewToOldMergeBudgetIsTwoEps) {
+// The new-to-old snap is gated by perpendicular incidence, not distance: a
+// generated crossing fuses into a nearby old corner only when one of its
+// transversal source edges actually passes through that corner within the
+// bounded eps perpendicular band. Mere proximity (within the 2*eps broad-phase
+// band) is not enough - distinct near-corner crossings whose source edges miss
+// the corner stay separate, because the along-line distance of a true
+// concurrency is unbounded as the crossing angle shallows and so cannot
+// separate must-fuse from must-not-fuse.
```
elalish: The wording here is unclear to me as to what is correct and desirable behavior vs what is a potential bug we're trying to catch. I would think in our algorithm we would want "distinct near-corner crossings whose source edges miss the corner" to break the edge and attach rather than stay separate, since that is necessary to get us into general position, right?

test/cross_section_test.cpp
```diff
@@ -406,18 +406,15 @@ TEST(CrossSection, TinyFeatureNearCornerHostFeatureSwap) {
...
-
-// Big-piece drop: the big piece's zero-radius star vertex makes it a thin spike
-// that collapses. Built from raw radii (StarRing's 0.1 floor would remove the
-// spike).
+// A tiny piece point-touches a big piece ~1e-9 from one of the big piece's
+// vertices; the two are disjoint, so the correct union keeps both. The 1024
+// case (below) holds; the 4096 case does not.
+
+// DISABLED: at offset 4096 the arrangement comes out non-planar and assertion
+// builds reject it ("retained directed edges must form closed walks"); release
```
elalish: What does non-planar mean in 2D? Also, this sounds suspiciously like it's non-manifold - what's the result of "swallowing" this?

test/boolean2_test.cpp
```diff
@@ -1175,3 +1196,82 @@ TEST(Boolean2, NonFiniteInputReturnsEmpty) {
...
+// The Add/Intersect variant of the short-edge fusion above. Independent
+// rectangle arithmetic: left/right are x-disjoint, each area 0.45h*1.11e-5 =
+// 4.995; rect area d*h = 2.6; union = 12.59 - 0.275 - 0.22 = 12.095;
+// union n (left u right) = left u right = 9.99.
+//
+// DISABLED: floor, not a regression. The two subtracted rectangles' inner edges
+// cross the rectangle's short bottom edge 0.4 eps apart (x0 = 1.1 eps, x1 = 1.5
+// eps) - a sub-eps gap. The eps-scale merge fuses those crossings (the same
+// fusion coincident-edge cancellation relies on), over-counting the gap.
+// Resolving sub-eps gaps would fix it but would break the cancellation.
```
elalish: What does "break the cancellation" mean? Do we have a sense of where the problem is for these three disabled tests? Do you have images?

My ethos with Manifold is that there are no edge cases - only counter examples. Our algorithm is not sound if any counter example exists. But then, we must also be very clear about what our metric is for testing this. It's always possible we need to adjust our eps if we haven't quite bounded our numerical errors.

elalish (changes requested):
Loving the simplification, now let's simplify some more. Can you summarize for me what your goals are as far as correctness of this positive fill rule algorithm? I think I would say it should:
- Contain no edges shorter than epsilon.
- No verts lie within epsilon of any edges, excluding their endpoints.
- Contain no segments that intersect, excluding the epsilon-region of their endpoints.
- Retained input verts that were already in general position are exactly reproduced in the output.
- Any point in R2 that is not within epsilon of the input edges should have a winding number of 1 in the output if its input winding number was >0, and 0 otherwise.

Do you think that's a necessary and sufficient set of conditions?

src/boolean2_winding.cpp
zmerlynn: Done in 312a06399. The non-incident ray crossing-x now goes through Interpolate (axes swapped so it solves at the segment's y); the only other FP is the incident-edge orient via the shadow. The hand a.x + t*(b.x-a.x) is gone.

src/boolean2_winding.cpp
zmerlynn: Done. FilterByWinding now evaluates the left-face winding at the start vertex under a symbolic perturbation Q = P + e*d + e^2*n; no more perpendicular clearance/midpoint sampling. Edges sharing the start vert are resolved by the Smith shadow of their other endpoint, as you suggested.

src/boolean2_winding.cpp
zmerlynn: Done - DistPointSeg and the nearest-edge clearance scan are deleted.

src/boolean2_winding.cpp
zmerlynn: Done - one winding now, rightW = leftW - mult. One nuance worth flagging: it's leftW - mult, not literally -1. The multiplicity is signed (a canonical sub-edge is +-1 depending on its direction), and coincident input can sum to |mult| > 1; a literal -1 would collapse coincident loops to zero area. So the invariant is that the two sides differ by the signed multiplicity, which is +-1 in general position. Happy to assert the +-1 form behind a DEBUG check if you'd prefer.

test/boolean2_test.cpp
zmerlynn: This pins both directions of the gate, which I think is the source of the confusion - both blocks assert desired behavior: block 1 (non-incident) crosses 1.5 eps off the corner so it must stay a distinct vertex (it still breaks the edge and attaches there, just not onto the corner); block 2 (incident) runs through the corner so it must fuse. The gate is perpendicular incidence, not proximity, because a true concurrency's along-line distance is unbounded as the angle shallows. I'll rewrite the wording to spell out must-fuse vs must-not-fuse and can rename the test. Does that resolve it, or are you seeing a genuinely incident crossing that fails to attach?

test/cross_section_test.cpp
zmerlynn: You're right, "non-manifold" is the accurate term. Concretely the retained directed-edge graph has a vertex with in != out degree, so the walk doesn't close - the "equal edges into and out of each vert" precondition failing. "Swallow": a Release build drops that non-closing fragment and emits the rest, and here the fragment is a sub-eps cluster of crossings at the corner, so the macro result is still right (the big piece survives, area ~30107); the assert build throws, hence it's disabled. Two corrections to my own comment: it's non-manifold not "non-planar", and it's offset-independent - identical at offset 0, so "4096" is a red herring.

On fixing it: I haven't found a clean local fix yet. The three near-corner crossings are genuinely distinct (~2-3 eps apart, not an FP artifact - higher precision doesn't move them). Treating them as one vertex is a tolerance-band call, and the three bands I tried each broke something: along-edge distance over-merges and drops the square (the case in `TinyEdgeFeatureKeepsSquare`, https://github.com/zmerlynn/manifold/blob/0583feaa3/test/cross_section_test.cpp#L477), perpendicular incidence under-merges a genuine 1.03-eps concurrency, and widening eps works for this input but is needle-threaded to the band - the test's own knobs re-thread any value. Every band I tried has a boundary an adversary can sit on, so this reads to me as below-resolution by your metric clause: the macro answer is already correct, only the sub-eps cluster is ambiguous. I'd leave it disabled and documented for now - unless you think the closed-walk assert should tolerate a sub-eps-confined cluster that Release already drops correctly.

test/boolean2_test.cpp
zmerlynn: "Break the cancellation": the eps-scale merge that fuses near-coincident crossings is the same mechanism that makes opposite-direction coincident sub-edges cancel (they need a shared vertex id for the multiplicities to sum to zero), so a naive tightening just trades one failure for another.

Measured against your metric:
- InclusionExclusionSliver: the triangles are actually disjoint and the engine gets the topology exactly right (union 2 contours, intersection empty); only the union area is off by 4.1e-6, a ~0.18-eps boundary shift levered by the 4096-long base - the same eps*length lever as CornerCrossingStrip/SubEpsRectangleOverlap. Below what (5) constrains, so I re-enabled it asserting the topology (NumContour + empty intersection) with a loose area bound, like those two.
- ShortEdgeFusion, Add: 1 contour, correct; area off 0.09 = a 0.09-eps shift x the 1e6 edge. Below the floor.
- ShortEdgeFusion, Intersect: the two probes are 0.4 eps apart and the merge fuses them, so it returns 1 contour where the disjoint truth is 2. Sub-eps, so I'm treating it as in-budget and re-enabled both ops with a loose, lever-aware bound - but here being below the metric surfaces as a contour merge, not just an area wobble, which is the case for your "adjust eps" clause.
- HostDrop: a resolution-floor case I haven't found a clean local fix for; stays disabled, reads as below-resolution.

So ShortEdge and InclExcl are re-enabled with loose, lever-aware bounds (InclExcl with exactly-correct topology as the load-bearing check); HostDrop stays disabled. Numbers are straight from the engine.

zmerlynn:
That lines up with how I've been thinking about it: (1)-(3) are general position of the output, (4) is fidelity, (5) is the actual fill correctness - the #289 I1-I3 precondition carried through to the output. Two things I'd pin down before calling it sufficient:
- (5) only constrains points farther than eps from the input edges, which is exactly what leaves a sub-eps feature free to be kept or dropped. I think that's intended, and it's where the "adjust eps" metric lives, but worth stating explicitly, since it's the line between a real counterexample and a below-resolution case.
- Is (4) a correctness condition or an idempotence one? It's what makes a second pass a no-op (which I want), but it reads orthogonal to the fill being correct.

What I can't yet show is that (5) alone forces the output to be closed, even-valence loops (a valid manifold), or whether that has to be stated separately - do you read it as implied?

src/boolean2_winding.cpp
```diff
@@ -52,503 +38,132 @@ bool IsInside(WindRule rule, int w) {
...
+      // Incident edge: it passes through P, so its crossing-x equals P.x to
+      // leading order and the perturbation decides. With w = O - P (O the
+      // non-P endpoint), the +x ray crosses to the right iff, after clearing
+      // the sign of w.y, -e*la::cross(d,w) - e^2*la::cross(n,w) > 0.
+      const int other = e.vMin == start ? e.vMax : e.vMin;
+      const vec2 w = verts[other] - P;
+      const double primary = -la::cross(d, w);
+      const double secondary = -la::cross(n, w);
+      const bool pos = primary != 0.0 ? primary > 0.0 : secondary > 0.0;
+      rightOf = w.y > 0.0 ? pos : !pos;
```
elalish: This branch should also be doable with just `Interpolate` and `Shadows`, without need for d or n. Whichever end vert is between P and the other end vert gets interpolated onto the other edge, and then shadowed.

test/cross_section_test.cpp
elalish: Yes, it feels like this test is already passing in Release mode, so we should focus on making it assertion-free without necessarily changing the output. It's a bit terrifying to me that a manifoldness assertion is failing, and even more surprising that somehow this is benign. It makes me wonder if something about that assertion isn't quite accurate? The main reason we're so picky about manifoldness is that generally errors of that kind are unrecoverable. How does the Release build manage if the number of in and out edges doesn't match at a vert? I think it's worth hunting this down and understanding it, since with this closed, the system is looking robust.

test/boolean2_test.cpp
elalish: That sounds reasonable. I would call an "along-line-distance" a projection, since distance and proximity to me mean point-to-point. And perpendicular incidence I believe is just point-to-line distance. It actually sounds like the along-line distance you're referring to is more of an inverse projection? Definitely not what we'd want to use, and maybe not even worth mentioning.

elalish (changes requested):
I'd say 1 and 2 remove most of the "interesting" classes of sub-eps features. I think all that's left is eps-collinear points, which I use 4 as the rule to say they stay. 4 is about "don't change more than you have to". It rules out things like auto-simplify and Clipper's snap-to-grid. Idempotence is good too. 

I did think that closed manifold loops went without saying (it's the name of the library! Plus our Polygons data structure can't actually represent a non-closed result). I would say 5 forces it in the sense that winding number is not even defined for a non-closed polyline. 

Looking good, let's just see if we can get to the bottom of that last disabled test.

src/boolean2_winding.cpp
elalish: Actually yes, asserting the multiplicity is one once all the cancellations are done seems like a good check. I'm guessing you've already swapped the start/end if it was negative anyway, so it's always +1 by this point?

zmerlynn:
That settles the two I'd left open: 5 forces closure because winding isn't defined on a non-closed polyline, and 4 ("don't change more than you have to" - no auto-simplify, no snap-to-grid, idempotent) is exactly the line I wanted. I'd call the five sufficient. On the last disabled test, I hunted it down - details in that thread.

src/boolean2_winding.cpp
zmerlynn: Done in 301acda. Reworked the incident branch to Interpolate+Shadows, dropping the d/n cross products: O (the edge's far vert) and the classified far vert share a y-side of P, so the side test reduces to interpolating whichever lies between them onto the other edge and Shadows-comparing - the same primitive the non-incident branch uses. Validated bit-identical to the prior winding across the full suite + corpus + a 40k scaled random/degenerate differential.

test/cross_section_test.cpp
zmerlynn: Dug in. `OutEdgesToPolygons` is a greedy tracer: it emits a loop only when the walk returns to its start; at an in != out vertex one edge can't continue, so in release that fragment is dropped and the macro loops still close - for 4096 the dropped pieces are 3 single-vertex dangling walks, so the area is right.

But I don't think we should bless the drop. I stress-tested it: a neighboring shape (two disjoint polygons point-touching ~2e-9 at a corner, one ~422k area) makes relaxing the assert silently delete the large polygon - the macro that traces first eats the shared corner edges and starves the other's walk, so the dropped fragment is a whole feature, not a sub-eps tangle. Added it as a disabled guard (`DISABLED_NearCornerTouchDropsMacroFeature`). So the assert is doing real work, not just 3D strictness.

On a source fix: I went through every pre-winding phase, staying FP. The crossings are 3-4 genuinely distinct points ~1 eps apart the merge can't collapse to even valence; the only thing that closes the walk is fusing them, and fusing by distance over-merges and relocates the imbalance to another shape. I haven't found one that's free elsewhere yet. Here's a trace of the 4096 arrangement so you can see the cluster and the degree imbalance: https://rawcdn.githack.com/zmerlynn/manifold/093375eeb38b0b32db35b180970df5265c517126/extras/boolean2_trace_viewer.html?trace=boolean2_traces/hostdrop_offset4096.json

test/boolean2_test.cpp
zmerlynn: Agreed - reworded to point-to-line distance vs point-to-point proximity, and dropped the along-line/inverse-projection aside (f7dfe53).

src/boolean2_winding.cpp
zmerlynn: The emitted OutEdge is already +1 and oriented interior-left (the start/end swap happens at retention). But the canonical sub-edge's mult isn't always 1 there - coincident same-direction input sums. I checked: `Polygons{square, square}` (a square unioned with itself) retains its boundary with mult 2 (leftW 2, rightW 0), correct output, so a `mult == 1` assert would trip on it. That's the case the `- mult` (vs a literal `- 1`) covers; the safe invariant at retention is just `leftIn != rightIn` (which already implies mult != 0). Happy to add that if useful.

test/cross_section_test.cpp
zmerlynn: Got HostDrop (3d110ca). The fix was to stop being a special case: infer eps from the absolute bbox scale (`EpsilonFromScale(bBox.Scale())`) and drop the local-origin recenter, so boolean2 matches `MaxEpsilon`/the triangulator instead of stripping position to stay translation-invariant. The position-scaled eps merges the near-coincident crossings instead of tangling into a non-closing walk - so HostDrop and the macro corner-touch both pass now, plus a bunch of far-from-origin degeneracies that used to come out invalid. And it's -14 LOC, recenter gone.

Tradeoff is eps isn't translation-invariant anymore, so four of our own tests don't hold (`TranslatedSmallPolygon`, `ShallowConcurrentEdges`, `SubEpsRectangleOverlap`, `OffsetIsInvariant`) - disabled each with a comment, since a feature way smaller than its distance from origin just isn't resolved, same as boolean3. Otherwise green, and a ~240k-pair fuzz/diff pass came back clean.

zmerlynn: Honest caveat: this moves the problem more than solves it. Position-inclusive eps is a no-op at the origin, so a centered near-degeneracy still produces a non-manifold arrangement that the closed-walk assert correctly catches (built one; fires with or without this change, Release area still correct). The assert's right to fire - the remaining work is the arrangement, separate from this. Still the right call for consistency with `MaxEpsilon`/the triangulator, and it kills the real large-coord drops.

elalish: Yes, you've discovered a classic truism of computational geometry - tolerances can't fix manifoldness problems, they can only move them to places that are more difficult to test. Boolean3 is robust because it is manifold by construction, so we should figure out how to do that here as well.

I looked at your link, but I couldn't quite see where the problem was. Also I couldn't find v1 - maybe it's identical to v5? It's hard to tell the topology because the edges say what positions they connect, but not which vert indices, which loses the connectivity info. Also not sure where the degree imbalance shows up.

Back to the algorithm, I'm trying to understand how a degree imbalance is possible to create? Every input vert has balanced degree (one in, one out). The only operations we can perform are: 
1. Merge verts - total degree is sum of vert degrees, so still balanced, minus the edge that collapsed, which is both in and out, so still balanced.
2. Break an edge at a vert and merge - just adds one in and one out to the vert, so still balanced. 
3. Intersect two edges - form a vert with two ins and two outs, so balanced.

Can you assert the degree balance at every stage of the algorithm and figure out where it's going wrong? I think that has to be a bug, because regardless of all tolerances, it should not be possible to make an unbalanced vert.

test/cross_section_test.cpp
```diff
@@ -471,9 +474,105 @@ TEST(CrossSection, DISABLED_TinyFeatureNearCornerHostDropAtOffset4096) {
...
+// arrangement is non-manifold (a retained vertex with in != out degree) and
+// OutEdgesToPolygons correctly hits the closed-walk assert. The assert is not
+// in question: a non-closing walk is a real defect. Position-inclusive eps is a
+// no-op at the origin, so it does not address this (it fixed HostDrop/Lane B
+// only by inflating eps at their large coords until the tangle merges). In
+// Release the macro area is incidentally correct (only a sub-eps sliver drops),
+// but the arrangement is still non-manifold - the fix belongs in the
+// arrangement, not the assert. Below-resolution floor, StressB/StressD class.
+// Re-enable when the arrangement is made manifold here.
+TEST(CrossSection, DISABLED_CenteredSubEpsNonClosingWalk) {
```
elalish, line 529: Great test - when this is working I think we'll be ready to merge.

test/cross_section_test.cpp
```diff
@@ -517,6 +616,89 @@ TEST(CrossSection, DISABLED_TinyFeatureNearCornerHostDropAtOffset1024) {
...
+  EXPECT_NEAR(u.Area(), a.Area() + b.Area() - inter.Area(), 5e-5)
+      << "inclusion-exclusion area outside the eps*length floor";
+}
+
+// A genuine ~4*eps-wide rectangle overlap; a sub-eps binary-incidence
+// knife-edge. DISABLED by the position-inclusive eps change: at these
+// off-origin (~2048) coordinates eps scales to bBox.Scale() (matching
+// Boolean3), ~2x the old span-based value, so this ~4*eps overlap now falls
+// under eps and merges to empty. See the InferEps contract in boolean2.cpp.
+TEST(CrossSection, DISABLED_SubEpsRectangleOverlap) {
```
elalish: Thanks for harmonizing the eps with the way we do in Boolean3. My argument for that approach is that floating point values really do lose precision far from the origin - adding and subtracting a huge translation will leave you with genuinely different values (truncation error), so it makes sense to me that it's not idempotent in that situation. Technically we should be doing a full bounding exercise on our floating point errors like Smith did to ensure our tolerances are properly conservative, but I'm okay with keeping them a bit extra-wide and verifying with fuzzing. 

Given that, I'd would either remove these DISABLED tests, or change what they EXPECT and enable them.

elalish (changes requested):

test/cross_section_test.cpp
zmerlynn: Did that - I added a per-vertex in==out assert and ran it at every stage. It holds through all three of your operations (merge, break+merge, intersect); the arrangement stays balanced right to the end. The imbalance comes from a 4th step that isn't on the list: FilterByWinding. The winding selection keeps edges one at a time by winding number with no global balance constraint, so it can leave a vert unbalanced.

The case: the inexact arrangement splits one near-degenerate corner into two crossings ~2 eps apart (just outside the merge band). FilterByWinding then correctly drops the seam between them - both incident faces are Inside, so it's interior, not boundary - but that seam was the only edge joining the two, so one is left with a spare in and the other a spare out.

I did look at making the selection balanced by construction, but as far as I can tell the imbalance is inherent to per-edge winding selection - any in-filter fix reduces to merging that split pair. So MergeWindingVerts does exactly that, after the selection: fuse the within-eps verts whose surpluses cancel back into one vertex. You're right that it's a tolerance, not construction - but the winding imbalance is the signal that the two are the same point, which is what makes it safe. Widening the merge to catch them pre-winding over-merges real tiny features; only the post-selection imbalance distinguishes a droppable seam from a kept boundary. Suite's green, and a 296k-case differential against base is byte-identical except where base hits the assert.

zmerlynn:
Heads up before you review: another near-degenerate case, with a disabled regression test pushed for it (`DISABLED_NearCoincidentCornersNonClosingWalk`).

Three ordinary triangles whose corners sit within eps of each other, where a real edge-edge crossing lands within eps of a neighboring corner. Insertion's on-edge-incidence rule treats the corner as the meeting point and drops the crossing, leaving a non-planar arrangement and an open walk. It's not degenerate input - it's realistic geometry on a knife-edge, firing only when a crossing happens to fall within eps of an existing corner. `MergeWindingVerts` doesn't catch this one (its imbalance is macro-separated, unlike the centered case that motivated the merge).

Three ways forward, and I'm happy with any:
1. Merge as-is - a narrow, loud, failure on inputs around a knife-edge of tolerance (move either way and the problem goes away).
2. Revert `MergeWindingVerts` (cb9c1abd3) and merge the rest. I'm honestly not sure the merge isn't papering over some other constraint violation, so dropping it keeps the PR to what I'm confident in - with the caveat that it reopens the centered case as an assert.
3. I keep going and try to fix this case first.

Either way I'll write up a detailed narrative with trace screenshots on #1707 for both this case and the centered one that motivated `MergeWindingVerts`.

What's your preference?

elalish:
I replied on the other thread - let's start with 2, then move on to 3. If this PR is getting ungainly for you we can merge it and continue in a follow-on, but I'm fine with doing the work here too.

test/cross_section_test.cpp
starseeker: Out of curiosity, what would a full bounding exercise (as opposed to extra-wide and fuzzing) look like?

zmerlynn:
Done - reverted `MergeWindingVerts` (efa495d), so the PR is the eps + winding wins without the merge hiding the open walk; both open-walk cases stay as disabled regression tests. Not abandoning the merge problem, just banking what's solid - I'll keep going on the insertion-side fix, here or in a follow-on, whichever's easier for you to review.

One flag: I suspect the centered and triangle cases may be two separate problems rather than one - I'll dig into that, with details, on #1707.

test/cross_section_test.cpp
elalish: Take a look at [Smith's thesis](https://github.com/elalish/manifold/blob/master/docs/RobustBoolean.pdf). I think it's an appendix - you can bound the error introduced in each add and multiply op and see how much can actually accrue as long as your operations are simple enough. He did that for the Boolean we implemented, and it's part of why we use his `Interpolate` and `Intersect` functions pretty much verbatim.

src/boolean2.cpp
```diff
@@ -592,7 +592,8 @@ void ProcessEdgePair(const std::vector<EdgeM>& edges,
                      const std::vector<EdgeGeom>& edgeG, double eps,
                      const std::pair<int, int>& pr,
                      std::vector<EdgeVertHit>* hits,
-                     std::vector<IntersectionPoint>* intersections) {
+                     std::vector<IntersectionPoint>* intersections,
```
elalish: Looks like a few `*`s are still hiding in here instead of `&`.

test/cross_section_test.cpp
```diff
@@ -343,7 +343,11 @@ TEST(CrossSection, ConcurrentIndependentEdgePairs) {
   }
 }
 
-TEST(CrossSection, TranslatedShallowConcurrentEdges) {
+// DISABLED by the position-inclusive eps change: InferEps now scales with
+// bBox.Scale() (absolute coordinate), matching Boolean3, so the shallow
+// concurrency of these ~1-unit rhombi is not preserved when translated to 2^40.
+// Translation invariance is intentionally given up; see the InferEps contract.
+TEST(CrossSection, DISABLED_TranslatedShallowConcurrentEdges) {
```
elalish: Let's remove all `DISABLED_` - I consider those as TODOs and look at them pretty hard during review, so leave only those that you want input from me on. Given the comment, it looks like this can just have its EXPECT updated.

elalish (commented):
Looking good! Interested to hear more about this last outstanding test failure you mentioned.

src/boolean2.cpp
zmerlynn: Done in 5c95152 - swept the PR: out-params now by reference across the boolean2 helpers and predicates (`ProcessEdgePair`, `RecordEdgeVertHit`, `SortIntersections`, `VESetInsert`, `RadixSortPairs`, `SortSmallInts`, `CollectIntersectionPairs`, `IntersectSegments`, `AddEdgeVertListAnnotations`). Kept the two `Trace*`s - one's nullable, the other's a stored member.

test/cross_section_test.cpp
zmerlynn: Done in b358443. Audited all five: four document intentional position-inclusive-eps behavior (a sub-eps-relative feature far from the origin isn't resolved) - un-disabled with EXPECTs flipped to assert that (`TranslatedShallowConcurrentEdges`, `TranslatedSmallPolygonDroppedBelowEps`, `SubEpsRectangleOverlap`, `OffsetNotInvariantUnderLargeTranslation`; renamed the two whose names asserted the old behavior). Kept only `DISABLED_CenteredSubEpsNonClosingWalk` - the open non-manifold case I'd want your input on (trace + diagnosis on #1707).

elalish (approved):
Thanks! I'm going to merge this because I'm heading on vacation and I think it'll be easier to get back into review from a clean slate. Also, this is looking quite good, even though we have a bit more algorithmic stuff to work out.

PR #1760, opened 2026-06-16
CrossSection: drop FillRule/Compose, add SetTolerance/GetTolerance

zmerlynn:
## Summary

Part of #1707. Breaks the CrossSection public API now that the boolean2 backend
is the only backend (#1751).

**Dropped:**
- `FillRule` enum and constructors that accepted it. The boolean2 backend
  supports only positive-winding fill; the enum had no effect since #1751, and
  passing any non-Positive value triggered a `DEBUG_ASSERT` in assert-enabled
  builds. Callers passing `FillRule::Positive` (the only valid value) just drop
  the argument.
- `Compose(const std::vector<CrossSection>&)` static. It was a batch boolean
  union - identical to `BatchBoolean(..., OpType::Add)`. Use `BatchBoolean` or
  the `+` operator instead.
- `ManifoldFillRule` enum from the C header (`types.h`).

**Added:**
- `GetTolerance()` / `SetTolerance(double)` - mirrors `Manifold::GetTolerance` /
  `Manifold::SetTolerance`. `GetTolerance` returns the cross-section's propagated
  drift budget. `SetTolerance` returns a copy with it adjusted: raising simplifies
  to the new budget; lowering floors at the geometry's natural epsilon.

**Changed:**
- `Simplify(double epsilon = 1e-6)` -> `Simplify(double tolerance = 0)`. Default
  `0` uses the section's own tolerance (from `GetTolerance()`). Callers passing
  an explicit non-zero value are unaffected.
- `Hull(SimplePolygon pts)` / `Hull(Polygons polys)` now take const-ref.
  `HullImpl` sorts in place; the old by-value signature obscured this.
- `Rect`, `Square`, and `Circle` constructors now set `tolerance_` from
  `InferEps` on the output geometry, consistent with the `Polygons` constructor.
  Previously `GetTolerance()` returned 0 for these paths.

**Fixed (pre-existing):**
- WASM `JoinType`: `'Bevel'` was missing from the TypeScript type and from
  `joinTypeToInt` in `bindings.js`, so passing `'Bevel'` silently fell through
  to `'Square'`.

C, Python, and WASM bindings updated throughout.

## Test plan

- [ ] 119 tests pass (58 CrossSection + 39 Boolean2 + 22 CBIND existing +
  1 new `CBIND.cross_section_tolerance`). 2 seed-queue CrossSection tests remain
  disabled pending fixes on a separate branch.


elalish (approved):
Thanks, looking much cleaner!

Issue #1761, opened 2026-06-18
Issues in Benchmark guard CI


AnshulPatil2005:
I found some issues in my previously written code of Benchmark guard CI #1680 while doing the weekly benchmark #1758 

1. The job has two steps marked continue-on-error: true:
Prepare worktrees —> creates wt-base and wt-head
Benchmark —> builds and runs perfTest inside those worktrees
If step 1 fails halfway (e.g. wt-base created, wt-head failed), step 2 still runs because of continue-on-error. It will then also fail because wt-head doesn't exist. But again continue-on-error hides it.

Then the final Compare step runs, finds no benchmark data, and prints:
Result: WARNING (benchmark data invalid/skipped)
Reason: No run*.txt files found in ./bench/head

we'd have no idea that the real problem was the worktree creation failing in step 1.

2. [scripts/run_pr_perf_guard.sh:14-21] passes no Clipper2 source dir and no MANIFOLD_CROSS_SECTION=OFF. If the default CMake config has cross-section ON, CMake will attempt a live FetchContent download of Clipper2 during each base and head build once per worktree, inside CI. Network fetches in benchmark builds also introduce unpredictable latency and also base and head could download different version of clipper2 flagging it falsely.

3. run_pr_perf_guard.sh builds with -DMANIFOLD_PAR=ON, which means TBB is enabled and perfTest runs across multiple threads. On a GitHub Actions shared runner, the number of available cores and how busy those cores are varies between runs — your base run might get 4 cores mostly to itself, your head run might get the same 4 cores while another job is also running on the same machine.

So you could get a result like:

Base:  0.012 sec   (runner was quiet)
Head:  0.018 sec   (runner was busy)
Delta: +50% WARNING

AnshulPatil2005:
@elalish @pca006132 
I opened this issue to track a few small problems I noticed in the benchmark guard CI while working on the weekly benchmark PR: #1761.
I will fix it soon in an upcoming PR

pca006132:
Btw, for markdown, remember to add a space after your dot. E.g., ``1. abc` not `1.abc`.

Also, I think my recommendation for benchmark is to force it on macos arm platform, since that should be using M cores and are probably more stable. Did you try that and see how well it works?

AnshulPatil2005:
> Also, I think my recommendation for benchmark is to force it on macos arm platform, since that should be using M cores and are probably more stable. Did you try that and see how well it works?

[...]

PR #1762, opened 2026-06-21
Fix benchmark guard and add memory guard

AnshulPatil2005:
ref - #1761


AnshulPatil2005:
I kept MANIFOLD_CROSS_SECTION=OFF for this specific perfTest guard. If later we include benchmark set that has CrossSection-heavy cases, then we can turn it back ON and pass -DFETCHCONTENT_SOURCE_DIR_CLIPPER2=.... so there is no version mismatch

elalish:
I believe Clipper2 was recently removed, along with the MANIFOLD_CROSS_SECTION flag.

AnshulPatil2005:
> I believe Clipper2 was recently removed, along with the MANIFOLD_CROSS_SECTION flag.

 Yes, you are right. I checked the current CMake options so Its removed.
 this makes it more simpler

.github/workflows/manifold.yml
```diff
@@ -649,13 +649,20 @@ jobs:
...
       shell: bash
       run: bash ./scripts/prepare_pr_perf_guard_worktrees.sh "${{ github.event.pull_request.base.sha }}" "${{ github.event.pull_request.head.sha }}"
     - name: Benchmark base and head
+      id: benchmark_perf
+      if: steps.prepare_worktrees.outcome == 'success'
       continue-on-error: true
       shell: bash
       run: bash ./scripts/benchmark_pr_perf_guard.sh "${PERF_REPEATS}"
     - name: Compare benchmark results
+      if: always()
```
pca006132: always?

.github/workflows/manifold.yml
```diff
@@ -636,7 +636,7 @@ jobs:
     timeout-minutes: 45
     runs-on: ubuntu-24.04
     env:
-      PERF_REPEATS: 3
+      PERF_REPEATS: 6
```
pca006132: how much difference are we seeing here? do you have statistics?

scripts/compare_pr_perf_guard.py
```diff
@@ -8,7 +8,7 @@
 import subprocess
 from pathlib import Path
 
-SCHEMA_VERSION = "1.0.0"
+SCHEMA_VERSION = "1.1.0"
```
pca006132: what is the use of this schema version? is it currently storing stuff for other scripts to parse in the future?

.github/workflows/manifold.yml
```diff
@@ -646,16 +646,23 @@ jobs:
...
+        packages: libgtest-dev cmake
         version: 1
     - name: Prepare base and head worktrees
+      id: prepare_worktrees
       continue-on-error: true
       shell: bash
       run: bash ./scripts/prepare_pr_perf_guard_worktrees.sh "${{ github.event.pull_request.base.sha }}" "${{ github.event.pull_request.head.sha }}"
     - name: Benchmark base and head
+      id: benchmark_perf
+      if: steps.prepare_worktrees.outcome == 'success'
```
pca006132: Ok I guess I should have questioned these previously when we merge this. But why `continue-on-error` and we now need to add these outcome detection?

It is fine for the CI to fail. If it is not a required check, github will not block us from merging the PR.

scripts/compare_pr_perf_guard.py
```diff
@@ -114,33 +114,42 @@ def build_summary(
...
         head_mean = head["benchmarks"][benchmark]["mean_sec"]
-        delta_sec = head_mean - base_mean
+        base_median = base["benchmarks"][benchmark]["median_sec"]
+        head_median = head["benchmarks"][benchmark]["median_sec"]
+        base_stdev = base["benchmarks"][benchmark]["stdev_sec"]
+        head_stdev = head["benchmarks"][benchmark]["stdev_sec"]
+        delta_sec = head_median - base_median
         delta_ms = delta_sec * 1000.0
-        pct = (delta_sec / base_mean * 100.0) if base_mean > 0 else 0.0
+        pct = (delta_sec / base_median * 100.0) if base_median > 0 else 0.0
```
pca006132: btw realistically, can we really get `base_median = 0`?

scripts/compare_pr_perf_guard.py
```diff
@@ -114,33 +114,42 @@ def build_summary(
...
                 "head_mean_sec": head_mean,
-                "delta_sec": delta_sec,
-                "delta_ms": delta_ms,
-                "delta_pct": pct,
+                "base_median_sec": base_median,
+                "head_median_sec": head_median,
+                "base_stdev_sec": base_stdev,
+                "head_stdev_sec": head_stdev,
+                "delta_median_sec": delta_sec,
+                "delta_median_ms": delta_ms,
```
pca006132: do we need a separate `ms` version?

scripts/prepare_pr_perf_guard_worktrees.sh
```diff
@@ -9,5 +9,15 @@ fi
 BASE_SHA="$1"
 HEAD_SHA="$2"
 
+if [ -z "$BASE_SHA" ]; then
```
pca006132: when would these happen in the CI?

scripts/compare_pr_perf_guard.py
```diff
@@ -104,7 +104,7 @@ def parse_suite(suite_dir: Path) -> dict:
 def build_summary(
     base: dict, head: dict, warn_pct: float, warn_abs_ms: float
 ) -> tuple[str, bool, dict]:
-    # compare benchmark means between base and head with dual-threshold warnings
+    # Compare benchmark medians between base and head with dual-threshold warnings.
```
pca006132: Since you are changing the statistics here, do you have a reason for this? Is the mean significantly different from the median for these cases? Do we have long tails?

Also, I remember people saying that if you are in a noisy environment, minimum is the best metric since noise can only make things run slower, not faster. We typically use mean or median (if outliers are common) because we want to measure the real-world performance in a typical user system (subject to various noises, e.g.,, memory allocation), but CI noise is typically worse than that and we don't want to measure that.

.github/workflows/manifold.yml
AnshulPatil2005: Yes it is much cleaner to remove continue-on-error fully.

This was due to my misunderstanding of "non-blocking" : during the making of benchmark guard, I thought every git CI job is blocking so i have to make it unblocking in my code.

scripts/compare_pr_perf_guard.py
AnshulPatil2005: I was not sure about the schema of the result.json so if in future we change the schema then it would be linked with schema version.

the bump from 1.0.0 to 1.1.0 was due the the change in schema mainly for mean to median change(and also some other changes).

> what is the use of this schema version? is it currently storing stuff for other scripts to parse in the future?

Yes that was the intent here

scripts/compare_pr_perf_guard.py
AnshulPatil2005: Realistically no, but not impossible, a timing bug or low-resolution clock could produce 0.0. but, the except Exception block will catches ZeroDivisionError while calculating delta_sec / base_median, so removing it should be fine.

scripts/compare_pr_perf_guard.py
AnshulPatil2005: The only reason it existed internally was for the threshold check (delta_ms >= warn_abs_ms), but that uses a local variable, not the dict field. Will remove it from the JSON output.

scripts/prepare_pr_perf_guard_worktrees.sh
AnshulPatil2005: yea i checked the trigger conditions, the workflow already has github.event_name == 'pull_request'
so its impossible for the SHAs to be empty, should've checked it earlier while adding it.

scripts/compare_pr_perf_guard.py
AnshulPatil2005: > Since you are changing the statistics here, do you have a reason for this? Is the mean significantly different from the median for these cases? Do we have long tails?

Yea i saw the data for recent benchmark runs, the median time was overall slightly lesser than mean time always so thought mean is a worse metric than median although the difference was neglieble 
0.5-1% difference.

mean was being pulled up by occasional slow runs. Switching to median was meant to reduce that.

> Also, I remember people saying that if you are in a noisy environment, minimum is the best metric since noise can only make things run slower, not faster. We typically use mean or median (if outliers are common) because we want to measure the real-world performance in a typical user system (subject to various noises, e.g.,, memory allocation), but CI noise is typically worse than that and we don't want to measure that.

You're right though that minimum takes the same logic further and has a theoretical justification for CI specifically. 
Also minimum seems to be quite obviously be better than both median and mean.
Will switch to minimum then.

.github/workflows/manifold.yml
AnshulPatil2005: increase from 3 to 6 was a judgment call to get a more stable median.

as now i will be comparing minimum across PERF_REPEATS, more repeats would rarely push the minimum lower hence will keep it 3.

scripts/compare_pr_perf_guard.py
pca006132: But the question here is, are we storing them now? If they are not stored now, we don't need a schema version change for compatibility.

scripts/compare_pr_perf_guard.py
pca006132: @elalish do you have any opinion about this?

scripts/compare_pr_perf_guard.py
AnshulPatil2005: Yes, that makes sense. Right now the weekly results are only emitted on my fork and its for testing, not stored as persistent storage. 

bringing in the change after making a storage would be fine.

scripts/compare_pr_perf_guard.py
elalish: Yes, median is less sensitive to outliers than mean is, but I agree that taking the minimum running time of a few runs is probably the most reliable in this case.

AnshulPatil2005: > Yes, median is less sensitive to outliers than mean is, but I agree that taking the minimum running time of a few runs is probably the most reliable in this case.

will switch to minimum then

AnshulPatil2005:
I think this is done as planned, Is there anything this pr should fix/contain ?

elalish:
Sounds reasonable to me - @pca006132 feel free to merge when you're happy.

pca006132:
@AnshulPatil2005 so what is the current status? ready for review?

AnshulPatil2005:
> @AnshulPatil2005 so what is the current status? ready for review?

Yes, just one confusion in decision about pinning it to macOS should i keep it that way to switch back to ubuntu ?

AnshulPatil2005:
Also, currently it only prints peak RSS per perfTest size for diagnostics; it does not guard on memory.

we can add a dedicated guard after this pr If that is okay ?

pca006132:
1. Judging from the graphs in https://github.com/elalish/manifold/pull/1758#issuecomment-4877691310, while macos timing is unstable due to P/E core migration, timing differences between different runner for linux is even higher (despite the variance within the same runner is extremely low). So macos may still be better.
2. Yes, just add the guard about memory.

AnshulPatil2005:
> 1. Judging from the graphs in [Weekly benchmarks #1758 (comment)](https://github.com/elalish/manifold/pull/1758#issuecomment-4877691310), while macos timing is unstable due to P/E core migration, timing differences between different runner for linux is even higher (despite the variance within the same runner is extremely low). So macos may still be better.

yes, also macos is giving the same machine every time which is Apple M1 (Virtual) / VirtualMac2,1.

I'm also looking into preferring P-cores to reduce the noise (via taskpolicy -c user-interactive or similar), will report the findings soon.

pca006132:
> yes, also macos is giving the same machine every time which is Apple M1 (Virtual) / VirtualMac2,1.

This is my initial hypothesis: This limits the possible runners we get. Everything can run linux, but there are only a few chips for macos runner.

.github/workflows/manifold.yml
```diff
@@ -634,28 +634,30 @@ jobs:
...
     steps:
     - uses: actions/checkout@v6
       with:
         fetch-depth: 0
-    - name: Install apt packages
-      uses: awalsh128/cache-apt-pkgs-action@v1.6.0
-      with:
-        packages: libgtest-dev libtbb-dev cmake
-        version: 1
+    - name: Verify macOS ARM runner
```
pca006132: needed?

.github/workflows/manifold.yml
```diff
@@ -634,28 +634,30 @@ jobs:
     name: perfTest base-vs-head guard
     if: github.event_name == 'pull_request'
     timeout-minutes: 45
-    runs-on: ubuntu-24.04
+    runs-on: macos-15
     env:
       PERF_REPEATS: 3
       REGRESSION_WARN_PCT: 20
       REGRESSION_WARN_ABS_MS: 10
+      MEMORY_WARN_PCT: 20
```
pca006132: btw, forgot to mention previously, we should just use `PERCENT` instead of `PCT`. Never see this abbreviation.

.github/workflows/manifold.yml
```diff
@@ -634,28 +634,30 @@ jobs:
...
+      MEMORY_WARN_ABS_MB: 50
+      CMAKE_OSX_ARCHITECTURES: arm64
```
pca006132: Do we need this?

scripts/compare_pr_perf_guard.py
```diff
@@ -46,6 +47,18 @@ def parse_run(run_path: Path, run_index: int) -> dict:
...
+            continue
+
+        peak_rss_match = PEAK_RSS_PATTERN.search(line)
+        if peak_rss_match:
+            ntri, size_index, peak_rss_mb, peak_rss_bytes = peak_rss_match.groups()
+            benchmark_key = (
+                f"nTri={ntri}" if ntri != "unknown" else f"size_index={size_index}"
+            )
+            peak_rss_by_benchmark[benchmark_key] = {
+                "peak_rss_mb": float(peak_rss_mb),
```
pca006132, line 70: why do we need both?

scripts/compare_pr_perf_guard.py
```diff
@@ -78,21 +96,42 @@ def parse_suite(suite_dir: Path) -> dict:
...
+            "mean_sec": statistics.fmean(samples),
             "median_sec": statistics.median(samples),
             "stdev_sec": stdev(samples),
+            "min_sec": min(samples),
             "max_sec": max(samples),
             "n_runs": len(samples),
         }
+        rss_samples = peak_rss_samples[benchmark]
+        if rss_samples:
+            benchmark_stats.update(
```
pca006132: I think this tells you a pattern, you should factor out these statistics stuff instead of copying the code over and over again. We can have a nested dictionary, e.g.,

```
timing_sec:
  mean
  median
  sd (btw, we never use the name stdev...)
  min
  max
peak_rss_bytes:
  ...
```

scripts/compare_pr_perf_guard.py
```diff
@@ -114,56 +159,175 @@ def build_summary(
...
+        if base_rss is not None and head_rss is not None:
+            rss_delta_mb = head_rss - base_rss
+            rss_delta_pct = rss_delta_mb / base_rss * 100.0 if base_rss else 0.0
+            memory_this_regressed = (
+                rss_delta_pct >= memory_warn_pct
+                and rss_delta_mb >= memory_warn_abs_mb
+            )
+            memory_regressed = memory_regressed or memory_this_regressed
+
+        # peak_rss_complete is False when a benchmark is missing an RSS sample
```
pca006132: why would we miss that?

scripts/compare_pr_perf_guard.py
```diff
@@ -212,16 +384,92 @@ def detect_compiler() -> str | None:
...
+def int_or_none(value: str | None) -> int | None:
+    if value is None:
+        return None
+    try:
+        return int(value)
+    except ValueError:
+        return None
+
+
+def default_cpu_model() -> str | None:
```
pca006132: IMO these should be in another script.

scripts/compare_pr_perf_guard.py
```diff
@@ -249,7 +499,14 @@ def main() -> int:
...
+        markdown, regressed, payload = build_summary(
+            base,
+            head,
+            args.warn_pct,
+            args.warn_abs_ms,
+            args.memory_warn_pct,
+            args.memory_warn_abs_mb,
+        )
         payload["data_valid"] = True
     except Exception as exc:
```
pca006132: generally we do not capture any exception, just capture targeted ones.

scripts/run_pr_perf_guard.sh
```diff
@@ -34,7 +43,40 @@ if [ -z "$BIN" ]; then
   exit 1
 fi
 
+run_measured_perf_size() {
+  local size_index="$1"
```
pca006132: ideally, do everything in python to avoid this complicated shell script. 
and thinking about it, maybe you can integrate these scripts into just one/multiple python files? e.g., `benchmark_pr_perf_guard`, `report_pr_perf_guard`, etc.

The goal here is to avoid having shell files that no one is familiar with (I guess you asked llms to generate these). Python files is more familiar to most of us. Also, we don't need so many files if there are not that many usecases.

pca006132 (changes requested):

.github/workflows/manifold.yml
AnshulPatil2005: oh, this was left over from when the job used macos-latest which was ambiguous, so the runtime check was a real guard back then. Now that we're pinned to macos-15 explicitly that ambiguity is gone.

.github/workflows/manifold.yml
AnshulPatil2005: noted

scripts/compare_pr_perf_guard.py
AnshulPatil2005: peak_rss_bytes still makes sense to keep in the shell script's log line since that's the raw, full-precision value before the 2-decimal MB rounding, useful when reading raw logs.

I'll remove it from the python parser tho

.github/workflows/manifold.yml
AnshulPatil2005: Yes I'll remove it, these are the dead configs left over when it was macos-latest.

I think there could be several more of such dead changes; ill review the code again and clean it

scripts/compare_pr_perf_guard.py
AnshulPatil2005: This might be a bit over-the-top but the scenario i thought was: the timing line gets printed by perfTest right when the op finishes, but the RSS line only gets written by /usr/bin/time after the process fully completes. If a run gets killed in between (job timeout or OOM), we'd get a timing sample with no matching RSS sample.

this could happen to the largest benchmark size as its most likely to hit memory pressure so it's a very narrow case.

anyways, I think it is better lightweight

scripts/compare_pr_perf_guard.py
AnshulPatil2005: yep, that makes sense

scripts/run_pr_perf_guard.sh
AnshulPatil2005: Yes I had planned this change but it was getting skipped/delayed due to other changes, I'll port to Python in this PR, converting them into fewer files.

Once this is merged, I'll do the same on the weekly-benchmark scripts (run_weekly_benchmarks.py, report_weekly_benchmarks.py, etc.) in that PR that side is already Python, so it's just the "fewer files" part that applies there.

AnshulPatil2005:
> I'm also looking into preferring P-cores to reduce the noise (via taskpolicy -c user-interactive or similar), will report the findings soon.

@pca006132 after looking more into it i found we cant actually force it to P-cores, as taskpolicy (Apple's scheduling-control CLI) can only push threads down to E-cores by tagging them "background" QoS so it has no mechanism to force high-priority/P-core.

AnshulPatil2005:
@pca006132 @elalish 
I think this pr is good to go, let me know if any other changes are needed.
Once this PR is merged, I can rebase the weekly benchmark PR and reuse the shared code from it, especially system metadata, perfTest execution, peak RSS parsing, and run*.txt parsing.

scripts/compare_pr_perf_guard.py
```diff
@@ -114,56 +153,175 @@ def build_summary(
...
+        if base_rss is not None and head_rss is not None:
+            rss_delta_mb = head_rss - base_rss
+            rss_delta_percent = rss_delta_mb / base_rss * 100.0 if base_rss else 0.0
+            memory_this_regressed = (
+                rss_delta_percent >= memory_warn_percent
+                and rss_delta_mb >= memory_warn_abs_mb
+            )
+            memory_regressed = memory_regressed or memory_this_regressed
+
+        # peak_rss_mb["complete"] is False when a benchmark is missing an RSS
```
pca006132: why is this possible?

pca006132 (approved):
Nice work.

scripts/compare_pr_perf_guard.py
pca006132: did we ever get this? I think timeout is for the whole job but not a step? and if we get OOM, shouldn't we just fail the job?

pca006132:
I think if the error situation does not happen or will fail the whole job, then there is no need to handle it. Keep the logic simple. I think this is also something I notice with LLMs: they tend to be a bit overly defensive. This is sometimes a good practice, but it may unnecessarily complicate the logic.

scripts/compare_pr_perf_guard.py
AnshulPatil2005: > did we ever get this? I think timeout is for the whole job but not a step? and if we get OOM, shouldn't we just fail the job?

```
void RunSizeIndex(int i) {
  Manifold sphere = Manifold::Sphere(1, (8 << i) * 4);
  Manifold sphere2 = sphere.Translate(vec3(0.5));
  auto start = ...;
  Manifold diff = sphere - sphere2;   // <-memory can cause oom
  diff.NumTri();
  auto end = ...;
  std::cout << "nTri = " << ... << ", time = " << ... << std::endl;   // <- time prints 
}
```
oom would cause the time to not print and timeout would fail the job in 45 minutes so mismatch isnt possible imo

AnshulPatil2005:
> I think if the error situation does not happen or will fail the whole job, then there is no need to handle it. Keep the logic simple. I think this is also something I notice with LLMs: they tend to be a bit overly defensive. This is sometimes a good practice, but it may unnecessarily complicate the logic.

Very fair point the coding agent are overly defensive for all logics and error handling, that is honestly the major part I tend to modify in the code before committing.

sometimes the coding agent gives a really good logic that i cant figure out on my own which seems really useful are the ones i keep in the code

Also i read the feedback I would review the code more strictly especially in the parts mentioned

scripts/compare_pr_perf_guard.py
AnshulPatil2005: after a closer look, that is not possible due to https://github.com/elalish/manifold/pull/1762#discussion_r3563185285

I'll just add something like 
```
       if not peak_rss_samples[benchmark]:
            raise RuntimeError(
                f"No peak RSS samples for benchmark {benchmark!r} in {suite_dir}"
            )
```
in case of parsing bug (unrelated of timeout or OOM)

scripts/compare_pr_perf_guard.py
pca006132: yeah, so no need for that logic. just simplifiy the code.

pca006132:
IMO a good way to control complexity is to cut everything by default, only keep them if you have a good reason to do so. This may not be the case when dealing with code from others, or when things are mission-critical, where a defensive check that looks redundant may save you. But this is not the case here.

pca006132:
Anyway, thanks for the fixes.

AnshulPatil2005:
> IMO a good way to control complexity is to cut everything by default, only keep them if you have a good reason to do so. This may not be the case when dealing with code from others, or when things are mission-critical, where a defensive check that looks redundant may save you. But this is not the case here.

Yea that makes sense

PR #1767, opened 2026-06-25
V3.5.2

pca006132:
dummy pr for test


pca006132:
@elalish what do you think about me just pushing a release from this branch without PR? The actual PR is elsewhere. So we do not pull in the breaking changes in master.

elalish:
> @elalish what do you think about me just pushing a release from this branch without PR? The actual PR is elsewhere. So we do not pull in the breaking changes in master.

Yeah, that sounds quite reasonable.

PR #1768, opened 2026-07-01
Add SolidSKeleton to users list

rogiergoossen:
SolidSKeleton is a lightweight data format. We’re using Manifold for boolean CSG in the Rust core used by our two reference packages, and it has worked great!


elalish (approved):
Thanks for sharing!

PR #1770, opened 2026-07-05
Boolean2: pivot the 2D arrangement to Smith's sweep line

zmerlynn:
Per #1707, where the plan landed on pivoting to Smith's sweep - guaranteed
robustness is the priority - this PR now builds Boolean2's arrangement with Smith's
2D simplification sweep (RobustBoolean.pdf ch 7) instead of the #289 point-order
approach the earlier revisions used.

## Why the pivot

The point-order arrangement (BVH crossing insertion + exact-distinct vertex
identity + ray-cast winding) regressed on dense near-concurrent fans - a handful
of irregular near-concurrent triangles that master resolves to one region but
point-order turns non-manifold: the closed-walk assert with asserts on, a silent
drop of the region in release. That's the trade we hit: master's eps-snap resolves
these fans but breaks near-coincident corners; keeping crossings exactly distinct
does the reverse, and fails silently on the fan class. The `DenseNearConcurrentFan`
regression test captures it.

I tried for a while to close the fan class without a sweep, since #289 chose the
BVH broad phase specifically to avoid one. The fuller log is on #1707; the short
version is that no local, order-free rule composes into a globally consistent
arrangement - local fusion is inert or over-merges, a repair pass closes
balanced-but-wrong, a statusless block rule is silently area-wrong, and you can't
box off a tangle and resolve it locally (the long edges through a concurrence are
split by far-away crossings that decide the local topology). As I read 7.6.3/7.8,
Smith reaches roughly the same conclusion: 7.6.3 ties the arrangement's
topological validity - the start-vertex == end-vertex balance the fan breaks - to
the maintained sweep order, and 7.8 reports his non-sweep alternatives either
iterated excessively or gave too much geometric error. The maintained order is
what carries the consistency.

## What this does

Replaces the point-order arrangement with Smith's sweep:

- a maintained-status Bentley-Ottmann sweep that discovers crossings as edges
  become adjacent and resolves near-concurrent events with the 7.6.2 block rule
  (the contiguous status block bracketing an event is forced through it in status
  order), and
- a second sweep pass that classifies the arrangement by winding, accumulating
  through each block in status order rather than by an independent per-vertex ray
  cast.

The BVH broad phase and the eps-scale incidence pre-split are kept; everything
downstream of them is now the sweep. The point-order machinery goes away
(`boolean2_winding.cpp`, the projected-order predicates, the crossing-insertion
path), so the pivot is a net reduction in code. `docs/Boolean2.md` is rewritten to
match, with a note on how the engine departs from the #289 sketch and why.

## Validation

- The `DenseNearConcurrentFan` regression goes green; the whole `CrossSection` +
  `Boolean2` suite is 110/110 with asserts on.
- Dense near-concurrent fans (the fuzz-found class) resolve to the reference
  region across unit-to-2^60 scale.
- Two library-reuse changes in the sweep (`la::cross` in the status comparator,
  and routing the crossing math through the shared `Interpolate`/`Intersect`
  kernels in `shared.h`) are byte-for-byte identical to the hand-rolled forms on
  the dense-fan corpus - verified inert, not taken on faith.

## Known tradeoffs

- This reopens what #289 set aside for the O(n log n) / parallel argument. The
  sweep is serial; 2D construction isn't perf-critical here (the old winding
  filter was already serial), and it's faster on dense soups, though ~1.3x slower
  on large sparse inputs - a real design change, not a drop-in.
- The sweep is topologically robust, not geometrically exact - Smith is explicit
  (7.7) that sub-rounding residuals can survive. The winding pass forces them
  through rather than leaving self-intersections, but I'm not claiming exactness.


elalish (approved):
Nice work. I was partway through suggesting something like this, but ended up writing [this comment](https://github.com/elalish/manifold/issues/1707#issuecomment-4829626701) instead. I was concerned this exact-equal thing might have trouble with three edges crossing where two are nearly vertical (or horizontal). Wasn't sure if there could still be some inconsistency between which of their exactly overlapping points were chosen as the intersection and how that would compare to the third edge. However, maybe this isn't actually a problem in practice? 

I take it you haven't managed to find any problem cases from fuzzing? Seems like a good sign. Do you know how to generalize this algorithm to 3D?

zmerlynn:
On fuzzing: it's been the driver throughout - several of the PR's regression tests are fuzz finds (the Extrude -> Project -> Slice one directly, the rest from targeted fuzzing against candidate fixes), and beyond this near-coincident-crossing class it hasn't turned up other classes for me so far. It did find this one, and I chased a targeted fix before concluding it needs a sweep (below). It's your #1707 near-vertical concern at higher edge count - a new red test (DenseNearConcurrentFanNonClosingWalk): seven irregular near-concurrent triangles that master resolves to one region (~1.01e29), but keeping crossings exactly distinct turns the arrangement non-manifold (the closed-walk assert with asserts on, a silent drop of the region in release). It reproduces from unit scale to 2^60; the irregularity is the trigger.

The bare three-edge case is fine (two straight segments cross once). It takes ~7+ irregular edges - a dense within-alpha crossing cluster that point-order insertion can't keep consistent, leaving the retained graph unbalanced.

So this is a trade, not a strict win: master's eps-snap resolves these fans but breaks near-coincident corners; keeping distinct does the reverse. The corner class motivated the PR, but the fan class fails silently in release, so I don't think it should land as-is without addressing that (more below).

On the targeted fix: I tried to close it without a sweep, since #289 chose the BVH to avoid one, but I couldn't find one that holds - local fusion is inert or over-merges, a repair pass is incomplete or closes balanced-but-wrong, a statusless block-rule over the BVH is silently area-wrong on part of the class, and you can't box off the tangle and resolve it locally (the edges are long, so their boxes span the input, and the far crossings that split them decide the local topology). I'm not sure it's impossible, but Smith seems to have hit the same wall: his section 7.6.3 argues the sweep is what maintains topological validity - the start-vertex == end-vertex balance the fan breaks - and he notes the alternatives he tried either iterated excessively or gave too much geometric error (7.8) - the one with the geometric error being a sweep that dropped the maintained data and ran on the original edges, which is exactly what separated the statusless attempt here (silently wrong) from the maintained-status one (correct). So the sweep reads as load-bearing for correctness there, not just the O(n log n) speedup.

For what it's worth, I do have a rough sweep-line prototype (Smith's block rule plus a global winding pass) that so far handles both the corner class and this fan class, matching master where it's right and tracking the truth where it isn't, across the fuzz corpus and dense cases up to a few hundred edges. It isn't wired into the engine, so I'd read it as a signal that the sweep can resolve this class, not a proposal - your call on how to weigh the two failure classes and whether the sweep is worth it.

I'll follow up on #1707 with a fuller log of what I've tried and the 3D stepping-stone question.

zmerlynn:
Per your note on #1707 that pivoting to Smith's sweep makes sense - I've pushed
the sweep to this PR, replacing the point-order engine with Smith's ch-7 sweep. It
turns the DenseNearConcurrentFan regression green and holds the
near-coincident-corner class, 110/110 with asserts on. I've updated the
description to match.

The honest tradeoffs (it reopens the O(n log n)/parallel question, and it's
topologically robust rather than geometrically exact) are in the description. I'll
follow up on #1707 on your two open threads - whether #6 matches your
shadow-against-originals idea (I don't think it does), and the 3D sweep-plane
analogy.

src/boolean2_sweep.cpp
```diff
@@ -0,0 +1,458 @@
...
+  void ProcessEvent(const vec2& p) {
+    const size_t n = status_.size();
+    std::vector<Side> cls(n);
+    for (size_t i = 0; i < n; ++i) cls[i] = Classify(status_[i], p);
+    size_t lo = 0;
+    while (lo < n && cls[lo] == Side::UNDER) lo++;
+    size_t hi = n;
+    while (hi > lo && cls[hi - 1] == Side::OVER) hi--;
+    std::vector<SweepEdge> reinsert;
+    long long w = 0;
```
elalish: Our style uses `size_t` or `uint64_t` instead of `long long`.

test/cross_section_test.cpp
```diff
@@ -568,6 +558,323 @@ TEST(CrossSection, NearCoincidentCornersNonClosingWalk) {
...
+       {545918584749244.25, -928345427098964.88},
+       {54081415250756.258, 528345427098965.06}},
+      {{392560430378768.94, -644468286981061},
+       {401641806989572.88, -642381746258070.88},
+       {198358193010426.28, 242381746258070.59}},
+  };
+  try {
+    const CrossSection u(fan);
+    EXPECT_GT(u.Area(), 0.0) << "dense near-concurrent fan dropped entirely";
+  } catch (const std::exception& e) {
```
elalish: Is there a need to try catch inside of a googletest? It already catches and reports the errors, doesn't it? Also in the test above.

elalish (approved):
A couple minor comments, but this looks great. Love that it is simpler.

src/boolean2_sweep.cpp
zmerlynn: Good catch. Switched to int64_t rather than uint64_t/size_t because these carry signed multiplicity/winding: Subtract feeds the second operand negative multiplicity, and the accumulated winding goes negative - a B-only region in A - B has w = -1, which has to read as outside. Unsigned would wrap it to a large positive and wrongly keep the region.

test/cross_section_test.cpp
zmerlynn: You're right, removed both - gtest catches the thrown assert and reports it as a failure.

PR #1775, opened 2026-07-24
document manifoldcad global settings

### LINKED ISSUE #1773: ManifoldCAD API Documentation skips functions
I've noticed this for `setCircularSegments`, which is used within the examples, and seems to carry the correct docstring, yet fails to show up in the [documentation](https://manifoldcad.org/docs/jsuser/functions/getCircularSegments.html).

tylergibbs1:
fixes #1773

the user guide only passed manifoldCAD.d.ts to typedoc, so the top level settings declared in manifoldCADGlobals.d.ts were never analyzed. this adds a docs entry point that reexports both declaration files while keeping typedoc on a single module.

that documents setCircularSegments and the other global animation and circular settings without changing any existing documentation urls. the existing package introduction is also preserved.

validation

* generated the user guide with npx typedoc@0.28.16 --options typedoc.user.json --skipErrorChecking
* compared the baseline and updated output paths and confirmed every existing url is retained
* confirmed all seven global setting pages are generated
* ran clang-format in dry run mode with warnings treated as errors

this sparse checkout does not contain the generated wasm declarations or full dependency tree, so typedoc type checking and the full npm build were not run.

### DIFF
--- a/bindings/wasm/tsconfig.json
+++ b/bindings/wasm/tsconfig.json
@@ -27,10 +27,10 @@
   ],
   "files": [
     "./types/manifoldCAD.d.ts",
-    "./types/manifoldCADGlobals.d.ts"
+    "./types/manifoldCADGlobals.d.ts",
+    "./types/manifoldCADUserGuide.d.ts"
   ],
   "exclude": [
     "./lib/*.test.ts"
   ]
 }
-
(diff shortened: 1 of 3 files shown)


elalish (approved):
Thanks for noticing and fixing!

PR #1776, opened 2026-07-24
Reuse triangulator scratch buffers

### LINKED ISSUE #1149: Optimize allocation in triangulator
We can probably optimize away some allocator calls in triangulator by reusing buffers across triangulation calls.

lite-tx:
## Summary

- reuse ear-clipping scratch buffers across face triangulations
- scope storage to one `Face2Tri` operation: `tbb::combinable` supplies one triangulator per worker in parallel builds, while non-TBB builds reuse one operation-local instance
- reset iterator-bearing state in a safe order while retaining vector and collider capacity
- add sequential and concurrent regressions for alternating polygon shapes

## Why

The ear-clipping path rebuilt its polygon and collider scratch buffers for every face. Reusing them removes repeated allocations without sharing mutable state between workers. Native `thread_local` storage is deliberately avoided, and public triangulation calls remain call-local.

## Results

Matched Linux/GCC 13 Release builds, one pinned core, seven interleaved runs after warm-up, forced ear clipping:

| Vertices | Base | This PR | Change |
|---:|---:|---:|---:|
| 8 | 2.252 us/call | 2.016 us/call | 10.5% faster |
| 16 | 7.116 us/call | 5.814 us/call | 18.3% faster |
| 64 | 34.791 us/call | 30.961 us/call | 11.0% faster |
| 256 | 201.023 us/call | 200.029 us/call | within noise (0.5% faster) |
| mixed 16/64/256 | 85.492 us/call | 84.882 us/call | within noise (0.7% faster) |

Steady-state global `new`/`new[]` instrumentation over 1,000 calls:

| Vertices | Allocations/call (base -> PR) | Allocated bytes/call (base -> PR) |
|---:|---:|---:|
| 8 | 58 -> 48 | 3,708 -> 2,276 |
| 16 | 119 -> 108 | 7,972 -> 5,180 |
| 64 | 457 -> 444 | 33,588 -> 22,700 |
| 256 | 1,803 -> 1,788 | 140,644 -> 97,340 |

Outputs matched in every benchmark run.

## Validation

- Linux/GCC 13, TBB ON, strict `-Werror`: 525/525 tests
- Windows/MSVC, TBB OFF, Release `/WX`: 507/507 tests
- Linux ASan+UBSan mixed-size reusable-path stress
- Linux TBB-TSAN reusable-path and representative Boolean-path tests
- clang-format 20 and `git diff --check`

Fixes #1149.

### DIFF
--- a/src/polygon.cpp
+++ b/src/polygon.cpp
@@ -15,6 +15,8 @@
 #include "manifold/polygon.h"
 
 #include <map>
+#include <memory>
+#include <optional>
 #include <set>
 #include <utility>
 
@@ -837,7 +839,7 @@ class EarClip {
   // Create a collider of all vertices in this polygon, each expanded by
   // epsilon_. Each ear uses this BVH to quickly find a subset of vertices to
   // check for cost.
-  IdxCollider& VertCollider(VertItr start) {
+  void BuildVertCollider(VertItr start) {
     ZoneScoped;
     collider_.points.clear(false);
     collider_.itr.clear();
@@ -848,7 +850,6 @@ class EarClip {
     });
 
     BuildTwoDTree(collider_.points);
-    return collider_;
   }
 
   // The main ear-clipping loop. This is called once for each simple polygon -
@@ -856,9 +857,9 @@ class EarClip {
   void TriangulatePoly(VertItr start) {
     ZoneScoped;
 
-    IdxCollider& vertCollider = VertCollider(start);
+    BuildVertCollider(start);
 
-    if (vertCollider.itr.empty()) {
+    if (collider_.itr.empty()) {
       PRINT("Empty poly");
       return;
     }
@@ -868,7 +869,7 @@ class EarClip {
     earsQueue_.clear();
 
     auto QueueVert = [&](VertItr v) {
-      ProcessEar(v, vertCollider);
+      ProcessEar(v, collider_);
       ++numTri;
       v->PrintVert();
     };
@@ -891,8 +892,8 @@ class EarClip {
       ClipEar(v);
       --numTri;
 
-      ProcessEar(v->left, vertCollider);
-      ProcessEar(v->right, vertCollider);
+      ProcessEar(v->left, collider_);
+      ProcessEar(v->right, collider_);
       // This is a backup vert that is used if the queue is empty (geometrically
       // invalid polygon), to ensure manifoldness.
       v = v->right;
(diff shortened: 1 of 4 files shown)


elalish:
Can you measure any speed difference? @pca006132 I'll leave this one to you.

lite-tx:
Yes — I measured this against the PR base `895bfc2` and head `4461bbf`.

Setup: Ubuntu 24.04, GCC 13.3, Ryzen 9 7940HX, matched static Release builds (`-O3 -DNDEBUG`, `MANIFOLD_PAR=OFF`, DEBUG/ASSERT off). The benchmark repeatedly calls `TriangulateIdx(..., allowConvex=false)` on concave alternating-radius star polygons. I pinned it to one core and took the median of 11 interleaved base/head runs after warm-up. Output checksums matched in every case.

| Vertices | Base | This PR | Change |
|---:|---:|---:|---:|
| 8 | 2.325 us/call | 2.149 us/call | 7.6% faster |
| 16 | 7.349 us/call | 6.132 us/call | 16.6% faster |
| 64 | 36.388 us/call | 32.836 us/call | 9.8% faster |
| 256 | 210.556 us/call | 211.040 us/call | within noise (-0.2%) |
| mixed 16/64/256 | 91.802 us/call | 90.939 us/call | 0.9% faster |

I also instrumented global `new`/`new[]` over 1,000 steady-state calls to verify the intended mechanism:

| Vertices | Allocations/call (base -> PR) | Allocated bytes/call (base -> PR) |
|---:|---:|---:|
| 8 | 58 -> 48 | 3,708 -> 2,276 |
| 16 | 119 -> 108 | 7,972 -> 5,180 |
| 64 | 457 -> 444 | 33,588 -> 22,700 |
| 256 | 1,803 -> 1,788 | 140,644 -> 97,340 |

So the reuse gives a measurable benefit for repeated small-to-medium triangulations; for larger polygons, ear-clipping work dominates and runtime is effectively unchanged, while allocated bytes still drop by about 31–39%. I can add the benchmark harness to the PR if that would be useful.

src/polygon.cpp
```diff
@@ -818,25 +837,26 @@ class EarClip {
   // Create a collider of all vertices in this polygon, each expanded by
   // epsilon_. Each ear uses this BVH to quickly find a subset of vertices to
   // check for cost.
-  IdxCollider VertCollider(VertItr start) const {
+  IdxCollider& VertCollider(VertItr start) {
```
pca006132: we don't need to return the collider now I guess.

src/polygon.cpp
```diff
@@ -939,8 +959,8 @@ HalfedgeTriangulation TriangulateIdxHalfedges(const PolygonsIdx& polys,
     if (allowConvex && IsConvex(polys, epsilon)) {  // fast path
       result = TriangulateConvex(polys);
     } else {
-      EarClip triangulator(polys, epsilon);
-      result = triangulator.Triangulate();
+      thread_local static EarClip triangulator;
```
pca006132: I think we want to use the tbb's thread local storage, and it is fine if we need an internal API that passes the triangulator in. IIRC, there are platform-specific limits for thread local so we want to avoid using that, but I may be wrong about that.

pca006132 (commented):
Nice. Originally I was thinking about reusing a memory resource, but iirc there is problem with Mac since it is unsupported there for some reason, and I never revisited reusing memory.

pca006132 (approved):
LGTM. Btw, I am curious if we can get better performance if we use a shared pmr allocator (still thread local).

For macos we need to set deployment target (https://developer.apple.com/xcode/cpp/), so if we want to use it we may need to conditionally compile it only when supported. But if the benefit is large, that will be worth it.

> Requires Xcode 15
>
> Minimum deployment target: iOS 17.0, macOS 14.0, watchOS 10.0 and tvOS 17.0.

lite-tx:
Thanks! I’ll benchmark a PMR version separately. If the gain is worthwhile, I’ll look into the macOS deployment-target guard.

pca006132:
so I think we can merge this for now?

lite-tx:
Sounds good, thanks!

Issue #1774, opened 2026-07-23
A pure-Rust port of Manifold — thank you for this library!


larsbrubaker:
Hi @elalish and Manifold contributors,

This isn't a bug report — just a note of appreciation and an FYI that I hope is welcome.

Over the past months I've been building **[manifold-rust](https://github.com/larsbrubaker/manifold-rust)**, a pure-Rust port of Manifold. Working through the codebase function by function gave me a deep appreciation for how much careful engineering has gone into this library — the ε-exact boolean algorithm, the halfedge invariants, the collider design, the attention to degenerate cases. It's genuinely impressive work, and the clean structure of the C++ source is what made porting it feasible at all. Thank you for building it and maintaining it in the open.

A few notes on the port, in case they're of interest:

- **Goal is exact behavioral match, not "inspired by":** identical results on identical inputs, validated by instrumenting both implementations and comparing outputs. Every C++ test has been ported or covered (currently 522 passing Rust tests), and the port tracks upstream releases (currently v3.5.2).
- **Performance is at parity with sequential C++**, measured with Rust ports of your own perf drivers (`extras/perf_test.cpp` and `extras/large_scene_test.cpp`), and peak memory is within ~10%. An optional rayon-based `parallel` feature parallelizes only determinism-preserving stages, so parallel results stay bit-identical to sequential.

  Best-of-several runs, same machine (Windows 10, i7-7660U, both sides sequential — C++ Release with `MANIFOLD_PAR=OFF`, MSVC 19.44; Rust with fat LTO). Triangle counts are identical on every benchmark:

  | Benchmark | input tris | C++ | Rust |
  |---|---|---|---|
  | sphere ∖ sphere | 32 768 | 48 ms | 43 ms |
  | sphere ∖ sphere | 131 072 | 170 ms | 177 ms |
  | sphere ∖ sphere | 524 288 | 659 ms | 621 ms |
  | sphere ∖ sphere | 2 097 152 | 2.61 s | 2.57 s |
  | sphere grid union, n=10 (999 spheres) | — | 2.29 s | 2.32 s |
  | sphere grid union, n=20 (7 999 spheres) | — | 13.5 s | 14.5 s |

- **No C++ dependency:** it's a from-scratch translation (Apache-2.0, same as upstream), so it runs anywhere Rust does, including a [WASM demo](https://larsbrubaker.github.io/manifold-rust/). Published on crates.io as [`manifold-rust`](https://crates.io/crates/manifold-rust).

[...]

PR #1777, opened 2026-07-24
Add manifold-rust (native Rust port) to Bindings & Packages

larsbrubaker:
As requested in #1774: adds [manifold-rust](https://github.com/larsbrubaker/manifold-rust) to the Bindings & Packages table, with a sentence in the section intro clarifying that it's a native port rather than a binding — no C++ dependency, but it passes Manifold's full test suite with identical results, so it behaves like a binding from a user's perspective.


elalish (approved):
Thanks!

PR #1778, opened 2026-07-24
Optimize MergeMeshGL open-edge discovery

lite-tx:
## Summary

- replace per-edge `std::multiset` updates with a compact encoded edge array
- populate the array in parallel and use Manifold's integral sort path
- cancel opposing edge counts in a linear scan while preserving duplicate, collapsed self-edge, and deterministic merge-root behavior
- add a regression for collapsed self-edge parity

## Performance

For a shuffled mesh of 1,000,000 disjoint triangles, measuring the complete `MeshGL::Merge()` call in matched Linux/GCC 13 Release builds with TBB, `MANIFOLD_DEBUG`, and `MANIFOLD_ASSERT` enabled:

- upstream: 2.394-2.408 seconds
- this branch: 0.575-0.623 seconds

Both versions returned `changed = true` with zero merge entries. This is a 74-76% reduction in elapsed time for this open-edge-heavy case.

## Validation

- exact output comparison against pristine upstream for 2,000 deterministic randomized meshes, including repeated, collapsed, and pre-merged edges
- Linux/GCC 13, TBB enabled, `MANIFOLD_DEBUG`, `MANIFOLD_ASSERT`, and `-Werror`: all 524 tests
- Windows/MSVC Release: all 507 tests
- ASan+UBSan randomized differential test
- clang-format and `git diff --check`

Addresses item 5 in #1474.

### DIFF
--- a/src/sort.cpp
+++ b/src/sort.cpp
@@ -13,7 +13,7 @@
 // limitations under the License.
 
 #include <atomic>
-#include <set>
+#include <cstdint>
 
 #include "disjoint_sets.h"
 #include "impl.h"
@@ -31,6 +31,23 @@ uint32_t MortonCode(vec3 position, Box bBox) {
   if (std::isnan(position.x)) return kNoCode;
 
   return Collider::MortonCode(position, bBox);
+}
+
+uint64_t EncodeOpenEdge(int first, int second) {
+  const uint32_t low = static_cast<uint32_t>(std::min(first, second));
+  const uint32_t high = static_cast<uint32_t>(std::max(first, second));
+  const uint64_t direction = first > second ? 1 : 0;
+  // Vertex indices are non-negative ints, so 31 bits each leaves one bit for
+  // direction. Keeping direction in the low bit groups opposing edges together
+  // when the encoded values are sorted.
+  return (static_cast<uint64_t>(low) << 32) |
+         (static_cast<uint64_t>(high) << 1) | direction;
+}
+
+int OpenEdgeFirst(uint64_t edge) {
+  const int low = static_cast<int>(edge >> 32);
+  const int high = static_cast<int>((edge >> 1) & 0x7FFFFFFFu);
+  return (edge & 1) == 0 ? low : high;
 }
 
 struct ReindexFace {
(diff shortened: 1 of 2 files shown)


elalish:
Thanks, that looks like a pretty impressive performance improvement for that function. @pca006132 any idea what's going on with the fuzzer? I feel like the error messages aren't especially illuminating.

pca006132:
Nice. I haven't yet looked too closely into the details, but the edge pairing code may be similar to https://github.com/elalish/manifold/blob/26fe6d8c4988bfcd12c3ce089c66b52b49cab0c3/src/impl.cpp#L456, where it is somewhat parallelized. Not sure if that helps.

@elalish it seems that the CI failure was due to a download error. I restarted it, and it seems to be building correctly.

lite-tx:
Thanks for the pointer. I tried the same range partitioning in a local TBB build. End-to-end Merge() was about 4-6% slower at 250k disjoint triangles and roughly neutral at 1M, so I kept the simpler scan. Edge generation and the integer sort are already using the parallel path.

pca006132:
Will it be simpler if we use `std::unordered_map<UnorderdEdge, std::pair<int, int>>` where `UnorderedEdge` is the edge representation you are using (except the forward bit), and the pair stores the count for forward and backward. I guess the code should be significantly simpler with that, but not sure about performance. Ideally it should be cheaper than a sort. For parallelization, we can use tbb's `concurrent_unordered_map`.

lite-tx:
I tried both versions. The sequential counting loop is shorter, but the parallel version needs atomic forward/backward counters because `concurrent_unordered_map` does not make writes to the mapped value safe by itself.

For 1M triangles in Windows/MSVC Release builds with TBB, the complete `Merge()` times were:

- disjoint triangles: sort 0.456 s, `std::unordered_map` 1.584 s, TBB map 1.608 s
- closed tetrahedra, where all edges cancel: sort 0.022 s, `std::unordered_map` 0.602 s, TBB map 0.677 s

The process peak on the disjoint case was about 684 MiB for the sort, 864 MiB for the standard map, and 841 MiB for the TBB map. In a sequential build, the standard map was also slower on these workloads. It only won on an artificial case with 1M triangles but three distinct edge keys.

Both prototypes matched the current output on 2,000 randomized cases and passed all 521 Windows tests. Given the performance and memory results, I think the current contiguous sort is the better tradeoff here.

pca006132 (approved):

PR #1779, opened 2026-07-26
Use PMR for reusable triangulator nodes

lite-tx:
## Summary

- allocate reusable ear-clipping node containers from a worker-local `std::pmr::unsynchronized_pool_resource`
- keep the pool alive across triangulation calls, so cleared nodes are reused instead of releasing the resource after every run
- let the pool grow through its upstream resource for larger polygons; there is no fixed 8 KiB ceiling
- keep the one-shot public path on `std::allocator`
- fall back automatically when `__cpp_lib_memory_resource` is unavailable, including unsupported Apple deployment targets

This is the PMR follow-up discussed in #1776. The shared pool backs `holes_`, `hole2BBox_`, and `earsQueue_`; it remains operation-local in serial builds and worker-local through `tbb::combinable` in parallel builds.

Apple's [C++ support matrix](https://developer.apple.com/xcode/cpp/) lists `<memory_resource>` as requiring Xcode 15 and minimum deployment targets of macOS 14, iOS/tvOS 17, and watchOS 10. Apple libc++ gates the standard feature-test macro by deployment target, so older targets retain the existing allocator path without raising Manifold's minimum target.

## Reuse and failure recovery

A counting upstream resource saw 7 allocations (30,576 bytes) during the first 1,024-vertex triangulation. A second identical call followed by 100 smaller calls made no additional upstream allocations. All blocks were returned when the triangulator was destroyed.

A separate probe forced an upstream `bad_alloc` during a 4,096-vertex call. The same triangulator then completed both a small input and a 4,096-vertex retry, with all blocks returned at destruction.

## Performance

The existing `perfTest` case with 524,288 triangles ran five times with a 0.246 s median (0.217-0.290 s range). The PR's base-vs-head performance guard will provide the authoritative comparison for the pushed revision.

## Validation

- Linux/GCC 13 strict Release, serial: 508/508 tests
- Linux/GCC 13 strict Release, TBB: 523/523 tests
- Windows/MSVC Release `/WX`: build plus 508/508 tests
- ASan+UBSan reuse/allocation probes, including 50 consecutive repetitions
- TSan: `TriangulatorReuse.ClearsState` and the 8-thread `IsThreadLocal` stress test
- forced no-feature-macro build of the allocator fallback path
- `git diff --check`

### DIFF
--- a/src/polygon.cpp
+++ b/src/polygon.cpp
@@ -14,11 +14,22 @@
 
 #include "manifold/polygon.h"
 
+#include <cstddef>
 #include <map>
 #include <memory>
+#if defined(__has_include)
+#if __has_include(<memory_resource>)
+#include <memory_resource>
+#endif
+#endif
 #include <optional>
 #include <set>
 #include <utility>
+
+// Apple libc++ defines this only for deployment targets with PMR support.
+#if defined(__cpp_lib_memory_resource) && __cpp_lib_memory_resource >= 201603L
+#define MANIFOLD_POLYGON_HAS_PMR
+#endif
 
 #include "manifold/manifold.h"
 #include "manifold/optional_assert.h"
@@ -233,8 +244,14 @@ HalfedgeTriangulation TriangulateConvex(const PolygonsIdx& polys) {
  * within epsilon.
  */
 
+template <typename NodeAllocator = std::allocator<std::byte>>
 class EarClip {
  public:
+  explicit EarClip(const NodeAllocator& allocator = NodeAllocator())
+      : holes_(MaxX(), NodeListAllocator(allocator)),
+        hole2BBox_(std::less<VertItr>(), HoleBoxAllocator(allocator)),
+        earsQueue_(MinCost(), NodeListAllocator(allocator)) {}
+
   HalfedgeTriangulation Triangulate(const PolygonsIdx& polys, double epsilon) {
     ZoneScoped;
 
@@ -271,8 +288,8 @@ class EarClip {
 
  private:
   struct Vert;
-  using VertItr = std::vector<Vert>::iterator;
-  using VertItrC = std::vector<Vert>::const_iterator;
+  using VertItr = typename std::vector<Vert>::iterator;
+  using VertItrC = typename std::vector<Vert>::const_iterator;
   struct MaxX {
     bool operator()(const VertItr& a, const VertItr& b) const {
       return a->pos.x > b->pos.x;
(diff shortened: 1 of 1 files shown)


pca006132:
I would probably make the buffer persisted (similar what the last PR is doing, shared between runs in the triangulation phase), and make it growable instead of merely 8KB. I don't think we are that limited on memory anyway.

elalish:
What makes this a draft PR? Is it ready for review?

lite-tx:
Yes, it is ready for review. I kept it in draft while completing the failure-recovery, thread-safety, and performance validation. The current revision also follows @pca006132's suggestion: the pool persists across triangulation calls and grows through its upstream resource, with no fixed 8 KiB ceiling. The full check rollup is green, so I'm marking it ready now.

elalish (approved):
LGTM - @pca006132 I'll let you merge this one if you're happy.

pca006132:
any performance numbers?

lite-tx:
Yes — sorry, I should have included the matched comparison when I marked this ready.

This is the incremental comparison for this PR: `f2b7582` (which already includes #1776's reusable scratch vectors) vs `11780bf`, so it isolates the PMR node-container reuse rather than counting the #1776 gain again.

Setup: Ubuntu 24.04, GCC 13.3, Ryzen 9 7940HX, matched static Release builds (`-O3 -DNDEBUG`, `MANIFOLD_PAR=OFF`). The benchmark keeps one `PolygonTriangulator` alive, forces ear clipping on alternating-radius star polygons (`allowConvex=false`), pins execution to one core, and takes the median of 11 interleaved base/head runs after warm-up. Output checksums matched in every case.

| Vertices | Base | This PR | Change |
|---:|---:|---:|---:|
| 8 | 2.03 us/call | 1.86 us/call | 8.4% faster |
| 16 | 5.82 us/call | 5.55 us/call | 4.6% faster |
| 64 | 30.55 us/call | 28.18 us/call | 7.8% faster |
| 256 | 194.42 us/call | 187.60 us/call | 3.5% faster |
| mixed 16/64/256 | 85.17 us/call | 80.43 us/call | 5.6% faster |

As a negative control, I repeated the same comparison through the one-shot public `TriangulateIdx` path, which this PR is not intended to optimize. It was effectively flat: results ranged from 0.3% faster to 0.8% slower, with the mixed case at 0.06% faster.

The pool-level allocation probe also confirmed the intended reuse: the first 1,024-vertex call requested 7 upstream blocks totaling 30,576 bytes; a second identical call plus 100 smaller calls requested no additional upstream blocks, and all blocks were returned when the triangulator was destroyed.

So the final persistent/growable pool gives a modest but repeatable 3.5–8.4% improvement on the reusable path in these cases (5.6% for the mixed workload), on top of the improvement already delivered by #1776, without materially changing the one-shot path.

src/polygon.cpp
```diff
@@ -271,8 +288,8 @@ class EarClip {
 
  private:
   struct Vert;
-  using VertItr = std::vector<Vert>::iterator;
-  using VertItrC = std::vector<Vert>::const_iterator;
+  using VertItr = typename std::vector<Vert>::iterator;
```
pca006132, line 291: are these `typename` required?

pca006132 (approved):
LGTM. @elalish what do you think about this change?

src/polygon.cpp
lite-tx: Yes, for C++17 they are required. `EarClip` is now a class template, so `std::vector<Vert>` is a dependent scope and `iterator` / `const_iterator` need the `typename` disambiguator. Omitting it in this alias context is permitted starting in C++20 (P0634R3).

I also rechecked with this PR's GCC 13 `-std=c++17` compile command: the current source compiles, while removing the two keywords fails on these lines. Some Clang configurations accept the omission as a C++20 extension, which may explain the local pass.

PR #1784, opened 2026-07-31
Fixes #1781 - Manifold::Transform() does not correctly handle mirror transforms

### LINKED ISSUE #1781: Manifold::Transform() does not correctly handle mirror transforms
I import MeshGL objects into Manifold objects, call Transform() on the manifolds, and combine the manifolds with boolean operations. I notice that when a transform has a negative determinant (i.e. a mirror transform), vertices in the output mesh seem to be duplicated, with incorrect normals. The image shows my problem, before and after my suggested fix below:

Image

Vertex normals are the red vectors. The "before fix" mesh has each vertex duplicated three times with different normals.

Another way to show the problem is:
```
const Manifold s = Manifold::Sphere(1.0, 32).CalculateNormals(0, 180);
printf("%u -> %u\n", s.GetMeshGL().NumVert(), s.Mirror({1, 0, 0}).GetMeshGL().NumVert());
```
This shows mirroring inflates the original 258 vertices to 3x that number.

The problem seems to be that `FlipTris` (`src/mesh_fixes.h`) reverses each triangle's winding correctly but permutes the corner `propVert` indices by one step relative to the vertices. Every corner ends up with a neighbor's property. The tangent path already seems to handle this correctly via `FlipHalfedge`.

My suggested fix is, in `FlipTris`, to replace
```
    halfedge.SetStart(3 * tri + i, face[i].startVert);
    halfedge.SetPair(3 * tri + i, face[i].pairedHalfedge);
    halfedge.SetProp(3 * tri + i, face[i].propVert);
  }
```
with
```
  const std::array<int, 3> props = {halfedge.Prop(3 * tri), halfedge.Prop(3 * tri + 2), halfedge.Prop(3 * tri + 1)};
    halfedge.SetStart(3 * tri + i, face[i].startVert);
    halfedge.SetPair(3 * tri + i, face[i].pairedHalfedge);
    halfedge.SetProp(3 * tri + i, props[i]);
  }
```

Evenedric:
Fixed code, added regression test.

### DIFF
--- a/src/mesh_fixes.h
+++ b/src/mesh_fixes.h
@@ -59,10 +59,11 @@ struct FlipTris {
       std::swap(face[i].startVert, face[i].endVert);
       face[i].pairedHalfedge = FlipHalfedge(face[i].pairedHalfedge);
     }
+    const std::array<int, 3> props = {halfedge.Prop(3 * tri), halfedge.Prop(3 * tri + 2), halfedge.Prop(3 * tri + 1)};
     for (const int i : {0, 1, 2}) {
       halfedge.SetStart(3 * tri + i, face[i].startVert);
       halfedge.SetPair(3 * tri + i, face[i].pairedHalfedge);
-      halfedge.SetProp(3 * tri + i, face[i].propVert);
+      halfedge.SetProp(3 * tri + i, props[i]);
     }
   }
 };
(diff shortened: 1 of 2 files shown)


elalish (approved):
Thanks!

PR #1787, opened 2026-08-08
versioned documentation deploy

### LINKED ISSUE #1731: Versioned docs
We have a lot of docs pages, for C++ and various pages for JS/TS. However, all of these are served off of master, which means if we make breaking changes (which we plan to do soon for v4.0), those become the only available public docs, which is confusing for anyone using the actual 3.5 release (or earlier). 

This is a very common thing, and plenty of other docs pages have some kind of version selector, or at least different URLs. I'm not looking for anything fancy, but at the very least keeping the latest release separate from master and making all public docs links point to the last release docs would be nice. Even better if we can forward old docs links to the release docs.

I hope there's some system or package we can use for this so we don't have to complicate our own systems too much. This may be good fodder for an LLM too. Any takers?

AnshulPatil2005:
Closes #1731

Replaces `force_orphan: true` with subdirectory deploys so doc versions accumulate in gh-pages instead of being wiped on every push.

- master push → `master/`
- release published → `vX.Y.Z/` + `latest/`
- CNAME stays at gh-pages root

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -1,11 +1,6 @@
 name: Deploy documentation
 
 on:
-  workflow_run:
-    workflows: [CI]
-    types:
-      - completed
-    branches: [master]
   release:
     types: [published]
   workflow_dispatch:
@@ -20,15 +15,11 @@ on:
 jobs:
   deploy_github_pages:
     runs-on: ubuntu-latest
-    if: >
-      github.event_name == 'workflow_dispatch' ||
-      github.event_name == 'release' ||
-      github.event.workflow_run.conclusion == 'success'
 
     steps:
     - uses: actions/checkout@v6
       with:
-        ref: ${{ github.event_name == 'workflow_dispatch' && inputs.ref || github.event_name == 'release' && github.ref || github.event.workflow_run.head_sha }}
+        ref: ${{ github.event_name == 'workflow_dispatch' && inputs.ref || github.ref }}
 
     - name: Doxygen Action
       uses: mattnotmitt/doxygen-action@edge
@@ -45,20 +36,19 @@ jobs:
       uses: dawidd6/action-download-artifact@v6
       with:
         workflow: manifold.yml
-        workflow_conclusion: completed
-        run_id: ${{ github.event.workflow_run.id }}
-        commit: ${{ github.event_name == 'release' && github.sha || '' }}
+        workflow_conclusion: success
+        commit: ${{ github.sha }}
         check_artifacts: true
         name: wasmpublic
         path: ./public
 
-    - name: Deploy to versioned subdirectory
+    - name: Deploy to vX.Y.Z/
       uses: peaceiris/actions-gh-pages@v3
       with:
         github_token: ${{ secrets.GITHUB_TOKEN }}
         publish_branch: gh-pages
         publish_dir: ./public
-        destination_dir: ${{ github.event_name == 'workflow_dispatch' && inputs.destination_dir || github.event_name == 'release' && github.ref_name || 'master' }}
+        destination_dir: ${{ github.event_name == 'release' && github.ref_name || inputs.destination_dir }}
         keep_files: true
 
     - name: Deploy to latest/
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -1,4 +1,4 @@
-name: Deploy documentation 
+name: Deploy documentation
 
 on:
   workflow_run:
@@ -6,13 +6,20 @@ on:
     types:
       - completed
     branches: [master]
+  release:
+    types: [published]
 
 jobs:
   deploy_github_pages:
     runs-on: ubuntu-latest
+    if: >
+      github.event_name == 'release' ||
+      github.event.workflow_run.conclusion == 'success'
 
     steps:
     - uses: actions/checkout@v6
+      with:
+        ref: ${{ github.event_name == 'release' && github.ref || github.event.workflow_run.head_sha }}
 
     - name: Doxygen Action
       uses: mattnotmitt/doxygen-action@edge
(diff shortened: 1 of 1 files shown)


pca006132:
Thanks. Have you tried several dummy version in your fork?

elalish:
Not a blocker, but just curious if there's an easy way to manually run this from a different branch? So that we could say, check out 3.4 and push docs for that onto the public website? Similarly, if we found after release that we'd messed up a doc, is there a way to update that version's docs without creating a new release?

AnshulPatil2005:
> Thanks. Have you tried several dummy version in your fork?

Setting up testing on my fork now with a workflow_dispatch trigger to simulate both master and release deploys, will update once I've verified the subdirectory structure looks correct.

AnshulPatil2005:
> Not a blocker, but just curious if there's an easy way to manually run this from a different branch? So that we could say, check out 3.4 and push docs for that onto the public website? Similarly, if we found after release that we'd messed up a doc, is there a way to update that version's docs without creating a new release?

yea,
Both can be solved by adding a workflow_dispatch with two inputs:

ref -> which branch or tag to checkout (v3.4.0, v3.5.0, a hotfix branch, etc.)
destination_dir -> where to deploy it on gh-pages (v3.4.0, v3.5.0, etc.)

I'll do it

.github/workflows/deploy.yml
```diff
@@ -8,18 +8,27 @@ on:
...
     if: >
+      github.event_name == 'workflow_dispatch' ||
       github.event_name == 'release' ||
       github.event.workflow_run.conclusion == 'success'
 
     steps:
     - uses: actions/checkout@v6
       with:
-        ref: ${{ github.event_name == 'release' && github.ref || github.event.workflow_run.head_sha }}
+        ref: ${{ github.event_name == 'workflow_dispatch' && inputs.ref || github.event_name == 'release' && github.ref || github.event.workflow_run.head_sha }}
```
elalish: How does one read these `&&` and `||` expressions without parentheses? I can't tell at all which things need to be true to activate this.

elalish (commented):
Can you show us what some example docs URLs will look like now? Will pages like https://manifoldcad.org/make-manifold.html change or not? Can we add some UI to Doxygen/Typedoc to list and select between versions?

.github/workflows/deploy.yml
AnshulPatil2005: right, will add parantheses so its more clear

AnshulPatil2005:
> Can you show us what some example docs URLs will look like now? Will pages like https://manifoldcad.org/make-manifold.html change or not? Can we add some UI to Doxygen/Typedoc to list and select between versions?


On URLs: existing ones like manifoldcad.org/make-manifold.html stay unchanged. New versioned URLs would be manifoldcad.org/v3.0.0/docs/html/ and manifoldcad.org/latest/docs/html/.

On the version selector: doable via Doxygen/Typedoc header injection, good follow up change after this 

also, 
Quick design question: if we remove the master push deploy and only deploy on release, should manifoldcad.org/ stay on the last release between releases, or do you want it to keep updating with every master commit?

elalish:
>Quick design question: if we remove the master push deploy and only deploy on release, should manifoldcad.org/ stay on the last release between releases, or do you want it to keep updating with every master commit?

This is a great question, and one @pca006132 and I have occasionally discussed. There was a time when we had much more frequent updates to the UI and such of manifoldCAD, but it's pretty stable now. Perhaps it's time to put it onto a proper release cadence. I suppose that might even simplify our deployment somewhat, since then it could point to npm instead of our own wasm? 

@tonious Perhaps we should consider adding a versioned import manifoldCAD line at the beginning of any new script, to aid with compatibility? Would be even better if we could add a little Monaco squiggle under it anytime the version isn't the latest to remind people to try updating.

AnshulPatil2005:
> > Quick design question: if we remove the master push deploy and only deploy on release, should manifoldcad.org/ stay on the last release between releases, or do you want it to keep updating with every master commit?
> 
> This is a great question, and one @pca006132 and I have occasionally discussed. There was a time when we had much more frequent updates to the UI and such of manifoldCAD, but it's pretty stable now. Perhaps it's time to put it onto a proper release cadence. I suppose that might even simplify our deployment somewhat, since then it could point to npm instead of our own wasm?

Right, If we move ManifoldCAD to a release cadence, I can drop the master push deploy from this PR entirely, which simplifies things considerably.

will update once decision is made.

elalish:
Sure, let's drop the master docs push from this PR. Perhaps your next PR can be to switch manifoldCAD to a release cycle?

AnshulPatil2005:
> Sure, let's drop the master docs push from this PR. Perhaps your next PR can be to switch manifoldCAD to a release cycle?

sounds good

AnshulPatil2005:
tested 
With the current setup there will be a latest/ it's deployed on every release. So manifoldcad.org/latest/ will always point to the most recent release docs.

AnshulPatil2005:
@elalish can you re-review this, I think its good for this pr after this is merged and verified on the master branch i can move on to the switching manifoldCAD to a release cycle

elalish (approved):
Thanks!

elalish:
@AnshulPatil2005 I noticed my quad smoothing fix didn't go into ManifoldCAD - did you already stop that push here, or was that going to be a follow on? Or am I missing something?

AnshulPatil2005:
> @AnshulPatil2005 I noticed my quad smoothing fix didn't go into ManifoldCAD - did you already stop that push here, or was that going to be a follow on? Or am I missing something?

The fix will appear when the next release is cut.

But I should have flagged a gap. The deploy only writes to vX.Y.Z/ and latest/, and v3.5.2 predates this change, so neither exists yet. The root of manifoldcad.org is stuck on the last master push from Aug 17 with no way to update until a release happens.

So what should own the root between releases? Either cut releases regularly, or publish master to a master/ subdirectory and point the root at it(basically similar to what we had before)

note: dispatching master into v3.5.2/ would file August docs under a June release

elalish:
Yeah, I'm thinking we should have a master / nightly version, so it's easier to share and test changes. Think you can throw that together quickly?

PR #1788, opened 2026-08-09
Add CrossSection::EvenOdd for even-odd input fill

zmerlynn:
Adds even-odd fill back for OpenSCAD compatibility, per the discussion on #1707.

`FillRule` came out of the public API in #1760, leaving positive fill as the only way to read
input contours, which makes porting even-odd geometry hard. This adds it back as a named
constructor rather than a constructor argument, so `FillRule` stays out of the API:

```cpp
CrossSection(contours)           // unchanged: filled where the winding is > 0
CrossSection::EvenOdd(contours)  // filled where the winding is odd
```

The rule is only selectable at construction, where the input's own winding is first read.
After positive fill the storage is winding 0/1, so a later even-odd pass would be a no-op.
Boolean, Offset and Warp stay positive, and no state is added to `CrossSection` - the winding
pass emits normal-oriented boundary either way, so nothing downstream changes.

Bindings, purely additive with no ABI break: C
(`manifold_cross_section_even_odd_polygons` / `_simple_polygon`, named after the
`manifold_cross_section_hull_polygons` peers), Python (`CrossSection.even_odd`), WASM
(`CrossSection.evenOdd`).

Tests cover where the two rules diverge: nested contours, a self-intersecting pentagram, a
bowtie, a lone clockwise contour, and three-deep nesting whose signed total also pins the
stored contours to alternating orientation.

Also drops two stale `'Positive'` arguments in `bindings/wasm/bindings.js` left by #1760,
silently ignored since the constructor takes one argument.

### DIFF
--- a/bindings/c/cross.cpp
+++ b/bindings/c/cross.cpp
@@ -40,6 +40,16 @@ ManifoldCrossSection* manifold_cross_section_of_polygons(void* mem,
   return to_c(new (mem) CrossSection(*from_c(p)));
 }
 
+ManifoldCrossSection* manifold_cross_section_even_odd_simple_polygon(
+    void* mem, ManifoldSimplePolygon* p) {
+  return to_c(new (mem) CrossSection(CrossSection::EvenOdd(*from_c(p))));
+}
+
+ManifoldCrossSection* manifold_cross_section_even_odd_polygons(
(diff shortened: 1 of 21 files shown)


elalish (approved):
Thanks!

Issue #1706, opened 2026-05-12
`Boolean` emits sliver tris on near-coincident inputs


zmerlynn:
### Summary

I've been seeing what look like near-degenerate sliver triangles in Boolean output when inputs have nearly-coincident surfaces (twin geometry, kit-bashed CSG, offset compositions), and I'd appreciate input on whether this is something the library wants to deal with or whether it falls under "input mesh is the caller's responsibility."

The triangles have three vertices all clustered within FP-arithmetic-noise distance, small area, all three edges short (`~1e-6` on a 25400-unit mesh) but no individual edge below `epsilon_`. `Status() == NoError` and topology stays manifold, so by the standard contract everything looks fine.

### Reproduction

```cpp
Manifold m1 = Manifold::ReadOBJ(/* test/models/Generic_Twin_7081.1.t0_left.obj */);
Manifold m2 = Manifold::ReadOBJ(/* test/models/Generic_Twin_7081.1.t0_right.obj */);
Manifold res = m1 + m2;
// 33230 tris; 615 with area below 1e-9; 21 with area below 1e-13
```

### Where they come from

`Kernel12` (`boolean3.cpp`) computes a new intersection vertex for each edge of P that pierces a face of Q. When inputs have nearly-coincident faces, multiple piercings in the same region produce multiple `v12` vertices within FP-arithmetic-noise distance of each other. `AddNewEdgeVerts` (`boolean_result.cpp`) distributes each new vertex into three `EdgePos` lists (the subdivided P-edge plus the two new face-face chord edges). The downstream `AppendPartialEdges` / `AppendNewEdges` then assemble these into halfedges and triangles without coalescing the near-duplicates, so the clusters are emitted as the three corners of the resulting sliver tris.

### Why existing cleanups don't seem to catch them

From reading the predicates: `ShortEdge` keys on `edge_length < epsilon_` and sliver edges seem to exceed `epsilon_`; `FlagEdge` keys on face-ref redundancy and each `v12` appears to carry a distinct ref; `SwappableEdge` requires `Is01Longest` which near-equilateral clusters seem to fail. I tried `SetTolerance` / `Simplify` at a range of values and they didn't clear the slivers (at larger values they destroyed legitimate mesh structure).

### Why I'm asking

[...]

PR #1790, opened 2026-08-11
Optional ID tracking and no degenerates

### LINKED ISSUE #1217: opt-in mesh id for primitives

### LINKED ISSUE #1280: Improve performance of MarkCoplanar
It is taking around 1/3 to 1/2 of total time when importing `MeshGL`. We should probably try to make it faster.

I checked it a bit. It seems that this is due to access being too random, so it is hard to optimize the sequential performance.
There are a few things we can do:

1. Optimize sorting, by making the `area` field single precision. I tried the second one and this reduced the time by about 5-10%.
2. Parallelize `MarkCoplanar`, maybe somehow do atomic CAS to make `coplanarID` the coplanar triangle with the greatest area.

2 may result in coplanar faces not being connected (if the middle is interrupted by another run of coplanar triangle that has a larger initial size, say), but we can use fix that using our union-find algorithm which is fast and parallel. For the connected components that are isolated from the original coplanar triangle, we still have a guarantee that the faces in that connected component are coplanar, so we can pick a random triangle inside and call it a day?

Another issue is determinism. I guess for this to be deterministic, we cannot stop early with `coplanarID != -1`.

elalish:
Fixes #1217
Fixes #1280 - by removing our coplanarity concept entirely.
Helps with #1706, but doesn't entirely fix it. I'll take a deeper look in a follow-on.

Redo of #1252. Turns out making the ID tracking optional is the easy part, but making it do something useful, e.g. collapse more coplanar triangles, is tricky. This turned into a complete overhaul of our degenerate triangle removal code. 

We used to have a single code path in charge of two things: removing the degenerate triangles that the Boolean created, and also general mesh decimation to a tolerance. Recent work gave us a new path for decimation that has dramatically better results, but is a bit slower, and hence was not appropriate for running after every Boolean op. Now I've here refactored the original degenerate removal step to be simpler and much more robust. 

The upshot: I can now guarantee Manifold produces output with zero degenerate triangles! `RemoveDegenerates` is now exposed as a public API in the spirit of doing less expensive work automatically during mesh import. So, since Booleans only remove degenerates that they produce, to have completely degenerate-free results, simply run `RemoveDegenerates` on any imported meshes from sketchy sources. 

Here's an example of how much it helps with a long string of ops to not track IDs you don't care about and allow all the coplanar stuff to collapse.
Before:
image

After:
image


test/boolean_test.cpp
```diff
@@ -525,6 +533,14 @@ TEST(Boolean, Perturb3) {
   }
 }
 
+TEST(Boolean, Perturb4) {
+  Manifold a = ReadTestOBJ("perturb3A.obj");
+  Manifold b = ReadTestOBJ("perturb3B.obj");
+  Manifold result = a + b;
```
elalish: @pca006132 This is really weird. When I run Boolean.Perturb3 on this branch, it reliably fails with a triangulation not CCW error and spits out the OBJ inputs. However, when I made this test to reproduce with those two files directly, it passes just fine. 

Okay, I just found the `MANIFOLD_OBJ_HEX_FLOAT` env variable, so now I think I've taken care of any rounding problems. However now I can't read the files at all.

test/test_main.cpp
```diff
@@ -531,8 +527,9 @@ MeshGL64 ReadTestMeshGL64OBJ(const std::string& filename) {
 #endif
   std::ifstream f;
   f.open(obj);
-  MeshGL64 a = ReadOBJ(f);
+  Manifold a = Manifold::ReadOBJ(f);
   f.close();
+  EXPECT_FALSE(a.IsEmpty());
```
elalish, line 533: I switched this to read more directly, but now it silently fails to read these hexfloat files at all (hence my added expect). It looks like the regex's aren't set up to read hex? Or did I miss something?

elalish: Actually, @AnshulPatil2005, it looks like you touched this code recently? Any of this sound familiar?

AnshulPatil2005: Yea, I did touch this code during  #1554 ("Fixes different verbose levels"),

looks like The FLOAT_PATTERN regex doesn't handle hexfloat format, so those vertex lines silently get skipped leaving an empty mesh. Both functions share the same parser so both are affected EXPECT_FALSE just exposed it.

currently it is 
`static const std::string FLOAT_PATTERN =
    "(-?\\d+(?:\\.\\d*)?(?:[eE][+\\-]?\\d+)?)";
`

should be something like afaik

`static const std::string FLOAT_PATTERN =
    "(-?(?:0[xX][0-9a-fA-F]+(?:\\.[0-9a-fA-F]*)?[pP][+\\-]?\\d+|\\d+(?:\\.\\d*)?(?:[eE][+\\-]?\\d+)?))";
`

src/boolean_result.cpp
```diff
@@ -950,16 +948,17 @@ Manifold::Impl Boolean3::Result(OpType op) const {
...
     DEBUG_ASSERT(outR.Is2Manifold(), logicErr,
                  "simplified mesh is not 2-manifold!");
-
-  // if (ManifoldParams().intermediateChecks)
-  //   DEBUG_ASSERT(outR.NumDegenerateTris() == 0, logicErr,
-  //                "simplified mesh has degenerate triangles!");
+  }
+  if (ManifoldParams().verifyNoDegenerates) {
+    DEBUG_ASSERT(outR.NumDegenerateTris() == 0, logicErr,
+                 "simplified mesh has degenerate triangles!");
```
elalish, line 955: This is a check I've wanted to add for years!

src/csg_tree.cpp
```diff
@@ -177,7 +177,7 @@ std::shared_ptr<CsgLeafNode> SimpleBoolean(const Manifold::Impl& a,
               << std::endl;
     std::cout << "RHS self-intersecting: " << b.IsSelfIntersecting()
               << std::endl;
-    if (ManifoldParams().verbose) {
+    if (ManifoldParams().verbose > 3) {
```
elalish, line 180: I switched this because it was insisting on dumping OBJs to the command line all the time when I just wanted to debug.

src/impl.cpp
```diff
@@ -822,7 +779,8 @@ static std::ostream& WriteOBJWithEpsilon(std::ostream& stream,
 static std::pair<MeshGL64, std::optional<double>> ReadOBJWithEpsilon(
     std::istream& stream) {
   static const std::string FLOAT_PATTERN =
-      "(-?\\d+(?:\\.\\d*)?(?:[eE][+\\-]?\\d+)?)";
+      "(-?(?:0[xX][0-9a-fA-F]+(?:\\.[0-9a-fA-F]*)?[pP][+\\-]?\\d+|\\d+(?:\\."
+      "\\d*)?(?:[eE][+\\-]?\\d+)?))";
```
elalish, line 783: This fix allows us to read back in our hex-float OBJ format.

src/manifold.cpp
```diff
@@ -448,14 +464,22 @@ int Manifold::OriginalID() const {
  * - these don't get joined at boundaries where originalID changes, so the
  * reset may allow triangles of flat faces to be further collapsed with
  * Simplify().
+ *
+ * @param id The ID to assign to this manifold. If negative (the default), a new
+ * ID is assigned. Use zero to match all default-constructed manifolds, thus not
+ * keeping track of the joints between input manifolds. Ensure separate IDs are
+ * used for mesh inputs containing properties, generally by calling this
+ * function without arguments just after construction.
```
elalish: I need to check this - I think it should be checking properties separately so calling AsOriginal isn't necessary for property meshes.

src/properties.cpp
```diff
@@ -230,7 +230,7 @@ int Manifold::Impl::NumDegenerateTris() const {
     for (int i : {0, 1, 2})
       v[i] = projection * vertPos_[halfedge_.Start(3 * face + i)];
 
-    const int ccw = CCW(v[0], v[1], v[2], tolerance_ / 2);
+    const int ccw = CCW(v[0], v[1], v[2], epsilon_ / 2);
```
elalish, line 233: I also need to think about epsilon vs tolerance. I think we might not need epsilon anymore? I believe we're much more robust to error stacking now, so tolerance may suffice. Then again, I wonder if tolerance itself should be simplified?

test/boolean_test.cpp
```diff
@@ -175,10 +176,7 @@ TEST(Boolean, Cubes2) {
 
   Manifold result = cube + cube.Rotate(0, 0, 45);
 
-  // has 14 verts instead of 12 because of symbolic perturbation making a jagged
-  // intersection, which is maintained to keep meshIDs separate. I don't love
-  // either of those behaviors by default...
-  ExpectMeshes(result, {{14, 24}});
+  ExpectMeshes(result, {{12, 20}});
```
elalish, line 179: Here's an example of the improvement - beyond just degenerate removal, it's also removing coplanar verts more effectively.

test/manifold_test.cpp
```diff
@@ -1623,18 +1622,6 @@ TEST(Manifold, MergeRefine) {
   EXPECT_NEAR(manifold.Volume(), 31.21, 0.01);
 }
 
-#ifdef MANIFOLD_DEBUG
-TEST(Manifold, OpenscadCrash) {
```
elalish, line 1627: Moved this test because it's a boolean.

test/properties_test.cpp
```diff
@@ -177,16 +177,6 @@ TEST(Properties, CalculateNormals) {
   }
   EXPECT_EQ(numBad, 0);
   EXPECT_EQ(numBad2, 0);
-  for (int tv = out2.runIndex[0]; tv < out2.runIndex[1]; ++tv) {
```
elalish, line 180: This is duplicate with the loop above.

test/samples_test.cpp
```diff
@@ -205,9 +197,6 @@ TEST(Samples, Sponge4) {
   EXPECT_EQ(rect.max.y, box.max.y);
   EXPECT_NEAR(projection.Area(), 0.535, 0.001);
   Manifold extrusion = Manifold::Extrude(projection.ToPolygons(), 1);
-  // This dense extrusion keeps a few sub-tolerance sliver tris; Genus (below)
```
elalish, line 208: Not anymore!

src/edge_op.cpp
```diff
@@ -139,24 +69,117 @@ void Manifold::Impl::CleanupTopology() {
...
+
+  int current = start;
+  while (current != tri0edge[2]) {
+    current = NextHalfedge(current);
+    vec3 pNext = vertPos_[halfedge_.End(current)];
+    const int tri = current / 3;
+
+    // TODO: use faceNormal offsets instead to avoid error stacking.
+    if (la::abs(la::dot(pNew - pOld, faceNormal_[tri])) > epsilon_)
+      return false;
```
elalish, line 168: Not sure error stacking can actually happen here...

src/edge_op.cpp
```diff
@@ -228,6 +251,19 @@ bool Manifold::Impl::Continuous(int edge) const {
...
+Manifold::Impl::TriResult Manifold::Impl::IsDegenerate(int tri) const {
+  const mat3 v = {vertPos_[halfedge_.Start(3 * tri)],
+                  vertPos_[halfedge_.Start(3 * tri + 1)],
+                  vertPos_[halfedge_.Start(3 * tri + 2)]};
+  const mat3 edgeVec = {v[0] - v[1], v[1] - v[2], v[2] - v[0]};
+  const vec3 edgeLen2 = {length2(edgeVec[0]), length2(edgeVec[1]),
+                         length2(edgeVec[2])};
+  const int edge = argmax(edgeLen2);
+  return {length2(cross(edgeVec[edge], edgeVec[Next3(edge)])) <=
+              edgeLen2[edge] * epsilon_ * epsilon_,
```
elalish, line 263: Switched from projected CCW to a 3D check.

PR #1791, opened 2026-08-14
Confine the atomic reinterpret_cast to one shim, use std::atomic_ref at C++20

zmerlynn:
`reinterpret_cast<std::atomic<T>&>` on a plain `T&` (#1153) is UB, and the sanctioned
`std::atomic_ref` is C++20 only. This doesn't remove the UB at C++17 - it moves the cast from nine
call sites (one of them dead) into one type whose preconditions (`sizeof`/`alignof` match,
`is_always_lock_free`) are static_asserts, and replaces it outright at C++20.

Also: `CMAKE_CXX_STANDARD` is now overridable (a plain `set()` shadowed both `-D` and parent
projects), and a new `GCC 14 (TBB:ON, C++20)` lane builds everything at C++20 under `-Werror`. It
immediately caught the deprecated shared_ptr `atomic_load`/`atomic_store` in `manifold.cpp`, now
behind a scoped pragma.

Suite green at C++17 and at C++20 under `-Werror`.

### DIFF
--- a/.github/workflows/manifold.yml
+++ b/.github/workflows/manifold.yml
@@ -91,24 +91,41 @@ jobs:
         name: meshes-windows
         path: meshes-windows/
   build:
-    name: GCC ${{matrix.gcc}} (TBB:${{matrix.parallelization}}${{matrix.std == 20 && ', C++20' || ''}})
+    name: ${{matrix.name}}
     timeout-minutes: 45
     strategy:
-      fail-fast: false
       matrix:
-        parallelization: [OFF, ON]
-        os: [ubuntu-24.04]
-        gcc: [13, 14]
-        std: [17]
         include:
-          - parallelization: ON
-            os: ubuntu-22.04
+          - name: GCC 11 (TBB:ON)
             gcc: 11
+            parallelization: ON
+            os: ubuntu-22.04
+            std: 17
+          - name: GCC 13 (TBB:OFF)
+            gcc: 13
+            parallelization: OFF
+            os: ubuntu-24.04
+            std: 17
+          - name: GCC 13 (TBB:ON)
+            gcc: 13
+            parallelization: ON
+            os: ubuntu-24.04
+            std: 17
+          - name: GCC 14 (TBB:OFF)
+            gcc: 14
+            parallelization: OFF
+            os: ubuntu-24.04
+            std: 17
+          - name: GCC 14 (TBB:ON)
+            gcc: 14
+            parallelization: ON
+            os: ubuntu-24.04
             std: 17
           # One strict lane at C++20, to catch what it deprecates or redefines.
-          - parallelization: ON
-            os: ubuntu-24.04
+          - name: GCC 14 (TBB:ON, C++20)
             gcc: 14
+            parallelization: ON
+            os: ubuntu-24.04
             std: 20
     runs-on: ${{ matrix.os }}
     env:
--- a/.github/workflows/manifold.yml
+++ b/.github/workflows/manifold.yml
@@ -91,17 +91,27 @@ jobs:
         name: meshes-windows
         path: meshes-windows/
   build:
-    name: GCC ${{matrix.gcc}} (TBB:${{matrix.parallelization}})
+    name: GCC ${{matrix.gcc}} (TBB:${{matrix.parallelization}}${{matrix.std == 20 && ', C++20' || ''}})
     timeout-minutes: 45
     strategy:
+      fail-fast: false
       matrix:
         parallelization: [OFF, ON]
         os: [ubuntu-24.04]
         gcc: [13, 14]
+        std: [17]
         include:
           - parallelization: ON
             os: ubuntu-22.04
             gcc: 11
+            std: 17
+          # One strict lane at C++20. Deprecations are silenced for now: the
+          # shared_ptr atomic free functions in manifold.cpp are deprecated
+          # there and removed in C++26. Drop the flag once those are replaced.
+          - parallelization: ON
+            os: ubuntu-24.04
+            gcc: 14
+            std: 20
     runs-on: ${{ matrix.os }}
     env:
       CC: gcc-${{ matrix.gcc }}
@@ -124,6 +134,8 @@ jobs:
       run: |
         cmake \
           -DCMAKE_BUILD_TYPE=RelWithDebInfo \
+          -DCMAKE_CXX_STANDARD=${{matrix.std}} \
+          -DMANIFOLD_FLAGS=${{matrix.std == 20 && '-Wno-deprecated-declarations' || ''}} \
           -DBUILD_SHARED_LIBS=ON \
           -DMANIFOLD_STRICT=ON \
           -DMANIFOLD_PYBIND=ON \
--- a/.github/workflows/manifold.yml
+++ b/.github/workflows/manifold.yml
@@ -105,9 +105,7 @@ jobs:
             os: ubuntu-22.04
             gcc: 11
             std: 17
-          # One strict lane at C++20. Deprecations are silenced for now: the
-          # shared_ptr atomic free functions in manifold.cpp are deprecated
-          # there and removed in C++26. Drop the flag once those are replaced.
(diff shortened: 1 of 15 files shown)


.github/workflows/manifold.yml
```diff
@@ -91,17 +91,25 @@ jobs:
         name: meshes-windows
         path: meshes-windows/
   build:
-    name: GCC ${{matrix.gcc}} (TBB:${{matrix.parallelization}})
+    name: GCC ${{matrix.gcc}} (TBB:${{matrix.parallelization}}${{matrix.std == 20 && ', C++20' || ''}})
     timeout-minutes: 45
     strategy:
+      fail-fast: false
```
pca006132: needed?

.github/workflows/manifold.yml
```diff
@@ -91,17 +91,25 @@ jobs:
         name: meshes-windows
         path: meshes-windows/
   build:
-    name: GCC ${{matrix.gcc}} (TBB:${{matrix.parallelization}})
+    name: GCC ${{matrix.gcc}} (TBB:${{matrix.parallelization}}${{matrix.std == 20 && ', C++20' || ''}})
```
pca006132: GitHub now has a `case` expression for ternary: https://docs.github.com/en/actions/reference/workflows-and-actions/expressions#case

.github/workflows/manifold.yml
```diff
@@ -91,17 +91,25 @@ jobs:
...
     timeout-minutes: 45
     strategy:
+      fail-fast: false
       matrix:
         parallelization: [OFF, ON]
         os: [ubuntu-24.04]
         gcc: [13, 14]
+        std: [17]
         include:
```
pca006132, line 98: I am thinking if it makes more sense if we just list the job configurations instead of relying on matrix. It is a bit more lines, but it also gives us much better idea about how many jobs we have.

src/atomic_compat.h
```diff
@@ -0,0 +1,449 @@
...
+template <typename T>
+using AtomicRef = std::atomic_ref<T>;
+
+#else  // C++17: everything below stands in for it
+
+namespace details {
+
+#if defined(_MSC_VER) && !defined(__clang__)  // MSVC backend
+
+// 32-bit MSVC keeps the pre-#1153 behavior. The intrinsics below are not all
```
pca006132: I wonder if we can just say that we do not support 32-bit MSVC... Do people still use that?
And honestly, I think we should just fail the compilation if the target does not support 64-bit atomics that we need, that is never correct anyway.

pca006132 (commented):
Thanks a lot for the PR, but my feeling around this is still like, this is mostly ceremony around something that C++20 should fix, so we probably shouldn't introduce that much code to handle these if the current, technically UB, reinterpret_cast works.

I do like the C++20 stuff though. I just mean maybe we can make the wrapper fallback to the technically UB reinterpret cast?

zmerlynn:
Fair - it is ceremony. The backends are deliberately just what libstdc++/libc++/MSVC STL
do, and the file is written to be deleted: at C++20 `AtomicRef` is
`using AtomicRef = std::atomic_ref<T>`.

I did check the cost. Built against master with identical flags, it emits the same atomics
(two fewer noop pads) and `libmanifold.so` comes out the same size, so it is
free at runtime. The real cost is the ~200 lines you have to read, which is fair not to
want.

That said, I'd still be inclined to remove the UB here despite the imitation.

.github/workflows/manifold.yml
zmerlynn: Dropped - that was just for debugging while I iterated on the new lane.

.github/workflows/manifold.yml
zmerlynn: (obviated by other change)

.github/workflows/manifold.yml
zmerlynn: Done - six entries, six jobs. Names are unchanged.

src/atomic_compat.h
zmerlynn: Done, it's an `#error` now. That drops 32-bit x86 and ARMv7 only - ARM64 and ARM64EC are 64-bit and unaffected.

starseeker:
@zmerlynn Not sure if this even makes sense to ask in the first place, but since you're looking at AtomicRef, would https://github.com/vtyulb/AtomicSharedPtr be at all interesting?  I have no idea if that hack they use would even be viable for our use case, or if it would offer any benefit - do you know anything about it?

zmerlynn:
Thanks, I read through it - it's the folly packed-pointer hack, refcount in the spare 16 bits.

I don't think it fits us. It isn't a `std::shared_ptr` but its own type with its own control block, and our `ctx_` shares ownership with `ExecutionContext::impl_`, a `std::shared_ptr` in a public header, so adopting it would put a third-party type on our public API and change ABI. It also leans on the top 16 bits of a pointer being dead, which isn't architectural, and the author says it isn't production-ready.

Mostly though there's little to win: our atomic access costs about 12 ns over a plain refcount bump, and `ctx_` is only touched at the entry of whole-mesh ops and in copy/assign, which already take a mutex.

pca006132:
For the atomic shared pointer thing, I don't think we use shared pointers in a way that impacts performance in the library, so we shouldn't need that.

For the atomic compat, I am fine with the GCC/Clang stuff, they are standard. My problem is really about the MSVC part: Saying that MSVC doesn't do TBAA and doing reinterpret casts (which is technically still UB, just that it is fine with that compiler), passing things as volatile (is the volatile doing anything, or just preventing the compiler from reordering stuff?), and realistically I don't want to check if all comments hold there...

pca006132:
Or does boost provide a small library for `AtomicRef`? If it is small and the license is compatible, maybe we can just copy it in? We can maybe trim it down if the library is large.

zmerlynn:
Backed off entirely, including GCC/Clang - no hand-written backends left. C++20 uses `std::atomic_ref`; C++17 does the same cast master already does, in one place instead of nine, with the preconditions as static_asserts (`sizeof`/`alignof`, `is_always_lock_free`). 425 lines down to 165. The 32-bit MSVC `#error` goes with it, so that restriction is lifted.

On Boost: `atomic_ref.hpp` is 2.8 KB but it's a facade over a `detail/` tree of 120 files and 1.2 MB. Not worth vendoring for this.

test/utils_test.cpp
```diff
@@ -0,0 +1,224 @@
+// Copyright 2026 The Manifold Authors.
```
pca006132, line 1: For these tests, I am not sure if we want to keep them, since I am pretty sure these are not sufficient to test atomic stuff since there is not enough work to make the threads run together, the OS very likely is launching and immediately finishing the thread. And since the whole thing is now just a wrapper, I think we can be quite confident that it should be doing the right thing.

pca006132 (approved):

pca006132:
> Backed off entirely, including GCC/Clang - no hand-written backends left. C++20 uses `std::atomic_ref`; C++17 does the same cast master already does, in one place instead of nine, with the preconditions as static_asserts (`sizeof`/`alignof`, `is_always_lock_free`). 425 lines down to 165. The 32-bit MSVC `#error` goes with it, so that restriction is lifted.

Thanks.
 
> On Boost: `atomic_ref.hpp` is 2.8 KB but it's a facade over a `detail/` tree of 120 files and 1.2 MB. Not worth vendoring for this.

I am quite surprised that it is this large, but boost probably has to support a lot of platforms and probably contains implementation using locks, so that is kind of understandable.

test/utils_test.cpp
zmerlynn: You're right - I checked, and a deliberately non-atomic version passes the same test, so it couldn't have caught a broken implementation. Dropped those four.

Kept the ones that test the wrapper rather than `std::atomic`: the per-width round-trips instantiate the static_asserts, and the const one guards a C++17/C++20 divergence.

elalish (approved):
Thanks!

PR #1801, opened 2026-08-21
deploy of master to master/

AnshulPatil2005:
Adds a nightly cron that publishes master to `master/`, so unreleased
changes are shareable and testable between releases. Follow-up to #1787

- `master/`  erased and rebuilt nightly
- `vX.Y.Z/`  created per release, never touched again
- `latest/`  erased and rebuilt on each release
- other version directories are never affected

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -3,6 +3,8 @@ name: Deploy documentation
 on:
   release:
     types: [published]
+  schedule:
+    - cron: '0 6 * * *'
   workflow_dispatch:
     inputs:
       ref:
@@ -19,7 +21,7 @@ jobs:
     steps:
     - uses: actions/checkout@v6
       with:
-        ref: ${{ (github.event_name == 'workflow_dispatch' && inputs.ref) || github.ref }}
+        ref: ${{ (github.event_name == 'workflow_dispatch' && inputs.ref) || (github.event_name == 'schedule' && 'master') || github.ref }}
 
     - name: Doxygen Action
       uses: mattnotmitt/doxygen-action@edge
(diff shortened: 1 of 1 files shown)


pca006132:
Have you tried it locally? Is it ready to be merged?

pca006132:
Thinking about it, why should this be a cron job? The documentation should only be updated when there is a merge to master, no?

AnshulPatil2005:
> Have you tried it locally? Is it ready to be merged?

not yet, I'll do it rn

AnshulPatil2005:
> Thinking about it, why should this be a cron job? The documentation should only be updated when there is a merge to master, no?

Actually yes, cron is wrong. It'd fire more often than commits do.

also it is also publishing the app with docs and the reason for master removal was actually this discussion (https://github.com/elalish/manifold/pull/1787#issuecomment-5264566164) but since both docs and app ship under the one trigger it did both at once.

Switching to workflow_run on master CI.

AnshulPatil2005:
> > Have you tried it locally? Is it ready to be merged?
> 
> not yet, I'll do it rn

tested, good to merge imo

elalish (approved):
Thanks! Let's try it out.

elalish:
Oh sorry, thought I already hit merge on this.

elalish:
Okay, either something isn't working or I'm missing something. When I do `npm run dev` on locally on master, and I paste this script in, I get a correct result (smooth fillet):
```
import { Manifold } from 'manifold-3d/manifoldCAD';

const depth = 3
const cylinder = Manifold.cylinder(40, 10, 10, 6, true).rotate(0, 30).calculateNormals(0, 80);
const section = cylinder.slice(0).simplify();
const chamfer = section.extrude(depth, 0, 0, [1.3, 1.2]).mirror([0, 0, 1]);
const base = Manifold.cube(40, true).translate([0, 0, -20 - depth]).calculateNormals(0);
const chamfered = cylinder.add(chamfer).add(base);

const test = chamfered.smoothByNormals(0).refineToTolerance(0.01);

export default test.translate([0, 0, depth]);
```
image
However, when I run the same script on manifoldcad.org, I get the wrong (old) result:
image

I've also tried it on manifoldcad.org/master - same result. I tried manifoldcad.org/latest and got a 404. I'm also curious, how does all this play with the service worker? I tried updating that to no avail, but how will it know the proper version to use when there are several deployed?

AnshulPatil2005:
Just looked into this The fix is deployed but unreachable.

/master/assets/manifold.wasm is 540,892 bytes (Aug 23, with your change) while /assets/manifold.wasm is 539,138 (Aug 15). Different files. But /master/index.html contains src="/assets/manifoldCAD.js" an absolute path so it loads the root bundle and the root's wasm. That's why / and /master/ give identical results.

Cause is that vite has no base configured, so it defaults to /. That was invisible while every deploy went to the site root, and #1787 exposed it. 

/latest/ 404s because no release has been done yet

On the service worker, you've found a second version of the same problem. Registration is fine, it uses new URL('./service-worker.js', ...) with scope: './', so at /master/ it correctly scopes to /master/. But inside, appShellAssets is a list of absolute paths ('/index.html', '/editor.css', ...) and cacheName is the single global 'manifoldCAD-cache-v4'. So every deployment caches the root's files into one shared cache. With several deployed they collide.

Three things to fix: set base: './' in the vite configs for the editor and the three examples, make appShellAssets relative, and derive cacheName per deployment (self.registration.scope would do it). I'll put up a PR.

PR #1805, opened 2026-08-23
Build native Windows ARM64 wheels

ssss141414:
## Summary

- add `windows-11-arm` to the existing cibuildwheel runner matrix
- rely on cibuildwheel's native architecture detection to build `win_arm64` wheels
- make no source or runtime behavior changes

## Validation

- verified cibuildwheel 3.4.0 selects `cp39`, `cp310`, `cp311`, `cp312`, `cp313`, `cp314`, and `cp314t` `win_arm64` builds on a native ARM64 host
- locally built and clean-installed a CPython 3.13 `win_arm64` wheel
- verified the extension PE machine is `AA64`
- the existing `run_all.py` extrusion test still fails on native ARM64; that runtime issue is intentionally out of scope for this one-line CI change
- hosted `build_wheels` execution is pending manual dispatch or a release event

### DIFF
--- a/.github/workflows/build_wheels.yml
+++ b/.github/workflows/build_wheels.yml
@@ -20,11 +20,6 @@ jobs:
           git config --global core.longpaths true
       - if: matrix.os == 'macos-15'
         run: echo "CIBW_ARCHS=arm64 universal2" >> "$GITHUB_ENV"
-      - if: matrix.os == 'windows-11-arm'
-        shell: pwsh
-        run: |
-          "CIBW_ARCHS=ARM64" >> $env:GITHUB_ENV
-          "CIBW_BUILD=cp311-* cp312-* cp313-* cp314-*" >> $env:GITHUB_ENV
       - uses: actions/checkout@v6
       - name: Build wheels
         uses: pypa/cibuildwheel@v3.4.0
--- a/.github/workflows/build_wheels.yml
+++ b/.github/workflows/build_wheels.yml
@@ -21,7 +21,7 @@ jobs:
       - if: matrix.os == 'macos-15'
         run: echo "CIBW_ARCHS=arm64 universal2" >> "$GITHUB_ENV"
       - if: matrix.os == 'windows-11-arm'
-        run: echo "CIBW_BUILD=cp311-* cp312-* cp313-* cp314-*" >> "$GITHUB_ENV"
+        run: echo "CIBW_BUILD=cp311-* cp312-* cp313-* cp314-*" >> $env:GITHUB_ENV
       - uses: actions/checkout@v6
       - name: Build wheels
         uses: pypa/cibuildwheel@v3.4.0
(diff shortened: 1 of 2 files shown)


elalish:
I take it this is fixing a bug in the Python bindings for Windows?

ssss141414:
> I take it this is fixing a bug in the Python bindings for Windows?

Hi elaish,

This PR adds native Windows ARM64 wheel builds and publishing support.

The **common.h** fixes a core math issue exposed by optimized MSVC ARM64 builds. The wheel builds and installs successfully without this change, but the existing cibuildwheel test command fails while running bindings/python/examples/run_all.py.

In particular, the **extrude.py** example expects a unit square extruded by 10 to have volume 10, but the Windows ARM64 build returns 8.333333333333334. This was traced to inconsistent quotient bits returned by std::remquo in sind(), which produces an incorrect rotation matrix.

Therefore, without this fix, the new Windows ARM64 cibuildwheel job cannot pass its existing test suite and the release workflow cannot complete successfully.

Thank you~

elalish:
Wow, that's a pretty serious compiler error! Is that a known issue for MSVC? This is still marked as draft - when will it be ready for review?

pca006132:
I think it is much simpler to just open another PR with the build matrix change and see what it goes. The other changes to the github action doesn't seem needed to me.

ssss141414:
> I think it is much simpler to just open another PR with the build matrix change and see what it goes. The other changes to the github action doesn't seem needed to me.

Sure. Let me only keep ARM64 publish related changes.

Please help review, thank you so much!

elalish:
I'm confused, now the PR doesn't change the remquo thing. Where's the evidence that's a bug, and what are we fixing here?

pca006132:
Kind of expecting this but I am also confused. AI stuff?

pca006132:
Anyway, I think this should just be a one line change, the writing to environment variable stuff is probably not needed as well.

ssss141414:
> I'm confused, now the PR doesn't change the remquo thing. Where's the evidence that's a bug, and what are we fixing here?

I just take pca006132's advice, revert bug related changes. The fix is in another PR: https://github.com/elalish/manifold/pull/1809

Current PR just focuses on ARM64 publish.

ssss141414:
> Anyway, I think this should just be a one line change, the writing to environment variable stuff is probably not needed as well.

Thank you so much for your comments. I tested the one-line runner change in my fork. Without a build filter, cibuildwheel also selects `cp39-win_arm64`, whose NumPy dependency has no compatible wheel and fails while building from source ([run](https://github.com/ssss141414/manifold/actions/runs/32947685786)).

The extra `CIBW_BUILD` line limits this new runner to CPython 3.11–3.14, where Windows ARM64 dependencies are available. With that filter, cibuildwheel successfully built, repaired, and installed the `cp311-win_arm64` wheel before reaching an existing runtime test failure ([run](https://github.com/ssss141414/manifold/actions/runs/32949925182)).

elalish (approved):
Thanks!

ssss141414:
Hi @elalish ,
May I ask the release plan of this package? So we can use ARM64 version... Thank you!

elalish:
Good question, we're in the middle of some churn at the moment, coming up towards v4.0, but this is small enough that we should probably just throw it on a patch for 3.5.3. @AnshulPatil2005, you were just getting our process set up to update old releases - this might be a good test.

PR #1806, opened 2026-08-24
Fix asset paths for subdirectory deploys

AnshulPatil2005:
https://github.com/elalish/manifold/pull/1801#issuecomment-5396695846

### DIFF
--- a/bindings/wasm/examples/editor/index.html
+++ b/bindings/wasm/examples/editor/index.html
@@ -14,8 +14,8 @@
   <meta property="og:image" content="https://elalish.github.io/manifold/samples/models/mengerSpongeSquare.png">
   <meta property="og:image:width" content="850">
   <meta property="og:image:height" content="850">
-  <link rel="shortcut icon" type="image/png" href="/icons/ManifoldIcon.png">
-  <link rel="manifest" href="/manifest.json">
+  <link rel="shortcut icon" type="image/png" href="./icons/ManifoldIcon.png">
+  <link rel="manifest" href="./manifest.json">
   <link type="text/css" href="editor.css" rel="stylesheet">
   <script type="application/ld+json">
     {
--- a/bindings/wasm/examples/editor/public/service-worker.js
+++ b/bindings/wasm/examples/editor/public/service-worker.js
@@ -12,33 +12,36 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+// The site is deployed to several paths at once (master/, latest/, vX.Y.Z/),
+// each registering this worker under its own scope. Key the cache by scope so
+// deployments do not share or evict each other's entries.
+const cachePrefix = `manifoldCAD-cache:${self.registration.scope}:`;
 // Increment version when updating CDN URLs to clean up cache.
-const cacheName = 'manifoldCAD-cache-v4';
+const cacheName = `${cachePrefix}v5`;
 
+// Relative to this worker's location, so they resolve within the deployment
+// this worker is scoped to rather than always hitting the site root.
 const appShellAssets = [
-  '/',
-  '/index.html',
-  '/editor.css',
-  '/editor.js',
-  '/editor-examples.js',
-  '/manifest.json',
-  '/fonts/orbitron-black-webfont.ttf',
-  '/icons/close.png',
-  '/icons/docs.png',
-  '/icons/manifoldCAD.png',
-  '/icons/manifoldCADonly.png',
-  '/icons/ManifoldIcon.png',
-  '/icons/mengerSponge192.png',
-  '/icons/mengerSponge512.png',
-  '/icons/mengerSponge64.png',
-  '/icons/pause.png',
-  '/icons/pencil.png',
-  '/icons/play.png',
-  '/icons/redo.png',
-  '/icons/share.png',
-  '/icons/star.png',
-  '/icons/trash.png',
-  '/icons/undo.png'
+  './',
+  './index.html',
+  './manifest.json',
+  './fonts/orbitron-black-webfont.ttf',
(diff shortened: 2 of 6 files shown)


AnshulPatil2005:
@elalish This should do it 
Sets `base: './'` in the editor and the three example configs so one
build works from any path, and makes the two hand-written absolute refs
in index.html relative.

Service worker had the same problem in two forms:

- `cacheName` was a single global string, so every deployment shared one
  cache, and `activate` deleted every cache whose key differed — meaning
  one deployment would evict another's. Now keyed by
  `self.registration.scope`, with eviction scoped to its own prefix.
- `appShellAssets` used absolute paths, so every deployment precached the
  root's files. Now relative.

AnshulPatil2005:
The service worker has a related problem that hasn't surfaced yet, so I fixed it pre-emptively. It's network-first, so it wasn't serving the stale content here, but with several versions deployed it would share one cache across all of them and evict other deployments' entries on activate.

I'll check whether more of these are lurking better to find them before the first release starts creating version directories.

elalish (approved):
Thanks! Please take a look once it's deployed and check that it's behaving as expected. I find these things are hard to check in a test environment.

AnshulPatil2005:
> Thanks! Please take a look once it's deployed and check that it's behaving as expected. I find these things are hard to check in a test environment.

Yep, I will actively check everything this time

AnshulPatil2005:
@elalish 
Deployed and verified. manifoldcad.org/master/ now loads its own assets rather than the root's.

/assets/        2b477588be9a      539,138   (Aug 15)
/master/assets/ ecc0816f16dd      540,892   (current)

Separate from this PR, but worth deciding: nothing writes site content to the gh-pages root any more. The release path only writes vX.Y.Z/, latest/, and the CNAME file. So manifoldcad.org/ stays on the Aug 15 build indefinitely. Options are pointing it at latest/, deploying master there as well as to master/, or leaving it frozen for now

elalish:
I'd like to have documentation on how to push a release to just our GitHub pages from a given branch. Use cases: when we need a quick fix, or just a docs update. If you can figure that out, please use it to push our prior release, so that we have something in latest. Then point manifoldcad.org at latest, and we can use master for testing. Maybe rename master to alpha in the URLs.

AnshulPatil2005:
> I'd like to have documentation on how to push a release to just our GitHub pages from a given branch. Use cases: when we need a quick fix, or just a docs update. If you can figure that out, please use it to push our prior release, so that we have something in latest. Then point manifoldcad.org at latest, and we can use master for testing. Maybe rename master to alpha in the URLs.

I'll write up the `workflow_dispatch` path from #1787. That already supports deploying any ref to any directory, it just isn't documented anywhere.

## Backfilling `latest/`

One snag: **v3.5.2 predates #1806**. It was tagged June 25, has no `base` configured, and still has absolute refs in `index.html`. Dispatching `ref: v3.5.2` directly would build a site whose HTML requests `/assets/manifoldCAD.js` from the site root, which is the stale Aug 15 build. So `latest/` would end up serving Aug 15 content rather than v3.5.2.

Plan to work around it:

1. Branch from the `v3.5.2` tag
2. Cherry-pick #1806 (`37e5b644`) onto it. Verified conflict-free, six files: four vite configs, `index.html`, `service-worker.js`
3. Dispatch the deploy twice from that branch, once with `destination_dir: v3.5.2` to create the version directory, and once with `destination_dir: latest`

I'll rehearse the whole thing on my fork first and report what it produces. Two things I'd need from you after that: the branch pushed to this repo, since `workflow_dispatch --ref` can't see a fork branch, and someone with write access to run the two dispatches.

One question: should the docs-link fix in #1807 ride along? Without it the Documentation button inside `latest/` points at the site root rather than `/latest/docs/`.

Renaming `master/` to `alpha/` is straightforward. Worth noting the old `/master/` directory would be orphaned, since the workflow never deletes directories it stops writing to, so it'd need removing by hand.

AnshulPatil2005:
On pointing manifoldcad.org at `latest/`, there are two ways to do it and they behave differently.

**A redirect at the root.** A small `index.html` that sends visitors to `latest/`. One file, stays clean.

**Copying `latest/` content to the root.** No redirect, the root serves real content. But a root deploy has no `destination_dir`, so the action's cleanup scope becomes the whole branch and it has to run with `keep_files: true`, otherwise it would delete every version directory. That means stale files accumulate at the root indefinitely.


Either way there's a leftover worth deciding on. The root currently holds the entire Aug 15 site: `index.html`, `assets/`, `three.html`, `model-viewer.html`, `icons/`, `samples/`. A redirect at `/` doesn't remove any of it, so `manifoldcad.org/three.html` and similar non-version-prefixed links would keep serving the old build indefinitely.

Do you want gh-pages pruned down to just the version directories, CNAME and the redirect? That's a manual cleanup, and it would risk breaking any deep link that isn't version-prefixed.

elalish:
I'd like to avoid redirects if possible. I suppose there's really no need for a `latest` directory - I want plain manifoldcad.org to be that already. So I guess just making a release copy its data to two places: e.g. v3.4.0 and the root. As for cleanup, it should mostly be the same files each time overwriting each other, so there's probably not too much to clean up? I suppose we can do it manually if it becomes a problem.

AnshulPatil2005:
> I'd like to avoid redirects if possible. I suppose there's really no need for a `latest` directory - I want plain manifoldcad.org to be that already. So I guess just making a release copy its data to two places: e.g. v3.4.0 and the root. As for cleanup, it should mostly be the same files each time overwriting each other, so there's probably not too much to clean up? I suppose we can do it manually if it becomes a problem.

Got it, that's simpler.

Issue #1796, opened 2026-08-18
Release pipeline issues 


AnshulPatil2005:
1. npm publishes master's version, not the tag's
When a release is published, the npm package is built from whatever is on master at that moment rather than from the tagged commit.

Master is currently 3.5.1 and v3.5.2 is already tagged, so publishing v3.5.2 today would push 3.5.1, which already exists on npm, and fail.

The download step keys on the branch rather than the commit:

- name: Download built examples
  uses: dawidd6/action-download-artifact@v6
  with:
    workflow: manifold.yml
    branch: master          # latest master, not the tag
    name: wasm
    path: ./bindings/wasm/
The wasm artifact is the whole bindings/wasm/ directory, package.json included, so it overwrites the one actions/checkout just pulled from the tag and the version comes with it.

PyPI is unaffected. build_wheels.yml builds from source at the tag and never touches a master artifact.

2. The version is hand-maintained in three files
CMakeLists.txt, bindings/wasm/package.json, and pyproject.toml each carry the version separately. 

3. The editor doesn't consume a published version
The editor and the three examples all depend on "manifold-3d": "file:../..", a symlink to the local tree. So nothing in the deployed site is tied to a released version.

4. Scripts can't pin a version(main blocker)
Once 1 through 3 are sorted, a user could write import ... from 'manifold-3d@3.4.1' in a ManifoldCAD script and know it won't break when a new version ships. That's the outcome #1721 is asking for, and 1 through 3 are what stand in the way of it.

elalish:
Yup, those sound like the right things to solve. Do you have a plan? Any questions?

AnshulPatil2005:
> Yup, those sound like the right things to solve. Do you have a plan? Any questions?

I had some questions after that I'll share a plan and start coding.

Q1(question on first point). To publish to npm you need the compiled WASM files. There are two ways to get them.

Option A: build it fresh. Inside the publish workflow, compile the WASM from the tagged commit right there. Takes about 15 extra minutes, but you know for certain it matches the tag.

Option B: reuse what CI already built. CI compiled the WASM when the commit was pushed to master and saved it. Just grab that saved copy, but point at the tagged commit instead of master. One line, instant.

The catch with B is that you're depending on something outside your control. That saved build has to exist and still be there. GitHub deletes them after 90 days, and if a release is ever tagged on a commit CI never ran on, there's nothing to grab.

So: certainty and 15 minutes, or speed and a dependency.

[...]

PR #1807, opened 2026-08-25
Versioned deploy follow-ups

AnshulPatil2005:
TBD 
#1806
refs:https://github.com/elalish/manifold/pull/1806#issuecomment-5402142914
https://github.com/elalish/manifold/pull/1806#issuecomment-5402172313

### DIFF
--- a/RELEASE_CHECKLIST.md
+++ b/RELEASE_CHECKLIST.md
@@ -12,4 +12,49 @@
 1. Check the Actions and verify that both PyPI and npm publishing actions ran successfully.
 1. Verify the npm [package](https://www.npmjs.com/package/manifold-3d?activeTab=code) looks good - unpacked size should be close to 1MB.
 1. Verify PyPI [package](https://pypi.org/project/manifold3d/#files) looks good - a bunch of built distributions ranging from ~600kB to ~1.1MB.
-1. If there's a problem with release deployment, the release workflows can be triggered separately, manually for any branch, under the Actions tab.
+1. If there's a problem with release deployment, the release workflows can be triggered separately, manually for any branch, under the Actions tab. See [Deploying manually](#deploying-manually) for the site deploy in particular.
+
+## Where the site is published
+
+`manifoldcad.org` is served from the `gh-pages` branch, which holds several
+builds side by side:
+
+| Path | Updated | Contents |
+| --- | --- | --- |
+| `alpha/` | every master commit that passes CI | unreleased build, for testing |
+| `latest/` | each published release | the newest release |
+| `vX.Y.Z/` | once, when that release is published | a frozen snapshot of that version |
+
+Each directory is a complete copy of the site: the ManifoldCAD editor, the
+examples, the benchmark dashboard, and both the C++ and TypeScript docs.
+Deploying to one never touches the others.
+
+## Deploying manually
+
+The Deploy documentation workflow can be run by hand from the Actions tab for
+a quick fix or a docs-only update, without cutting a release. It takes two
+inputs:
+
+- **ref** - the branch or tag to build from, e.g. `v3.4.0`
+- **destination_dir** - the directory on `gh-pages` to publish into, e.g. `v3.4.0`
+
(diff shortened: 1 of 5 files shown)


elalish (approved):
Thanks!

elalish:
Is this ready for review? Also, when pushing a release update for an older release, is there a way to ensure it doesn't overwrite the most recent release at the root directory?

AnshulPatil2005:
> Is this ready for review? Also, when pushing a release update for an older release, is there a way to ensure it doesn't overwrite the most recent release at the root directory?

Yes, ready for review. Taking it out of draft now.

On overwriting the root: a manual dispatch can't. The two steps are gated differently.

```yaml
- name: Deploy to vX.Y.Z/
  if: github.event_name != 'workflow_run'    # fires on dispatch and release

- name: Deploy to gh-pages root
  if: github.event_name == 'release'          # release only
```

So dispatching a fix for an older release writes only to that version's directory. The root is untouched, because the root deploy is gated on the event being an actual release.

One edge case that isn't covered: if an old release is *re-published* through the GitHub UI, `release: published` fires with that old tag and both steps run, so v3.4.0 would land at the root and clobber the current release. Editing a published release fires `edited` rather than `published`, so normal note-tidying is safe. The risk is really publishing an old draft, or unpublishing and republishing.

AnshulPatil2005:
also for this plan https://github.com/elalish/manifold/pull/1806#issuecomment-5402142914 which i posted 
I'd suggest we skip it and let the next release handle it.

The site's editor, examples and TypeScript docs aren't built by the deploy. They're built by CI and saved as the `wasmpublic` artifact, which only exists for commits pushed to master:

```yaml
if: github.event_name == 'push' && matrix.parallelization == 'OFF' && matrix.debug == 'OFF'
```

Backfilling means deploying a ref that was never on master, so there's nothing to download. Making it work means `deploy.yml` builds the whole site itself on a dispatch: emsdk, TBB and gtest, the full WASM compile, then the editor, examples and typedoc builds. Around 30 lines duplicating `manifold.yml`'s build config, with the two free to drift apart, and roughly 20 minutes per run.

That's a fair amount of new machinery for a one time thing.

## plan 

1. Merge #1807
2. PR for #1796, so releases publish correctly and versions stay in sync
3. Cut a release, which fixes the root, creates the version directory, and gets npm current

## After #1807 merges

one manual step: **delete `master/` from gh-pages** once `alpha/` has been created

RELEASE_CHECKLIST.md
```diff
@@ -12,4 +12,54 @@
...
+
+Subdirectory deploys erase their target before writing, so each holds exactly
+one build. The root deploy cannot, because clearing it would take the
+subdirectories with it, so there it overwrites in place instead.
+
+## Deploying manually
+
+The Deploy documentation workflow can be run by hand from the Actions tab for
+a quick fix or a docs-only update, without cutting a release. It takes two
+inputs:
```
elalish, line 41: >The site's editor, examples and TypeScript docs aren't built by the deploy. They're built by CI and saved as the wasmpublic artifact, which only exists for commits pushed to master

So, does a deploy documentation workflow update the TypeScript docs or not?

elalish (approved):
Let's keep this moving and continue the conversation - seems like we're heading in a good direction.

RELEASE_CHECKLIST.md
AnshulPatil2005: Only sometimes, and that's the bug here actually.

The deploy generates the C++ docs itself, so those always match the ref you asked for. But the editor, examples and TypeScript docs are built by CI and just downloaded as the `wasmpublic` artifact.

The problem is which artifact it grabs. On a manual dispatch it looks for the one belonging to the branch you clicked from, not the ref you asked to deploy. So dispatching `v3.4.0` from master gives you v3.4.0's C++ docs sitting next to master's editor and TypeScript docs. And if no artifact matches at all, `continue-on-error` skips it silently and publishes a site with no editor, showing green.

will be fixing it in the next pr

PR #1809, opened 2026-08-26
Fix degree sine reduction on Windows ARM64

ssss141414:
## Summary

Use explicit angle reduction in `sind()` only on Windows ARM64, where the existing `std::remquo` path produces incorrect extrusion geometry.

Other platforms retain the existing `std::remquo` implementation.

Add a regression test for extruding a unit square to verify the expected volume and surface area.

## Validation

Tested natively on Windows ARM64:

- built and installed a CPython 3.13 `win_arm64` wheel
- ran `bindings/python/examples/run_all.py`: all 12 examples passed
- ran the new `Manifold.ExtrudeSquare` regression test
- ran the complete native C++ test suite: all 544 tests passed

Before this fix, the extrusion test produced a volume of `11.666666666666668` instead of the expected `10.0`.

### DIFF
--- a/include/manifold/common.h
+++ b/include/manifold/common.h
@@ -109,8 +109,14 @@ constexpr double smoothstep(double edge0, double edge1, double a) {
 inline double sind(double x) {
   if (!la::isfinite(x)) return NAN;
   if (x < 0.0) return -sind(-x);
+#if defined(_WIN32) && defined(_M_ARM64)
+  x = std::fmod(x, 360.0);
+  const int quo = static_cast<int>(std::floor(x / 90.0 + 0.5));
+  x -= quo * 90.0;
+#else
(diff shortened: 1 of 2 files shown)


elalish (approved):
Is there an issue filed on Window's ARM64 compiler somewhere? Seems like there ought to be.

pca006132:
Are we even sure it breaks without the patch in the CI?

elalish:
> Are we even sure it breaks without the patch in the CI?

Yes, it's the second link in this [comment](https://github.com/elalish/manifold/pull/1805#issuecomment-5421306170), hence why I merged this.

pca006132:
OK interesting, this sounds like a serious bug that should be reported to MSVC...

PR #1810, opened 2026-08-28
Fix the release pipeline

AnshulPatil2005:
mainly focuses on fixing issues in #1796 (point 1 and 2)

follow up to #1807


AnshulPatil2005:
@elalish some decisions in here I'd rather have you check 

## Only bumping one of the two versions in `test-cmake.sh`

That file has two, and they mean different things:

```bash
find_package(manifold "3.5.0" REQUIRED)                    # a minimum
#if MANIFOLD_VERSION < MANIFOLD_VERSION_NUMBER(3, 5, 1)    # the current release
```

The script bumps the second and leaves the first alone. The `find_package` minimum is currently one release behind, which reads as deliberate, since a floor that always equals the current version isn't testing anything. Is that right, or should both track the release?

## Opening a PR rather than pushing commit to master

The workflow could commit the bump straight to master and save a step but 
Going through a PR keeps a human review in front of that, and matches the existing checklist.

AnshulPatil2005:
also this is ready for review

elalish:
@pca006132 do you remember the story with the two cmake versions? It escapes me at the moment.

pca006132:
I forgot that as well, I think maybe it is to test if the version reported is correct or something... Maybe we can just remove the check.

elalish:
@AnshulPatil2005 yeah, looks vestigial to me too - let's simplify. I don't want to keep multiple version strings hanging around anyway. And yes, I think the workflow pushing a PR is a good, especially if the workflow itself is fast (doesn't build or run tests) since that can all happen in the PR instead.

AnshulPatil2005:
done

`set_version.py` now bumps both the `find_package` minimum and the `MANIFOLD_VERSION_NUMBER` check, so the file only ever carries one version.

also,
Running the script at the *current* version, 3.5.1, still changed six files. The lockfiles were all sitting at 3.5.0 and `test-cmake.sh`'s `find_package` was too, so they'd been quietly drifting for a release. Those corrections are in this PR, which is a decent demonstration that the drift was real.

.github/workflows/prepare_release.yml
```diff
@@ -0,0 +1,64 @@
...
+          exit 0
+        fi
+        git config user.name 'github-actions[bot]'
+        git config user.email '41898282+github-actions[bot]@users.noreply.github.com'
+        git checkout -b "release/v$VERSION"
+        git commit -am "Set version to $VERSION"
+        git push origin "release/v$VERSION"
+        # Heredoc so the body is not indented by the YAML block scalar,
+        # which markdown would render as a code block.
+        cat > /tmp/pr-body.md <<EOF
```
elalish, line 52: This is cute - didn't know that was possible.

elalish (approved):
Thanks! Can we cherry-pick this and #1809 onto the 3.5.2 branch and try a 3.5.3 release to test this out?

AnshulPatil2005:
> Thanks! Can we cherry-pick this and #1809 onto the 3.5.2 branch and try a 3.5.3 release to test this out?

Branch is ready: `v3.5.3-release`, cut from `v3.5.2-release`.

- #1809 cherry-picked cleanly (ARM64 sine fix)
- From #1810, only the release-publishing parts: `publish_npm.yml` building from the tagged commit, plus `prepare_release.yml` and `set_version.py`

I left `deploy.yml` alone deliberately. #1810's changes to it assume the post-#1807 file, and master's version does `mv ./dashboard ./public` and `dashboard/` only arrived in #1758, after v3.5.2, so it would fail on this tree. Also skipped `RELEASE_CHECKLIST.md`, since master's copy documents the `alpha/` and versioned-directory layout that doesn't exist here.

also the site won't deploy for this release, and that's fine: `deploy.yml` at v3.5.2 only has a `workflow_run` trigger, since `release` was added in #1787. The v3.5.2 release confirms, it fired Release File, publish_npm and build_wheels, and no deploy. So 3.5.3 publishes to npm and PyPI without touching gh-pages, and `alpha/` is unaffected.

Verified `set_version.py` against this tree: 3.5.2 → 3.5.3 across all ten files, both `test-cmake.sh` strings and every lockfile. I left the version at 3.5.2 on the branch so running `prepare_release` is itself part of the test.

Sequence once you're happy with it: dispatch Prepare release with `3.5.3`, review and merge the PR it opens against this branch, then tag and publish from the merge commit.

AnshulPatil2005:
https://github.com/elalish/manifold/pull/1813

PR #1813, opened 2026-09-05
cherry-pick release pipeline for 3.5.3

AnshulPatil2005:
Branch is ready: `v3.5.3-release`, cut from `v3.5.2-release`.

- #1809 cherry-picked cleanly (ARM64 sine fix)
- From #1810, only the release-publishing parts: `publish_npm.yml` building from the tagged commit, plus `prepare_release.yml` and `set_version.py`

I left `deploy.yml` alone deliberately. #1810's changes to it assume the post-#1807 file, and master's version does `mv ./dashboard ./public` and `dashboard/` only arrived in #1758, after v3.5.2, so it would fail on this tree. Also skipped `RELEASE_CHECKLIST.md`, since master's copy documents the `alpha/` and versioned-directory layout that doesn't exist here.

also the site won't deploy for this release, and that's fine: `deploy.yml` at v3.5.2 only has a `workflow_run` trigger, since `release` was added in #1787. The v3.5.2 release confirms, it fired Release File, publish_npm and build_wheels, and no deploy. So 3.5.3 publishes to npm and PyPI without touching gh-pages, and `alpha/` is unaffected.

Verified `set_version.py` against this tree: 3.5.2 → 3.5.3 across all ten files, both `test-cmake.sh` strings and every lockfile. I left the version at 3.5.2 on the branch so running `prepare_release` is itself part of the test.

Sequence once you're happy with it: dispatch Prepare release with `3.5.3`, review and merge the PR it opens against this branch, then tag and publish from the merge commit.


elalish:
Thanks, but please remember to never open PR with an empty description - [This info](https://github.com/elalish/manifold/pull/1810#issuecomment-5552662966) should be in this PR's description instead of where it is.

>I left deploy.yml alone deliberately. https://github.com/elalish/manifold/pull/1810's changes to it assume the post-https://github.com/elalish/manifold/pull/1807 file, and master's version does mv ./dashboard ./public and dashboard/ only arrived in https://github.com/elalish/manifold/pull/1758, after v3.5.2, so it would fail on this tree.

Would it be better to cherry-pick these commits into 3.5.3 as well? Or is it better to keep it this way?

AnshulPatil2005:
> Thanks, but please remember to never open PR with an empty description - [This info](https://github.com/elalish/manifold/pull/1810#issuecomment-5552662966) should be in this PR's description instead of where it is.

done 

> > I left deploy.yml alone deliberately. #1810 changes to it assume the post-#1807 file, and master's version does mv ./dashboard ./public and dashboard/ only arrived in #1758, after v3.5.2, so it would fail on this tree.
> 
> Would it be better to cherry-pick these commits into 3.5.3 as well? Or is it better to keep it this way?

yes i actually tested this. I think its better to keep it as is. Replaying #1787, #1801, #1806 and #1807 gives three conflicts, all in `deploy.yml`, since this branch has the pre-#1787 version and each expects the one before it. There is a clean path cherry-pick #1806 on its own, which applies without conflict, then take master's `deploy.yml` whole minus the `mv ./dashboard ./public` line. Seven files, no conflict resolution. So it's doable.

But I don't think it's worth it

The point of 3.5.3 is proving the npm publish works, and the branch already does that with two commits and no deploy changes. Adding the deploy stack means exercising a path for the first time during a release, and #1806 has to come along or `v3.5.3/` would have the absolute-path bug and load the root's assets instead of its own. So it's all of it or none, not a partial.

There's also a visible consequence. The root currently serves the Aug 14 master build, including smooth quads and improved decimation. Deploying 3.5.3 there would replace it with v3.5.2-era code. That's correct under the model where the root is the current release, but it would read as the site losing features.

elalish:
Okay, that sounds reasonable, thanks for the explanation. I see this PR is to merge into `[elalish:v3.5.2-release](https://github.com/elalish/manifold/tree/v3.5.2-release)` - is that correct? Or should we have a new 3.5.3 branch? What are the next steps after we merge this and which can you do and which do I need to do?

pca006132:
I'm thinking about if there is a way to make release nicer. E.g., if I want to publish a certain branch as certain version (may not be the master, since we can have backport release to v3.5.x), how should I do that? Merge the PR not touching versions, and then publish a version, and it will automagically update the versions for me?

AnshulPatil2005:
> Okay, that sounds reasonable, thanks for the explanation. I see this PR is to merge into `[elalish:v3.5.2-release](https://github.com/elalish/manifold/tree/v3.5.2-release)` - is that correct? Or should we have a new 3.5.3 branch? What are the next steps after we merge this and which can you do and which do I need to do?

Yes, `v3.5.2-release` is the right base that branch becomes the 3.5.x maintenance line rather than a snapshot of one patch. A separate 3.5.3 branch would mean a new branch per patch.

Worth renaming it to something like `v3.5.x` at some point, since the name goes stale the moment 3.5.3 lands on it. Not urgent, and it can be renamed later without disruption.

**After this merges**

You need to do all of it, since every step needs write access here:

1. Actions → Prepare release → Run workflow, select branch `v3.5.2-release`, version `3.5.3`
2. It opens a PR bumping the version in all ten files. I can review that.
3. Merge it
4. Tag the merge commit `v3.5.3` and publish the release

Then three workflows fire off the release, all reading the tagged commit: `publish_npm` (which finally gets npm off 3.5.1), `build_wheels` for PyPI, and `release_file`. `deploy` won't fire, since `deploy.yml` on that branch has no `release` trigger, so the site is untouched and `alpha/` is safe.

One thing to watch on the first run: `workflow_dispatch` uses the workflow file from the branch you select, and `prepare_release.yml` only lands on `v3.5.2-release` when this PR merges. So run it after merging, not before.

AnshulPatil2005:
> I'm thinking about if there is a way to make release nicer. E.g., if I want to publish a certain branch as certain version (may not be the master, since we can have backport release to v3.5.x), how should I do that? Merge the PR not touching versions, and then publish a version, and it will automagically update the versions for me?

right, that's roughly what `prepare_release` does now, and it already works on any branch, it takes the version as an input and runs against whichever ref you select, so a backport to `v3.5.x` works the same as a release from master.

The flow is: merge your PRs normally without touching versions, then dispatch Prepare release with the branch and version. It bumps `CMakeLists.txt`, `pyproject.toml`, `flake.nix`, `package.json`, `test-cmake.sh` and all five lockfiles, and opens a PR. Merge that, tag it, publish.

The part that isn't automatic is tagging, and I think deliberately. Going fully automagic would mean tagging first and updating versions after, but then the tag wouldn't contain the version bump and the published artifacts wouldn't match the commit they claim to come from.

elalish:
Okay, I ran the workflow (nice UX!), but it's trying to merge into master, which doesn't look right: #1815.

AnshulPatil2005:
> Okay, I ran the workflow (nice UX!), but it's trying to merge into master, which doesn't look right: #1815.

That's my bug, sorry. gh pr create targets the repository's default branch when --base isn't passed, so the branch was created correctly off v3.5.2-release but the PR went to master.

That also explains why #1815's diff is so big it's the full difference between the release branch and master. Please don't merge it.

Fixed in #1817 for master and #1818 for the release branch. It passes --base "$GITHUB_REF_NAME" so the PR lands on whichever branch it was dispatched from, and force-pushes the generated branch so a re-run replaces a failed attempt instead of being rejected.

Both are needed, since workflow_dispatch uses the workflow file from the branch you select, master's copy alone wouldn't fix the 3.5.3 run.

To retry: merge #1818, close #1815, then re-run Prepare release against v3.5.2-release with 3.5.3. The old PR has to be closed first

PR #1814, opened 2026-09-06
Remove duplicate Smoothing APIs

elalish:
This is a major API simplification for smoothing, part of the breaking changes for v4.0. This was motivated by #1790, which will make it possible to completely remove our internal coplanar concept, except that our oldest smoothing API was still using it. 

Smoothing has gone through a lot of improvements over the years and I feel like `SmoothByNormals` is the modern version I'm happier with, as it handles a lot more cases. I regretted the `Smooth` constructors I made - their only purpose was to allow sharpened edges, but the more I look at it, the more I feel like this is a strange mid-level API. I think we're better served with just the high level based on normals, and the low-level where tangents are supplied directly. I've updated the Scallop example to demonstrate the lower level API and found it produced a better result anyway without that much extra code.


elalish:
@pca006132 I still need to fix a bug and update a few smoothing tests, but I wanted to run it past you early in case you have an opinion here.

pca006132:
I am happy about removing that, I didn't use it much anyway (I probably should do more modelling and try new stuff).

Icheb3891:
My framework removes all pathologies and can give you the gravitational load constants along with the orientation expression necessary for the dimensional modeling you're attempting for the mesh. You want to cut Python script down to a simplified as possible? What if I were to tell you no PDE differential equations needed? You just simply measure the curvature deficit needed compared to the emergent, convergent hypersphere geometry and *poof* your 40 pages of script collapse to a simple expression. Don't believe me? Gibbs: 2(pi-3)/pi.  Gibbs was 0.065% accurate to true convergence. Want to know more? Come play in my Dimensional Calculus repository.

PR #1820, opened 2026-09-07
fix the five high-severity zizmor findings

### LINKED ISSUE #1799: Static analysis findings in GitHub Actions workflows (zizmor)
I ran [zizmor](https://github.com/zizmorcore/zizmor) as suggested by @pca006132 over .github/workflows/. Reproducible with:

uvx zizmor --no-online-audits .github/workflows/

It reports 126 findings, but that number needs a caveat up front: 74 are marked high severity and 69 of those are a single rule, unpinned-uses, which flags every action not pinned to a commit SHA. That's a blanket policy

The five important ones 

1 and 2. Template injection: release_file.yml:14 and :29

- run: |
    ref=${{ github.ref_name }}

${{ }} is substituted into the script as text before the shell runs, so the value isn't a shell variable. Only people with write access can push tags, so this isn't reachable from outside. It's hardening against a malformed tag rather than a live hole.

Routing the value through env: clears all 7 template-injection findings, since the other 5 are the derived env.release_name. zizmor auto-fixes this one.

3. Cache poisoning: publish_npm.yml:18

actions/setup-node v6 has a package-manager-cache input defaulting to true, which auto-enables caching when package.json has a top-level packageManager field. Since this workflow publishes to npm, a poisoned cache restored here could reach the published package.

afaik fix should be package-manager-cache: false, one line.

4. Overly broad permissions: weekly_benchmark.yml:11

contents: write is granted at workflow level, so every job inherits repo write access

5. Unpinned container image:  manifold.yml:436

The MXE job runs openscad/mxe-x86_64-gui:latest. Because latest is mutable, a later run can pull different content than an earlier one.

The rest
52 findings are medium and low: mostly excessive-permissions on jobs that only read

logs attached 

[report.txt](https://github.com/user-attachments/files/31200512/report.txt)

AnshulPatil2005:
Fixes #1799. Verified with `uvx zizmor --no-online-audits .github/workflows/` — template-injection, cache-poisoning and unpinned-images are all at zero.

- **`release_file.yml`** — the tag goes through `env:` and later steps read `$release_name`, clearing all seven template-injection findings since six were the derived value.
- **`publish_npm.yml`** — `package-manager-cache: false` on setup-node. Also swapped `cache-apt-pkgs-action` for a plain `apt-get`, since it "always restores from cache" and this job publishes to npm. Slightly slower, negligible against the emscripten build.
- **`weekly_benchmark.yml`** — `contents: read` at workflow level, `write` only on the job pushing to `benchmark-data`. The `sanitizer` job only uploads artifacts.
- **`manifold.yml`** — MXE container pinned by digest, resolved from `:latest` on 2026-09-07.
- **`prepare_release.yml`** — `inputs.version` through `env:`.

Two of these weren't in the issue: the apt caching action and the `prepare_release` injection. Both only surfaced once the others were fixed, and both came in with #1810.

Worth knowing: updating the container digest is manual. `.github/dependabot.yml` only covers `pip`, so nothing bumps it automatically. Adding the `github-actions` ecosystem there would also make SHA-pinning the 69 `unpinned-uses` findings much less painful, if that's ever worth revisiting.

### DIFF
--- a/.github/workflows/manifold.yml
+++ b/.github/workflows/manifold.yml
@@ -459,7 +459,8 @@ jobs:
       matrix:
         parallelization: [OFF, ON]
     runs-on: ubuntu-latest
-    container: openscad/mxe-x86_64-gui:latest
+    # Resolved from openscad/mxe-x86_64-gui:latest on 2026-09-07.
+    container: openscad/mxe-x86_64-gui@sha256:2df7d1c09a50192a523b4f9ea1f6155228a90e405545149d5bb3b41907f61b55
     steps:
     - uses: actions/checkout@v6
     - name: Build
--- a/.github/workflows/prepare_release.yml
+++ b/.github/workflows/prepare_release.yml
@@ -29,7 +29,9 @@ jobs:
     # The lockfiles are edited in place rather than regenerated with npm
     # install, which keeps this fast and limits the commit to the version.
     - name: Set the version
-      run: python3 scripts/set_version.py "${{ inputs.version }}"
+      run: python3 scripts/set_version.py "$VERSION"
+      env:
+        VERSION: ${{ inputs.version }}
 
     # gh rather than a third-party action; release_file.yml already uses it.
     - name: Commit and open a pull request
--- a/.github/workflows/publish_npm.yml
+++ b/.github/workflows/publish_npm.yml
@@ -27,15 +27,14 @@ jobs:
       with:
         node-version: 25
         registry-url: 'https://registry.npmjs.org/'
+        # No shared cache on the publish path.
+        package-manager-cache: false
 
     - name: Update npm # Ensure npm 11.5.1 or later for trusted publishing
       run: npm install -g npm@latest
 
-    - name: Install apt packages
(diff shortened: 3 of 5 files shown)


elalish (approved):
Thanks!

AnshulPatil2005:
>Worth knowing: updating the container digest is manual. .github/dependabot.yml only covers pip, so nothing bumps it automatically. Adding the github-actions ecosystem there would also make SHA-pinning the 69 unpinned-uses findings much less painful, if that's ever worth revisiting.

@elalish need your opinion here, It was latest earlier but it was a security threat according to zizmor findings so i changed it to manual ie. pinning by digest fixes that but means we only pick up new images when someone updates the line by hand.

elalish:
Hmm, my instinct is to prefer automation to strict security in this setting. @pca006132 what do you think?

AnshulPatil2005:
also it might not be a tradeoff, pinning plus adding the github-actions ecosystem to dependabot would give reproducible runs and a PR when the image updates, so you'd still get the new one, just with a commit attached.

worth looking at this approach too

PR #1821, opened 2026-09-07
use node 24 lts

AnshulPatil2005:
`npm@latest` is now npm 12, which dropped Node 25 — it's odd-numbered so not LTS. Required is `^22.22.2 || ^24.15.0 || >=26.0.0`, which is why the v3.5.3 publish failed at `Update npm`.

Once this is in, 3.5.3 can be published without re-tagging: dispatch publish_npm **from this branch** with `ref: v3.5.3`. The workflow comes from the branch, the code from the tag.

Same fix for master separately.

### DIFF
--- a/.github/workflows/manifold.yml
+++ b/.github/workflows/manifold.yml
@@ -277,7 +277,7 @@ jobs:
     - uses: actions/checkout@v6
     - uses: actions/setup-node@v6
       with:
-        node-version: 25
+        node-version: 24
         cache: 'npm'
         cache-dependency-path: bindings/**/package-lock.json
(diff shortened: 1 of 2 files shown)


elalish:
Nice, looks like that worked! See anything problematic in the publish?

AnshulPatil2005:
> Nice, looks like that worked! See anything problematic in the publish?

Publish looks good. 

One thing that looked odd but is correct : the wasm is byte-identical to 3.5.1. #1809's fix sits behind #if defined(_WIN32) && defined(_M_ARM64), so it compiles out under emscripten.

PR #1826, opened 2026-09-15
track github actions with dependabot

### LINKED ISSUE #1823: Add the github-actions ecosystem to dependabot
Following on from the discussion in
https://github.com/elalish/manifold/pull/1820#issuecomment-5568508539, where
the container pinning question came up and it turned out pinning plus
dependabot might not be a tradeoff at all.

when the update happens we get a say if we want it or no.

`.github/dependabot.yml` only covers `pip` today, so nothing tracks our
GitHub Actions versions. They've drifted a long way:

| Action | Pinned | Latest |
| --- | --- | --- |
| `actions/checkout` | v6 | v7.0.1 |
| `actions/upload-artifact` | v4 | v7.0.1 |
| `actions/download-artifact` | v4 | v8.0.1 |
| `actions/setup-python` | v5 | v7.0.0 |
| `actions/setup-node` | v6 | v7.0.0 |
| `actions/cache` | v4 | v6.1.0 |
| `peaceiris/actions-gh-pages` | v3 | v4.1.0 |
| `codecov/codecov-action` | v5 | v7.0.0 |
| `dawidd6/action-download-artifact` | v6 | v24 |
| `mymindstorm/setup-emsdk` | v14 | v16 |

Four lines would surface these as PRs instead:

```yaml
  - package-ecosystem: "github-actions"
    directory: "/"
    schedule:
      interval: "weekly"

### PR DESCRIPTION
**Pull request by AnshulPatil2005**
Closes #1823

Adds the `github-actions` ecosystem so action versions get tracked. Minor and patch updates are grouped into one PR; majors stay separate so each can be reviewed against its own changelog.

## What to expect

The first run will open roughly ten PRs, since that is how far behind things are. That is the config working rather than misfiring.

Two of those need reading properly rather than merging on sight:

- `peaceiris/actions-gh-pages` v3 to v4 changes `force_orphan` behaviour. `deploy.yml` relies on `keep_files: true` at the site root, because clearing there would delete every version directory.
- `dawidd6/action-download-artifact` v6 to v24 is eighteen majors, so input names have almost certainly moved. Used in `deploy.yml`.

Also worth fixing while triaging: `actions/cache` is currently at both `@v4` and `@v5` in different workflows.

## What this does not cover

Five references are not semver, so there is no version for dependabot to compare:

```
DeterminateSystems/nix-installer-action@main
DeterminateSystems/magic-nix-cache-action@main
mattnotmitt/doxygen-action@edge
psf/black@stable
pypa/gh-action-pypi-publish@release/v1
```

Those stay manual regardless.

## Follow-up

The container digest pinned in #1820 is still updated by hand. If dependabot's `docker` ecosystem picks up a `container:` line inside a workflow, adding it would close that out too. I could not confirm either way, so it seems worth trying once this is running.


elalish (commented):
LGTM, thanks. Why is it a draft?

AnshulPatil2005:
> LGTM, thanks. Why is it a draft?

Ready now, thanks.

elalish (approved):
Thanks

Issue #1825, opened 2026-09-15
Windows ARM64 support


ssss141414:
Hi manifold maintainers @elalish @AnshulPatil2005 ,

To support win_arm64 version package, I've merged this PR: https://github.com/elalish/manifold/pull/1805

I notice that in latest 3.5.3 release https://pypi.org/project/manifold3d/3.5.3/#files, it still doesn't have the win_arm64 version.... May I ask if we still need some code change or some process to support win_arm64? Appreciate for any insight!

Thanks!

AnshulPatil2005:
Good catch, and it's on me.

#1805 didn't make it into 3.5.3. (missed it during cherry pick)
only the release-pipeline fix and #1809 were cherry-picked across.

So the ARM64 wheel job simply never ran. No further code changes needed #1805 is complete.

Worth noting #1809, your degree-sine fix, *is* in 3.5.3, so only the wheel build is missing.

@elalish we could add the wheel to 3.5.3 directly, but it'd mean temporarily relaxing the publish gate and leaving a version whose files changed after release so I was thinking shall I just cherry-pick #1805 onto v3.5.2-release and cut 3.5.4 instead?

also maybe if we are cutting a release anyway, it's worth checking whether anything else on master should ride along ?

elalish:
Thanks for checking this - welcome to why I avoid cherry-picking into releases when I can. If it's just the wheel, can we trigger just that job, or even run it manually from the command line? I'd prefer adding the missing piece to putting out a 3.5.4.

AnshulPatil2005:
Yes, and no workflow changes needed for the build part. `actions/checkout` in `build_wheels.yml` takes no `ref`, so a dispatch builds whatever branch you point it at.

Plan: branch off the `v3.5.3` tag, cherry-pick just #1805's three lines, push, then dispatch `build_wheels` against that branch. `pyproject.toml` there already says 3.5.3, so cibuildwheel produces `manifold3d-3.5.3-*-win_arm64.whl` 

The only snag is publishing. `pypi-publish` is gated on `github.event_name == 'release'`, so a dispatch builds the wheel as an artifact but doesn't upload it. Two ways round that:

1. Download the `python-windows-11-arm` artifact and `twine upload` it yourself. Nothing in the repo changes.
[...]

PR #1827, opened 2026-09-15
Publish the missing 3.5.3 win_arm64 wheel

AnshulPatil2005:
Adds the `win_arm64` wheel missing from 3.5.3, per #1825. **Not for merging** — this branch exists only so `build_wheels` can be dispatched against it.

3.5.3 was cut from `v3.5.2-release` before #1805 reached master, so the tag's matrix has no `windows-11-arm` entry. PyPI's 3.5.3 has 43 wheels and none of them is native Windows ARM64.

Three changes on top of `v3.5.2-release`:

- Cherry-picks #1805, which adds `windows-11-arm` to the matrix and limits it to cp311-cp314.
- Drops `if: ${{ github.event_name == 'release' }}` from `pypi-publish`, so a manual dispatch reaches it. The 3.5.3 release is already published, so that event cannot fire again.
- Adds `skip-existing: true`, so the run uploads the one new wheel and skips the files already on PyPI rather than failing on them.

`pyproject.toml` on this branch is already `3.5.3`, so the wheel is built and uploaded under the right version.

To use it: push this branch to this repo, then run `build_wheels` from the Actions tab with the branch selected. The dispatch dropdown only lists branches here, so a fork branch cannot be selected. `workflow_dispatch` reads the workflow file from the selected branch, so the relaxed gate and `skip-existing` apply to that run and nowhere else. Neither master nor `v3.5.2-release` gets them, and the branch can be deleted once the wheel is up.


AnshulPatil2005:
@elalish 
#1827 gets the win_arm64 wheel onto 3.5.3 without cutting a release, but it needs a throwaway branch to do it.

I had a thought and would like to propose making this a normal operation instead.

**Today**

`pypi-publish` runs only on a release event:

```yaml
if: ${{ github.event_name == 'release' }}
```

That is the only path to PyPI. A dispatch builds the wheels and then skips the publish job, so there is no supported way to add a file to a version that already shipped. That is why #1827 needs a branch.

**Proposed**

Allow two paths instead of one.

- Release event: unchanged. This is what creates a version.
- Dispatch: allowed to publish, but only to add missing files to a version that
  already shipped.

The dispatch path would be add-only. It cannot create a version, because the tag
has to already be a published release. It cannot replace a file, because
`skip-existing` skips anything already there. Worst case it uploads nothing.

That needs a `ref` input, a `publish` boolean, and one check on the publish job:

```yaml
- name: Only backfill an existing release
  if: github.event_name == 'workflow_dispatch'
  env:
    REF: ${{ inputs.ref }}
    GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
  run: |
    gh release view "$REF" --json tagName > /dev/null
    version=$(grep -m1 '^version = ' pyproject.toml | cut -d'"' -f2)
    [ "v$version" = "$REF" ]
```

With that, this issue would have been one dispatch against `v3.5.3`. No branch, and the matrix comes from master so newly added platforms get picked up automatically.

**Two notes**

Worth keeping this in `build_wheels.yml` rather than a new workflow. PyPI trusted publishing is configured against a specific workflow filename, so a new file would need a second publisher added on pypi.org first.

`skip-existing: true` is worth having regardless of the rest. Today if the publish job dies partway through the upload, a re-run hard fails on the files that already landed.

also Worth adding: this is already the pattern on the npm side, so it would not be a new idea in the repo.

`publish_npm.yml` on master:

```yaml
  workflow_dispatch:
    inputs:
      ref:
        description: 'Branch or tag to publish from (e.g. v3.5.2)'
        required: true
```

```yaml
    - uses: actions/checkout@v6
      with:
        ref: ${{ (github.event_name == 'workflow_dispatch' && inputs.ref) || github.ref }}
```
worth doing ?

elalish:
I think that's fine - honestly I'm pretty sure it'll be a lot quicker to read the two lines in the form of a PR than this essay. I hope you didn't take the time to write all that - it has the *very long-winded* sound of AI about it. Let's try to stay pithy. 

>push this branch to this repo
It's late, I'm tired, and that's a lot of "this". I think you're telling me not to actually merge this PR? So can you just give me the commands you need me to run?

AnshulPatil2005:
> I think that's fine - honestly I'm pretty sure it'll be a lot quicker to read the two lines in the form of a PR than this essay. I hope you didn't take the time to write all that - it has the _very long-winded_ sound of AI about it. Let's try to stay pithy.
> 
> > push this branch to this repo
> > It's late, I'm tired, and that's a lot of "this". I think you're telling me not to actually merge this PR? So can you just give me the commands you need me to run?

my bad for the text, I had a hard time explaining that thing, anyways Right, don't merge it. Closing #1827 in favour of the ref-input change, which
handles the backfill without a branch. Opening that now.

PR #1833, opened 2026-09-16
let build_wheels add a missing file to a released version

AnshulPatil2005:
3.5.3 shipped without a `win_arm64` wheel, because #1805 landed after it was branched (#1825). `pypi-publish` only runs on a release event, so there is no way to add it short of cutting 3.5.4.

Adds a `tag` input and a `publish` boolean. A dispatch with both set publishes, but only after checking the tag is an already-published release and that `pyproject.toml` matches it. With `skip-existing`, a dispatch can add files to an existing release and can never replace one.

Fixes 3.5.3 once this is in:

    gh workflow run build_wheels.yml -f tag=v3.5.3 -f publish=true

`skip-existing` is worth having regardless: a release publish that dies partway through the upload is currently not re-runnable.


.github/workflows/build_wheels.yml
```diff
@@ -57,12 +68,38 @@ jobs:
...
+    - uses: actions/checkout@v6
+      with:
+        ref: ${{ inputs.tag || github.sha }}
+    - name: Check the tag is an already-published release
+      if: github.event_name == 'workflow_dispatch'
+      env:
+        GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        TAG: ${{ inputs.tag }}
+      run: |
+        if [ -z "$TAG" ]; then
```
elalish, line 85: What is STAG?

elalish (approved):

.github/workflows/build_wheels.yml
AnshulPatil2005: `$TAG`, the env var set from the `tag` input three lines up.

It goes through `env:` rather than `${{ inputs.tag }}` inline because
inserting an input straight into a `run:` block is the template-injection
pattern which zizmor flags

AnshulPatil2005:
works 
PyPI 3.5.3 went from 43 files to 47

PR #1834, opened 2026-09-16
fix faceID round trip

elalish:
I realized my earlier refactor of faceID had made a round-trip problem, where triangle IDs were automatically getting copied into faceIDs on meshGL export, but that would mean on import you'd get a different result that would now stop coplanar triangles from being collapsed. This is fixed here, with an added testing hook function `TriID2FaceID` that does that behavior on demand, since it's necessary to run our RelatedGL check.


src/manifold.cpp
```diff
@@ -416,6 +416,7 @@ Manifold Manifold::RemoveDegenerates() const {
     return PropagateStatus(leafImpl->status_);
   auto impl = std::make_shared<Impl>(*leafImpl);
   impl->RemoveDegenerates();
+  impl->SortGeometry();
```
elalish, line 419: This was a bug - this function is necessary because it also removes all the verts that have been set to NaN to indicate they're not part of the mesh anymore.

PR #1835, opened 2026-09-17
track cache-apt-pkgs-action by major tag

AnshulPatil2005:
Per https://github.com/elalish/manifold/pull/1828#discussion_r4030211397: `awalsh128/cache-apt-pkgs-action` publishes a moving `v1` tag, so pointing at it keeps us current without a PR per patch.

The other three pinned actions stay as they are, since none of them publish a major tag to point at:

- `pypa/cibuildwheel@v3.4.0` (has `3.3.x` branches, no `v3`)
- `lukka/get-cmake@v4.4.2`
- `DoozyX/clang-format-lint-action@v0.20` (0.x, where a minor bump is the breaking one)


elalish (approved):
Thanks!

PR #1837, opened 2026-09-18
docs: add Manifold 3D MCP to users

zhicwan:
## Summary

Add [Manifold 3D MCP](https://github.com/zhicwan/manifold3d-mcp) to the README users list. It provides MCP tools and a Copilot Canvas extension for AI-assisted parametric CAD and 3D printing, powered by the `manifold-3d` TypeScript/WASM package.

## Testing

Documentation-only change.


pca006132:
Nice project but the name sounds too similar to our package...

zhicwan:
> Nice project but the name sounds too similar to our package...

Thanks for reviewing! That's fair — I didn't put much thought into the name when I started the project. It began as a side project built heavily on Manifold, wrapping it with MCP tools and a viewer to provide a connected create, inspect, annotate, and refine workflow.

I can see how the current name could be confused with the `manifold-3d` package or imply an official affiliation. I'll come up with something more distinct and update the PR.

pca006132:
Yeah something a bit more distinct will be nicer. But also, maybe we can integrate your tool into our manifoldcad as well?

Btw, I don't recommend stl format. 3mf, gltf/glb, obj, any of these is nicer than stl since stl loses topology.

zhicwan:
I see ManifoldCAD as a powerful, professional, code-first tool, while my project explores how AI can make precise 3D part modeling accessible to more people. Users shouldn’t need to understand the underlying code or representation—they can simply point to an area and describe what they want changed. I’m also exploring visual annotations and WebXR for inspecting and commenting on models. I’d like to keep the experiences distinct, but I’m open to lightweight integration.

Thanks for pointing out the limitations of STL as well. I’ll look into adopting 3MF and GLB where appropriate.

zhicwan:
Feel free to play around with it. I’d love to hear what you think! 🙂 I’ll keep the PR as a draft while I figure out a better name.

elalish:
Very cool! Honestly, if it really is built entirely on top of Manifold, I think it's reasonable to include Manifold in the name if you want. However, in reading about it I had absolutely no idea what MCP was - had to google it. Even the acronym isn't very descriptive. I might suggest something like Manifold AI. 

I appreciate your focus on the vibe-coding aspect, but I bet a fair number of users would also appreciate being more in-the-loop and using your tool in the context of something like ManifoldCAD. Worth considering if it could be built in a such a way that parts could be reused across projects.

PR #1839, opened 2026-09-18
track manifold.wasm size on the dashboard

AnshulPatil2005:
Tracks `manifold.wasm` size on the dashboard, as two series.

A `wasm_size` job builds it weekly with `publish_npm.yml`'s flags.
`collect_wasm_release_sizes.py` reads the published size of every release from
jsDelivr, so the chart has history back to 1.0.0 instead of starting empty.

its a separate ubuntu job so the pinned mac runner stays free for timing, and a
broken wasm build doesn't cost the week's benchmark data.

Tested on a fork:
https://github.com/AnshulPatil2005/manifold/actions/runs/35394788323


scripts/collect_wasm_release_sizes.py
```diff
@@ -0,0 +1,88 @@
...
+        key=published.get,
+    )
+
+    releases = []
+    skipped = []
+    for version in versions:
+        size = wasm_size(version)
+        if size is None:
+            # 2.3.0 is published but ships no wasm. Recording it as zero would
+            # put a false cliff in the chart.
```
elalish: 👍

elalish (commented):
This is great, thank you! One question: what is the metric for the size being recorded / reported? I believe the most useful metric is the gzipped size, as this is what you need to wait for over the network, rather than the uncompressed size. Either would work for tracking, but if someone compares to other web libraries, generally the smaller number is shown.

AnshulPatil2005:
> This is great, thank you! One question: what is the metric for the size being recorded / reported? I believe the most useful metric is the gzipped size, as this is what you need to wait for over the network, rather than the uncompressed size. Either would work for tracking, but if someone compares to other web libraries, generally the smaller number is shown.

Uncompressed right now. yea gzipped is the better number, switching it.

also we can keep the uncompressed value alongside it if useful ?

it's free once the file is built

elalish (approved):

